PR #2189, opened 2014-04-07
Load EDD on `plugins_loaded`

danielbachhuber:
If EDD is included as an MU plugin, this will prevent it from instantiating itself too early.

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -334,4 +334,4 @@ function EDD() {
 }
 
 // Get EDD Running
-EDD();
+add_action( 'plugins_loaded', 'EDD' );


pippinsplugins:
Let's test this for 2.6. If we can test 50-100 extensions and don't find any examples of break points, let's consider putting it in.

ghost1227:
Note that this will also break our installation/upgrade routines. We run the install routine using `register_activation_hook`, which can't be called after... well... anything. Even if we split the activation function into a separate function after the class instantiation, it depends on other things (like our constants) that don't exist until after instantiation. I'd LOVE it if someone can show me a way to do it that's both sane and doesn't break anything though...

pippinsplugins:
@ghost1227 Do you mean the ones in core or the ones in other extensions?

ghost1227:
i mean the ones in core. right now, we’re including install.php from inside the `Easy_Digital_Downloads` class. Since we're instantiating the class at runtime (instead of from inside an action like `plugins_loaded`), the `register_activation_hook` function works as expected. However, if we move to loading EDD on `plugins_loaded`, that function will no longer work. From the codex:

<blockquote>Registering the hook inside the 'plugins_loaded' hook will not work. You can't call register_activation_hook() inside a function hooked to the 'plugins_loaded' or 'init' hooks (or any other hook). These hooks are called before the plugin is loaded or activated.</blockquote>

The most obvious solution is to instantiate the main class at `plugins_loaded` (as suggested), and simply move the activation hook outside of the class. This, however, introduces its own set of problems...

1. We use the constants `EDD_PLUGIN_DIR`, `EDD_PLUGIN_FILE` and `EDD_VERSION` during the install process. These will not have been defined yet.
2. We use several functions which we will not have access to given they are loaded through the main class:
  - `edd_setup_edd_post_types`
  - `edd_setup_download_taxonomies`
  - `edd_get_option`
  - `edd_get_registered_settings`
  - `edd_create_protection_files`
  - `edd_set_upgrade_complete`
3. We also reference several methods instantiated during the instantiation of the main class:
  - `EDD()->customers->create_tables()`
  - `EDD()->sessions->use_php_sessions()`
  - `EDD()->roles->add_roles()`
  - `EDD()->roles->add_caps()`
  - `EDD()->api->get_version()`

We could, obviously, load all the dependencies during the installation process... but that'd be a lot of recursion... We'd have to instantiate EDD so we could install and instantiate EDD!

pippinsplugins:
Ah! Yes, I see what you mean. I've had to do that for AffiliateWP. Thanks for the reminder.

I'd like to try it, though we may find it's not worth the change.

ghost1227:
What if we moved all the relevant stuff to a new class `EDD_Loader` and instantiate that during both the install process and from the main `Easy_Digital_Downloads` class? Wouldn't be that ridiculous a conversion and would make life easier.

pippinsplugins:
Closing as maybe later.

PR #2276, opened 2014-05-08
Custom extra columns for view order details

bseddon:
I've updated the code to accommodate the omission you mentioned and made the changes against a fresh fork.

I took the opportunity to modify the implementation so it's more analogous to adding a column to a list table: there's a filter to request an array of additional columns; one to return a value for a row; and an action to create an input element of the user's choosing.  As long as the user follows the pattern an input field will be displayed within the 'inside' row and it will be recorded as a 'hidden' element like all the other fields.  The user can then use the 'edd_updated_edited_purchase' filter to process their extra fields.

The admin-script.js has been updated to add 9 lines to handle processing of 'extra' columns.  Because I don't have a JavaScript compressor, the -min version has been updated but is not minified.

Here is an example implementation of the two filters and one action:
```php
/** Return an array of extra column names to be presented. 
 *  The array key will be the css selector to use while the value will be the field label
 *
 */
add_filter( 'edd_view_order_details_extra_columns', 'view_order_details_extra_columns' );
function view_order_details_extra_columns( $extra_columns )
{
	// Should check if the extra_columns parameter is an 
	// array and, if it is, then add to the array.  In this 
	// example it's good enough to return a new array.
	return array( 'currency' => 'Currency' );
}
```
```php
/** Used to get a value for a row
 * @param int $item_id Is the EDD id of the row
 * @param int $key Is ordinal number of the row for which a value is being requested
 * @param string $extra_column Is the extra column for which a value is being requested
 * @param array $cart_item Is the cart content for the row being rendered
 */
add_filter( 'edd_view_order_details_downloads_extra_column_value', 'view_order_details_downloads_extra_column_value', 10, 4 );
function view_order_details_downloads_extra_column_value( $item_id, $key, $extra_column, $cart_item ) {
	// This value is hard-coded but in reality it would 
	// be a dynamically generated value recovered from 
	// wherever the value is stored - may be the cart item
	return esc_attr( 'XXX' );
}
```
```php
/** Used to render the element to capture the value for a new row
 * @param int $item_id Is the EDD id of the row
 * @param int $key Is ordinal number of the row for which a value is being requested
 * @param string $extra_column Is the extra column for which a value is being requested
 * @param array $cart_item Is the cart content for the row being rendered
 * @param string $name The value that should be used as the 'name' attribute's value on the rendered element
 * @param string $id The value that should be used as the 'id' attribute's value on the rendered element
 */
add_action( 'edd_view_order_details_download_extra_column_render', 'view_order_details_download_extra_column_render', 10, 6 );
function view_order_details_download_extra_column_render( $item_id, $key, $extra_column, $cart_item, $name, $id ) {
	// This example only adds one additional column
	if ( $extra_column !== "currency" ) return;
?>
	<input type="text" name="<?php echo $name; ?>" id="<?php echo $id; ?>" value="" placeholder="" class="small-text edd-order-download-currency">
<?php
}
```

Extra columns will require additional css.  Here is an example

```css
#edd-edit-order-form .columns-5 .column, #edd-purchased-files.columns-5 .row > ul > li,
#edd-purchased-files.columns-5 .inside > ul > li { width: 18%; margin-right: 2%; }
#edd-edit-order-form .columns-5 .column:nth-child(4n+1),
#edd-purchased-files.columns-5 li.download { width:33%; }
#edd-purchased-files.columns-5 li.price,
#edd-purchased-files.columns-5 li.actions { width: 20%; margin-top: 0; margin-right: 0; }
```

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -389,7 +389,7 @@ jQuery(document).ready(function ($) {
 					var value = $( this ).val();
 					name = name.replace("edd-order-download-", "");
 					clone.find( 'input.edd-payment-details-download-' + name ).val( value );
-					clone.find( '.currency' ).text( value );
+					clone.find( '.extra-column' ).text( value );
 				});
 
 				// Replace the name / id attributes
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -353,6 +353,8 @@ jQuery(document).ready(function ($) {
 				var price_name     = $('.edd_price_options_select option:selected').text();
 				var quantity       = $('#edd-order-download-quantity').val();
 
+				var extra_columns = $('.extra-column input');
+
 				if( download_id < 1 ) {
 					return false;
 				}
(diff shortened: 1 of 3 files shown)


pippinsplugins:
Closing due to lack of feedback. Feel free to re-open.

PR #2381, opened 2014-06-17
Global ajax functions

somewebmedia:
Created one global function edd_ajax() that receives data from all other
functions and delivers the request. Each ajax call has it's own function
(edd_add_to_cart, edd_remove_from_cart, edd_update_quantity...), and
accepts it's properties, than calls edd_ajax(). Each of them accepts two
extra parameters, success and fail. They are the ajax callback functions
and are optional.

This way developers can use those functions without the need to know or
write their own ajax requests, so they can manipulate HTML on callback
in their own way. Default EDD button events and callbacks stayed the
same, but also go through the new functions.

<strong>!important this should be more properly tested!</strong>

### DIFF
--- a/assets/js/edd-ajax.js
+++ b/assets/js/edd-ajax.js
@@ -1,314 +1,462 @@
+
+/*
+ * Main ajax request
+ */
+function edd_ajax(postData, success, fail){
+
+	if (!postData) return false;
+
+	if (typeof postData.action == 'undefined') return false; // we need to have action
+
+	var url;
+
+	// check for url
+	if (typeof postData.url != 'undefined') {
+		url = postData.url;
+		delete postData.url
+	} else {
+		if (typeof edd_scripts != 'undefined')
+			url = edd_scripts.ajaxurl;
+		else if (typeof edd_global_vars != 'undefined')
+			url = edd_global_vars.ajaxurl;
+		else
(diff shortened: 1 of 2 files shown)


pippinsplugins:
@somewebmedia sorry about this getting lost.

I'd love to hear @easydigitaldownloads/core-devs opinions on this. There's some tremendous value in this for sure.

spencerfinnell:
Any advantage to this over using `wp.ajax.post`or `wp.ajax.send`, which accepts success/error callbacks, and a few more options?

https://github.com/WordPress/WordPress/blob/master/wp-includes/js/wp-util.js#L72

pippinsplugins:
@spencerfinnell It's more the functions that let developers easily add an item to the cart, remove a discount, etc, from JS.

I would like to see them use the `wp` functions though.

spencerfinnell:
Yeah I like splitting out all the reusable stuff, but most of the things in `edd_ajax` are already handled by `wp.ajax.send` -- could still be used as a wrapper though to set a few options (especially the URL)

somewebmedia:
@pippinsplugins @spencerfinnell I can try replacing jQuery ajax with wp's ajax and see how it goes.

pippinsplugins:
Great. When you do, let's please open this as a new issue and a new PR (both issue and PR please).

Issue #1756, opened 2013-11-26
Force EDD test mode with secret URL


sumobi:
There should be a secure URL that we could type in or append to the URL of a site to force the site into test mode and be able to test purchases without disrupting real sales.

This would be extremely helpful for

1. EDD support to diagnose customer's payment issues, such as "pending payments" without disrupting real sales.
2. Customers wanting diagnose their own payment issues while their site is live, and don't have a development site.

For example, if EDD support needs to diagnose why a payment is "pending" after coming back from PayPal, it's much easier to turn on EDD's test mode and hook up a PayPal sandbox account. This of course allows testing of purchases without spending actual money. 

Currently we need to turn EDD test mode on, which will force PayPal's URL to be the sandbox URL. If a real customer should be on the live site at the same time and makes a purchase, it would not work, and a sale would be lost.

Pippin has suggested:

1. Appending a security token
2. Displaying this URL somewhere in the admin to be used
3. URL must be unique to ensure no one abuses it
4. Will need to do cap checks

cklosowski:
@ghost1227 Can you make the suggested changes and submit a new PR against ```release/2.3```?

pippinsplugins:
@ghost1227 Could you update your PR again?

ghost1227:
@pippinsplugins @cklosowski Updated

pippinsplugins:
Anyone have any further thoughts on this?

ghost1227:
Happy to do updates again if needed...

pippinsplugins:
The proposed code changes look good. I'm more looking for feedback on whether we definitely want to do this.

pippinsplugins:
Per the discussion in the dev meeting today, we have decided to not go ahead with this.

PR #2411, opened 2014-06-29
Initial pass at adding support mode

ghost1227:
Ref: #1756

This is a first-round patch for adding the ability to force a 'support mode' to EDD.


cklosowski:
Closing this out so it can be resubmitted against 2.3.

PR #2418, opened 2014-06-30
Restrict loading EDD scripts

### LINKED ISSUE #2364: loading admin styles/scripts only when necessary
As per this ticket https://easydigitaldownloads.com/support/topic/admin-scripts-not-enqueued-to-custom-post-type/#post-255502 It wouldn't hurt to check and see if any of this could be tightened up.

evertiro:
Fixes #2364

I THINK this covers all possible scenarios for pages requiring EDD scripts without loading it anywhere crazy. Removed `page-new.php` and `page.php` as those are covered by the existing `'download' == $typenow` check and moved `'index.php' == $pagenow` to a separate check with the condition that the query string is empty so that it ONLY loads on the dashboard homepage. Seems loading it on all children of index.php was causing issues with other plugins...

### DIFF
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -56,12 +56,16 @@ function edd_is_admin_page() {
 		return false;
 	}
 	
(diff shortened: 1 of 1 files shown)


JJJ:
> This won't work because we need EDD styles on the post.php and post-new.php screens for our "Insert Download" modal.

#6848 addresses this by introducing post-type support, to define what types allow for the "Insert Download" UI. `post` and `page` are supported by default, and others are easily added.

PR #2594, opened 2014-09-08
Fixed incorrect remove by ajax cart IDs and amounts

mintplugins:
When removing an item from the cart widget, it was removing the item we wanted it to, but not resetting the cart ID (data-cart-item). This means that if you had 3 items in the cart(0, 1, 2) and removed item 1, the items left were still labelled as 0 and 2, rather than 0 and 1. This meant that if you, then, removed item 3, it didn't exist. It would get removed from the screen using javascript BUT it didn't actually get removed from the PHP session. If you then refreshed the page you would see that the item is still in your cart - even though you thought you had removed it. This solves that problem by re-assigned the right data-cart-item ids after an item is removed.

Signed-off-by: Phil Johnston <support@mintplugins.com>

### DIFF
--- a/assets/js/edd-ajax.js
+++ b/assets/js/edd-ajax.js
@@ -33,25 +33,26 @@ jQuery(document).ready(function ($) {
 
 	                // Remove the selected cart item
 	                $('.edd-cart').find("[data-cart-item='" + item + "']").parent().remove();
-
+					
+					//Reset the data-cart-item attributes to match their new values in the EDD session cart array
+					var cart_item_counter = 0;
+					$('.edd-cart').find("[data-cart-item]").each(function(){
+						$(this).attr('data-cart-item', cart_item_counter);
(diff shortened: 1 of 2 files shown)


cklosowski:
@mintplugins this looks good to me, I could replicate the bug, and verified your changes fix it.

Can you remove the changes to the .min.js? We don't typically pull those in via PR. Just a common place for conflicts, like we've got here.

We'll re-minify it once merged.

mintplugins:
Should be good to go. I re-tested it on my end and it appears to be working. 

I found/opened another issue I found while testing this - somewhat related but when adding items to the cart:

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/2926

cklosowski:
Awesome thanks @mintplugins will take a look at that new issue too.

cklosowski:
Looks good to me.

PR #2779, opened 2014-11-15
Adding multiple product variations to cart

ericnicolaas:
This fixes a bug that made it impossible to add multiple variations of the same product to the cart at once.

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -129,13 +129,15 @@ function edd_add_to_cart( $download_id, $options = array() ) {
 	} else {
 		$quantity = 1;
 	}
-
+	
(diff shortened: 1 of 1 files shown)


cklosowski:
I'm not having any issues on this either @pippinsplugins

chriscct7:
This is regression #3070 also fixes.

chriscct7:
To be more specific the regression is the commit (see #3070 for a link) that @cklosowski did. Before you could pass in an array of variation id's and since that commit you can't. 

As #3070 is more complete (it addresses and improves the regression), and since this pull request doesn't merge cleanly, closing in favor of #3070

PR #2793, opened 2014-11-18
Add edd_items_in_cart cookie on add to cart

davekiss:
If this cookie doesn't exist before the next page loads, users could potentially see a cached page on systems that check for the cookie's presence before serving uncached content.

Also needs to be added to the minified js file.


pippinsplugins:
@davekiss Were you still interested in adding the localized approach?

pippinsplugins:
Closing due to lack feedback. @davekiss if you're still interested in this, let me know!

grappler:
@pippinsplugins I am interested in looking into this. If the cookie is not set when adding the download to the cart then the site will continue to serve the cached pages till the visitors views a non-cached page which will set the cookie.

pippinsplugins:
@grappler I'd be more than happy to revisit it.

Issue #2794, opened 2014-11-18
Remove download attribute on file download links


pippinsplugins:
In #2393 we added the `download` attribute to file download links.

This is a nice attribute that helps instruct browsers to download the linked file.

Unfortunately, I discovered a really negative side-effect of it.

While working with a user on [this ticket](https://easydigitaldownloads.com/support/topic/setting-has_access-to-false-in-edd_process_download-blocks-display-of-error/), I discovered that it was not possible to disable file download access and have it display an error.

Our `process-download.php` file contains a filter called `edd_file_download_has_access` that is designed to allow developers to prevent a file from being downloaded if some supplemental validation fails.

For example, a developer might want to restrict file downloads to logged in users, so they could do this:

``` php
add_filter( 'edd_file_download_has_access', 'is_user_logged_in' );
```

With that in place, only logged in users would be able to download files.

The user in the ticket linked above couldn't make this work, so I started working on replicating it.

As a simple test, I added this to disable all file downloads:

``` php
add_filter( 'edd_file_download_has_access', '__return_false', 10 );
```

What I should receive is a wp_die() error message, like so:
![image](https://cloud.githubusercontent.com/assets/1034109/5096607/4bf2ae66-6f39-11e4-9cc8-dccfd1442b8c.png)

Instead, I get a file download that fails due to a server error:
![image](https://cloud.githubusercontent.com/assets/1034109/5096619/6576854c-6f39-11e4-8a22-e98e76af1375.png)

Turns out that when the `download` attribute is used, the browser (or at least Google Chrome) doesn't give two f**ks about what the link does, it just downloads the contents of the link any how.

This means that no error messages can _ever_ be displayed during file download.

I'm going to remove the `download` attribute entirely.

PR #2795, opened 2014-11-18
Reverse how we handled #2794 by reimplementing the download attribute and checking the filter

chriscct7:


cklosowski:
@chriscct7 Can you resubmit this against ```release/2.3``` and we can go from there?

chriscct7:
Resubmitted as #3090

PR #2811, opened 2014-11-26
Change all uses of $_REQUEST in the SL Updater Changelog to $_GET.

jjeaton:
http://stackoverflow.com/questions/2142497/whats-wrong-with-using-request

While I haven't run into an issue with the updater particularly, I came across this while fixing issues on a PHP 5.2 site where `$_COOKIE` data was included in `$_REQUEST` for a different plugin.

Since `view_plugin_changelog` is never handled via POST, `$_GET` should be sufficient.

### DIFF
--- a/includes/EDD_SL_Plugin_Updater.php
+++ b/includes/EDD_SL_Plugin_Updater.php
@@ -289,15 +289,15 @@ class EDD_SL_Plugin_Updater {
     public function show_changelog() {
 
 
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@jjeaton if you'd still like to see this addressed. Let me know, otherwise closing as wontfix.

jjeaton:
@pippinsplugins no worries. Not a huge deal.

pippinsplugins:
:+1:

PR #2831, opened 2014-12-02
Rewrite PHP session detection method. No need to store result in DB opti...

dannyvankooten:
...on as it's fast already.

Follow-up of #2767 .

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -212,9 +212,6 @@ class EDD_Session {
 	}
 
 	/**
-	 * Starts a new session if one hasn't started yet.
-	 *
-	 * @return null
 	 * Checks to see if the server supports PHP sessions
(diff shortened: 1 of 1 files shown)


dannyvankooten:
Could we have a new look at this, perhaps @cklosowski could take a look? My main issue with the code is still the same: the `use_php_sessions()` method is overly complex and its logic is faulty.

For example, these two lines right after each other make no sense.

`
$this->set( 'edd_use_php_sessions', 1 );
if( $this->get( 'edd_use_php_sessions' ) ) {
`

This sets an instance property and then reads it right after as a way of testing whether PHP sessions are necessary. This has nothing to do with PHP sessions and will always work. 

I know this somehow works for most right now, but not for the right reasons.

pippinsplugins:
I'm all for rewriting this, but let's get it opened as an issue for discussion then we can get a PR together.

PR #2878, opened 2014-12-16
Validate that the user is already in the customers database.

ericnicolaas:
As it stands, this code throws notices if the passed `$user` does not exist in the customers table. This ensures that a sensible value is always returned.


cklosowski:
I'll merge in and fix the 'empty'

Issue #2880, opened 2014-12-19
Refresh the EDD Admin Menu Items


cklosowski:
Originally we were thinking of 2 menu items, much like WooCommerce does for products and store management. After some discussion below, we're all kind of in the idea that there might be some backwards compatibility issues, so we'll stick with 1 menu, but come up with a way to better organize these items within our one menu.

Orignial proposal:
~As we grow the plugin, we're creating a monolith of an admin menu area. We should break this down into a more intuitive management experience for our users, proposed in #2548 originally:
I'm not opposed to having two top level menu items. What about: <organization redacted since we don't want it>~

cklosowski:
Here's what I've found:
Pages that give ```You do not have sufficient permissions to access this page.```
- [ ] Viewing Payment Details - I'd assume saving will be the same thing.
- [ ] Searching for payments on the Payments List Table
- [ ] Bulk Actions on the Payments List Table
- [ ] Applying a Date Ranges in the Payments List Table
- [ ] Searching Discounts
- [ ] Editing/Adding a Discount
- [ ] Saving Banned Emails in Tools -> General
- [ ] Generating new API Keys for a user in Tools -> API Keys
- [ ] Clicking ```Download System Info``` on Tools -> System Info
- [ ] Clicking ```Import``` on the Tools -> Import/Export
- [ ] Filtering Earnings of Time

Also, choosing a 'Report Type' from the drop down on Reports and clicking ```Show``` is giving ```Cannot load edd-reports.```

Those are the broken ones I found now. Once the reports dropdown is fixed, you'll need to make sure each one of those work if they if they have any filter controls.

ghost1227:
Ok... why can't I find where I need to update for searches??

pippinsplugins:
It's in admin/payments/

pippinsplugins:
2.4 is too close without enough movement on this. Punting to 2.6.

cklosowski:
@easydigitaldownloads/core-team Do we still want to do this?

JJJ:
How about we keep 1 top level menu item (to maintain backwards compatibility) but break our sub-menus up into sections?

We could use horizontal lines, or sub-headings, or maybe something else? I haven’t seen anyone do this yet, but it wouldn’t be hard to do.

JJJ:
For separators, we could just use CSS or JS to cludge them in, or link’less sub-items and CSS to style them appropriately.

[...]

PR #2885, opened 2014-12-21
Split EDD menu into Store and Downloads menu items

ghost1227:
Issue #2880

### DIFF
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -31,12 +31,13 @@ function edd_add_options_link() {
 
 	$edd_payment            = get_post_type_object( 'edd_payment' );
 
-	$edd_payments_page      = add_submenu_page( 'edit.php?post_type=download', $edd_payment->labels->name, $edd_payment->labels->menu_name, 'edit_shop_payments', 'edd-payment-history', 'edd_payment_history_page' );
-	$edd_discounts_page     = add_submenu_page( 'edit.php?post_type=download', __( 'Discount Codes', 'edd' ), __( 'Discount Codes', 'edd' ), 'manage_shop_discounts', 'edd-discounts', 'edd_discounts_page' );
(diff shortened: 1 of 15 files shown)


danieliser:
Just a thought.

You could even go as far as to make a 3rd Top level menu item for extensions to use that only shows if an extension has added a menu item to it.

You would end up with

Store
Downloads
Extra Settings / Store Extensions

This further extrapolates them. If they don't really have a place semantically in Store or Downloads they can easily fit into menu 3 which is basically misc settings. Social Login & Wishlists for example.

ghost1227:
I don't really think that's necessary. Conceptually it's not a horrible idea, but I think it'll be more complicated than it's worth given the few extensions that actually need to provide a dedicated menu item. Plus, the few that DO provide a dedicated menu item are already registering their own.

danieliser:
I see your point, but It's more about the future growth of EDD in my opinion. What may not exist now could easily be there tomorow with the growth rate of EDD extensions. That said for me its more semantics & organization. I hate cluttered menus, and some things just don't feel like they belong under either of those menus.

As for existing extensions, they will all hook the 'downloads' menu anyways and will require modification to move them to 'Store' menu any ways, giving them a 3rd option now means less maintanance in the future.

As I said this is purely based on my opinion, and the fact that EDD is growing rapidly.

pippinsplugins:
It's definitely something to consider.

cklosowski:
@ghost1227 If you could please bring the unit tests up to speed so they are passing, not sure why so many are failing here...have you merged ```release/2.3``` in recently?

Also, can we clean up the spacing on this list:
screen shot 2015-02-11 at 5 39 19 pm

Assignments like this should be aligned with spaces, no tabs at all (based on WP Coding standards)
I know this existed prior to your updates, but let's just clean it up while we're here.

ghost1227:
no problem

ghost1227:
@cklosowski Fixed the indentation issues... Not sure what you mean by things failing though, it's showing 100% passing as far as I can see...

cklosowski:
Yep initial commits were failing for some reason. Will test this.

ghost1227:
Roger that!

pippinsplugins:
I'm going to close this PR for now. We need to decide exactly how we're splitting this up before we do too much with actual changes.

Issue #2872, opened 2014-12-13
Polylang support


grappler:
I am looking to make [Polylang](https://wordpress.org/plugins/polylang/) work with EDD.

One thing I need to make multilingual is the checkout page. I created on in English and one in German.

So that [`edd_get_checkout_uri()`](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/checkout/functions.php#L118) returns the correct link acording to the language I would need to run `$edd_options['purchase_page']` though `pll_get_post()`. The function `pll_get_post()` returns the post id depending on the language of the current page.

So something like this
```php
$edd_options = edd_get_settings();

// Translate post_id for pages in options
$edd_options['purchase_page'] = pll_get_post( $edd_options['purchase_page'] );
```

Is it possible to add a filter which I could hook into?

----------------

Two posts(downloads) are created so to be able to translate the content in two langues. At the moment depending on which language the site being viewed in that version of the download would be added to the basket.

A solution would be to only allow one version of the downloads to be added to the cart. I see that you use [`edd_add_to_cart()`](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/cart/functions.php#L121) to add to the cart. If I were able to run this line of code then I would be able to achieve this.

```php
$download_id = pll_get_post( $download_id, pll_default_language() );
```

Is it possible to add a filter in this place?

I hope I am making sense. I can make a PR to add these two filters. I just wanted to discuss this before make it.

pippinsplugins:
All commits (and more) from #2888 are now in `issue/2872` branch.

pippinsplugins:
Build is failing:  https://travis-ci.org/easydigitaldownloads/Easy-Digital-Downloads/builds/46222620

pippinsplugins:
Tests passing now! https://travis-ci.org/easydigitaldownloads/Easy-Digital-Downloads/builds/46229014

pippinsplugins:
@cklosowski and @sdavis2702 Could you please give this some really thorough testing? Testing it just involves using the store. Purchase products, view reports, export data, etc.

pippinsplugins:
@grappler Could you also see if these changes allow you to do what you need with Polylang?

grappler:
Yes, I will have a look. I have checked out branch `issue/2872`

grappler:
It seems to work fine. Thanks :smile:

[...]

PR #2888, opened 2014-12-22
Convert to edd_get_option

ghost1227:
Helps with #2872 among other things. In hindsight, this probably should have been to release/2.3, but given the sheer amount of time it took... thought I'd let you voice your opinion before redoing that much work. Additionally, this is going to need a LOT of testing... works fine on my first run-through though!


pippinsplugins:
I've manually merged this into `issue/2872` so that we can double check the unit tests before merging into master.

PR #2911, opened 2015-01-01
Include item quantities when calculating cart quantity.

stephenharris:
This might be intended behaviour but `edd_get_cart_quantity()` doesn't take item quantities into account. So if I have a cart containing one product with quantity 2, `edd_get_cart_quantity()` returns 1 rather two. 

This pull request changes that behaviour, summing the quantities of each item in the cart.

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -103,7 +103,17 @@ function edd_get_cart_content_details() {
  * @return int Quantity of items in the cart
  */
 function edd_get_cart_quantity() {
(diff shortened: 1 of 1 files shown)


pippinsplugins:
I need to look back on the history for `edd_get_cart_quantity()`. I remember there being a reason it doesn't include individual item quantities.

I'll tag for 2.3 so we can look it up.

cklosowski:
@pippinsplugins we actually merged something for this already...for issue #2931

Issue #2839, opened 2014-12-08
Add PHP 5.5, 5.6 and HHVM  to the travis-ci matrix


cklosowski:
Anyone opposed to adding PHP 5.5 to the Matrix? I know a couple of our local environments are using this version. Might be nice to get ahead and make sure we're testing for it as hosts are starting to adopt it.

cklosowski:
So I've been getting some good community feedback about optimizing the build matrix from the community after my blog post about this. 2 suggestions stand out, what do you think?

1) Only run Multisite on one PHP version (suggested by @danielbachhuber)
2) Only run code coverage on one matrix build (probably only latest & 5.5?) as suggested by @bradt

pippinsplugins:
Sounds good to me.

cklosowski:
I'm much happier with this build setup, good compatibility checking, ~17 minutes to full build with code coverage run on a single mode and Multisite on a single mode.

Any objections to closing this one out @chriscct7 @pippinsplugins ?

pippinsplugins:
I'm happy!

bradt:
Nice, thanks for sharing!

PR #2922, opened 2015-01-05
Further optimizations for Travis Build Matrix

cklosowski:
Addresses the last comments on #2839 

Reduces to only testing multisite on one config (PHP 5.3 and latest WP)

Also only runs code coverage on 5.5, latest, non-multisite. This will run coverage on all commits, now that matrix is smaller and the builds faster, no harm in that.


cklosowski:
Failing test at this point for this branch is the hard coded year dopdown that's fixed in master. Merging should cause no issues.

PR #2825, opened 2014-12-01
Add error check for invalid email in profile editor

ghost1227:
Fixes issue #2783... I overcomplicated the CRAP out of that!

PR #2924, opened 2015-01-05
Add check for invalid email in profile editor. Props @ghost1227. Fixes #2783

### LINKED ISSUE #2783: Profile editor template allows email address to be deleted
The template used by the `edd_profile_editor` shortcode allows a customer to remove their email address completely. Since WP requires an email address on an account, I think it makes sense to prevent a blank value from being submitted here. In our case, someone was able to erase their email address from their profile, because customers do crazy things, then couldn't use the password reset link after forgetting their password.

mindctrl:
See #2825


pippinsplugins:
Thanks!

PR #2937, opened 2015-01-07
[i18n] - [l10n] > updates v2.2.2 from TX

fxbenard:
93 languages updates from TX with latest pot + French 100%


pippinsplugins:
Amazing. Thank you so much!

PR #2945, opened 2015-01-08
Set post_date_gmt from post_date if a date is supplied.

### LINKED ISSUE #2944: Properly set payment GMT date
When `edd_insert_payment()` is called, the `post_date` and `post_date_gmt` fields are both set to the same value, if a date is supplied in the `$payment_data` array. If no date is supplied, WordPress handles both dates.

If a date is supplied, the GMT date gets set improperly since both `post_date` and `post_date_gmt` get set to the same value. We need to make sure that `post_date_gmt` is properly set when a custom date is supplied.

pippinsplugins:
Resolves #2944


cklosowski:
Looks good.

PR #2956, opened 2015-01-12
Quantities on cart widget are incorrect when adding with quantities > 1

### LINKED ISSUE #2931: Cart Widget item count does not respect quantities
When adding multiples of an item to the cart, using the 'quantities' field, the cart widget will only show 'Items in cart 1'.

Widget shows 3 items in cart:
screen shot 2015-01-06 at 11 29 58 am

Actual Cart:
screen shot 2015-01-06 at 11 30 10 am

cklosowski:
Fixes #2931 and updates the ```edd_get_cart_quantity()``` to return the sum of the  cart quantities instead of a sum of the different products in the cart.


pippinsplugins:
Looks good to me.

PR #2961, opened 2015-01-15
Added new filter edd_reciept_no_files_found_text.

danieliser:
Added new filter edd_reciept_no_files_found_text allowing filtering of the "No downloadable files found." message per $download_id.


danieliser:
strange, coulda swore i copy/pasted that one. Will make a new PR in a minute. Not sure I can modify this one.

sumobi:
If you push to the same branch as before it will add a new commit to the PR

PR #2972, opened 2015-01-18
Add install tests

JeroenSormani:
Hey,

Adding tests for `includes/install.php`

Almost 100% testing the file, just not two lines; http://cl.ly/image/1t3F242Y0L08

I can't think of a way to test this, you guys have an idea?

Cheers,
Jeroen

### DIFF
--- a/tests/tests-install.php
+++ b/tests/tests-install.php
@@ -5,14 +5,208 @@
  * @group edd_activation
  */
 class Tests_Activation extends WP_UnitTestCase {
+
+	/**
+	 * SetUp test class.
+	 *
+	 * @since 2.1.0
+	 */
 	public function setUp() {
 		parent::setUp();
 	}
 
+	/**
+	 * Test if the global settings are set and have settings pages.
+	 *
+	 * @since 2.1.0
+	 */
 	public function test_settings() {
 		global $edd_options;
 		$this->assertArrayHasKey( 'purchase_page', $edd_options );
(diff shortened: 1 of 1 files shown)


chriscct7:
Nice work!

You can  run a $this->assertFileExists( ) to ensure the .htaccess file exists inside the wp-content/uploads/edd/ folder

pippinsplugins:
@JeroenSormani You are quickly moving up the list of people I owe drinks to. Thank you so much for this!

JeroenSormani:
Nice :smile:, I try to do as much as I can.

@chriscct7 I see the assertFileExists is already done here; https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/tests/tests-easy-digital-downloads.php#L42

I don't think that will cover that line in the coverage report, right?

pippinsplugins:
For testing that file, are you just trying to test that it exists?

JeroenSormani:
Not even specific (since its already tested before), I just want those two lines covered for the 100% coverage.

So the challenge is to get inside that `if` statement. (Where I don't see a way for)

pippinsplugins:
Oh I see what you mean. Hmm. That function already exists so the file doesn't need to be loaded.

This might just be a case where we can't get to 100%.

pippinsplugins:
@cklosowski I'm happy with this if you are too.

cklosowski:
Yeah looks solid. I'll go ahead and merge. Thanks @JeroenSormani!

PR #2980, opened 2015-01-20
Add a prefix to session index if on multisite. #2951

### LINKED ISSUE #2951: EDD Session crossing WordPress multi-site blogs?
A recent user of my plugin complained that VAT is being added to a handling fee created by the EDD fee plugin.  To test this I created an action to add a fee:

```
function add_handling_fee() {
	// Add a $10 handling fee without tax
	
	EDD()->fees->add_fee(
		array(
			'amount'      => 10,
			'label'       => 'Handling Fee',
			'id'          => 'handling_fee',
			'no_tax'      => false
		)
	);
}
add_action( 'init', 'add_handling_fee' );
```

It turns out there's nothing wrong with the plugin and showing the user the 'no_tax' option works for the user's scenario.  However, using a fee shows something that may be a problem with EDD.

I use the multi-site version of WordPress and I added the fee using an action implemented in the functions.php file of the theme of one site I use for testing.  The sites in the WPMU installation share no common files except the WordPress files.  However when I add a fee using the action above on one site, the fee appears in the carts on all sites for the same user.

I can see that the fees class uses the EDD session to store fee definitions.  Since the session is PHP wide, my guess is this is the reason the fee is being applied to carts in any site not just the one in which it is defined.  If my assessment is right, it will be great if the fees class can be updated to record and check for the blog number, or some other unique blog identifier, of the site defining the fee.

pippinsplugins:
Fixes #2951

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -40,6 +40,15 @@ class EDD_Session {
 	 */
 	private $use_php_sessions = false;
 
+	/**
+	 * Session index prefix
+	 *
+	 * @var string
+	 * @access private
+	 * @since 2.3
+	 */
+	private $prefix = '';
+
 
 	/**
 	 * Get things started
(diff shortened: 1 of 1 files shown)


cklosowski:
Looks solid and build passes. merging.

PR #2993, opened 2015-01-24
Wrap price option in <span>

JeroenSormani:
Hey,

Did two things in here;
1. Changed echo'ing HTML to plain HTML, this is easier to read and IDE's are able to recognise the syntax
2. Wrapped the edd_price_options in an extra div for more control. 

Reason for 2. is because in themes like Stocky/Marketify long names will fall under the check/select-box like this:
screen shot 2015-01-24 at 10 27 41
which might not be desired. With the existing html elements it is not possible to style that appropriately.

Here's an example of the output, you'll see its exactly the same, but it will give theme authors the tools to change it.
screen_shot_2015-01-24_at_10_20_05
(displaying the current version, and my edited code - side by side)

Let me know your thoughts!
Cheers!

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -964,22 +964,23 @@ jQuery(document).ready(function ($) {
 
 		misc : function() {
 
+			var downloadMethod = $('select[name="edd_settings[download_method]"]');
+			var symlink = downloadMethod.parent().parent().next();
+
 			// Hide Symlink option if Download Method is set to Direct
-			if( $('select[name="edd_settings[download_method]"]:selected').val() != 'direct' ) {
-				$('select[name="edd_settings[download_method]"]').parent().parent().next().hide();
-				$('select[name="edd_settings[download_method]"]').parent().parent().next().find('input').attr('checked', false);
+			if( downloadMethod.val() == 'direct' ) {
+				symlink.hide();
+				symlink.find('input').prop('checked', false);
 			}
 			// Toggle download method option
-			$('select[name="edd_settings[download_method]"]').on('change', function() {
-				var symlink = $(this).parent().parent().next();
+			downloadMethod.on('change', function() {
 				if( $(this).val() == 'direct' ) {
 					symlink.hide();
+					symlink.find('input').prop('checked', false);
 				} else {
 					symlink.show();
-					symlink.find('input').attr('checked', false);
 				}
 			});
-
 		}
 
 	}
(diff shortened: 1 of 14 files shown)


pippinsplugins:
Thanks for the suggestion! Initially I think it looks good but we'll review and test further for 2.3.

cklosowski:
This is one of those that we need to make sure that Extensions developers are aware of.

@ghost1227 and I discussed this at Pressnomics and it will more than likely affect Purchase Limits, I'm sure there are a few more.

chriscct7:
This might conflict with Discounts Pro as well

cklosowski:
@chriscct7 Ok, if you could give it a look over in testing that would be awesome. If not, let me know and I'll squeeze in some testing with it.

chriscct7:
Something merge conflicted this. @JeroenSormani would you mind fixing that?

JeroenSormani:
@chriscct7 ~~Just rebased and pushed.~~ Typed that yesterday, but never pushed comment as it was running tests :stuck_out_tongue: 

Well it looks to be good now :)

pippinsplugins:
We need to test this with the following:
- SquareCode theme
- Marketify theme
- Stocky theme
- Digital Store theme
- Purchase Limits extension
- Any other extension we can think of that interacts with variable prices

chriscct7:
Discounts Pro and maybe variable price descriptions?

pippinsplugins:
Those are good ones.

JeroenSormani:
I've tested Marketify and Stocky when making the edit (screenshots above are from those themes). Am not in possession of others mentioned.

Where would you say it needs to be tested on. I went on visual changes (or the lack of it).

chriscct7:
@cklosowski I've haven't tested it yet. If you get time go for it. You'll need to flip the display settings on for Discounts Pro and use either a flat rate or percentage discount

pippinsplugins:
@JeroenSormani would you be able to bring this up to date so it merges cleanly?

JeroenSormani:
Yes I'll look into this hopefully in the evening, otherwise in the weekend.

On Thu, Jun 18, 2015 at 3:49 PM, Pippin Williamson <notifications@github.com
> wrote:

> @JeroenSormani <https://github.com/JeroenSormani> would you be able to
> bring this up to date so it merges cleanly?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/2993#issuecomment-113162569>
> .
>



-- 
Jeroen Sormani
http://www.jeroensormani.com
en
http://www.nederlandersfietsen.nl

JeroenSormani:
Ugh, rebasing gone wrong. I'm gonna make a new branch/PR thats easier.

chriscct7:
See #3520

PR #2995, opened 2015-01-24
Theme compat tests

JeroenSormani:
Hey,

Adding some tests for the theme-compatibility.php file.

Cheers!

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -5,7 +5,7 @@
  * Description: Serve Digital Downloads Through WordPress
  * Author: Pippin Williamson
  * Author URI: http://pippinsplugins.com
- * Version: 2.2.3
+ * Version: 2.2.4
  * Text Domain: edd
  * Domain Path: languages
  *
@@ -25,7 +25,7 @@
  * @package EDD
  * @category Core
  * @author Pippin Williamson
- * @version 2.2.3
+ * @version 2.2.4
  */
 
 // Exit if accessed directly
@@ -187,7 +187,7 @@ final class Easy_Digital_Downloads {
 
 		// Plugin version
 		if ( ! defined( 'EDD_VERSION' ) ) {
-			define( 'EDD_VERSION', '2.2.3' );
+			define( 'EDD_VERSION', '2.2.4' );
 		}
 
 		// Plugin Folder Path
--- a/readme.txt
+++ b/readme.txt
@@ -7,7 +7,7 @@ Tags: download, downloads, e-store, eshop, digital downloads, e-downloads, ecomm
 Requires at least: 3.9.2
 Tested up to: 4.1
 
-Stable Tag: 2.2.3
+Stable Tag: 2.2.4
 
 License: GNU Version 2 or Any Later Version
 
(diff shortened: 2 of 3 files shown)


cklosowski:
Looks like we're getting a build failure on this one:
1) Tests_Theme_Compatibility::test_edd_responsive_download_post_class_post
Failed asserting that an array contains 'post'.

The failing line is:
```
$this->assertContains( 'post', get_post_class() );
```

Can you take a look and see what's going on with that? Once this passes we can merge.

JeroenSormani:
Hmm, currently not getting in on my local system.

Will look into it more tomorrow/later :+1:

cklosowski:
Np. I'll pull it down too and take a look. Pressnomics ends today so I'll have a few to look.

JeroenSormani:
@cklosowski Running some tests here; https://travis-ci.org/JeroenSormani/Easy-Digital-Downloads/builds

You see that test Nr. 2 builds successfully and has some debug info in the test (it contains the 'post' class)
After rebasing/syncing you will see that Nr. 11 doesn't have the 'post class' anymore. Do you have any clues on why? Looked through the EDD code but found nothing, the test code should be fine IMO. The  classes should contain 'post'.

Easiest fix would be to remove the 'post' assert as it is not required, but I still wonder whats up with it.

cklosowski:
I think I found it...when it gets a class it actually gets ```post-#``` where ``#``` is the post ID:

```
array(6) {
  [0] =>
  string(8) "post-490"
  [1] =>
  string(9) "type-post"
  [2] =>
  string(14) "status-publish"
  [3] =>
  string(15) "format-standard"
  [4] =>
  string(6) "hentry"
  [5] =>
  string(22) "category-uncategorized"
}
```

Can you try making line 36 look like this:
```
$this->assertContains( 'post-' . $post_id, get_post_class() );
```

That should be a more consistent check as I know it get's that...not always ```post```.

This is pulled from one of my sites, with no EDD Being used.
```
class="post-1948 post type-post status-publish format-standard has-post-thumbnail hentry category-geek tag-easy-digital-downloads tag-php tag-software-development tag-travis-ci tag-unit-tests"
```

I'd be curious, what do you get if you run the ```get_current_theme()``` function during that test? I'm getting:
```string(14) "Twenty Fifteen"```

I do know that we should always get ```post-#```

**Comment by JeroenSormani** [comment]
I've changed the test to what you've mentioned to fix the test. (But still doesn't explain why it happens ;-)

Should the theme matter? `post_class` isn't tied to a theme if i'm correct.

Anyway, the checks are still running now, but looking good so far. Let me know if you guys prefer a new clean PR, as this one got some rebase changes.

Hope you enjoyed PressNomics :smile:

**Comment by cklosowski** [comment]
@JeroenSormani My guess is it has something to do with the order of tests...or something in a race condition...not 100% positive...but ```post``` would only be added if we're not in admin. Here's the line that would add ```post``` to the unit test:

```
	$classes[] = 'post-' . $post->ID;
	if ( ! is_admin() )
		$classes[] = $post->post_type;
```

Since one is a conditioned statement, I'd say let's go with the class that gets returned no matter what, just to be safe.

cklosowski:
If you could submit a clean PR (since we pushed a new version between you starting and now) that'd be awesome. Just go ahead and close this one when you are all set and link to the new PR.

Thanks, these tests you are writing up are super helpful. Appreciate that.

JeroenSormani:
Opened #3000 in favor of this one.

PR #2996, opened 2015-01-24
Query filter tests

JeroenSormani:
Hey :smile_cat: 

Unit tests for the query-filter.php file, covers it 100% :smile: 

Found this (nice?) construction to test the wp_die function https://github.com/JeroenSormani/Easy-Digital-Downloads/compare/easydigitaldownloads:master...JeroenSormani:query-filter-tests?expand=1#diff-64cd20412ba2e7f38a137e8500112304R233

Let me know what you guys think about that...

Cheers!
Jeroen

### DIFF
--- a/tests/tests-query-filters.php
+++ b/tests/tests-query-filters.php
@@ -58,7 +58,9 @@ class Tests_Query_Filters extends WP_UnitTestCase {
 
 		$this->assertNull( edd_block_attachments() );
 
+		// Reset to origin
 		wp_delete_attachment( $attach_id, true );
+		$this->go_to( '' );
 
 	}
 
@@ -98,8 +100,10 @@ class Tests_Query_Filters extends WP_UnitTestCase {
 
 		$this->assertNull( edd_block_attachments() );
 
+		// Reset to origin
 		wp_delete_post( $parent_post_id, true );
 		wp_delete_attachment( $attach_id, true );
+		$this->go_to( '' );
 
 	}
 
@@ -157,6 +161,7 @@ class Tests_Query_Filters extends WP_UnitTestCase {
 		// Reset to origin
 		wp_delete_post( $parent_post_id, true );
 		wp_delete_attachment( $attach_id, true );
+		$this->go_to( '' );
 
 	}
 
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@JeroenSormani This all looks great! The only issue I see with it is some formatting (ex: https://cloudup.com/cfbhf98e8Vt). Could you update it to correct the formatting? All variables should be aligned with spaces, not tabs.

JeroenSormani:
Nooooooooo :scream: *drama off*

Can I lobby to go over to tabs instead of spaces? 
(or let other people do it for me as this has been discussed many times before; http://goo.gl/zP1ygk)

I will even volunteer to change every single file in EDD ;-)


Anyways, I have updates the file (for now :smiling_imp:)
//cc @pippinsplugins @cklosowski @chriscct7

pippinsplugins:
Tabs are for indention, spaces are for alignment. Not changing, sorry :)

cklosowski:
:) What @pippinsplugins said:
https://make.wordpress.org/core/handbook/coding-standards/php/#indentation

```
Exception: if you have a block of code that would be more readable if things are aligned, use spaces:
```

JeroenSormani:
:crying_cat_face: Ok.

So how are you guys putting in that indentation? I just converter that document and already got sick of all those spaces.. :stuck_out_tongue_winking_eye: 
(tabs are faster to put in and to go through your code, instead of going left 16 times, you can go left 4 times if you use tabs)

pippinsplugins:
Tabs don't work for alignment because editors can change the tab size.
Spaces are the only way to make it consistent. When aligning vars, we do it
manually. We are meticulous; takes extra time but is better in the long run.

JeroenSormani:
Well, its your choice.

Its can be a endless discussion and I think for both sides can be made a strong case.
So it'll be spaces.

pippinsplugins:
Do you want to update the formatting? If not, we can once it's merged.

JeroenSormani:
@pippinsplugins already did yesterday ;-)

8992067

pippinsplugins:
Ah! I missed that, thanks!

Issue #2882, opened 2014-12-19
Do not show bundled products in the products drop down (if on a bundle)


pippinsplugins:
Bundles should not be allowed to be assigned to another bundle.

ghost1227:
Still struggling with this one @pippinsplugins @cklosowski Love to be able to close it though... either of you mind giving me a hand?

pippinsplugins:
Sure. Tell me what you have so far and what you've tried, then I can provide feedback.

ghost1227:
I don't really have anything beyond what's already been merged, because I simply don't understand the ajax necessary. My JS skills are limited to the type of stuff I write for my extensions, and I have a VERY hard time understanding most of the EDD core JS, particularly when it starts involving AJAX.

pippinsplugins:
Ok let's start simple.

Do you see and understand the JS that currently triggers the search?

ghost1227:
I know where it IS, though I may not have been able to find it easily if you hadn't already pointed me to it. I understand parts of it in theory, but not necessarily in practice. I understand that when you start typing, it triggers `edd_download_search` and passes it the value of whatever is in the text field, and I understand most of the success function just from my own work with Select2 (it's not terribly dissimilar). Totally lost on how any of that function is related to a GET query (probably just my lack of understanding regarding AJAX generally).

I DO understand much of the `edd_download_search` function, though the use of the `wp_ajax` actions is beyond me still. Speaking of that function... why are we using direct wpdb calls? And the PHPDOC is wrong for that function too... it says states, not downloads.

pippinsplugins:
Great, you're most of the way there.

Two exclude the current post ID from it, all you need to do is pass another parameter here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/assets/js/admin-scripts.js#L1045

`s: val` is the current search term that's being entered. The `s` becomes `$_POST['s']` in the PHP. Simply add another key/value that is set to the ID of the current post, which is available in the EDD localized vars.

For example, if you added `current_id: edd_vars.post_id`, you could access that with `$_POST['current_id']` in the PHP.

Does that make sense?

ghost1227:
A quick glance at the `edd_download_search` function looks like `s` becomes `$_GET['s']`, not `$_POST['s']`, right? Also, that all makes sense, but how do you exclude that? Would it just be:

[...]

PR #3008, opened 2015-01-26
Hide bundles from products dropdown when editing a bundle

ghost1227:
Attempting to finish up #2882. Thoughts on what I'm missing would be appreciated.


pippinsplugins:
Mobile so can't test.

Does it work for you with this? Looks good at first glance.

ghost1227:
It does work, thanks to some assistance from @cklosowski

pippinsplugins:
Note: please include meaningful PR titles, even if it is just Issue/#

pippinsplugins:
Since we're now including semi-user-defined data (post_id) in the WPDB query, the queries need to be properly sanitized and passed through `$wpdb->prepare`.

ghost1227:
Does sanitization require anything other than running through `$wpdb->prepare`?

pippinsplugins:
You should run the post_id through `array_map( 'absint', $ids )`

ghost1227:
How's that?

pippinsplugins:
Tip: read this http://codex.wordpress.org/Data_Validation

ghost1227:
Other than inverting the wpdb functions (I blame a lack of coffee), what am I missing?

pippinsplugins:
You need to sanitize the post IDs that are passed into the query. If you
don't know how to do that, please read the article I linked. If you still
don't, read
http://codex.wordpress.org/Class_Reference/wpdb#Protect_Queries_Against_SQL_Injection_Attacks

It's really, really simple.

ghost1227:
Maybe I'm completely missing something. Other than the ABSINT mapping, what else can you do to sanitize IDs? I've read both and I'm not seeing anything else...

pippinsplugins:
Two things:

1. Sanitize by mapping all array values to absint
2. Pass the query through $wpdb->prepare()

ghost1227:
I'm already doing that though... https://github.com/Section214/Easy-Digital-Downloads/commit/71526096e6c4640496f8d2c495aa2c6e13977301#diff-3c3876298821010282ba08aa4229d521R449

ghost1227:
That completely makes sense, dunno how I missed that. I think that commit fixes it, and it works as expected, but let me know if I'm wrong or there's a better way to do it.

pippinsplugins:
There you go! Haven't tested yet but will soon.

cklosowski:
Ok i only see one more issue (sorry for not catching it earlier), and that's if the array of ```$excludes``` is empty ```$exclude``` ends up being an empty string, which breaks the mysql query because you cannot do ```NOT IN ()```

```mysql> SELECT ID,post_title FROM wp_2_posts WHERE `post_type` = 'download' AND `post_title` LIKE '%bundle%' and ID NOT IN () LIMIT 50;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') LIMIT 50' at line 1```

We shouldn't be including the ```NOT IN``` clause if the ```$exclude``` is empty.

**Comment by cklosowski** [comment]
Looks good @ghost1227 thanks. Going to merge this bundles thing in.

File: includes/ajax-functions.php [hunk 1]
```diff
@@ -437,14 +437,42 @@ function edd_ajax_get_states_field() {
 function edd_ajax_download_search() {
 	global $wpdb;
 
-	$search  = esc_sql( sanitize_text_field( $_GET['s'] ) );
+	$search   = esc_sql( sanitize_text_field( $_GET['s'] ) );
+	$excludes = ( isset( $_GET['post'] ) ? (array) $_GET['post'] : array() );
```
cklosowski: @ghost1227, @pippinsplugins is correct, `post` here needs to change to `current_id`, that's what's being sent in the JavaScript above: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3008/files#diff-4e07975920906985a5fbbe3afd720629R1046

This won't ever occur on bundles, as they are excluded in the dropdown via the html element, but in the event that it's on a download page that's not a bundle and we do have a product search...this would prevent it.

PR #3026, opened 2015-01-28
Remove notice on removing last file

### LINKED ISSUE #3019: Remove JS "you must have at least one file" alert
When you already have at least 1 file added to a download and you try and remove it you get a JS alert with "you must have at least one file". 

Considering you can add a download without adding file downloads I believe you should be able to remove the last file and save the download without receiving the JS alert. Currently the solution is to wipe out the file field and save the download.

ghost1227:
Fixes #3019


pippinsplugins:
We still need to prevent the last row from being deleted. If the admin tries to delete the last option, let's clear out the fields.

ghost1227:
Gotcha.

ghost1227:
Done

PR #3027, opened 2015-01-29
Add Content-Length header for #2983

### LINKED ISSUE #2983: File size not sent to the browser during file download
It seems that the size of the file is not being properly sent to the browser when a file is being downloaded, as reported [here](https://easydigitaldownloads.com/support/topic/server-isnt-reporting-file-size-for-downloads/).

We should see if we can fix this.

pippinsplugins:
Fixes #2983

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -168,6 +168,9 @@ function edd_process_download() {
 
 				}
 
+				// Set the file size header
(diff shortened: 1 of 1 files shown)


cklosowski:
Looks good to me:
screen shot 2015-02-06 at 10 15 02 am

PR #3036, opened 2015-01-30
Issue/2872

### LINKED ISSUE #2872: Polylang support

pippinsplugins:
Fixes #2872


pippinsplugins:
Resubmitting to `release/2.3`.

Issue #3031, opened 2015-01-29
Update edd_get_users_purchases() to pull all payments from payment_ids column


pippinsplugins:
The `edd_get_users_purchases()` function pulls all payments that match a user ID or email.

Sometimes a customer purchases while logged in, sometimes they purchase while logged out. With the current iteration, `edd_get_users_purchases()` will only pull payments made while logged in (if searching on user ID).

To ensure that all payment records get queried for a customer, we should pull both the payment IDs from `payment_ids` column of the customers table and also search by the email and then merge the results.

pippinsplugins:
I'm going to take this one.

pippinsplugins:
I think the changes in #3038 fixes this but we need to be careful with it. When we first pulled payment IDs from the customer table, we had issues with some payment records not having the proper customer ID attached to them. It was a big problem before because there was no good way to attach a payment to a customer ID. With the introduction of the new customer UI in #2548 , this is much more negligible.

pippinsplugins:
Unit tests are failing for this so proceed with caution.

pippinsplugins:
https://travis-ci.org/easydigitaldownloads/Easy-Digital-Downloads/jobs/48956983

pippinsplugins:
Related #3209

cklosowski:
Closing this for now, if we need to we can re-open for discussion.

ineagu:
@cklosowski @pippinsplugins I have some issues with this thing, here is what they are : 

1. Here are the customer details : http://www.awesomescreenshot.com/image/258394/2d3a1d9e6c097733c2109233fbd09365 ( 4 payments , 1st is complete )

2.After that in code I do : $purchases = edd_get_users_purchases( get_current_user_id(), 30, true, 'any' );

However what this is listing is only the 1st payment ( showed as pending ) - 

[...]

PR #3038, opened 2015-01-30
Pull payment IDs from customer table for edd_get_users_purchases()

pippinsplugins:
#3031


cklosowski:
If we hold off on this, we can use ```new EDD_Customer( $_id_or_email )``` to pull these.

pippinsplugins:
Smart :+1:

pippinsplugins:
This will fail if you pass a user ID. `edd_get_users_purchases()` is designed to take a user ID, not a customer ID.

cklosowski:
Ah, yes @pippinsplugins I'll change it to the get_by method in `EDD_DB_Customers`. We should come up with a way to instantiate `EDD_Customer` via the User ID...but there's just not a good way.

We could add a parameter to the `__construct` of `EDD_Customer` that flags the input as a `user_id` which can default to false. So if you say:
`EDD_Customer( $user_id, true )` it'll do the lookup by the `user_id`column. Otherwise by `email` or `customer_id`

pippinsplugins:
I like the idea of a `$by_user_id = false` parameter.

PR #3039, opened 2015-01-30
Allow administrators to run upgrades without install_plugins

johnjamesjacoby:
Allow administrators to run upgrades when install_plugins is explicitly denied. Ones ability to upgrade databases should not be dependent on their ability to write to the plugins directory.


pippinsplugins:
I'd vote that we use manage_shop_settings for this instead of
manage_options.

@cklosowski?

johnjamesjacoby:
That seems fine to me also, looking at the map.

pippinsplugins:
@johnjamesjacoby If you want to update with `manage_shop_settings`, I'll go ahead and merge. I can do it too if you'd like.

chriscct7:
manage_options is usually those who can manage the whole site. In a case where you have a "web guy" in a company, he might want to lock it down so only he can update it. `manage_shop_settings` says they can manage the store, not necessarily the whole website. I'd vote for manage_options

cklosowski:
I'm with @pippinsplugins on this one ```manage_shop_settings``` seems like the best option here. If they really want to restrict this further, they can do it with a capability plugin.

pippinsplugins:
Resolved with https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/adf0f9d566c3437afeecb36773449e3b9aa5c270

johnjamesjacoby:
Thanks y'all. Looks good here.

PR #3041, opened 2015-01-31
Unit helpers

JeroenSormani:
Hey,

Added some unit test helpers for;
- Downloads
- Discounts
- Payments

Note; I've used mostly the same data as in the original setUp() methods were used, but not always. Like here, I've used different data because it actually adds up (and thus makes more sense) and is easy to calculate yourself. https://github.com/JeroenSormani/Easy-Digital-Downloads/compare/unit-helpers?expand=1#diff-dc9f94e691a2043185baa9b59764b916L153

Let me know what you guys think :-)

Cheers,
Jeroen


pippinsplugins:
I really like this. @cklosowski ?

tests/tests-downloads.php
```diff
@@ -150,8 +108,8 @@ public function test_edd_download() {
 		$this->assertNotEmpty( $download3->prices );
 		$this->assertEquals( $prices, $download3->prices );
 		$this->assertEquals( $prices, $download3->get_prices() );
-		$this->assertEquals( 59, $download3->sales );
-		$this->assertEquals( 129.43, $download3->earnings );
```
cklosowski: Actually using a decimal value here is testing that our rounding is working correctly. It needs to remain that one of these tests uses a decimal value so we continue to test that we aren't breaking rounding.

JeroenSormani: I increased the earnings for this download specific for testing (and also a value below).

JeroenSormani:
Hey guys,

Whats the current status, are you expecting anything from me?

Have a nice weekend!
Jeroen

pippinsplugins:
We just need to test it again :)

cklosowski:
This is perfect thanks @JeroenSormani.

Super helpful in reducing a TON of duplicated code. Merging.

Issue #3018, opened 2015-01-28
Introduce EDD_Customer class


cklosowski:
In working on #2548, I'm finding that we end up using the ```EDD_DB_Customer``` class to retrieve a customer, which is intended to be an abstraction layer for the custom table, not a true ```Customer``` Class.

Initially I'm looking to have a class that allows directly accessing the customer properties, as well as updating key attributes of the customer.

This is a blocker now for #2548.

pippinsplugins:
This should mimic EDD_Download and EDD_Payment (WIP) as much as possible.

cklosowski:
@pippinsplugins That's the plan.

pippinsplugins:
This is looking really good.

First thing I notice that I think I'd like changed is how the `global $edd_customers_db` is setup in `EDD_Customer`. Instead of setting it up inside each method, how about just setting it up as a private or protected property and then setting it up once in the constructor.

spencerfinnell:
I thought the same thing as @pippinsplugins 

Also should passing no ID set up a new customer similar to how the downloads class works? If no customer is found and `$field` is an email one could be generated.

pippinsplugins:
We discussed creating new customers like that and ultimately decided it was
too risky.

cklosowski:
@spencerfinnell Yeah, we determined that we'd rather have the creation be explicit, not implied.

@pippinsplugins I had it that way initially, trying to remember why I changed it. Let me do it again and see if it comes back.

cklosowski:
Huh, seems the way I'm setting up the customer cleared up the initial reason for it being a ```global```. Yay for refactoring :)

Committing up with ```$this->db``` and no global, as a protected.

pippinsplugins:
Woot!

pippinsplugins:
How do you guys feel about the `EDD()->customers` object? Keep it?

cklosowski:
@pippinsplugins I'm good with keeping it. It makes sense to have it there for some of the use cases, like dealing with the database directly. This is an ```EDD_Customer``` class, not ```Customers```.

My vote, it stays.

pippinsplugins:
The reason I brought it up is because I'm thinking about the reason for having a global DB class to access customers. Since we have it, and we use the `EDD()->{object}` method elsewhere, I think we should be consistent and use it through `EDD_Customer` as well.

cklosowski:
[...]

Issue #2548, opened 2014-08-31
New customer management UI


pippinsplugins:
With the introduction of the new customer API in #2256, we need to now look at dramatically improving our customer management UI.

A few things I can think of that we need:

1. a better list table for customers that is _not_ buried in the Reports page
2. a new Customer Details screen
3. a visible link between the `View Order Details`  screen and the `Customer Details` screen
4. the ability to more easily change the customer a purchase belongs to
5. reporting on customers

pippinsplugins:
Let's flesh out the main customer management UI before we touch the `View Order Details`.

cklosowski:
How is this looking for a 'view/edit' format. Currently it's just a view, no edit just yet. Thoughts?

![image](https://cloud.githubusercontent.com/assets/1390124/5765889/291a1f04-9ccb-11e4-9298-61dc3684cc64.png)

pippinsplugins:
This is looking great!

chriscct7:
Do we want to have per-user reporting, like graphs + exports?

pippinsplugins:
Not for the first iteration.

cklosowski:
What elements of a customer should be allowed to be edited? And also, if there is a user tied to a customer, should editing the user edit the customer as well (name, email)?

cklosowski:
Here's the concept I have for the customer notes:
![image](https://cloud.githubusercontent.com/assets/1390124/5770623/2dec6714-9cfb-11e4-8494-1db11487cbf8.png)

cklosowski:
![image](https://cloud.githubusercontent.com/assets/1390124/5770656/a4097d60-9cfb-11e4-91a5-7db11eedd08b.png)

pippinsplugins:
To be editable:

Email
Name
Known ip addresses (maybe)
Connected user ID
Stored billing address

I'd vote that changing name / email also changes it on the connected user
account.

cklosowski:
Sounds good. I'll start the editing tomorrow. 

What do you think of a custom post type for the customer notes.

pippinsplugins:
No custom post type. If they go in a default WP table, they should be in
comments or user meta.

cklosowski:
Ok. I'm can live with those options. Will probably use comments.

cklosowski:
Ok Just pushed some changes that I think should be reviewed. Note that this DOES NOT include EDITING a customer just yet. Want to get the interface down first:
[...]

PR #3048, opened 2015-02-02
Introducing EDD_Customer class

cklosowski:
This is the initial scope of an EDD_Customer class that should help us interact with a customer much easier.

#3018


cklosowski:
Closing this PR as this is now merged into #2548

Issue #3045, opened 2015-02-01
Add Ability to Resend Receipt to Shop Accountant Role


chriscct7:
Per https://easydigitaldownloads.com/support/topic/shop-accountant-cannot-resend-receipt/ I'm wondering if we shouldn't allow Shop Accountants to resent receipts, something they can't do right now. I could see this being something an accountant could want to do, for like end of year tax purposes or something like that. Thoughts?

chriscct7:
So this is the cap check: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/c09023ab8e36f29363cd4aab3a8f712f5b3f1142/includes/emails/actions.php#L47

I disagree slighly with the error. Someone shouldn't need to have access to edit the payments in order to resend the receipt, because resending the reciept can be done regardless of whether the payment is edited or not.

chriscct7:
I almost want to say that could be it's own cap. If you wanted to have like a support user role that can only login and resend receipts for example.

chriscct7:
For those who want to bypass the check for now: https://gist.github.com/chriscct7/d051cfcd4b9b204247ab

pippinsplugins:
I think shop accountants should have the option. I don't see why this is
limited to ones that can edit payments. On that note, shop accounts should
probably be able to edit payments too.

pippinsplugins:
Actually, shop accountants already have this:
```php
$wp_roles->add_cap( 'shop_accountant', 'edit_shop_payments' );
```
My guess is the user that reported this just needs to re-activate EDD.

pippinsplugins:
Nevermind, bug is happening to me. So it's a bug, not a change.

PR #3049, opened 2015-02-03
Correct capability check for editing payments. #3045

pippinsplugins:
#3045


cklosowski:
So, we just misspelled the capability? I'm good with this. Shall I merge for further testing in ```release/2.3```

pippinsplugins:
Yep, good to go. Having it in master is fine. If we end up not pushing out 2.2.8, we'll just update all milestoned issues for 2.3.

Issue #3061, opened 2015-02-07
Add WP CLI command for creating customers


pippinsplugins:
We should introduce a WP CLI command for creating / updating customer records. It should support setting all DB columns and should also support creating X arbitrary customer accounts, allowing us to more easily test things with sites that have 10k, 20k, or even 100k customer records.

PR #3062, opened 2015-02-07
Issue/3061 - Allow customers to be created via WP CLI

pippinsplugins:
#3061


pippinsplugins:
Everything works great for me on this. @cklosowski could you test so we can merge?

Issue #3057, opened 2015-02-05
Discounts that don't meet product requirements get submitted when clicking Purchase


mindctrl:
Steps to reproduce:

1. Create discount code and set Download Requirements to PRODUCT ABC
2. Add PRODUCT XYZ to cart and go to checkout
3. Try to apply discount code created in step 1 and get error "The product requirements for this discount are not met." 
4. Fill out remaining purchase details and click Purchase.

At this point, the checkout page reloads. You'll notice the code is attached to the cart, with the error messages at the bottom stating the discount code doesn't meet requirements.

If the customer doesn't go back and clear the coupon code field, it gets submitted, which probably makes some sense in case someone forgets to apply a coupon. But, it would be nice if the coupon code fails during apply, we blanked the field and/or did another check.

[Reported here](https://easydigitaldownloads.com/support/topic/user-cannot-purchase-after-discount-code-fails/)

mindctrl:
It might be worth noting the customer details like email, name, etc. get cleared when this happens. If that needs to be in another issue, I can do that.

pippinsplugins:
Sounds like the discount code is getting set in the cart session even when it's invalid. That's definitely a bug.

cklosowski:
@pippinsplugins This _might_ have come in when we were fixing #2916, might be a good place to check.

pippinsplugins:
Pretty sure you're right.

cklosowski:
@mindctrl You are right, it's happening because if someone forgets to hit apply, we still try and attach it. I think we'll have to silently fail this though and just never attach it.

cklosowski:
@mindctrl I'm not able replicate it, If I follow the steps you listed:

| If the customer doesn't go back and clear the coupon code field, it gets submitted, which probably makes some sense in case someone forgets to apply a coupon.

It clears for me.

cklosowski:
Ok @mindctrl Can you check branch ```issue/3057``` and see if that fixes it for you?

mindctrl:
@cklosowski this fix seems to be good. Tested various scenarios and all worked.

cklosowski:
Awesome thanks I'll get a PR together.

PR #3069, opened 2015-02-09
Only apply a posted discount if it's valid

cklosowski:
Only add the posted discount if it's valid, as the setting of the discount does not verify this. #3057


chriscct7:
@pippinsplugins  This is actually causing the issue with EDD Same Page Checkout that you were looking at in Pressnomics

pippinsplugins:
@chriscct7 That's good to know! Hopefully there are not adverse consequences to it.

PR #3070, opened 2015-02-09
Add the ability to add multiple price_ids to the cart at a time

chriscct7:
**Update: Per the comment below, this actually fixes a regression**

Right now if you wanted to add multiple price_ids to the cart in a single add_to_cart call you can't.

For example, let's assume I have a variable priced product, where 2 of 30 price options are often purchased together. I want to make a call to add_to_cart that adds these both to the cart at the same time.

Right now, you can't do something like this:
```php
$options = array(
	'price_id' => array( 0, 1 )
);
edd_add_to_cart( $post_id, $options );
```

As opposed to calling edd_add_to_cart individually for every option I want to add, I'd really like EDD to do that work for me.

With this pull request, you can now do the above.

In addition, you can also pass in an array for the quantities and price_id and EDD will match them up and add them to the cart:

For example:
```php
$options = array(
	'price_id' => array( 0, 1 ),
	'quantity' => array( 2, 3 )
);
edd_add_to_cart( $post_id, $options );
```

Adds variable option 1 with a quantity of 2, and option 1 with a quantity of 3

Checks have been put in place to ignore extra quantity array items (the array of quantities is larger than the array of price ids). Further, if the quantity array is smaller than the price_id, the remaining items are mapped to the last quantity.

So lets say I want to add 3 options with a quantity of 2, and 1 option with a quantity of 4.
```php
$options = array(
	'price_id' => array( 4, 1, 2, 3 ),
	'quantity' => array( 4, 2 )
);
edd_add_to_cart( $post_id, $options );
```

In this situation, price_id 4 will have a quantity of 4, and price_ids 1, 2, and 3 will have a quantity of 2.

Unit tests have been added for over and undersized quantities, mapped quantities to price_ids (arrays are same size) and the lack of any declared quantity. If no quantity index is passed, it is assumed the quantity requested is 1, which is the EDD default for add_to_cart with 1 price_id

Further, tests have been added for the new helper function implemented that resizes the array as a helper function to PHP's `array_combine` function, so that you don't have to manually unset array items to match the sizes so you can `array_combine` them.

This was inspired by https://twitter.com/nikolovtmw/status/564837836411920384 who was trying to do https://gist.github.com/nikolov-tmw/700e2599147cb836c710


nikolov-tmw:
Hi @chriscct7, thanks for the hard work! I don't want to be picky, but I envisioned this as a single $item in the cart, just having multiple options attached to it. Let me give you some background into why I'm trying to do that. 

I'm working on a sports site, that allows you to purchase(amongst other things like sports equipment) tickets for events(as a player) and sports courses. For those events, they need extra information about the player. 
Initially they were using the EDD Gravity Forms plugin, but the problem with that is that you're directly checked out, instead of adding the product to your cart so you can get other things if you want(like sports equipment). 
The idea is to keep using Gravity Forms for collecting the required data, but also keep the shopping cart experience. 

The screenshots I shared with you earlier ( http://screens.paiyakdev.com/Screenshot-at-2015-02-09_12:04:57.png ), shows the three courses for their spring session - they want to allow people to sign up for all courses at the same time.

The problem with having each variation as a separate cart item would be that the form submission will now need to be linked to three separate cart items, instead of only one(they need to be able to edit their information during the checkout page as well) - we'd have to keep track of whether any of those items is still in the basket, etc. 

I realise that this is a pretty much an edge-case, but I would like you to consider the possibility(is there one?) of adding multiple price_id options as a single item, instead of multiple items(because, well - they should be all the same product to some extent, right).

I'd be definitely interested in contributing to this(although sadly my schedule will allow me to do so beginning from March) and won't mind if it waits a little longer before it's implemented(since it seems like no one else has complained about it, right? :smile: ).

Let me know what you think and once again thank you for your time!
Nikola

chriscct7:
Those would probably be better off added using the fees API instead of as a product variation, which we should look at beefing up reporting for. Fees can be directly attached to a product.

nikolov-tmw:
Oh, I haven't seen the fees API - thanks for pointing that out. I'll definitely look into it, rather than the price options. 

Ok, no worries - backwards compat is very important, so I definitely wouldn't want to break that.

chriscct7:
Not a problem! 

This PR is similar to your Twitter comment but like you pointed out not exactly, because it works on solving a problem.

chriscct7:
Actually this is a regression. EDD used to allow this (see https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/76d7232f929e3362db039071d09b34b91d6009b9/includes/cart/functions.php#L139 )

chriscct7:
@cklosowski inadvertently removed multiple price_id support in https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/2e4c9d1370f6a44d97e06c471417524716f9fe42 it seems

pippinsplugins:
I do like the idea of supporting multiple price IDs per add to cart action.

@nikolov-tmw We won't be able to support multiple price IDs per item, sorry.

There is lot going on here, so this will need really extensive testing.

@chriscct7 There is some code cleanup that needs done before this can be merged (obviously not until it's super thoroughly tested). Formatting, spacing, etc.

chriscct7:
@pippinsplugins fair enough. If you can tell me what needs to be fixed  can fix it in terms of formatting. I'd think we should test the heck out of this first then work out the formatting, as the code might have to change if we find a bug with it

pippinsplugins:
@chriscct7 You just have a not of non-standard formatting. For example, brackets must always be on the same line as the conditional.

nikolov-tmw:
@pippinsplugins and @chriscct7 
Sorry for dropping-in, but I'd recommend checking out the WordPress Coding Standards Code Sniff: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
You'll need to set-up PHP Code Sniff in order to use it, but it's great because it highlights any errors right away.
I've set it up with CI on my WP projects so that any violations result in failed builds.

pippinsplugins:
:+1:

chriscct7:
The formatting issues can and will be dealt with. I already have Code Sniff on my computer, this PR was written in notepad on a locked down university computer. I'll try to get the formatting fixed this weekend but again we should start testing this now.

chriscct7:
To view the actual non whitespace changes use https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3070/files?w=1

chriscct7:
This is formatted, done, and with extensive unit tests. Just needed to be tested since this is a mission critical function

pippinsplugins:
Note: this is submitted against `master` so even once it is ready to be merged , if cannot be merged until we're ready to send it out.

chriscct7:
Its deliberately submitted against master,so we can send it out as soon as possible, so we aren't waiting for a feature release. This pull request fixes a functionality regression that so far at least 3 developers have run into so far

pippinsplugins:
I know, that was more of a precautionary note to whoever ends up doing the
final merge.

pippinsplugins:
We can't do this: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/feature/add-to-cart-multiple-price-ids/includes/cart/functions.php#L190

That's going to prevent the proceeding do_action() hooks from firing.

chriscct7:
Which do_actions. With that pr all do_actions that were previously fired are still fired, nothing changed there

pippinsplugins:
screen shot 2015-02-15 at 9 25 20 pm

chriscct7:
Yeah, but when you add 2 price_ids to cart, each one calls add_to_cart in the loop, so each one does all those do_ actions below it. Its sort of like a recursive loop

pippinsplugins:
Oh I see. That makes sense, but damn, I really don't like that, sorry. Doing it recursively is just asking for trouble.

This shouldn't do it recursively, instead it should just set the session data as appropriate.

pippinsplugins:
I do get why it's setup as a recursive function, but I don't think that's the best way to go about it.

cklosowski:
Just running through testing on this. At this point it's possible to pass in options of price IDs that don't exist:

```
function ck_test_multiples() {
	$options = array( 'price_id' => array(100,99999) );
	var_dump(edd_add_to_cart( 10262, $options));
}
add_action( 'init', 'ck_test_multiples' );
```

screen shot 2015-02-16 at 1 34 26 pm

cklosowski:
In the event that someone submits options like:
```
$options = array(
    'price_id' => array( 1, 2 ),
    'quantity' => 2
);
```
Should we default each of the price ids at a value of 2?

chriscct7:
That's what it does automatically (the 2 question)
As for the other one, what happens right now in release/2.3 if you send in an invalid price_id atm?

includes/cart/functions.php
```diff
@@ -122,73 +122,118 @@ function edd_add_to_cart( $download_id, $options = array() ) {
...
+			$quantity = array( 1 );
+			$options[ 'quantity' ] = $quantity;
+		}
+
+		if ( is_array( $options[ 'quantity' ] ) ) {
+			// if the number of quantities is the same as the number of price options, match them up
+			if ( count( $options[ 'quantity' ] ) == count( $options[ 'price_id' ] ) ) {
+				foreach ( edd_combine_array( $options[ 'price_id' ], $options[ 'quantity' ] ) as $price => $quantity ) {
+					$local_options = $options;
+					$local_options[ 'price_id' ] = preg_replace( '/[^0-9\.-]/', '', $price );
```
cklosowski: Instead of regular expressions here. I'd say it's safer to run an `is_numeric` check on this, and make sure that `absint` matches the `(int)`.

I'd act on caution when parsing to get a price ID if a string was sent in. Worse case scenario their items are not added to the cart.

includes/cart/functions.php
```diff
@@ -122,73 +122,118 @@ function edd_add_to_cart( $download_id, $options = array() ) {
...
 
 	do_action( 'edd_pre_add_to_cart', $download_id, $options );
 
 	$cart = apply_filters( 'edd_pre_add_to_cart_contents', edd_get_cart_contents() );
 
-	if ( edd_has_variable_prices( $download_id )  && ! isset( $options['price_id'] ) ) {
+	if ( edd_has_variable_prices( $download_id )  && ! isset( $options[ 'price_id' ] ) ) {
 		// Forces to the first price ID if none is specified and download has variable prices
-		$options['price_id'] = '0';
+		$options[ 'price_id' ] = '0';
```
cklosowski: `0` is no longer the first price ID. May want to use the default price ID for a download?

`edd_get_default_variable_price( $download_id )`

includes/cart/functions.php
chriscct7: I was just doing what we already had. We can absolutely do that, but I think alot of people pass in strings. I mean we do in some places I think

includes/cart/functions.php
chriscct7: sure I can fix that

cklosowski:
@chriscct7 The 2nd one isn't working. If i give an array of price ids, and only a single quantity, i just get 1 of each price ID.

As far as the first one, invalid price IDs, it does the same your does, but if you're refactoring this thing to work, we should validate a price ID exists before adding it. Can just be to get the price IDs and see if array key exists.

includes/cart/functions.php
cklosowski: Yeah makes more sense...I'd rather see us not add an invalid price ID, than add one if there was something else weird in there.

includes/cart/functions.php
cklosowski: :+1:

cklosowski:
@pippinsplugins @chriscct7 I think the logic is a bit confusing...but it's pretty solid logic. Maybe if we just had some comments calling out, specifically, that we're recursively calling this function to complete the add to cart. My biggest concern is that the logic can be somewhat confusing if you're trying to quickly debug.

pippinsplugins:
After talking with @cklosowski some more, I think this should work differently. Instead of having price IDs passed in as an array,I think `edd_add_to_cart()` should allow an array of downloads to be passed, with each one having it's one options.

For example:
```php
// to add multiple items
edd_add_to_cart( array( 
    2215 => array(
        'price_id' => 2,
        'quantity' => 2
    ),
    2215 => array(
        'price_id' => 1,
        'quantity' => 2
    )
) );

// To add a single item
edd_add_to_cart( 2215, $options );
```

pippinsplugins:
It's a much simpler logic, both in the code and the usage.

chriscct7:
That's how I originally wanted to do it as well, but the problem is that doesn't solve the regression. The way the data is currently setup in the PR is how we had it before we accidentally removed the functionality. Thoughts?

pippinsplugins:
I'm confused. What's the regression again?

chriscct7:
We used to let people add multiple price_ids to the cart using a format like this:

``` php
$options = array(
    'price_id' => array( 1, 2 ),
    'quantity' => 2
);
edd_add_to_cart( 2215, $options );
```

chriscct7:
@pippinsplugins thoughts on the previous comment?

pippinsplugins:
@chriscct7 We've never allowed people to add multiple price options.

chriscct7:
We did, whether we realized it or not. If you go to any commit before https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/76d7232f929e3362db039071d09b34b91d6009b9/includes/cart/functions.php#L139 when @cklosowski removed it, you can.

chriscct7:
Ticket opened for this as #3192

pippinsplugins:
Closing in favor of the logic proposed above (and discussed curing Prestige):
```php
// to add multiple items
edd_add_to_cart( array( 
    2215 => array(
        'price_id' => 2,
        'quantity' => 2
    ),
    2215 => array(
        'price_id' => 1,
        'quantity' => 2
    )
) );

// To add a single item
edd_add_to_cart( 2215, $options );
```

Issue #2984, opened 2015-01-21
EDD_Payment_Stats triggers errors with no date passed.


danieliser:
The following example returns errors with all debug enabled.

	$stats = new EDD_Payment_Stats;
	$total = $stats->get_earnings( 67 );

Warning: array_key_exists(): The first argument should be either a string or an integer in \wp-content\plugins\easy-digital-downloads\includes\class-edd-stats.php on line 141

The problem seems to be that EDD_Payment_Stats::get_earnings() defaults parameters for $start_date & $end_date to false.

After some digging I found this https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/class-edd-stats.php#L107

		if( empty( $_start_date ) ) {
			$this->start_date = 'this_month';
		}

		$this->start_date = $_start_date;

Which looks like it should be:

		if( empty( $_start_date ) ) {
			$_start_date = 'this_month';
		}

		$this->start_date = $_start_date;

This is just prior to the call to convert_date() which triggers the error when `false` is passed to array_key_exists.

danieliser:
This fixes my issue.

danieliser:
Looking at the rest of that same function where the error begins, it could also be replaced as:

		if( empty( $_start_date ) ) {
			$this->start_date = 'this_month';
		}
		else {
			$this->start_date = $_start_date;
		}

Your preference.

pippinsplugins:
We can get this fixed for 2.2.5.

danieliser:
Ok, NP. I will get a PR submitted.

danieliser:
Preference for fix? Assuming the second.

pippinsplugins:
Mobile so haven't read too closely. Pick one and I'll let you know when testing :)

pippinsplugins:
@danieliser Still interested in writing up a PR for this?

danieliser:
Absolutely. Will submit it today.

danieliser:
Is there any reason we can't simplify this function? By that i mean currently it sets $this->start_date and then 2 lines later sets $this->start_date again using $this->convert_date.

I feel like it could be simplified to check for empty values and then set $this->start_date at the end only.

danieliser:
I wrote it up in a more simplified manner. Since there are no other methods called between setting $this->start_date and setting it again later in the function there is no need to set the class variables twice.

I rewrote it to simply check for empty values before setting the class variables with convert_date.

[...]

PR #3075, opened 2015-02-10
Issue/2984

danieliser:
#2984


cklosowski:
This looks fine, however it's tagged for ```2.3```. Can you close this and resubmit against ```release/2.3``` branch?

That way if we have to drop a ```2.2.8 or 2.2.9``` we're not pushing something along with it.

danieliser:
Sure thing. Maybe a stupid question but the process is to just create a new PR but not to the master, rather to the release/2.3 correct?

Just wanna make sure its in right place.

pippinsplugins:
That's correct.

cklosowski:
:+1: Thanks @danieliser !

PR #3091, opened 2015-02-12

chriscct7:

#2794


chriscct7:
Will refresh when I get time

cklosowski:
:+1: Thanks @chriscct7

chriscct7:
Resubmited as https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3255

PR #3095, opened 2015-02-12
Add support for banning email domains

### LINKED ISSUE #3094: Add support for banning email domains
As suggested [here](https://easydigitaldownloads.com/support/topic/idea-for-banned-emails/), it'd be cool for people to proactively disallow email domains, not just retroactively ban individual emails

ghost1227:
Fixes #3094

### DIFF
--- a/includes/checkout/functions.php
+++ b/includes/checkout/functions.php
@@ -282,11 +282,12 @@ function edd_is_email_banned( $email = '' ) {
 	}
 
 	$banned_emails = edd_get_banned_emails();
-	$ret = false;
 
 	foreach( $banned_emails as $banned_email ) {
-		if( $banned_email == trim( $email ) || stristr( trim( $email ), $banned_email ) ) {
-			$ret = true;
+		if( is_email( $banned_email ) ) {
+			$ret = ( $banned_email == trim( $email ) ? true : false );
+		} else {
+			$ret = ( stristr( trim( $email ), $banned_email ) ? true : false );
 		}
 	}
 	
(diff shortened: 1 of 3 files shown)


cklosowski:
@ghost1227 Minor thing of making the tests more readable...can we change the domains to:
```is-banned.com``` and ``is-no-banned.com```.

Then any email that should return ```true``` should be on ```is-banned.com``` and anything that should return ```false``` is on ```is-not-banned.com```. Allows people contributing to quickly read the unit tests and know what's expected without having to do too much reading.

I like this idea, @pippinsplugins, thoughts?

**Comment by ghost1227** [comment]
@pippinsplugins those changes work for you?

File: includes/checkout/functions.php [hunk 1]
```diff
@@ -281,8 +281,16 @@ function edd_is_email_banned( $email = '' ) {
 		return false;
 	}
 
-	$ret = in_array( trim( $email ), edd_get_banned_emails() );
+	$banned_emails = edd_get_banned_emails();
 
+	foreach( $banned_emails as $banned_email ) {
+		if( is_email( $banned_email ) ) {
+			$ret = ( $banned_email == trim( $email ) ? true : false );
```
pippinsplugins: This can be simplified to `$ret = $banned_email == trim( $email )`

ghost1227:
Crap... now the build is failing

includes/checkout/functions.php
cklosowski: @pippinsplugins I think it simplifies it, but for what it's worth, the readability of it as it stands is better I think.

ghost1227: @pippinsplugins gets final say. Say the word and I'll switch it

pippinsplugins: It's fine. Revert back to how it was, that's fine.

cklosowski:
:+1: thanks @ghost1227

PR #3096, opened 2015-02-12
Address Undefined Indexes and only for users

cklosowski:
Addresses further issues on #2548


cklosowski:
@pippinsplugins When you are ready.

Issue #3130, opened 2015-02-15
Invalid $startdate in CLI and improper $cart_details placement in payments() method


chriscct7:
There's a syntax error, a misnamed variable, and a missing aray declaration

pippinsplugins:
@chriscct7 Make sure you always use a proper title that describes the issue. Please don't just use "has bugs, must fix" type of titles.

PR #3131, opened 2015-02-15
EDD CLI class fixes

chriscct7:
Solves #3130

### DIFF
--- a/includes/class-edd-cli.php
+++ b/includes/class-edd-cli.php
@@ -542,7 +542,6 @@ class EDD_CLI extends WP_CLI_Command {
 			'discount'      => 'none'
 		);
 
-		$cart_details = array();
 		for( $i = 0; $i < $number; $i++ ) {
 
 			$products = array();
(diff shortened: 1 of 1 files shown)


includes/class-edd-cli.php
```diff
@@ -602,7 +602,8 @@ public function payments( $args, $assoc_args ) {
 					'quantity' => 1,
 					'options'  => $options
 				);
-
+				
+				$cart_details = array();
```
cklosowski: This is going to reset the `$cart_details` as an array for every item in this `foreach`. It means that we'll never be able to have anything but the last key in this array...were you meaning to set `$cart_details[$key] = array()`? Or maybe put `$cart_details = array();` outside of the foreach?

Foreach starts here:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/issue/3130/includes/class-edd-cli.php#L574

chriscct7: That was supposed to be before the foreach. I'll move it.

PR #3139, opened 2015-02-15
edd_get_cart_item_price called with too many arguments

chriscct7:
Solves #3138 and some broken docblocs


cklosowski:
Yes, the definition of `edd_get_cart_item_price` only accepts 2 args:
`function edd_get_cart_item_price( $download_id = 0, $options = array() )`

Merging.

Issue #3144, opened 2015-02-15
The download class doesn't have a post_status key as used in edd_get_purchase_link


chriscct7:

pippinsplugins:
Anytime creating an issue like this, please link to the actual error on scrutinizer so others can see it.

pippinsplugins:
Is there a way we could automatically setup the properties of `WP_Post` without manually declaring every one of them? We can't extend WP_Post since it is marked as `final`

pippinsplugins:
By the way, this isn't just for `post_status`, it's for all properties on `WP_Post` that we don't explicitly setup.

chriscct7:
https://scrutinizer-ci.com/g/easydigitaldownloads/Easy-Digital-Downloads/issues/release/2.3/files/includes/template-functions.php?orderField=path&order=asc#inspectioncomment-557442657

We should set them all explicitly up, so that other developers looking at that class understand they are there

pippinsplugins:
I'm fine with just adding them.

PR #3145, opened 2015-02-15
Fixes the use of the post_status key in edd_get_purchase_link

chriscct7:
Solves #3144 and corrects a docbloc

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -71,13 +71,14 @@ function edd_get_purchase_link( $args = array() ) {
 
 	$args = wp_parse_args( $args, $defaults );
 
-	$download = new EDD_Download( $args['download_id'] );
+	$download      = new EDD_Download( $args['download_id'] );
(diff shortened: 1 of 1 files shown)


cklosowski:
There is zero need for this:
```$download = WP_Post::get_instance( $_id );``` is used when instantiating the `EDD_Download` class, and by that, brings in the `post_status` attribute.

Tested it out myself:
```
$download = new EDD_Download( 74 );
var_dump($download->post_status);
```

Result: `string(7) "publish"``

It's set in this process:
```
foreach ( $download as $key => $value ) {

	$this->$key = $value;

}
```

chriscct7:
Scutinizer doesn't seem to like the fact that the key isn't declared for the class. I think we should add the key declaration to the class so other developers looking at that class know its there.

pippinsplugins:
What's the actual error scrutinizer is giving?

pippinsplugins:
Closing this and fixing in a new PR.

PR #3147, opened 2015-02-15
add filter for edd_cart_item_price()

JeroenSormani:
Hey guys,

Just a small PR to add a filter to the checkout price label.

Cheers!

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -398,7 +398,7 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
 
 	$price = edd_currency_filter( edd_format_amount( $price ) );
 
-	return $price . $label;
+	return apply_filters( 'edd_cart_item_price_label', $price . $label, $item_id, $options );
 }
 
 /**


pippinsplugins:
Looks good to me! I Both @cklosowski and I feel like there may already be a way to do this, so we're going to do some digging before merging. Assuming we don't find a way, we'll merge it in.

Second note: since this against `master`, please do not merge until 2.3 is ready to be merged into `master.

JeroenSormani:
There is a filter here to filter the cart item price; https://github.com/JeroenSormani/Easy-Digital-Downloads/blob/new-filter/includes/cart/functions.php#L447

But that doesn't allow to change the complete cart price label.

This where I use the filter for, it is not possible without; http://cl.ly/image/1O0R0c3h1I1n
(free 'EDD Sale Price' plugin)

pippinsplugins:
Ah! Excellent example. Ok merging.

pippinsplugins:
Well merging once 2.3 is ready to go into `master`.

cklosowski:
:+1: Thanks @JeroenSormani :)

Issue #3104, opened 2015-02-13
Another rounding issue with product prices when prices are inclusive of tax


pippinsplugins:
We have an extra penny showing up due to rounding issues.

To replicate:

- Set tax rate to 20%
- Set prices inclusive of tax
- Create a Download with a price of 9.99
- Add item to the cart and select appropriate country / state to be charged 20% tax
- Final price should be 9.99 but is actually 10.00

![screen shot 2015-02-13 at 1 47 15 pm](https://cloud.githubusercontent.com/assets/1034109/6194129/ddee4bb8-b386-11e4-8855-c17cf343396f.png)

The issue was noticed via this [ticket](https://easydigitaldownloads.com/support/topic/price-calculated-incorrectly/).

cklosowski:
I believe the issue lies in `edd_calculate_tax()`.

For a value of `9.99` it's returning that the tax amount should be `1.665`, which is then getting increased to `1.67` and then setting the `item_price` at the wrong amount putting us at `10.00` instead of `9.99`.

If we use the `edd_currency_decimal_filter()` function to take the decimal count for the shop, we'll get rid of this.

Does it seem right tot do the following before retuning from the `edd_calculate_tax` function?
```$tax = round( $tax, edd_currency_decimal_filter() );```

pippinsplugins:
Yeah I think that is right

cklosowski:
Ok verified sales record properly, and when sending to paypal that the amount is sent to them correctly as well.

pippinsplugins:
:+1:

pippinsplugins:
@cklosowski I don't think the proposed changes in #3169 work actually. We used to round the taxes there, https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/ee46b2ca4e55e2f22d006844334eb1bb1c2dc615#diff-e1900e6f7d703ca91cbd7c62d0c28eee,  but it caused a problem when there are multiple items and quantities: #1881

cklosowski:
Not sure where else we would round taxes...but in the tax calculation...but we can go back to the drawing board. This tells me we have some DEEP DEEP seeded issues with rounding that need to be resolved.

pippinsplugins:
Oh I guarantee there are some deep, deep, deep issues with taxes. Welcome to my *favorite* API in EDD.

chriscct7:
I think there's a very simple solution to this

[...]

PR #3169, opened 2015-02-19
Rounding edd_calculate_tax to the store decimal filter

cklosowski:
Addresses tax/rounding issue in #3104


cklosowski:
Closing as there was a previously known issue with this type of rounding

Issue #3083, opened 2015-02-11
Enhancement: Give Option Name field focus when clicking Add New Price


mindctrl:
When creating new variable price options, it would be nice if the Option Name field received focus when clicking Add New Price.
![image](https://cloud.githubusercontent.com/assets/1254870/6150760/469119c4-b1df-11e4-8793-20d661d0ff21.png)

pippinsplugins:
:+1:

danieliser:
I can tackle this one.

danieliser:
Got a PR for this. Would like to discuss a rewrite of the JS for the future though. See a lot of ways we can improve on whats there.

danieliser:
Also not sure what you use to minify your JS currently so I am not submitting a minified version with this as to not break your routine.

danieliser:
Ignore the first PR, had some code that PHPStorm decided to reformat for me. Didn't notice till I submitted the PR.

pippinsplugins:
When submitting PRs, do not worry about .min files. We take care of those just before release.

danieliser:
Good to know.

danieliser:
PR has been updated to reflect the requested changes. Should be good for merge now.

pippinsplugins:
Thanks @danieliser !

PR #3178, opened 2015-02-21
Issue/3083

danieliser:
#3083


danieliser:
This had changes that were not made intentionally due to autoformatting.

PR #3179, opened 2015-02-21
Option name focus when adding new price.

danieliser:
#3083


danieliser:
Updated.

pippinsplugins:
Looks great.

Issue #3181, opened 2015-02-22
Introduce edd_update_option


chriscct7:
We desperately need an edd_update_option function as many extensions and even some core unit tests need to adjust the value of options. 

Just a simple function that takes in the key and the values and returns the result.

pippinsplugins:
Yep, we've needed one for a while now.

pippinsplugins:
@cklosowski #3182 is finished and the build is passing (I updated all unit tests to use the new functions). Could you review it and merge if you approve?

cklosowski:
Reviewing now @pippinsplugins

PR #3182, opened 2015-02-22
Adds edd_update_option

chriscct7:
Makes #3181 a reality

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -79,6 +79,7 @@ function edd_update_option( $key = '', $value = false ) {
  * @return boolean True if updated, false if not.
  */
 function edd_delete_option( $key = '' ) {
+
 	// If no key, exit
 	if ( empty( $key ) ){
 		return false;
@@ -88,14 +89,18 @@ function edd_delete_option( $key = '' ) {
 	$options = get_option( 'edd_settings' );
 
 	// Next let's try to update the value
-	unset( $options[ $key ] );
+	if( isset( $options[ $key ] ) ) {
+
+		unset( $options[ $key ] );
+	
+	}
+
 	$did_update = update_option( 'edd_settings', $options );
 
 	// If it updated, let's update the global variable
 	if ( $did_update ){
 		global $edd_options;
-		unset( $options[ $key ] );
-
+		$edd_options = $options;
 	}
 
 	return $did_update;
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -37,6 +37,11 @@ function edd_get_option( $key = '', $default = false ) {
  * @return boolean True if updated, false if not.
  */
 function edd_update_option( $key = '', $value = false ) {
+	// If no key, exit
+	if ( empty( $key ) ){
+		return false;
+	}
+
 	// First let's grab the current settings
 	$options = get_option( 'edd_settings' );
 
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -29,6 +29,35 @@ function edd_get_option( $key = '', $default = false ) {
 }
 
 /**
+ * Update an option
+ *
+ * Updates an edd setting value in both the db and the global variable.
+ *
+ * @since 2.3
+ * @return boolean True if updated, false if not.
+ */
+function edd_update_option( $key = '', $value = false ) {
+	// First let's grab the current settings
+	$options = get_option( 'edd_settings' );
+
+	// Let's let devs alter that value coming in
+	$value = apply_filters( 'edd_update_option', $value, $key );
+
+	// Next let's try to update the value
+	$options[ $key ] = $value;
+	$did_update = update_option( 'edd_settings', $options );
+
+	// If it updated, let's update the global variable
+	if ( $did_update ){
+		global $edd_options;
+		$edd_options[ $key ] = $value;
+
+	}
+
+	return $did_update;
+}
+
+/**
  * Get Settings
  *
  * Retrieves all plugin settings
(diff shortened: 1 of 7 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -29,6 +29,35 @@ function edd_get_option( $key = '', $default = false ) {
...
 /**
+ * Update an option
+ *
+ * Updates an edd setting value in both the db and the global variable.
+ *
+ * @since 2.3
+ * @return boolean True if updated, false if not.
+ */
+function edd_update_option( $key = '', $value = false ) {
+	// First let's grab the current settings
```
pippinsplugins: This needs some defensive coding. If `$key` is empty, return false.

includes/admin/settings/register-settings.php
```diff
@@ -29,6 +29,35 @@ function edd_get_option( $key = '', $default = false ) {
...
+	$options = get_option( 'edd_settings' );
+
+	// Let's let devs alter that value coming in
+	$value = apply_filters( 'edd_update_option', $value, $key );
+
+	// Next let's try to update the value
+	$options[ $key ] = $value;
```
pippinsplugins: If `$value` is `empty()` and it exists in `$options`, we need to `unset()` it.

tests/tests-misc.php
```diff
@@ -477,4 +477,22 @@ public function test_edd_is_host() {
...
+		$this->assertFalse( $isset );
+
+		$updated = edd_update_option( $key, $value );
+
+		// The option should have succesfully updated
+		$this->assertTrue( $updated );
+
+		// The option retrieve should be equal to the one we set
+		$this->assertEquals( $value, edd_get_option( $key, false ) );
+
```
pippinsplugins: Let's add a test for removing an option, like this: `edd_update_option( 'test_mode', false );`

includes/admin/settings/register-settings.php
chriscct7: That sounds really dangerous. If you set a value to false, which alot of people do, including core, doing that means the index keys won't exist.
Before edd_get_option, extension developers used to use the edd_option global. Let's say we have a key called 'something'.

My code from before edd_get_option could have reasonably been:

if ( $edd_options['something] ){

}
because even if the value was set to false, I still expected the array key to be there.
Suddenly if we remove it we're causing errors. Sure, is it their responsibility to do an isset check? Maybe, but we can't assume people have

includes/admin/settings/register-settings.php
chriscct7: we can do that

includes/admin/settings/register-settings.php
chriscct7: Futhermore, if we wanted to unset or remove a setting, we should have an edd_delete_option. WordPress doesn't remove an option if you update_option with a value of false

tests/tests-misc.php
chriscct7: See https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3182#discussion_r25135927

includes/admin/settings/register-settings.php
pippinsplugins: Not dangerous at all. It's exactly how settings get removed right now: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/admin/settings/register-settings.php#L741

chriscct7: Here's an example of two plugins. One plugin, Plugin 1, is newer and uses the new function. The second plugin, Plugin 2, uses a setting from Plugin 1, and like many extensions it uses the $edd_options global variable without doing an isset check:

https://gist.github.com/chriscct7/b47d1307e503cb464c8d
https://gist.github.com/chriscct7/b544548ef1ab796df770

Testing:
Install and activate plugins 1 and 2
After activating both. Comment out the upper part of plugin 1 (to simulate it updating to using this new function) and uncomment the bottom.

includes/admin/settings/register-settings.php
```diff
@@ -29,6 +29,78 @@ function edd_get_option( $key = '', $default = false ) {
...
+
+/**
+ * Remove an option
+ *
+ * Removes an edd setting value in both the db and the global variable.
+ *
+ * @since 2.3
+ * @return boolean True if updated, false if not.
+ */
+function edd_remove_option( $key = '' ) {
```
chriscct7: make edd_delete_option

tests/tests-misc.php
```diff
@@ -477,4 +477,37 @@ public function test_edd_is_host() {
...
+		// The option shouldn't exist
+		$this->assertFalse( $isset );
+
+		$updated = edd_update_option( $key, $value );
+
+		// The option should have succesfully updated
+		$this->assertTrue( $updated );
+
+		// The option retrieve should be false since it doesn't exist
+		$this->assertFalse( edd_get_option( $key, false ) );
```
chriscct7: this test needs to be fixed still

includes/admin/settings/register-settings.php
```diff
@@ -29,6 +29,78 @@ function edd_get_option( $key = '', $default = false ) {
...
+ * @since 2.3
+ * @return boolean True if updated, false if not.
+ */
+function edd_update_option( $key = '', $value = false ) {
+	// If no key, exit
+	if ( empty( $key ) ){
+		return false;
+	}
+
+	if ( ! $value ){
```
pippinsplugins: Use `empty()` here.

includes/admin/settings/register-settings.php
```diff
@@ -29,6 +29,78 @@ function edd_get_option( $key = '', $default = false ) {
...
+		$remove_option = edd_remove_option( $key );
```
pippinsplugins: @chriscct7 This should be called edd_delete_option()

pippinsplugins:
@chriscct7 I've gone ahead and made the updates. Just waiting for unit tests to pass then we can merge this in.

chriscct7:
Ok cool

cklosowski:
Looks good, merging into `release/2.3`

Issue #2955, opened 2015-01-12
Introduce edd_item_supports() function


pippinsplugins:
We should introduce an `edd_item_supports()` function that would allow developers to determine if a product supports various features.

For example:
```php
edd_item_supports( 'quantities' );
```
```php
edd_item_supports( 'taxes' );
```
```php
edd_item_supports( 'shipping' );
```
This should be introduced as a method in the `EDD_Download` class that then has a global wrapper function.

ghost1227:
oooh! I like that idea!

cklosowski:
:+1:

danieliser:
How would you implement it? I'm assuming in the Download class you would have a filter that developers could hook into to check meta and enable features on the fly storing them in the Download class?

Sounds relatively straight forward. I can take this one.

pippinsplugins:
The plan is to have it work just like add/remove_theme_support() and add/remove_post_type_supports().

PRs always welcome :)

danieliser:
So the question becomes, since both of those are based on static items. Post Types & Themes aren't quite as dynamic as say download 242 supports feature x.

So this becomes a template function that uses the current query args similar to how get_the_content( $id = 0 ) works. 

In which case I would just make functions to wrap new methods in EDD_Download class.

Unless I misunderstood the usage. But I can't quite see a general download_supports('key') functions usage.

But a download_support( $id, $key ) sounds like something I would use during templating or in a loop mainly.

If this is correct by your original idea then I can take this one as well.

pippinsplugins:
We need to introduce both a method in `EDD_Download` and a global helper
function that calls the method.

danieliser:
OK, that's how I had it worked in my head. I can get this done. Just wanted to be sure it matched your original Idea, in that it is a function to check that each individual download supports a feature.

pippinsplugins:
If you end up working on it, post a proof of concept before going all the way through it, just to ensure everyone is on the same page and work doesn't get wasted.

danieliser:
No problem. What do you suggest? The functionality itself is rather simple. You want an outline of how it will be integrated and used? Can do that for sure.

pippinsplugins:
Just a basic proposal of the API itself.

danieliser:
Sure thing. WIll post it here.

pippinsplugins:
I'm going to go ahead and punt this to 2.4 since it's still a `nice to have`.

[...]

PR #3184, opened 2015-02-22
Add edd_download_supports to EDD_Download class w/ helpers

danieliser:
Feature #2955
Added private $supports array to EDD_Download class.
Added public supports() method to the EDD_Download class, accepts a string or array of features to check against $supports array.
Added 'edd_download_supports' filter allowing addition and setting of features to the $supports array.
Added edd_download_supports function for usage outside the loop.
Added edd_item_supports for use inside the loop.

### DIFF
--- a/includes/class-edd-download.php
+++ b/includes/class-edd-download.php
@@ -101,6 +101,17 @@ class EDD_Download {
 	private $button_behavior;
 
 	/**
+	 * Default download support options
+	 *
+	 * @since 2.4
+	 */
+	private $supports = array(
+		'quantities' => false,
+		'taxes'      => false,
+		'shipping'   => false,
+	);
+
+	/**
 	 * Get things going
 	 *
 	 * @since 2.2
@@ -628,6 +639,33 @@ class EDD_Download {
 
 		return (bool) apply_filters( 'edd_is_free_download', $is_free, $this->ID, $price_id );
 
+	}
+
+	/**
+	 * Determine if the download supports a feature
+	 * 
+	 * @since  2.4
+	 * @param  string $feature   The feature being checked for support
+	 * @return bool              Whether the download supports the feature.
+	 */
+	public function supports( $features = '' ) {
+		$supports = apply_filters( 'edd_download_supports', $this->supports, $this->ID );
+
+		if( empty( $features ) ) {
+			return false;
+		}
+
+		if ( ! is_array( $features ) ) {
+			return ! empty( $supports[ $features ] ) && $supports[ $features ];
+		}
+
+		$support_check = true;
+		foreach ( $features as $feature ) {
+			if ( empty( $supports[ $feature ] ) || ! $supports[ $feature ] ) {
+				$support_check = false;
+			}
+		}
+		return $support_check;
 	}
 
 }
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -831,3 +831,23 @@ function edd_add_download_post_classes( $classes, $class = '', $post_id = false 
 	return $classes;
 }
 add_filter( 'post_class', 'edd_add_download_post_classes', 20, 3 );
+
+/**
+ * Checks if download supports specific feature in the loop
+ *
+ * @since 2.4
+ * @author Daniel Iser
+ * @param string $feature The feature being checked.
+ * @return bool $supports True if the product supports the feature, false if the product does not or no ID is passed
+ */
+function edd_item_supports( $feature ) {
+	global $post;
+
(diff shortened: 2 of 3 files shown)


includes/class-edd-download.php
```diff
@@ -101,6 +101,17 @@ class EDD_Download {
 	private $button_behavior;
 
 	/**
+	 * Default download support options
+	 *
+	 * @since 2.4
+	 */
+	private $supports = array(
```
pippinsplugins: Defaults should be setup at the beginning of the `supports()` method.

includes/class-edd-download.php
```diff
@@ -630,4 +641,31 @@ public function is_free( $price_id = false ) {
...
 
+	/**
+	 * Determine if the download supports a feature
+	 * 
+	 * @since  2.4
+	 * @param  string $feature   The feature being checked for support
+	 * @return bool              Whether the download supports the feature.
+	 */
+	public function supports( $features = '' ) {
+		$supports = apply_filters( 'edd_download_supports', $this->supports, $this->ID );
```
pippinsplugins: `apply_filters` should be one of the very last lines of the function and should return a boolean value.

includes/template-functions.php
```diff
@@ -831,3 +831,23 @@ function edd_add_download_post_classes( $classes, $class = '', $post_id = false
...
+
+/**
+ * Checks if download supports specific feature in the loop
+ *
+ * @since 2.4
+ * @author Daniel Iser
+ * @param string $feature The feature being checked.
+ * @return bool $supports True if the product supports the feature, false if the product does not or no ID is passed
+ */
+function edd_item_supports( $feature ) {
```
pippinsplugins: Let's combine this and `edd_download_supports` into a single function. The download ID can be passed as the second parameter. If no ID is passed, we look for one. If no ID is found (so not in the loop), then we return false.

includes/class-edd-download.php
danieliser: I was planning to add the checks for each of the builtin features in the support method once we nailed down functionality. So this works for me. Do you simply want to remove the $supports array entirely?

includes/class-edd-download.php
danieliser: Assuming we set it up that way, how would an extension add extra features? Unless we make it a static/global set of download supported features. Maybe make it a static $supports array variable, and static add_support_feature/remove_support_feature method as well to modify that array.

Then we just pull that static array into the support() method. This would more closely resemble the add/remove_post_type_support functions you mentioned before.

includes/template-functions.php
danieliser: Ok no problem, making a second function in general felt a bit wrong anyways, but passing the download_id as the second parameter didn't seem to fit the design pattern set by the other functions available.

So is the preference to make it edd_download_supports or edd_item_supports? And would you classify it more a template function or download function?

pippinsplugins: I prefer `edd_download_supports( $feature = '', $download_id = 0 )` in download-functions.

includes/class-edd-download.php
pippinsplugins: Ah yes, you're right. Do it that way.

includes/class-edd-download.php
pippinsplugins: It'd work the same way. It needs to be last so that extensions can see which features are already supported and then remove support for certain features if necessary too.

pippinsplugins:
@danieliser Would you be up for bringing this branch up to date with `release/2.4`?

danieliser:
@pippinsplugins Sure thing. I saw your new notes just now, so other than moving that filter everything else looks solid? Need to do the built in checks but need a list of what you want as the default or _built_in $supports items.

pippinsplugins:
Not 100% sure if it's ready to go in but it's definitely ready for further testing.

danieliser:
@pippinsplugins did you see my last message? I can get this updated asap, just need a full list of built in supported items.

pippinsplugins:
@danieliser Initially it will just be `taxes`, `quantity`, `variable-prices`, `buy-now`.

Also, could you resubmit this to `issues/2955`. We're too close to the 2.4 release (beta today) to get this in, but I'd really like to get it into 2.5. Along with getting the API in, we'll also make sure the API is used throughout the plugin when this is released.

pippinsplugins:
@danieliser can you re-submit this to `issue/2955`?

pippinsplugins:
Manually merged into `issue/2955`

PR #3194, opened 2015-02-24
Update test purchase key for 2.3

ghost1227:
Issue #1756


pippinsplugins:
Can you resubmit to `issue/1756`? We don't know for sure when this will go in so best to get it in an issue branch first.

PR #3196, opened 2015-02-24
Quick fix for #3080

### LINKED ISSUE #3080: No action hooks available to add "Actions" to items in the cart
Right now the only elements available in the "Actions" column of the Cart table are the quantity input(if enabled) and the "Remove" link for removing the item from the cart. 

It would be great if there's at least an action that allows themes/plugins to hook in and add extra actions there(for instance if you have some extra customizations, you can show controls to remove those). Ideally EDD would handle (to some extent) the rendering of those actions(think the row actions in posts tables - "Quick Edit", "Edit", "Trash", etc.), but this would be far more work and might be over-complicating it.

Given the current situation, my options are limited to:
 - introducing a new column(not ideal, because of #3079 )
 - hooking to the `get_title` filter(probably won't work, because it'd be difficult to figure out which cart item we're displaying the title for)
 - Adding the actions via JS(far from ideal)
 - Output buffering triggered at the right action hooks, followed by some DOM manipulations in PHP(overkill)

Well, there are plenty of options(and I'm sure there are maybe one or two I've missed), but none of them seem very feasible to me.

Again the context for this is that I'm trying to attach Gravity Form submissions to specific cart items and I'd like to give users a way to edit those entries before checking out.
For this particular issue, I might go with sort of duplicating things(having an interface that lists all items with form submissions and a way to edit that submission).

ghost1227:
Fixes #3080

### DIFF
--- a/templates/checkout_cart.php
+++ b/templates/checkout_cart.php
@@ -43,7 +43,7 @@ global $post; ?>
 							<input type="hidden" name="edd-cart-download-<?php echo $key; ?>-options" value="<?php echo esc_attr( serialize( $item['options'] ) ); ?>"/>
 						<?php endif; ?>
 						<a class="edd_cart_remove_item_btn" href="<?php echo esc_url( edd_remove_item_url( $key, $post ) ); ?>"><?php _e( 'Remove', 'edd' ); ?></a>
-						<?php do_action( 'edd_cart_actions', $item ); ?>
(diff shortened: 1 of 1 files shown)


cklosowski:
@ghost1227 can you add the `$key` back in as a hook argument and commit up? Otherwise looks good.

PR #3203, opened 2015-02-27
Fallback check for buy now support

### LINKED ISSUE #3201: Buy Now setting for downloads should check payment gateway functionality
Some gateways support Buy Now functionality, so the button behavior setting will display in the Download Settings meta box. With that gateway activated, you can change the behavior to Buy Now. If you deactivate that gateway and only use one that does **not** support Buy Now functionality (Test Payment gateway is perfect to test this), the buy now meta stays in place and there's no obvious way to change that behavior back to Add to Cart. You have to reactivate the old gateway just to switch back.

ghost1227:
Fixes #3201

### DIFF
--- a/includes/class-edd-download.php
+++ b/includes/class-edd-download.php
@@ -481,7 +481,7 @@ class EDD_Download {
 
 			$this->button_behavior = get_post_meta( $this->ID, '_edd_button_behavior', true );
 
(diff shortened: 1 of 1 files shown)


cklosowski:
Looks good to me. merging.

Issue #3215, opened 2015-03-02
Don't check if ajax calls work if Airplane Mode plugin is active


chriscct7:
If the Airplane mode plugin is installed and enabled we should ignore the test. It's kind of annoying to see the error all the time.

chriscct7:
Pull request submitted: #3216

cklosowski:
We've set feature freeze for 2.3, punting to 2.4.

cklosowski:
Strike that, we'll go 2.3.3

pippinsplugins:
Oops, resubmitting to `master`.

chriscct7:
Implemented in #3281

PR #3216, opened 2015-03-02
Ignore the ajax works test if Airplane mode is active

chriscct7:
Solves #3215


pippinsplugins:
I don't think this is done right. It shouldn't set the transient if airplane mode is active. Instead it should short circuit the remote request and just return `true`.

chriscct7:
If you do that then Airplane mode will be checked on every single admin page viewing for those using Airplane mode. That would slow the localhost down alot, even though we know it works

chriscct7:
Unless what you want to do is move the airplane mode check above the remote request. We could do that and solve the problem.

pippinsplugins:
I'm doing the airplane mode check before the request is made. I have an updated commit almost ready.

chriscct7:
Ok

pippinsplugins:
@cklosowski Take a look at this and let me know if you are happy with it.

pippinsplugins:
@chriscct7 too

pippinsplugins:
Resubmitting against master.

Issue #2657, opened 2014-09-25
Add admin page helper function for extension developers


chriscct7:
Enqueuing scripts for backend EDD work right now could be made easier for extension devs.

I'm thinking we could do something like edd_is_admin_page() but the problem with that function is it would let scripts load on all pages

What I'm thinking is something like edd_is_admin( $page ) where you pass in a param

The param could be a string, or an array of strings.
Possible values: 'download' (edit download page), 'order' (edit order page)', 'coupon' (edit coupon page), download_table (download table), etc

If the param is a string, for example, download, and we're on the edit download page it would return true, else return false.

If the param is an array, it would see if any are true, and if so return true, else false. 

To reduce the chance of issues with users on old versions of EDD not having that function, we could actually just impement this into edd_is_admin_page, and if theres a param do the new logic ^, else do the old one

This would make it easier to enqueue scripts just to specific EDD admin views
thoughts?

chriscct7:
This is ready for merge in #3088

pippinsplugins:
This shouldn't be a separate function. It should be part of the same function we already have.

chriscct7:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/55c0e6cdadcd74f53d6c1615b9e90c104bde8444

cklosowski:
Punting until 2.4, we need some more time to make sure this logic is solid.

PR #3231, opened 2015-03-09
edd_is_admin_page to support $page and $view

chriscct7:
Implements `edd_is_admin` as defined in #2657

Supports new customer management ui and reports page

### DIFF
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -45,37 +45,311 @@ function edd_add_options_link() {
 add_action( 'admin_menu', 'edd_add_options_link', 10 );
 
 /**
- *  Determines whether the current admin page is an EDD admin page.
- *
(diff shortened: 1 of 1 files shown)


chriscct7:
@pippinsplugins can this be merged now?

pippinsplugins:
No, absolutely not.

Not until @cklosowski and I have personally tested it extensively and we've evaluated it to see if it's the best approach.

chriscct7:
I didn't mean merged as in like merge it now, was about 1am when I wrote that. I meant does anything else need to be done for it so we can move it along. I'd really like to use this in several of my extensions and if there's something that needs to be done, I just wanted to know so I could do it now, so it hopefully doesn't miss yet another release  *crosses fingers for 2.4*

chriscct7:
Resubmitted as #3384

PR #3234, opened 2015-03-09
Moved Logged In Check To Template

fpcorso:
As the templates can be overrode using this process http://docs.easydigitaldownloads.com/article/266-customizing-the-purchase-history-template, it makes more sense to have the check in the template so the user can customize the entire template.

For example, I am customizing the template to show a certain message if the user is not logged in. To do that now, I would have to remove the shortcode and create a new shortcode function without the logged in check. Then, I would have to replace the template.

Using my new way, I can just replace the template.


pippinsplugins:
Makes sense to me.

@cklosowski Any objections?

cklosowski:
I'm good with this, can you please update the PR from the most recent master so it merges cleanly @fpcorso ?

fpcorso:
@cklosowski  Alright, I am still fairly new with github but I updated my master and had some merge conflicts which I followed a guide to resolve. Now I see a third file in the pull request (includes/admin/customers/class-customer-table.php). Is there something I need to do to change that?

cklosowski:
That was a file we added in 2.3, but it looks like your local git configuration is set to commit file permission changes...

Best to set your local development environment to ignore commiting file permission changes:
`git config core.fileMode false`

cklosowski:
You may try changing the permissions back to `755` and then committing, then setting that config...

fpcorso:
Ah, there we go. Back to my original two files changed. Sorry for being such a novice with git.

pippinsplugins:
Looks good!

Issue #3259, opened 2015-03-13
File Download Limit not observed in 2.3


mindctrl:
It looks like the File Download limit isn't working in 2.3.

* Create new download, set file download limit to 1.
* Purchase and download the item.
* Download item again. And again.

The limit worked in 2.2.8.

[Reported here](https://easydigitaldownloads.com/support/topic/file-download-limit-a-problem-after-upgrade-to-version-2-3/). Customer mentions link expiration doesn't work either. Will test that shortly.

chriscct7:
It looks like function that checks to see if the download count is checked edd_is_file_at_download_limit is only called via a deprecated function for legacy downloads. @mindctrl can you test if legacy links respect the limit?

I think we forgot to implement the checks for expiration and download limits for the new download links

mindctrl:
It looks like the old style links are limited as they should be.

chriscct7:
Looks like both expiration and download count checks are missing for new urls. Wrote #3259 to fix

pippinsplugins:
The expiration is checked in `edd_validate_url_token()`

chriscct7:
File download count check implemented in https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3260

PR #3260, opened 2015-03-13
Checks for download limit for new urls #3259

chriscct7:
The file download count check which for legacy links are done in the deprecated edd_verify_download_link function are not done for the new download links. This pr solves that issue.

Solves #3259

This needs to be tested

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -693,7 +693,7 @@ function edd_process_signed_download_url( $args ) {
 	$order_parts = explode( ':', rawurldecode( $_GET['eddfile'] ) );
 	
 	// Check to make sure not at download limit
-	if ( edd_is_file_at_download_limit( $order_parts[1], $order_parts[0], $order_parts[2], $order_parts[3] ) ){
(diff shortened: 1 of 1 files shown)


includes/process-download.php
```diff
@@ -691,7 +691,21 @@ function edd_process_signed_download_url( $args ) {
...
-
+	
+	// Check to make sure not at download limit
+	if ( edd_is_file_at_download_limit( $order_parts[1], $order_parts[0], $order_parts[2], $order_parts[3] ) ) {
+		wp_die( apply_filters( 'edd_download_limit_reached_text', __( 'Sorry but you have hit your download limit for this file.', 'edd' ) ), __( 'Error', 'edd' ), array( 'response' => 403 ) );	
+	}
+	
+	// Make sure the link hasn't expired
+	$expire = $_GET['ttl'];
+	if ( base64_encode( base64_decode( $expire, true ) ) === $expire ) {
```
pippinsplugins: We can't use base64 functions here. They can only be used in `deprecated-functions.php`.

pippinsplugins: `ttl` is not a base64 value any more either.

chriscct7: Fixed

PR #3267, opened 2015-03-17
Correct db customers orderby formatting

chriscct7:
Corrects formatting in https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/ac8dcc4f8fdaef4b0eb4d03fa91bb7058b6a98d9

### DIFF
--- a/includes/class-edd-db-customers.php
+++ b/includes/class-edd-db-customers.php
@@ -458,7 +458,7 @@ class EDD_DB_Customers extends EDD_DB  {
 
 		}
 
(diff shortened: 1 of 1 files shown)


cklosowski:
Actually we're doing it wrong in other places of EDD, to follow WP Coding standards with arrays, the space should only be used when a variable is used as the array key:
image

chriscct7:
Fixed all of them the other way so other people wouldn't be confused here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3268

PR #3268, opened 2015-03-17
Fix all misformatted array spacing and fix them

chriscct7:
Finds all mis-formatted array keys and fixes them
$x = $foo['bar']; // correct
$x = $foo[ 'bar' ]; // incorrect


cklosowski:
:+1:

Issue #3275, opened 2015-03-18
All add to cart buttons for same product changed to checkout when one is clicked


pippinsplugins:
If you have multiple `[purchase_link]` short codes for a product on the same page, they all get changed to `Checkout` when one is clicked.

When the buttons are all for the same price ID, it is the expected behavior but when each button has a different price ID, this should not happen.
![screen shot 2015-03-18 at 11 30 00 am](https://cloud.githubusercontent.com/assets/1034109/6713717/36a95a00-cd62-11e4-86ee-b11aa54c1c87.png)
![screen shot 2015-03-18 at 11 30 07 am](https://cloud.githubusercontent.com/assets/1034109/6713719/36ab8096-cd62-11e4-8254-a1d3325a4b00.png)
![screen shot 2015-03-18 at 11 30 13 am](https://cloud.githubusercontent.com/assets/1034109/6713718/36aad902-cd62-11e4-8c9c-c9ad25361acf.png)

mindctrl:
Works, but if you go to checkout, or another page, then come back, they all say Checkout.

cklosowski:
I'm having an issue with multiple price IDs for the same product. When I have two price ID purchase_link shortcodes, it's only adding 1 price ID, no matter what button I click:

```
[purchase_link id="514" text="Add to Cart" style="button" color="blue" price_id="1"]
[purchase_link id="514" text="Add to Cart" style="button" color="blue" price_id="2"]
```

Should be first button adds `Song 2` and second button adds `Song 3`
![image](https://cloud.githubusercontent.com/assets/1390124/6716156/a14f5a14-cd60-11e4-9b12-25fa07bc2575.png)

Only one button changes to checkout when clicking them, but they are both putting `Song 2` into the cart:
![image](https://cloud.githubusercontent.com/assets/1390124/6716168/bd149ad4-cd60-11e4-9882-6c7fdc8493ae.png)

Then when removing from the cart I see:
![image](https://cloud.githubusercontent.com/assets/1390124/6716179/cc51e39e-cd60-11e4-9b91-3b6fbb44f3a4.png)

![image](https://cloud.githubusercontent.com/assets/1390124/6716185/d92e147a-cd60-11e4-9c14-0e760c51b5f9.png)

cklosowski:
And...now I see that #3270 is set to resolve this....can we just get them in the same PR so we can test this properly?

pippinsplugins:
[...]

PR #3276, opened 2015-03-18
All add to cart buttons for same product changed to checkout when one is clicked  (#3275)

pippinsplugins:
Solves #3275

If you have multiple [purchase_link] short codes for a product on the same page, they all get changed to Checkout when one is clicked.

When the buttons are all for the same price ID, it is the expected behavior but when each button has a different price ID, this should not happen.


cklosowski:
Testing now.

Issue #3270, opened 2015-03-17
Wrong price_id gets added to cart using multiple [purchase_link] shortcodes for same product


mindctrl:
When there are multiple `[purchase_link]` shortcodes for the same product on the same page, each with its own `price_id`, adding any of them to the cart results in the first one on the page being added instead of the one clicked. Example shortcodes below. Clicking the one for `price_id="2"` results in `price_id="3"` being added to the cart.

```
[purchase_link id="718" text="Buy Now" style="button" price_id="3" color="grey"]

[purchase_link id="718" text="Buy Now" style="button" price_id="2" color="grey"]

[purchase_link id="718" text="Buy Now" style="button" price_id="1" color="grey"]
```

mindctrl:
Looks good here.

PR #3277, opened 2015-03-18
Wrong price_id gets added to cart using multiple [purchase_link] shortcodes for same product (#3270)

pippinsplugins:
Solves #3270

When there are multiple `[purchase_link]` shortcodes for the same product on the same page, each with its own `price_id`, adding any of them to the cart results in the first one on the page being added instead of the one clicked. Example shortcodes below. Clicking the one for `price_id="2"` results in `price_id="3"` being added to the cart.


cklosowski:
Looks good :+1:

PR #3281, opened 2015-03-18
Don't check if ajax calls work if Airplane Mode plugin is active (#3215)

pippinsplugins:
Solves #3215

If the Airplane mode plugin is installed and enabled we should ignore the test. It's kind of annoying to see the error all the time.

### DIFF
--- a/includes/ajax-functions.php
+++ b/includes/ajax-functions.php
@@ -35,6 +35,16 @@ function edd_is_ajax_enabled() {
  */
 function edd_test_ajax_works() {
 
+	// Check if the Airplane Mode plugin is installed
+	if ( class_exists( 'Airplane_Mode_Core' ) ) {
+
+		global $Airplane_Mode_Core;
+
+		if ( $Airplane_Mode_Core->enabled() ) {
+			return true;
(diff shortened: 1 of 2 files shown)


chriscct7:
We can make a simple unit test that declares the airplane mode class and then times it. The test call to edd_is_ajax_working shouldn't take more than 1 second if this pr works correctly

pippinsplugins:
It'd be better to test if the transient is set instead. Using the amount of time to process isn't accurate enough.

chriscct7:
But we're returning true. The transient won't be set.

Technically there's 2 tests that should be here:
1. asserting that when we instantiate the airplane mode class and then call edd_test_ajax_works that the transient (which shouldn't be there before we run this test) isn't set after we call this.
2. asseting that the time to call edd_test_ajax_works is really small (larger than 1 second indicates that we're bypassing the return true_

pippinsplugins:
That's my point @chriscct7

pippinsplugins:
@chriscct7 ^

chriscct7:
Got it. That works :)

Issue #3172, opened 2015-02-19
Cache busting cookies: edd_items_in_cart


clifgriffin:
I've started using fastcgi caching with Nginx over caching plugins.  The down downside of this is that Nginx will not cache any page or serve a cached page for any request with a Set-Cookie, and unlike Varnish, you can't exclude cookies from the cache.

I've noticed that every page request on sites with EDD includes a Set-Cookie header from EDD in the form of `edd_items_in_cart`.  It defaults to `deleted`, but if any items are in the cart it is set to that number.

I can't figure out why it would be necessary to set this cookie on every single request, especially when a user hasn't initiated any shopping actions.  Nor does it feel like the kind of data that needs to be in a cookie. 

With EDD's strong AJAX coverage, I'd think there would be plenty of ways to avoid this entirely. 

(I'm not sure if EDD uses PHP sessions, but if so the same would apply for first page requests that cause a new session. That's harder to solve though.)

pippinsplugins:
This is a bug and is unintentional.

pippinsplugins:
We definitely need to fix it.

cklosowski:
With the PR attached here ( #3197 ) we'll have to do some good testing. @clifgriffin can you submit this to the `release/2.4` branch instead of master? With 2.3 just going out we're trying to keep merges into master at a minimum as to allow for quick bug fixes.

pippinsplugins:
Things to test before merge:
- [ ] WPEngine
- [ ] Pagely
- [ ] W3 Total Cache
- [ ] WP Super Cache
- [ ] WP Rocket

pippinsplugins:
We have merged in for testing but all cases above must be tested before closing this.

cklosowski:
I've tested GoDaddy Managed and it's working fine on there.

pippinsplugins:
Works great on Pagely.

cklosowski:
If someone has a WPEngine account they can test with, I'm having issues with mine showing empty carts, on `master` even.

pippinsplugins:
I'll be able to test on mine.

pippinsplugins:
Testing on WPEngine now.

pippinsplugins:
It works on my localhost with WP Super Cache.

mindctrl:
Setting up for W3TC testing now.

pippinsplugins:
It's failing on WPEngine for me.

clifgriffin:
[...]

PR #3289, opened 2015-03-20
Cache busting cookies: edd_items_in_cart

clifgriffin:
Resubmit against `release/2.4` of PR #3172 

I've started using fastcgi caching with Nginx over caching plugins.  The down downside of this is that Nginx will not cache any page or serve a cached page for any request with a Set-Cookie, and unlike Varnish, you can't exclude cookies from the cache.

I've noticed that every page request on sites with EDD includes a Set-Cookie header from EDD in the form of `edd_items_in_cart`.  It defaults to `deleted`, but if any items are in the cart it is set to that number.

I can't figure out why it would be necessary to set this cookie on every single request, especially when a user hasn't initiated any shopping actions.  Nor does it feel like the kind of data that needs to be in a cookie. 

With EDD's strong AJAX coverage, I'd think there would be plenty of ways to avoid this entirely. 

(I'm not sure if EDD uses PHP sessions, but if so the same would apply for first page requests that cause a new session. That's harder to solve though.)

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -198,7 +198,9 @@ class EDD_Session {
 			if( $set ) {
 				@setcookie( 'edd_items_in_cart', '1', time() + 30 * 60, COOKIEPATH, COOKIE_DOMAIN, false );
 			} else {
-				@setcookie( 'edd_items_in_cart', '', time() - 3600, COOKIEPATH, COOKIE_DOMAIN, false );
+				if ( isset($_COOKIE['edd_items_in_cart']) ) {
+					@setcookie( 'edd_items_in_cart', '', time() - 3600, COOKIEPATH, COOKIE_DOMAIN, false );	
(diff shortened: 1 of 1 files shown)


pippinsplugins:
We'll need to test this extensively with WP Engine as the cookie was originally setup this way specifically to prevent empty carts on WP Engine.

clifgriffin:
@pippinsplugins Yeah, WP Engine caches all cookies by default (except for the native PHP session cookie).  I've found that even when they exclude a specific cookie from their cache, some visitors still receive the cached version. 

If this technique is a workaround for such a scenario, it would probably be best to make this either an option you enable by adding a constant to your config, or by detecting a WP Engine environment and enabling it magically (probably overkill).

pippinsplugins:
WPE is super easy to detect and we already have functions for doing that,
so we may need to go that route.

Issue #3294, opened 2015-03-20
Add new general logging system


ghost1227:
Currently, we don't have a good method of logging general errors in EDD. Things like gateway and API errors are easily tracked, but what happens when you integrate with a third-party service and that service throws an error? Right now, there's not any easy way of notifying the site operator that there has been a problem. You could always display the error on the frontend (in the case of checkout errors), but usually that kind of error isn't relevant to the end user.

I propose the addition of a new logging system based on the current gateway log structure. The gateway logs are ideal due to the nature of the currently tracked data (identification of the transaction which caused the error, date, time, etc), and given it's already structured to handle multi-type data (strings or arrays being the most commonly returned data). The new system would also necessitate addition of a new function `edd_log_extension_error( <Extension Name>, $error );`.

ghost1227:
@pippinsplugins has suggested using the function `edd_log_error` instead of `edd_log_extension_error`. Given we are currently using `edd_log_gateway_error`, and given this function would be exclusively used by extensions, I think inclusion of the extra word makes sense. Thoughts?

cc: @cklosowski

chriscct7:
I like the idea of an edd_log_extension_error function (even more if we have an output of all the errors in a tab of the tools section, with a line for "has extension errors" on the system info)

cklosowski:
I'd actually say we go a step further and `edd_log_error` can take a parameter of what type...`gateway` or `extension`. That can then make it much easier for us to control how logging happens.

pippinsplugins:
I like the idea of adding a new parameter. The `edd_log_gateway_error()` function can just become a wrapper then.

pippinsplugins:
I'm not sure it should be called `Extension Error`. That really doesn't mean much to an end user.

ghost1227:
So tell me what you want me to do :grinning:

pippinsplugins:
I think I'd just call it `edd_log_error`

ghost1227:
@pippinsplugins how about `edd_log_error( $type, $title, $description, $data, $payment_id )` where `$type` is a string identifying 'gateway' or a specific plugin name?

pippinsplugins:
Sure

spencerfinnell:
Should `$payment_id` be its own argument? What if I want to log an error about a failed user creation, failed cart recovery, etc.

cklosowski:
[...]

PR #3295, opened 2015-03-20
First pass at adding general/extension error logging

ghost1227:
Issue #3294

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -697,6 +697,22 @@ function edd_set_upload_dir( $upload ) {
 }
 
 
+/**
+ * Record an extension error
+ *
+ * A simple wrapper function for edd_record_log()
+ *
+ * @since 2.4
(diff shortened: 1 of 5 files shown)


includes/misc-functions.php
```diff
@@ -697,6 +697,22 @@ function edd_set_upload_dir( $upload ) {
...
+ *
+ * A simple wrapper function for edd_record_log()
+ *
+ * @since 2.4
+ * @param string $title Title of the log entry (default: empty)
+ * @param string $message Message to store in the log entry (default: empty)
+ * @param int $parent Parent log entry (default: 0)
+ * @return int ID of the new log entry
+ */
+function edd_record_extension_error( $title = '', $message = '', $parent = 0 ) {
```
chriscct7: We should take in a 4th argument for $type, so if an extensions wants they can send in a display ready name of the extension, which could be used to make tabs showing errors for just that extension, and EDD could slug it for them.

PR #3299, opened 2015-03-22
Adding support for Nepali Currency - NPR

wenthemes:
This pull request will add Nepali currency (NPR) with its currency symbol. Test function is updated now.


chriscct7:
Our policy for currencies is to only list those that are supported by our core gateway, PayPal standard. I would recommend releasing this as a simple extension on WordPress.org (just use a filter to add it), instead.

PR #3307, opened 2015-03-26
Checks for discount property to avoid index not found notice.

ericnicolaas:
Minor fix to avoid a index not found notice.


pippinsplugins:
Thanks!

Issue #3312, opened 2015-03-28
Prevent New Download Drafts by introducing a `create` method.


dancameron:
Here's the backtrace on one of the incidents...

```
[28-Mar-2015 00:48:02 UTC] backtrace: Array
(
    [0] => EDD_Download->__construct
    [1] => edd_is_bundled_product
    [2] => EDD_Software_Licensing->activate_license
    [3] => EDD_Software_Licensing->remote_license_activation
    [4] => call_user_func_array
    [5] => do_action('edd_activate_license')
    [6] => edd_get_actions
    [7] => call_user_func_array
    [8] => do_action('init')
    [9] => require_once('wp-settings.php')
    [10] => require_once('wp-config.php')
    [11] => require_once('wp-load.php')
    [12] => require('wp-blog-header.php')
)
```

Looks like edd_is_bundled_product doesn't sanity check the id and a new download is created just by passing any string/int.

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/99f7940daa4d3e009e801bd1f20d7c4b9b3c7259/includes/download-functions.php#L487

I'll create a PR.

dancameron:
Added Pull Request 3313 to address this specific issue, however...

I think this is a rather big problems:

1. There are  a lot of functions that could cause a drafted download, it's why [this issue](https://easydigitaldownloads.com/support/topic/new-drafts-of-new-download-product-created-automatically/page/2/) has been around for so long. 
2. The solution I used for that function isn't idea because you don't want to force a query on every function call. Although it's very likely the $post object is cached at that point.

I'll look over the __construct and see if I can think of a better solution since that's where the focus should be.

cklosowski:
I think we should use this issue as the basis for making a `create` method within the `EDD_Download` class, and then remove the `false` to create a draft part. This leaves us with an EXPLICIT call to create a new download, instead of automatically doing it when the conditions are right.

cklosowski:
@pippinsplugins ^^^

pippinsplugins:
I agree @cklosowski

pippinsplugins:
@dancameron I like the start you have in #3318.

One thing we really need to support is adding prices, files, etc. We could do this either by allowing additional arguments to be passed to the `create()` method, or we could do it similar to how `EDD_Customer` works and introduce methods for `set_price()`, `set_files()`, etc.

I'm in favor of the second option.

[...]

PR #3313, opened 2015-03-28
Add Sanity Checks for some functions

dancameron:
Check to make sure the download_id passed to the bundled functions are actual downloads, otherwise a new download will be created/drafted if the id is not passed, invalid, or not a real download.


cklosowski:
If it's not of the `download` post type we'll be ok as we have the checks here in the construct of the `EDD_Download` class:
```
		if( ! is_a( $download, 'WP_Post' ) ) {
			return false;
		}

		if( 'download' !== $download->post_type ) {
			return false;
		}
```

I could see changing it to making sure it's not empty so we're not creating drafts, but I think it'd be redundant to check it's post type twice.

Thoughts @pippinsplugins?

dancameron:
Chris please refer to #3312 for the history. But....

I do get your point though and I agree these are a bit of bandaids to fix the issue that's arising (for me and others). 

Maybe we should talk about the conditional `false === $_id` in the __construct. I don't know why EDD should accept FALSE to create a drafted post. I had thought originally (because I skimmed the method) that a draft was created to prevent returning a FAIL instead of a graceful dummy object...now that I see FALSE being returned on a bunch of different conditions I don't understand the circumstances why a draft should ever be created.

Thanks!

cklosowski:
I agree with you on the faults creating a new draft @dancameron. I've talked to Pippin about that and I'm hoping we can try and change that

dancameron:
🍻




—
Dan Cameron
Sprout Venture Inc. & Sprout Apps

pippinsplugins:
The post type checks should definitely stay within the `EDD_Download` class.

I'd like to go ahead and see `EDD_Download` get a `create()` method, just like we did with the `EDD_Customer` class.

dancameron:
@pippinsplugins Thanks but for the sake of clarity does that mean a draft will not be created if "FALSE" is passed as the id? 

Cause the real issue isn't that the post_type checks are in the construct, it's that for some reason `false` is being passed and drafts are being created.
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/99f7940daa4d3e009e801bd1f20d7c4b9b3c7259/includes/class-edd-download.php#L140

pippinsplugins:
Yes, drafts will not be created when passing `false` in.

pippinsplugins:
At least that's my initial thought.

dancameron:
Closing PR so that another could be used to address the direction discussed.

PR #3314, opened 2015-03-28
Dulicate variable declaration; no need for this.

ericnicolaas:
Noticed that this line is exactly the same as line 225. :)

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -228,7 +228,6 @@ function edd_purchase_variable_pricing( $download_id = 0, $args = array() ) {
 		return;
 	}
 
-	$prices = apply_filters( 'edd_purchase_variable_prices', edd_get_variable_prices( $download_id ), $download_id );
 	$type   = edd_single_price_option_mode( $download_id ) ? 'checkbox' : 'radio';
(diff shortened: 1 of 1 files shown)


cklosowski:
Nice Catch. Some feedback,

As long as we're fixing this...I think this would be better if it was 2 checks:

1) Check if it has variable pricing, if not return
2) Since we know we have variable pricing, pull the prices, and then do the `false !== $args['price_id'] && isset( $prices[$args['price_id']]` checks, and return if that condition is met.

that way we are not doing an unnecessary lookup on prices if the download doesn't even have variable pricing enabled.

does that make sense @ericnicolaas ? If you just update the `Studio164a:duplicate-prices` branch the PR should update itself and we can review again.

ericnicolaas:
Updated. Took me a bit to figure out you're returning if the price ID *is* set. 

It seems a little strange that when passing an invalid price ID (i.e. one that doesn't exist), the fallback behaviour is to show all the prices. In cases like this, I would think that no prices should be displayed at all, possibly with a `_doing_it_wrong` notice. 

Just my 2 cents :)

pippinsplugins:
Looks good to me, merging.

PR #3318, opened 2015-03-30
New drafts of New Download Product Created Automatically

dancameron:
Here's a start of a create method for the EDD_Download class.

I can't find anywhere that false or null would be intentionally passed when instantiating the download so I'm not sure if a conditional in the `__construct` is necessary.

#3312


dancameron:
This related to PR #3313

pippinsplugins:
This is looking good. I'm going to post a few thoughts on the issue.

cklosowski:
Closing this PR out as we're merging it into `issue/3312` to wrap up development.

Thanks @dancameron this is looking like it'll work perfectly

PR #3326, opened 2015-04-02
Ensure meta data contains tax key. Fixes #3324. 

### LINKED ISSUE #3324: Taxes backwards comparability script "loses" tax data
Apologies if this bug has been reported; I couldn't see anything in the 'bug' issues.

I have a script which effectively updates payment meta (it's collecting evidence for VATMOSS):

    $current_meta = edd_get_payment_meta( $payment_id );
    $current_meta['user_info']['psp_country'] = $paypal_cert_country; //Sets country of PayPal account
    edd_update_payment_meta( $payment_id, '_edd_payment_meta', $current_meta );

After this script runs, however, the `_edd_payment_tax` meta key field is erroneously set to 0. The tax amount in the serialized field is the correct amount. Here's a 'beautified' copy of that data:

```
Array(
     [currency] => GBP
     [downloads] => Array( ... )
     [user_info] => Array(...)
     [cart_details] => Array(
          [0] => Array(
               [name] => Sandbox Product
               [id] => 7174
               [item_number] => Array(
                    [id] => 7174
                    [options] => Array(...)
                    [quantity] => 1
               )
               [item_price] => 120
               [quantity] => 1
               [discount] => 0
               [subtotal] => 120
               [tax] => 24
               [fees] => Array(... )
               [price] => 144
          )
     )
     [key] => ...
     [email] => ...
     [date] => 2015-04-01 20:20:24
)
```

Whenever `_edd_payment_meta` or `_edd_payment_tax` meta values are updated, the other is also updated to keep them sync (`edd_update_payment_backwards_compat()`). However the part of that routine which runs when `_edd_payment_meta` is updated appears to assume that `$metadata['tax']` is populated (see https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.3.5/includes/payments/actions.php#L297-L298). 

As demonstrated above, this isn't necessarily the case. And indeed `edd_insert_payment()` seems to check both `tax` and `cart_details` (see https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.3.5/includes/payments/functions.php#L152-L158)

Because `$metadata['tax']` is not populated, the `_edd_payment_tax` meta value is set to 0.

Happy to submit patch, just wanted to double check this is a bug.

stephenharris:
Fixes bug with payment meta/tax backwards compatability code. Adds unit test. Fixes #3324.

Passes all tax unit tests. However I was only able to test 60% of all unit tests before I encountered an error (unrelated to this pull request):

 > PHP Fatal error:  Call to undefined method Tests_Activation::assertNotFalse() in /var/www/html/plugins/edd/tests/tests-install.php on line 76

(Seems the phpunit version set by `composer.json` is out of date? I think it requires 4.4+)

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -161,7 +161,8 @@ function edd_insert_payment( $payment_data = array() ) {
 			'currency'     => $payment_data['currency'],
 			'downloads'    => $payment_data['downloads'],
 			'user_info'    => $payment_data['user_info'],
-			'cart_details' => $payment_data['cart_details']
+			'cart_details' => $payment_data['cart_details'],
+			'tax'          => $cart_tax,
 		);
 
 		$mode    = edd_is_test_mode() ? 'test' : 'live';
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -162,7 +162,6 @@ function edd_insert_payment( $payment_data = array() ) {
 			'downloads'    => $payment_data['downloads'],
 			'user_info'    => $payment_data['user_info'],
 			'cart_details' => $payment_data['cart_details'],
-			'tax'          => $cart_tax,
 		);
 
 		$mode    = edd_is_test_mode() ? 'test' : 'live';
--- a/tests/tests-tax.php
+++ b/tests/tests-tax.php
@@ -265,5 +265,18 @@ class Tests_Taxes extends WP_UnitTestCase {
 	public function test_get_payment_tax() {
 		$this->assertEquals( '0.36', edd_get_payment_tax( $this->_payment_id ) );
 	}
+	
+	/**
+	 * @see https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3324
+	 */
+	public function test_backwards_compat_bug(){
+		
+		$this->assertEquals( '0.36', edd_get_payment_tax( $this->_payment_id ) );
+		
+		$current_meta = edd_get_payment_meta( $this->_payment_id );
+		edd_update_payment_meta( $this->_payment_id, '_edd_payment_meta', $current_meta );
+		
+		$this->assertEquals( '0.36', edd_get_payment_tax( $this->_payment_id ) );
+	}
 
 }
--- a/includes/payments/actions.php
+++ b/includes/payments/actions.php
@@ -295,7 +295,12 @@ function edd_update_payment_backwards_compat( $meta_id, $object_id, $meta_key, $
 
 		case '_edd_payment_meta':
 			$meta_value   = maybe_unserialize( $meta_value );
-			$tax_value    = ! empty( $meta_value['tax'] ) ? $meta_value['tax'] : 0;
+			
+			if( !isset( $meta_value['tax'] ) ){
+				return;
+			}
+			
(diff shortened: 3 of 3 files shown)


includes/payments/functions.php
```diff
@@ -161,7 +161,8 @@ function edd_insert_payment( $payment_data = array() ) {
 			'currency'     => $payment_data['currency'],
 			'downloads'    => $payment_data['downloads'],
 			'user_info'    => $payment_data['user_info'],
-			'cart_details' => $payment_data['cart_details']
+			'cart_details' => $payment_data['cart_details'],
+			'tax'          => $cart_tax,
```
pippinsplugins: This was intentionally removed so that we can cut down on the amount of meta data stored. It's stored in its own key now.

Looks like we need to fix our backwards compat rather than re-store this.

stephenharris: Good point. I've updated the PR so to update the backwards compat function. It now only copies tax in `_edd_payment_meta` to `_edd_payment_tax` if tax exists (rather than defaulting it to 0). I think that makes more sense than the original patch.

includes/payments/functions.php
```diff
@@ -161,7 +161,7 @@ function edd_insert_payment( $payment_data = array() ) {
 			'currency'     => $payment_data['currency'],
 			'downloads'    => $payment_data['downloads'],
 			'user_info'    => $payment_data['user_info'],
-			'cart_details' => $payment_data['cart_details']
+			'cart_details' => $payment_data['cart_details'],
```
cklosowski: remove the trailing `,` please

tests/tests-tax.php
```diff
@@ -265,5 +265,18 @@ public function test_download_is_exclusive_of_tax() {
 	public function test_get_payment_tax() {
 		$this->assertEquals( '0.36', edd_get_payment_tax( $this->_payment_id ) );
 	}
+
```
cklosowski: This test is redundant to what we're already running here:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/tests/tests-payments.php#L214

stephenharris: No it isn't - the unit test there is setting a value for `$current_meta['tax']`. If you do not set a value, it defaults to 0 and over-rides `_edd_payment_tax` value, and by default the tax key of `_edd_payment_meta` is not set.

includes/payments/functions.php
stephenharris: @cklosowski I thought EDD followed [WP coding standards](https://make.wordpress.org/core/handbook/coding-standards/php/#indentation), so I left the trailing coming when reverting the first commit.

tests/tests-tax.php
cklosowski: Ok, can you please move it to the area where the rest of the unit tests for testing payment meta (tax) are located?

includes/payments/functions.php
cklosowski: I stand corrected, sorry didn't know that one was added to their list. Carry on.

cklosowski:
going to merge and make the update.

stephenharris:
Thanks!

PR #3329, opened 2015-04-04
Update widget-cart-item.php

brashrebel:
Closing `<span>` tag needed for `edd-cart-item-quantity` element. Discovered by a customer in HS ticket #7754.


pippinsplugins:
Thanks!

PR #3352, opened 2015-04-17
Add filter for join clause in edd_count_payments function

### LINKED ISSUE #3344: Provide filter for both `where` and `join` in `edd_count_payments` function
Right now the `edd_count_payments` function provides filter only for the `where` function and not for the `join` clause. Relevant code.

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/payments/functions.php#L506

`where` clause without `join` will be useless in most scenarios when we filter things in the payment history screen.

Example: I need this filter payments by campaign in my EDD Campaign Tracker extension

We have two options.

- Use a single filter, but pass two parameters. In that case the name of the filter should be changed.
- Use two filters.

I am ready to send a pull request once a decision is made on one of the above two options.

sudar:
Fix #3344

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -504,6 +504,7 @@ function edd_count_payments( $args = array() ) {
 	}
 
 	$where = apply_filters( 'edd_count_payments_where', $where );
+	$join  = apply_filters( 'edd_count_payments_join', $join );
 
 	$cache_key = md5( implode( '|', $args ) . $where );
 


sudar:
The Travis build is failing because there was some problem in downloading WordPress when it tried, not because there is some issue with the code.

Is there a way to restart the travis build?

sunnyratilal:
@sudar I've restarted the build for you.

sudar:
Thanks @sunnyratilal

All is well now :)

PR #3354, opened 2015-04-20
Rename "Cyprus Island" to just "Cyprus".

dannyvankooten:
There's no "Cyprus Island", just "Cyprus". :)

### DIFF
--- a/includes/country-functions.php
+++ b/includes/country-functions.php
@@ -199,7 +199,7 @@ function edd_get_country_list() {
 		'HR' => 'Croatia/Hrvatska',
 		'CU' => 'Cuba',
 		'CW' => 'Cura&Ccedil;ao',
-		'CY' => 'Cyprus Island',
+		'CY' => 'Cyprus',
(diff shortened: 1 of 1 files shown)


chriscct7:
Technically it's the "Republic of Cyprus" officially. I think we should rename it to that

dannyvankooten:
Sorry but I definitely disagree. That would mean we'd have to change all countries as A LOT of them are republics, technically speaking. 

I think we should go with the most commonly used names. The list is fine, really, it's just that "Cyprus Island" [is not a thing](http://europa.eu/about-eu/countries/member-countries/).

pippinsplugins:
Considering it's listed as `Cyprus` [here](http://europa.eu/about-eu/countries/member-countries/), I 'd say we leave it at that. Doesn't get much more official :)

pippinsplugins:
We'll need to fix the unit tests on this.

cklosowski:
I'm going to merge, and then fix the unit test on this

Issue #3360, opened 2015-04-22
edd_get_file_download_method() called too late during download process


pippinsplugins:
`edd_get_file_download_method()` is called during the file download process in order to determine which file download method to use. This includes a filter that allows extensions to change the method. The problem, however, is that it gets too early for an extension to properly hook into the download process in most cases.

We need to make the call later. Moving it down allows us to resolve https://github.com/easydigitaldownloads/edd-amazon-s3/issues/8.

cklosowski:
:+1: looks good to me. If you want to PR it I'll review/test/and merge

pippinsplugins:
I need to work on it a bit more. I haven't come up with a way to fix the S3 issue yet.

If you have any ideas, post them to S3 repo.

PR #3370, opened 2015-04-27
Issue/3360

pippinsplugins:
#3360


cklosowski:
Tested with Amazon S3 update and is working as expected. S3 items delivered via `redirect` even with `forced` is the selected deliver method.

Issue #3367, opened 2015-04-27
Race condition with edd_get_next_payment_number()


pippinsplugins:
Through two separate support tickets, I've discovered that `edd_get_next_payment_number()` is susceptible to race conditions.

We use `EDD_Payments_Query` to determine the next payment number by pulling up the latest payment number.

If a site is processing a large number of payments, has a strong database cache, or a combination of both, it's possible for two payments to get assigned the same number since the number returned by `EDD_Payments_Query` is actually one (or more) numbers behind the actual latest.

To resolve this, we should look at making the query more efficient. I suspect we could resolve this by getting rid of `EDD_Payments_Query` and instead using a direct query on the `wp_postmeta` table to find the row with a key of `_edd_payment_number` and grabbing just the most recent row.

Tickets: 
https://easydigitaldownloads.com/support/topic/duplicate-invoice-numbering/
https://easydigitaldownloads.com/support/topic/urgent-duplicated-payment-ids/

chriscct7:
What if we just stored the last number as an option?

pippinsplugins:
That could work though we'd still run into potential race conditions there.
Definitely faster than querying wp_postmeta though!

chriscct7:
My thoughts exactly

cklosowski:
I agree, this should reduce it enough. Not remove it entirely, but it's a good start.

cklosowski:
Ok @pippinsplugins I think this should help avoid the race condition.

This allows a backwards compatibility check and if the option doesn't exist, creates it from the last payment, or if no past payments exist, starts at the `$start` value.

chriscct7:
@cklosowski see the comments on that pr

PR #3375, opened 2015-04-29
Reduce chances of race condition in getting next payment number

cklosowski:
Helps address #3367

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -213,6 +213,7 @@ function edd_insert_payment( $payment_data = array() ) {
 
 		if( edd_get_option( 'enable_sequential' ) ) {
 			edd_update_payment_meta( $payment, '_edd_payment_number', $number );
+			update_option( 'edd_last_payment_number', $number );
 		}
 
 		// Clear the user's purchased cache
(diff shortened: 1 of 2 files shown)


includes/payments/functions.php
```diff
@@ -1225,6 +1224,34 @@ function edd_get_next_payment_number() {
...
+	// Ensure it's a whole number
+	$number = intval( $number );
+
+	// Increment the payment number
+	$number++;
+
+	// Re-add the prefix and postfix
+	$number = $prefix . $number . $postfix;
+
+	return $number;
```
chriscct7: There should be a filter here, so extensions can come in and say not to increment. Specific use case that comes to mind is in the EU, some countries cannot increment numbers on pending, they have to do it on completed. Not sure how the extension would work, but a filter can't kill us

includes/payments/functions.php
```diff
@@ -1225,6 +1224,34 @@ function edd_get_next_payment_number() {
...
+ * @since  2.4
+ * @param  string $prefix  The Payment Number Prefix
+ * @param  string $number  The formatted Current Number to increment
+ * @param  string $postfix The Payment Number Postfix
+ * @return string          The new formatted Payment number
+ */
+function edd_increment_payment_number( $prefix, $number, $postfix ) {
+
+	// Remove prefix and postfix
+	$number = str_replace( $prefix, '', $number );
```
chriscct7: This seems like a bad idea. What if your prefix is "1", and the current order number is 11?

pippinsplugins: I don't think the prefix or postfix should be passed into this function.

Let's say a user decides to change their prefix or postfix after making some sales. The new prefix and postfix wouldn't be used after 2 more sales had been recored, since the first one would pull from the options table.

I vote we only store the number itself and then apply the prefix and post fix in `edd_get_next_payment_number()`.

chriscct7: +1

cklosowski: You are both forgetting we've used this method before. I simply extracted it from the function.

pippinsplugins: It's a change of scope for the ticket, but I really don't think changing prefixes should break anything. Let's fix both issues.

cklosowski: Ok @chriscct7 yes, changing the prefix breaks the current functionality. But @pippinsplugins is right, we need to try and fix both the race condition and the prefix/postfix change issue. I'm going to close this PR, rewrite what I can, and we'll go from there.

PR #3377, opened 2015-04-29
Improve sequential order numbers performance

cklosowski:
This addresses the concerns in #3367 by reducing the risk of a race condition as well as fully supporting changing the prefix/postfix.


pippinsplugins:
Works great!

PR #3384, opened 2015-05-04
edd_is_admin_page updates see #2657

chriscct7:
Implements `edd_is_admin` as defined in #2657

Supports new customer management ui and reports page

### DIFF
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -46,284 +46,292 @@ add_action( 'admin_menu', 'edd_add_options_link', 10 );
 
 /**
  *  Determines whether the current admin page is a specific EDD admin page.
- *  
- *  Only works after the `wp_loaded` hook, & most effective 
+ *
+ *  Only works after the `wp_loaded` hook, & most effective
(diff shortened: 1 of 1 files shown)


includes/admin/admin-pages.php
```diff
@@ -45,37 +45,309 @@ function edd_add_options_link() {
...
+					$found = false;
+				}
+			} elseif ( in_array( $pagenow, $admin_pages ) ) {
+				$found = true;
+			}
+			break;
 	}
 
-	return (bool) apply_filters( 'edd_is_admin_page', $ret );
+	return (bool) apply_filters( 'edd_is_admin_page', $found, $page, $view );
```
pippinsplugins: `$passed_page` and `passed_view` need to be accessible here.

pippinsplugins:
Something in here is failing.

Go to the main Downloads screen (post_type=download) and then run `var_dump( edd_is_admin_page() );`.

It will return `false`. It should be true.

cklosowski:
@pippinsplugins looking

pippinsplugins:
Nevermind, it's working. I forgot that Debug Bar Console fires through ajax so doesn't return expected results.

Issue #3386, opened 2015-05-04
Add item_id support to EDD_License


pippinsplugins:
The EDD_License handler currently passes `item_name` to the Software Licensing API.

This limits us on the ability to change Download product names on the EDD site, since the name passed to `EDD_License` is used to look up the license key.

We need to add support for passing the product ID (as `item_id`) so that we can start updating extensions.

Two extensions (as examples) that need updated:

- Wysija needs to be updated to MailPoet (formerly Wysija)
- Mail Chimp needs to be updated to MailChimp

The product names can't be changed until a few weeks after an update is pushed out with item_id support.

cklosowski:
:+1: Been thinking about this a lot lately.

PR #3394, opened 2015-05-09
Add support for item_id in EDD_License. #3386

pippinsplugins:
#3386


cklosowski:
As long as we're aware that this makes it so the download name cannot be a number, then I'm good with this.

pippinsplugins:
I thought of that but can't think of a single scenario where we would sell an official extension with a name like that.

cklosowski:
Agreed :+1:

Issue #3337, opened 2015-04-08
Version EDD API


chriscct7:
We should version our API.

Right now we use site.com/edd-api/{endpoints}
If we move to a system like site.com/edd-api/v{number}/{endpoints} we can improve our api iteratively without needing to worry about breaking things relying on our api (for example #3338 ).

We should alias the current api to site.com/edd-api/v1/{endpoints} and come up with a good system so that extensions and core can register the same endpoints across multiple versions easily.

The sooner this is in the better. Not only can EDD core use it, but also extensions wishing to register an action to a specific version endpoint could utilize it.

chriscct7:
Actually, now that I think of it, revisions might be better.

The problem with versions is it becomes a big problem quickly to iterate or add new methods. Like if #3338 becomes the only change we do for 2.4, does that constitute version 2 of the api? I don't think it does but its a needed improvement that would get delayed till we have enough to push as a v2. Instead I think we could do something like revision numbers, so specific functionality is available from a specific revision forward. Maybe for example, instead of v1, v2, etc, we use EDD's version number. Thoughts?

cklosowski:
Let me think on the `Version vs Revision` decision. There are pros and cons to each side of these.

chriscct7:
Yeah. Let's talk it over at Loop

chriscct7:
A third option would be instead of doing global numbers, we could version each endpoint.

So like /edd-api/v1/sales would pull version 1 of sales
           /edd-api/v2/earning would pull version 2 of earnings

That might be a really appealing option. It lets us version each endpoint without an extension having to wait for an EDD core release to go out, and it lets us version each endpoint individually.

chriscct7:
If we did that, we don't even necessarily have to do /edd-api/{num}/{endpoint}

We could do /edd-api/sales/?version=2 where version tells you which version of the api to use. By default it always could use the latest version, and then people could use that query arg to change to an older version and set it there.

cklosowski:
I side we table this for an in-person discussion we have at Loop.

pippinsplugins:
Agreed, let's talk on it more in person.

@chriscct7 Note, please don't add `Needs Unit Tests` to an issue until it actually has code written for it. A unit test without the code is silly.

chriscct7:
Fair enough

pippinsplugins:
We'll need to work with @mps to get the EDD iOS app updated at the same time as this goes out, or shortly afterwards.

[...]

Issue #3407, opened 2015-05-15
Improvements to Products Endpoint in API


cklosowski:
The products endpoint in the EDD API has quite a bit of duplicated code, and no places to filter the data returned for each product, making it impossible for extensions liked Software Licensing to return information in the Products Endpoint.

pippinsplugins:
Anything else you'd like to change with it?

cklosowski:
Not at this time, i'm good with it as is.

PR #3411, opened 2015-05-15
Issue/3337 - API Versionsing

pippinsplugins:
#3337


chriscct7:
@pippinsplugins looks like there's some merge conflict now. Will test when I get a chance this week.

includes/api/class-edd-api-v1.php
```diff
@@ -0,0 +1,24 @@
...
+ * @copyright   Copyright (c) 2015, Pippin Williamson
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       2.4
+ */
+
+// Exit if accessed directly
+if ( ! defined( 'ABSPATH' ) ) exit;
+
+/**
+ * EDD_API_V13 Class
```
chriscct7: Shouldn't this be: EDD_API_V1?

includes/api/class-edd-api.php
```diff
@@ -92,7 +92,43 @@ class EDD_API {
...
+	private $queried_version;
+
+	/**
+	 * All versions of the API
+	 *
+	 * @var string
+	 * @access public
+	 * @since 2.4
+	 */
+	protected $versions = array();
```
chriscct7: if this is gonna be protected, there needs to be a getter for this. Some extensions might want to use this

includes/api/class-edd-api.php
```diff
@@ -92,7 +92,43 @@ class EDD_API {
...
+	public $override = true;
+
+	/**
+	 * Version of the API queried
+	 *
+	 * @var string
+	 * @access public
+	 * @since 2.4
+	 */
+	private $queried_version;
```
chriscct7: Same for this. Particularly useful if you wanted to say write an extension that logs api requests outside of the edd core one.

includes/api/class-edd-api.php
```diff
@@ -163,6 +209,65 @@ public function query_vars( $vars ) {
...
+	 * @return string
+	 */
+	private function get_default_version() {
+
+		$version = get_option( 'edd_default_api_version' );
+
+		if( defined( 'EDD_API_VERSION' ) ) {
+			$version = EDD_API_VERSION;
+		} elseif( ! $version ) {
+			$version = 'v1';
```
chriscct7: this should be 1, not v1 though right? We should define the version as a number, even if we display it as v{num}

includes/api/class-edd-api.php
```diff
@@ -304,33 +416,41 @@ private function invalid_key() {
 	 * @return void
 	 */
 	public function process_query() {
+
 		global $wp_query;
 
 		// Check for edd-api var. Get out if not present
-		if ( ! isset( $wp_query->query_vars['edd-api'] ) )
+		if ( ! isset( $wp_query->query_vars['edd-api'] ) ) {
```
chriscct7: could be empty for slighly better performance if there's a set ?edd-api= where the = is empty string that would cut it off early before running all the validity checks

includes/api/class-edd-api-v1.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * Easy Digital Downloads API V13
```
chriscct7: Same here

includes/api/class-edd-api.php
```diff
@@ -101,6 +137,15 @@ class EDD_API {
 	 * @since 1.5
 	 */
 	public function __construct() {
+
+		$this->versions = array(
+			'v1' => 'EDD_API_V1',
```
chriscct7: versions should be defined with just the number. The "v" part should only be for stylistic purposes added to the endpoint at the end. API endpoint versions should always only be defined by an integer number >=1 This prevents future issues and allows people to rename the api endpoints if they wanted to do that.

includes/api/class-edd-api.php
```diff
@@ -101,6 +137,15 @@ class EDD_API {
...
+		);
+
+		foreach( $this->versions as $version => $class ) {
+			require_once EDD_PLUGIN_DIR . 'includes/api/class-edd-api-' . $version . '.php';
```
chriscct7: This should do a file_exists before trying this. Bad things happen if hosts delete files or if we mess up for a release.

includes/api/class-edd-api.php
pippinsplugins: This is the map from URL to class, so the `v` is correct since you'r do /edd/v1/, not /edd/1/

includes/api/class-edd-api.php
pippinsplugins: Considered that but it's faster to not do it, and this is no different that our main EDD includes. This files are not filterable so the only reason one would disappear is if it is deleted, which would result in bad things even with the check.

includes/api/class-edd-api-v1.php
pippinsplugins: Yes.

pippinsplugins:
Merge conflict due to the recent filters for filtering product data: #3407

includes/api/class-edd-api.php
chriscct7: @pippinsplugins this still needs to be addressed

pippinsplugins: No, see my comments on the main PR. Having the V is correct.

Issue #3416, opened 2015-05-18
EDD Log records do not correctly record post_date_gmt


ericnicolaas:
When a log is inserted for a payment, the `post_date_gmt` is not set correctly; rather, it is the same as `post_date`. A by-product of this is that it can result in a `post_status` of `future` for the log. 

Here's a database snapshot from a test payment: 

![screen shot 2015-05-18 at 12 51 52 pm](https://cloud.githubusercontent.com/assets/676582/7673999/fbd585c2-fd5c-11e4-988e-327b910f925f.png)

Pull request forthcoming. 

Eric

PR #3417, opened 2015-05-18
Set post date GMT for EDD Log

ericnicolaas:
#3416 

Correctly sets the GMT for the edd log. See below record of test payment:

screen shot 2015-05-18 at 1 01 59 pm


pippinsplugins:
Looks good to me.

PR #3419, opened 2015-05-19
Add filter on cart item quantity ajax event.

ericnicolaas:
Just adding a filter that is useful if you're doing anything in the checkout that needs to respond to changes in the cart. I'm working on an application that could really benefit from this. My current workaround is to fire off a second AJAX request after this one returns, which results in an odd delay and is certainly far from ideal.

Happy to explain in further detail if you want a better understanding of the use case.


pippinsplugins:
I like this.

@cklosowski any objections or suggestions?

cklosowski:
I like it as well :+1:

PR #3424, opened 2015-05-21
Provide $price to callback functions

ericnicolaas:
It can be helpful to pass `$price` to the `edd_purchase_link_args` filter, particularly since the price is used to modify the button text in certain cases. This adds it as an optional second argument to the callback functions. 

It also cleans up the conditional check on line 127 since `$price` is always defined; it just needs to check that it isn't `false`.


pippinsplugins:
Price is included inside of args. Why not just access it there?

ericnicolaas:
The price in args is used to determine whether to show the price; it doesn't store the price. `$price` on the other hand is the calculated price for the product. Note below the var_dump output of `$args` and `$price`: 
screen shot 2015-05-22 at 11 16 10 am

cklosowski:
Yeah the actual price isn't passed in `$args`. I'm trying to see if there's a better solution to including this in the `$args`. Problem is we'd have to name it something other than `price` since that's used as a `bool` of it to show the price.

Once we have a price could we send it as maybe `$args['display_price']`?

pippinsplugins:
I'm :+1: for `display_price`

pippinsplugins:
@cklosowski @ericnicolaas thoughts on `display_price`?

ericnicolaas:
@pippinsplugins @cklosowski `display_price` sounds good to me. I'll work on an update to the PR.

pippinsplugins:
:+1:

ericnicolaas:
Looks like somehow I've managed to add a whole stack of other commits into this PR. Happy to just start with a fresh PR if you'd prefer. Otherwise it looks like it's mostly your stuff anyway, @pippinsplugins

cklosowski:
@ericnicolaas yeah if you could just resubmit it. I'd appreciate it, will keep the merge history cleaner. I'll get an issue written up so we can track it as well.

cklosowski:
New issue created in #3849. If you could submit a new PR against `release/2.5` that'd be prefect. I'll slate the issue to be fixed there.

PR #3430, opened 2015-05-23
Remove edd_version option on uninstall

DevinWalker:
I noticed this option was being left behind if the user chooses to do a clean uninstall. I couldn't find any reason it's not being deleted.


pippinsplugins:
Thanks1

Issue #3163, opened 2015-02-19
Tools tabs display poorly on mobile


pippinsplugins:
As shown [here](https://s3.amazonaws.com/f.cl.ly/items/2G0210102T0w3I45210T/IMG_0077.jpg), the tools tab displays poorly on small screens. Let's fix that.

ghost1227:
Any recommendations?

chriscct7:
When you fix this, can you please @chriscct7 me. Gotta fix it in another extension as well

pippinsplugins:
:+1:

pippinsplugins:
Improvement for tabs:
![screen shot 2015-05-25 at 1 49 00 pm](https://cloud.githubusercontent.com/assets/1034109/7801143/c038dbb6-02e5-11e5-8930-a9130dd85ec1.png)

pippinsplugins:
![screen shot 2015-05-25 at 2 12 45 pm](https://cloud.githubusercontent.com/assets/1034109/7801298/261460a2-02e8-11e5-9c33-8a7035dcabb3.png)

pippinsplugins:
Anyone care to review?

ghost1227:
Reviewed... Looks beautiful to me! Everything seems to work well... anything specific beyond testing on various devices for functionality I should be looking for?

pippinsplugins:
@ghost1227 mostly just testing. Also pay attention to general usability and see if you notice any areas that could be improved further.

pippinsplugins:
Closing as done. Re-open if you notice any issues.

PR #3432, opened 2015-05-25
Issue/3163

pippinsplugins:
#3163


chriscct7:
@cklosowski I think it needs more struts. Thoughts?

Issue #3339, opened 2015-04-09
Add shortcode to publish metabox


sumobi:
Here's a nice way of showing the shortcode (From GiveWP, see far right). We should consider doing this for EDD and possibly even go as far as removing the shortcode from the main downloads page. Would make it cleaner.
 
![screen shot 2015-04-09 at 3 25 40 pm](https://cloud.githubusercontent.com/assets/52581/7076118/c331a128-decc-11e4-8c85-76e12d858c94.png)

sumobi:
Or do it like this at least which is much nicer IMO

![screen shot 2015-04-09 at 3 36 32 pm](https://cloud.githubusercontent.com/assets/52581/7076329/3e72afca-dece-11e4-988b-a6a6b6a29f8a.png)

chriscct7:
+1

pippinsplugins:
I really like this.

sunnyratilal:
:+1:

cklosowski:
Yes, I personally would lot to remove it from the list table, or make it hidden until a click. We're wasting a ton of space with that column.

spencerfinnell:
Agreed @cklosowski -- even removing the text/style/color attributes would help. 

I know I'm not the average user but I know I've never used it. Might be better as a quick plugin or code snippet to add it to the columns for those who tend to use it a lot.

ghost1227:
Something like this?

![](http://f.cl.ly/items/1U0K1I2P0E2Z1F0E043S/Screen%20Shot%202015-05-27%20at%202.56.23%20PM.PNG)

spencerfinnell:
Maybe a "View" link to toggle the input? Similar to the Edit/Browse links above it.

ghost1227:
I just copied Andrew's example... but...

ghost1227:
Something like this @spencerfinnell?

![](http://cloud.section214.com/image/471k0w3K0R0t/Screen%20Shot%202015-05-27%20at%203.30.03%20PM.PNG)

![](http://cloud.section214.com/image/1a0Q423S2n1o/Screen%20Shot%202015-05-27%20at%203.30.31%20PM.PNG)

mindctrl:
This is cool, though you could argue a shortcode that displays a purchase link has nothing to do with publishing the content and is unrelated to everything else in that box.

ghost1227:
[...]

PR #3448, opened 2015-05-27
Move purchase shortcode to publish meta box

ghost1227:
Issue #3339


pippinsplugins:
@ghost1227 since it's done already, could you go ahead and submit to release/2.4? We'll put t in early.

PR #3457, opened 2015-05-28
Add more info to tax rate response

dannyvankooten:
This is useful (necessary) for VAT plugins to update the displayed tax rate everytime `edd_recalculate_taxes` is called, without having to perform an unnecessary additional AJAX request.

PR #3458, opened 2015-05-28
Create window.EDD_Checkout object and expose some functions to it.

dannyvankooten:
- Bind `EDD_Checkout` object to window
- Expose `recalculate_taxes` method so other plugins can call it
- Bind events in one place, for a clearer overview and better usage of variables etc.
- Don't use inefficient $('body') 
- Less anonymous functions

Together with #3457 this helps other plugins recalculate & show tax rates based (like after entering a VAT number in a field).

### DIFF
--- a/assets/js/edd-checkout-global.js
+++ b/assets/js/edd-checkout-global.js
@@ -1,336 +1,360 @@
-jQuery(document).ready(function($) {
-    var $body = $('body'),
-		$form = $("#edd_purchase_form"),
-        $edd_cart_amount = $('.edd_cart_amount');
+window.EDD_Checkout = (function($) {
+	'use strict';
 
-    // Update state/province field on checkout page
-    $body.on('change', '#edd_cc_address input.card_state, #edd_cc_address select', function() {
-        var $this = $(this);
-        if( 'card_state' != $this.attr('id') ) {
+	var $body,
+		$form,
+		$edd_cart_amount,
+		before_discount,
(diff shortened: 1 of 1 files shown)


pippinsplugins:
Holy crap I love this so much. Obviously need to extensively test it but this solves a huge problem that I ran into just yesterday.

pippinsplugins:
@cklosowski I'd like extra eyes testing this.

chriscct7:
Plugins that rely on the JS for the checkout page already should be tested as well, just in case

chriscct7:
Off the top of my head, EU VAT + Discounts Pro

dannyvankooten:
To be fair, not much changed in this PR and the chance of breaking other extensions is very close to 0, I'd say 0. 

Nothing changed in the way the code runs, the only thing that changed is the binding of the function to `window` and the exposure of some methods which were previously only available in the scope of the anonymous function itself.

pippinsplugins:
Yep I see that, I just like to be super safe when changing something that will break 100% of everything if it does break :)

cklosowski:
tested this out with a few JS required items on the checkout. Seems to be fine. Mainly tested with Stripe and Discounts Pro. Go ahead and merge if your testing passed @pippinsplugins

pippinsplugins:
Beautiful.

pippinsplugins:
We'll merge as soon as we're ready to put this into `release/2.4`.

Issue #3444, opened 2015-05-27
Export Payment History shows wrong price in Products column when discount applied


mindctrl:
When an item has been purchased with a discount, the export report shows the wrong price in the product column item description. Steps to reproduce:
- Purchase $10 product using a 50% discount code, leaving final sale price at $5.
- Confirm things look right on Payment History screen.
- Go to Reports > Export > Export Payment History and export the payment history.

At this point you'll notice the Products column, which appends the price to the description, has doubled the coupon amount, resulting in the price saying $2.50 instead of $5.

![Screenshot](http://f.cl.ly/items/351U310V1d123p1l1f3B/Image%202015-05-27%20at%202.41.43%20PM.png)

[Reported here](https://easydigitaldownloads.com/support/topic/bug-in-csv-export-of-payment-history/#post-533258).

pippinsplugins:
Let's get this fixed in 2.5.

pippinsplugins:
@mindctrl could you test @ghost1227 's PR?

mindctrl:
@ghost1227 @pippinsplugins still seeing the same issue here

![image](https://cloud.githubusercontent.com/assets/1254870/9476739/5b9a4f60-4b3c-11e5-8e1b-3b48e2d3e976.png)

pippinsplugins:
@ghost1227 want to look at it again?

cklosowski:
Looks like this is specific to % based discounts. The top row is like @mindctrl describes, but the 2nd row is a flat `$10` discount, ant it appears to be applying correctly.

![screen shot 2015-09-28 at 8 46 47 pm](https://cloud.githubusercontent.com/assets/1390124/10154889/2610f2da-6622-11e5-814b-45b572bf2cfe.png)

I'm going to try and tie in #3185 as well as fix that entities issue.

cklosowski:
How do we feel about this @mindctrl?
![screen shot 2015-09-28 at 9 07 30 pm](https://cloud.githubusercontent.com/assets/1390124/10155085/fd30871a-6624-11e5-9339-09e55081e761.png)

Adds the `(X)` for the quantity, uses the sale price, and the sale total stays in tact.

cklosowski:
My initial thought was to do `3 @ $X.XX` but due to rounding, it left a non-equal total when you added up everything (off by a cent here and there)

pippinsplugins:
I like the screenshot above.

mindctrl:
[...]

PR #3464, opened 2015-05-30
Base variable price override on subtotal, not price

ghost1227:
Issue #3444


includes/admin/reporting/class-export-payments.php
```diff
@@ -130,8 +130,8 @@ public function get_data() {
 					$id = isset( $payment_meta['cart_details'] ) ? $download['id'] : $download;
 
 					// If the download has variable prices, override the default price
-					$price_override = isset( $payment_meta['cart_details'] ) ? $download['price'] : null;
-
+					$price_override = isset( $payment_meta['cart_details'] ) ? $download['subtotal'] : null;
```
pippinsplugins: This will need an isset() check on the `subtotal`. Old payment records didn't have that key.

Issue #3485, opened 2015-06-08
Add generic way to prefil checkout fields from session data


pippinsplugins:
We need to have a way to prefill checkout fields from session data to help facilitate plugins that need to set up customer details before reaching the checkout screen.

This is necessary to complete #3223

pippinsplugins:
When logged in, user details aren't always getting pre-filled.

mindctrl:
Getting undefined variables for the address fields. Fails here as it's never empty. https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/8f4ff659e4a6460cc2ac56932430e718e8376544/includes/checkout/template.php#L303

pippinsplugins:
@mindctrl Could you test it again now?

PR #3486, opened 2015-06-08
#3485

pippinsplugins:
#3485


cklosowski:
Do we have some sort of sanitization on this? I'm thinking we should at least run an `array_map` using `sanitize_text_field`.

pippinsplugins:
That's a good idea.

PR #3490, opened 2015-06-10
Remove inline styles

### LINKED ISSUE #3489: Remove inline styles from [downloads] shortcode
I propose we remove the inline styles from the [downloads] shortcode. 

By that I mean the `style="width: <?php echo $column_width; ?>; float: left;">` here:

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/shortcodes.php#L489

CSS should be kept to stylesheets. For theme developers it means filtering the entire shortcode just to remove the hardcoded styles. I for one do not want to use `float: left`. Yes, you can override the float and width properties but I believe it shouldn't be there in the first place and means adding more CSS.

We don't need the `width` CSS rule either either. A CSS class is already applied to the wrapper div which lists the column count so this can be used: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/shortcodes.php#L484-L486

Compatibility wise it shouldn't affect many themes. Since the float is just moved to the stylesheet, and the columns all have the same widths, I don't see many layouts breaking. The only layouts I can think that will break are those who have filtered the entire shortcode (I know I did in Shop Front) using the `downloads_shortcode` filter:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/shortcodes.php#L567

PR coming soon to show how I think we should tackle this, we can remove a bunch of code and it will look exactly the same.

sumobi:
Fix #3489


sumobi:
@pippinsplugins ^ passed in empty string

pippinsplugins:
@sumobi can you resubmit to `release/2.5`?

PR #3491, opened 2015-06-10
Add filter to purchase summary

ericnicolaas:
Would be nice to be able to filter the purchase summary (we need to include certain "fees" in the summary sent to the gateway).

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -767,11 +767,13 @@ function edd_get_purchase_summary( $purchase_data, $email = true ) {
 		$summary .= $purchase_data['user_email'] . ' - ';
 	}
 
-	foreach ( $purchase_data['downloads'] as $download ) {
-		$summary .= get_the_title( $download['id'] ) . ', ';
-	}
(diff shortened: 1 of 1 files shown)


includes/cart/functions.php
```diff
@@ -750,7 +750,7 @@ function edd_get_purchase_summary( $purchase_data, $email = true ) {
 
 	$summary = substr( $summary, 0, -2 );
 
-	return $summary;
+	return apply_filters( 'edd_get_purchase_summary', $summary, $purchase_data );
```
pippinsplugins: Add in the `$email` parameter here and then we can merge it in.

ericnicolaas:
Done

pippinsplugins:
We'll include it shortly after 2.4

ericnicolaas:
@pippinsplugins I have added one additional modification in here, which is to check for downloads before looping over the array. Without this, you run into a PHP Warning when making a purchase with a cart that only has fees in it (no downloads).

pippinsplugins:
@ericnicolaas Thanks, we'll get this reviewed for 2.4.2

Issue #2935, opened 2015-01-07
Show message on items that are in the cart


cklosowski:
In #2788 we've modified the way that the checkout/add to cart button works on variable price products without multi option mode checked.

When the page is refreshed, if it's in the cart, we just show the checkout button without options. This might look slightly jarring to a user, so we should add some sort of messaging about the item being in the cart.

We held off on this for `2.2.3` as it's possible it could break some layouts. So we'll push it during a larger point release.

We may want to build some unit tests around this logic too, to see what results we get with the item in the cart.

pippinsplugins:
Tagging for 2.3 so it gets some attention.

pippinsplugins:
I'm going to go ahead and punt this to 2.4.

cklosowski:
This gives us, with a variable priced item, a notice when it's in the cart:
![image](https://cloud.githubusercontent.com/assets/1390124/8092181/f9b64cb4-0f6c-11e5-8f6e-5e0dc1a9c44a.png)

pippinsplugins:
What about non-variable items?

chriscct7:
I strongly disagree with the idea of an item is in cart message. I also disagree with changing the button to say checkout (on adding item to cart, a link could be added below the button to take you to checkout). If you look at studies done by Amazon, telling people items are already in the cart is a terrible way to improve ecommerce conversion rates. Amazon for example, doesn't change the button when you add an item to the cart, because some people avoid using the quantity fields and expect that they can just push the add to cart button twice to get the same effect. As Amazon wrote about it in one of their studies, they'd rather people correct over adding of items to cart on the checkout page than discouraging multiple purchase buys.

cklosowski:
@chriscct7 nothing changed with the checkout, that's remained the same the whole time...

@pippinsplugins The scope isn't for those, as nothing radically cahnges. The point of this ticket is that:
1) Add variable priced item to cart - The button changes to checkout, and the rpices still show
2) Refresh teh page...just the checkout button.

This was the jaring UI.

cklosowski:
@chriscct7 Cite your sources please. I don't want references to reports done by Amazon, without...reports done by amazon ;)

cklosowski:
We can just leave the UX as it is right now, in that a page refresh is going to have a variable priced item with no price IDs, just a checkout button.

pippinsplugins:
@cklosowski Thanks for the clarification, I missed that.

[...]

PR #3496, opened 2015-06-10
Add notice when variable priced item is in the cart

cklosowski:
Addresses #2935


cklosowski:
Closing for more research as noted in #2935

PR #3499, opened 2015-06-12
Validate credit card number using Luhn

benjaminprojas:
@pippinsplugins This validates whether the credit card number is valid based on the luhn algorithm.  This also validates the expiration date, making sure that the month + year are >= the current month.  

This can be used by all gateways using the default purchase form.

### DIFF
--- a/includes/process-purchase.php
+++ b/includes/process-purchase.php
@@ -778,7 +778,21 @@ function edd_purchase_form_validate_cc() {
 		}
 	}
 
-	// This should validate card numbers at some point too
+	// Validate the card number
+	$card_type = edd_purchase_form_cc_card_type( $card_data['card_number'] );
+	$valid = empty( $card_data['card_number'] ) ? false : true;
+	$valid &= edd_purchase_form_validate_cc_luhn( $card_data['card_number'] );
+	$valid &= edd_purchase_form_validate_cc_length( $card_type, $card_data['card_number'] );
+
+	if ( ! apply_filters( 'edd_purchase_form_valid_cc', $valid, $card_type ) ) {
+		edd_set_error( 'invalid_cc_number', __( 'The credit card number you entered is invalid', 'edd' ) );
+	}
+
+	// Validate the card expiration date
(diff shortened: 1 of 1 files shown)


cklosowski:
We should get some unit tests around these added functions if we are going to merge them.

benjaminprojas:
@cklosowski sounds good.  @pippinsplugins asked me to make this PR so whatever needs to happen is fine with me.

cklosowski:
:+1:

benjaminprojas:
@pippinsplugins actually, just realized that this will fail for any gateway who does not use the purchase form. With PayPal Express I am not displaying the CC form like so:

```php
add_action( 'edd_paypalexpress_cc_form', '__return_false' );
```

Would something similar work to avoid validation here? I imagine we would want to bypass the entire edd_purchase_form_validate_cc() function.

pippinsplugins:
@benjaminprojas This should probably get called by the payment gateway and not automatically by core. It's really here as a helper function so that gateways don't have to do it themselves, but the gateway will still need to call it.

benjaminprojas:
@pippinsplugins that totally makes sense.  I'll make the changes and push it out.

pippinsplugins:
Great, ping me when ready.

benjaminprojas:
@pippinsplugins okay I pulled out the automatic validation.  Now just the helper functions remain... should be good to go.

cklosowski:
Once merged, let's make sure #3502 get's completed

benjaminprojas:
@pippinsplugins I see this isn't in the beta 1 for EDD 2.4.  Are we still planning this for 2.4?

chriscct7:
@benjaminprojas yeah @cklosowski is working on this and #3502

cklosowski:
@benjaminprojas I've taken your basis and extended it, there were a few issues with regular expressions that weren't working on a couple card types. I've also made it a single function to call for developers to use that goes through the process. Also shortens the loop of card types once once is found and returns it, so we don't continue looping if the first card type is found.

cklosowski:
Once it's approved in the PR I've submitted we'll Props you on it

pippinsplugins:
Closing since it's being done through #3502

Issue #3076, opened 2015-02-11
Buy Now buttons does not support purchase form quantity field


pippinsplugins:
The `Buy Now` buttons do not currently support the quantity field that is shown on the purchase form, as introduced in #2125.

Pre-click: https://cloudup.com/cMztW9MCF9X
Post-click: https://cloudup.com/cJx39uiL6mz

pippinsplugins:
@cklosowski It's close but not quite: 
![screen shot 2015-06-12 at 4 41 19 pm](https://cloud.githubusercontent.com/assets/1034109/8140525/e9fc1b12-1121-11e5-8957-379926cc6cb3.png)

The item price is 0.50, so the total should be 1.00.

cklosowski:
@pippinsplugins Ah, yes, fixing

cklosowski:
![image](https://cloud.githubusercontent.com/assets/1390124/8142083/ef50fa10-1126-11e5-8a86-b4dc3cb5ddbe.png)

pippinsplugins:
Hmm, not sure I think something has broken with buy now buttons, at least with variable prices enabled. The button starts as Buy Now but then changes to Add to Cart as soon as a price option is selected. Instead of processing buy now, it's going to checkout.

pippinsplugins:
I don't think the break is related to changes here, just noting it here because this is where I noticed it.

cklosowski:
I'm getting an Add to Cart button, but it's acting like a buy now and going to PayPal.

pippinsplugins:
With variable prices?

cklosowski:
@pippinsplugins Yes, with variable prices.

PR #3500, opened 2015-06-13
Also update subtotal to reflect quantity

cklosowski:
#3076

### DIFF
--- a/includes/gateways/functions.php
+++ b/includes/gateways/functions.php
@@ -238,7 +238,7 @@ function edd_build_straight_to_gateway_data( $download_id = 0, $options = array(
 			'tax'         => 0,
 			'discount'    => 0,
 			'item_price'  => $price,
-			'subtotal'    => $price,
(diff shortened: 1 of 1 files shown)


includes/gateways/functions.php
```diff
@@ -238,7 +238,7 @@ function edd_build_straight_to_gateway_data( $download_id = 0, $options = array(
 			'tax'         => 0,
 			'discount'    => 0,
 			'item_price'  => $price,
-			'subtotal'    => $price,
+			'subtotal'    => ( $price * $quantity ),
 			'price'       => $price,
```
pippinsplugins: `price` also has to reflect quantity.

cklosowski: Not according to the PayPal checkout. It showed item price based off `price` and then total based off `subtotal`, see attached screenshot.

cklosowski: Oh, I see what you mena, ignore my deleted comment.

Issue #3364, opened 2015-04-24
Update customer email when user email is updated


pippinsplugins:
If user that is connected to a customer account updates their email, the customer email does not get updated. Nor do the payments attached to the customer.

We should update the customer and payments when the user account email is changed.

Ticket: https://wordpress.org/support/topic/bug-when-user-changed-email

chriscct7:
Also for name?

pippinsplugins:
Yes.

chriscct7:
And I assume you also want the reverse as well? Change email on EDD Customer page changes the WordPress user email for the customer?

pippinsplugins:
Yes.

chriscct7:
There's a couple more potential issues I want to consider and there's a minor conflict issue I need to fix. Should have this done within next 48, prob within next 24

cklosowski:
@chriscct7 how are we coming on this? I'd like to try and get 2.3.8 out this week.

pippinsplugins:
I don't think this is a 2.3.8 issue. It could have some larger implications. Let's punt to 2.4.

chriscct7:
After spending quite a bit of time on it, I would like to second the move to 2.4. I've got a list of possibilities and potential issues I'm still running down. I think this is something though that I can have sorted out and done this weekend, and then we can play around with and test my proposal with Loop if that's agreeable. I do echo the concern that this can't be a point release even though its a bug, because the implications if something goes wrong here are far reaching and potentially catastrophic.

cklosowski:
Yep, i'm 100% good with that. Thanks @chriscct7. Appreciate the due diligence on it.

pippinsplugins:
Things that have to be updated when user email is updated:
- file download logs have to have their email updated
- payment records need their email updated
- customer records need their email updated

Any others?

mintplugins:
Might be a related bug here. I have a customer who mistyped their email so it created a new "customer" but not a new "user". The user id is 299 for both purchases. But the customer email is different (because of the typo). Because of that typo, the customer doesn't see their purchase in their account. I also appear to have no way of adding that purchase to the customer's account because I can't fix the typo in the email address. If I try, I get this error:

"WordPress database error:  [Duplicate entry 'user@domain.com' for key 'email']

:(

cklosowski:
[...]

PR #3505, opened 2015-06-15
Issue/3364

pippinsplugins:
#3364

### DIFF
--- a/includes/class-edd-db-customers.php
+++ b/includes/class-edd-db-customers.php
@@ -274,7 +274,7 @@ class EDD_DB_Customers extends EDD_DB  {
 
 		$user = get_userdata( $user_id );
 
-		if( $user->user_email !== $customer->email ) {
+		if( ! empty( $user ) && $user->user_email !== $customer->email ) {
 
 			if( ! $this->get_customer_by( 'email', $user->user_email ) ) {
 
--- a/includes/class-edd-db-customers.php
+++ b/includes/class-edd-db-customers.php
@@ -294,6 +294,8 @@ class EDD_DB_Customers extends EDD_DB  {
 						
 					}
 
+					do_action( 'edd_update_customer_email_on_user_update', $user, $customer );
+
 				}
 
 			}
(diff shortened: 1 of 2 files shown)


includes/class-edd-db-customers.php
```diff
@@ -257,6 +259,50 @@ public function decrement_stats( $customer_id = 0, $amount = 0.00 ) {
...
+						}
+						
+					}
+
+				}
+
+			}
+
+		}
+
```
chriscct7: There should be an action here for extensions that now need to update data as well

chriscct7: Also does this break download links that were already sent and old style download links which some extensions still rely on (PDF vouchers comes to mind)

includes/class-edd-db-customers.php
```diff
@@ -257,6 +259,50 @@ public function decrement_stats( $customer_id = 0, $amount = 0.00 ) {
...
+		$customer = new EDD_Customer( $user_id, true );
+
+		if( ! $customer ) {
+			return false;
+		}
+
+		$user = get_userdata( $user_id );
+
+		if( $user->user_email !== $customer->email ) {
```
chriscct7: $user should not be assumed to be a user object here. We should ensure it is not empty, that it is an object and that it is a user object before assuming it is a user object (it could be int 0, int 0, or a WP error object respectively)

includes/class-edd-db-customers.php
pippinsplugins: The download URLs do not rely on the email addresses any more, so this shouldn't be an issue.

I do agree on the action. I'll see if I can work one in.

includes/class-edd-db-customers.php
pippinsplugins: Good catch.

includes/class-edd-db-customers.php
chriscct7: @pippinsplugins did you test specifically PDF Vouchers? That extension manually generates an old style download url that it uses to issue the pdf vouchers

pippinsplugins: I'll test that one. Thanks.

chriscct7: @pippinsplugins what was the result?

pippinsplugins: It works perfectly.

Issue #2798, opened 2014-11-19
Re-add download attribute


pippinsplugins:
The `download` attribute for download links was added in #2393 then removed in #2794.

I'd like to see it added back in.

@chriscct7 has made a good start in #2795

pippinsplugins:
This is not happening for 2.2.3.

pippinsplugins:
#3091

pippinsplugins:
#3255

chriscct7:
#3512

pippinsplugins:
Spent some time chatting with @cklosowski about this and we both agree that it needs to be done but are going to hold off a bit because we are tentatively planning on refactoring the download processing to abstract it a bit more, primarily for #3002

chriscct7:
Can we go ahead and merge this into the 2.5 branch please? This pr affects some of the most volatile (in terms of code changes) area of EDD, and we've already had to resubmit this pr 5 times

pippinsplugins:
No. The entire point of not merging it now is so that we can determine if
any changes here need to be made as we refactor the whole download
processor.

pippinsplugins:
This is getting marked as `wontfix` because of issue detailed here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3681#issuecomment-137552441

Issue #2393, opened 2014-06-22
Add download attribute to download links


chriscct7:
Now that the `download` attribute is officially part of the HTML5 spec, we should add it to all download links. This will help prevent even more cases of browsers inline displaying instead of force downloading media items.

Spec: http://www.w3.org/TR/html5/links.html#downloading-resources
Better explanation of usage: http://www.sitepoint.com/new-html5-attributes-hyperlinks-download-media-ping/

PR #3512, opened 2015-06-16

### LINKED ISSUE #2794: Remove download attribute on file download links

chriscct7:
YAR of + resolves #2794 and #2798

Refs:
#2393
#2794
#2795
#2798
#3091
#3255
#3512

### DIFF
--- a/includes/download-functions.php
+++ b/includes/download-functions.php
@@ -1022,13 +1022,13 @@ function edd_get_file_download_method() {
  *
  * @since 2.4
  * @author Chris Christoff
- * @param bool $bool Default for has_access
+ * @param bool $has_access Default for has_access
  * @param mixed $purchase_data Array of purchase data 
  * @param mixed $args Array of arguments
  * @return nool If a user can access the file
  */
-function edd_can_access_download( $bool = true, $purchase_data = array(), $args = array() ) {
-	return apply_filters( 'edd_file_download_has_access', $bool, $purchase_data, $args );
+function edd_has_file_download_access( $has_access = true, $purchase_data = array(), $args = array() ) {
+	return apply_filters( 'edd_file_download_has_access', $has_access, $purchase_data, $args );
 }
 
 /**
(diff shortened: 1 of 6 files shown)


includes/download-functions.php
```diff
@@ -1018,6 +1018,40 @@ function edd_get_file_download_method() {
...
+ * Returns if a user can access a download of a purchase
+ *
+ * @since 2.4
+ * @author Chris Christoff
+ * @param bool $bool Default for has_access
+ * @param mixed $purchase_data Array of purchase data 
+ * @param mixed $args Array of arguments
+ * @return nool If a user can access the file
+ */
+function edd_can_access_download( $bool = true, $purchase_data = array(), $args = array() ) {
```
cklosowski: `$bool` should be `$has_access`, if someone stumbles upon the function, seeing `$bool` gives no context.

pippinsplugins: Can you rename this to `edd_has_file_download_acces()`?

mindctrl: ^ missing an s.   `edd_has_file_download_access()`

pippinsplugins: Thanks

chriscct7: Both done in 9269c32b1c6351c84f9a2614a20144aef362be4d

includes/download-functions.php
```diff
@@ -1018,6 +1018,40 @@ function edd_get_file_download_method() {
...
 
 /**
+ * Returns if a user can access a download of a purchase
+ *
+ * @since 2.4
+ * @author Chris Christoff
+ * @param bool $has_access Default for has_access
+ * @param mixed $purchase_data Array of purchase data 
+ * @param mixed $args Array of arguments
+ * @return nool If a user can access the file
```
mindctrl: bool

chriscct7: fixed

pippinsplugins:
This needs resubmitted to the 2.5 branch.

chriscct7:
Resbumitted in #3681

pippinsplugins:
:+1:

Issue #2519, opened 2014-08-16
Add Earnings by Category report


pippinsplugins:
There needs to be a way to view the aggregated earnings per category.

Something like this:
![image](https://cloud.githubusercontent.com/assets/1034109/3942309/cc5fc026-255f-11e4-92e4-49d7140d0945.png)

cklosowski:
This could use some pretty heavy testing. But I've got a flot.js pie graph showing for the earnings and sales totals as well.

cklosowski:
![screen shot 2015-06-10 at 4 58 09 pm](https://cloud.githubusercontent.com/assets/1390124/8096977/eab37fd2-0f91-11e5-96c6-125be672a081.png)

Some notes on this:
1) The parent categories contain the sales/earnings/averages of all their children as well (uniquely of course) so you cannot just add `Child 1 + Child 2 = Parent`.
2) The graph is filtered out to show only items with `> 1%` input into the charts

cklosowski:
Oh
3) The Averages are calculated by adding up all of their average monthly sales and earnings of the downloads in the category, and then dividing by the number of downloads in the category

boluda:
Wow, it looks great!

chriscct7:
@cklosowski we could/should consider doing a key instead of labels. If you have 50 categories that each take up 2%, we're in trouble. Even 20 at 5% each looks messy. I'd say just do a key with all values over 1%.

chriscct7:
That or a hover label thing. Where label is shown as you hover over it. Would be cool down the road for someone to write a little plugin that when you click on a slice it shows more details about it and maybe shows a pie breakdown for products in the selected category slice. Would be really slick if it was all js animated.

chriscct7:
Hover doesn't work on mobile though. Grr. Key is probably better

pippinsplugins:
Such an awesome improvement.

cklosowski:
@Christoff. Still getting into query.flot's options. Have a bit of ways to go

chriscct7:
@cklosowski it's @chriscct7. @christoff is another developer.

cklosowski:
@chriscct7 Yeah stupid mobile app didn't auto complete for me.

cklosowski:
![image](https://cloud.githubusercontent.com/assets/1390124/8113283/c87cbf10-1021-11e5-9ebc-de2b47695e28.png)

Updated to show data in the legends.

[...]

PR #3514, opened 2015-06-16
Introduce Earnings by Category report

pippinsplugins:
#2519


includes/admin/reporting/class-categories-reports-table.php
```diff
@@ -0,0 +1,332 @@
...
+	 *
+	 * @since 1.5
+	 * @see WP_List_Table::__construct()
+	 */
+	public function __construct() {
+		global $status, $page;
+
+		// Set parent defaults
+		parent::__construct( array(
+			'singular'  => edd_get_label_singular(),    // Singular name of the listed records
```
pippinsplugins: These actually shouldn't be labels. These are for defining nonce name fields.

includes/admin/reporting/graphing.php
```diff
@@ -760,3 +760,129 @@ function edd_parse_report_dates( $data ) {
...
+ * Output a Pie graph using jQuery.flot
+ *
+ * @since  2.4
+ * @param  string $id   Unique ID for the DIV, alphanumeric and '_' only
+ * @param  array  $data The Data set in key/value form
+ * @param  array  $args List of options for the pie graph
+ * @return string       Returns the outputted JS and HTML for the graph display
+ */
+function edd_pie_graph( $id = '', $data = array(), $args = array() ) {
+	global $id_count;
```
pippinsplugins: Could we prefix this please?

includes/admin/reporting/graphing.php
```diff
@@ -760,3 +760,129 @@ function edd_parse_report_dates( $data ) {
...
+		'label_formatter'  => false,
+		'height'           => '300',
+	);
+
+	$args = wp_parse_args( $args, $defaults );
+
+	ob_start();
+	?>
+	<script type="text/javascript">
+		var <?php echo $id; ?>_data = [
```
pippinsplugins: Since $id can be passed to the function, let's escape it.

includes/admin/reporting/graphing.php
```diff
@@ -760,3 +760,129 @@ function edd_parse_report_dates( $data ) {
...
+		<?php foreach ( $data as $label => $value ) : ?>
+			<?php echo '{ label: "' . esc_attr( $label ) . '", data: "' . $value . '" },' . "\n"; ?>
+		<?php endforeach; ?>
+		];
+
+		var <?php echo $id; ?>_options = {
```
pippinsplugins: Escape it here too

pippinsplugins:
@cklosowski You may hate me for this, but the `edd_pie_graph()` function should really be part of `EDD_Graphs`, not a global function.

cklosowski:
I'll see what I can do to get those fixed up.

includes/admin/reporting/class-categories-reports-table.php
cklosowski: Not following, this is how we define most of these in any other list table for downloads.

includes/admin/reporting/graphing.php
cklosowski: We'll just escape it early. We use it a few times.

cklosowski:
@pippinsplugins moving to issue for discussion, Have a few things about the `EDD_Graphs` class

PR #3519, opened 2015-06-18
Given our themes are now free and the code doesn't work on bundles

chriscct7:
Remove the confusing wording

### DIFF
--- a/includes/admin/tracking.php
+++ b/includes/admin/tracking.php
@@ -243,7 +243,7 @@ class EDD_Tracking {
 			$optout_url = add_query_arg( 'edd_action', 'opt_out_of_tracking' );
 
 			echo '<div class="updated"><p>';
-				echo __( 'Allow Easy Digital Downloads to track plugin usage? Opt-in to tracking and our newsletter and immediately be emailed a 20% discount to the shop for <a href="https://easydigitaldownloads.com/extensions" target="_blank">Extensions and Themes</a>. No sensitive data is tracked.', 'edd' );
+				echo __( 'Allow Easy Digital Downloads to track plugin usage? Opt-in to tracking and our newsletter and immediately be emailed a 20% discount to the shop for <a href="https://easydigitaldownloads.com/extensions" target="_blank">non-bundle extension purchases</a>. No sensitive data is tracked.', 'edd' );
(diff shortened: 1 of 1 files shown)


cklosowski:
I agree that `themes` needs to be removed, but I think the `non-bundle` part get's a little wordy for people.

`extension purchases` would be just as accurate (since bundles aren't extensions)

We should also add GA tracking on that extensions link:
`https://easydigitaldownloads.com/extensions?utm_source=notice&utm_medium=admin&utm_content=extensions&utm_campaign=edd-tracking`

pippinsplugins:
I'd agree with changing it to just `extension1

cklosowski:
@pippinsplugins I talked with @chriscct7 and I do think maybe we can word this better to reduce confusion, but I don't want to make any large changes without having a baseline on the current wording.

pippinsplugins:
I vote we don't change it actually, for a couple of reasons:

1. There's nothing to say we won't have paid themes in the future
2. All translations would need to be updated

Does anyone know if there is evidence (such as a support ticket) of someone being confused why they couldn't apply the discount to a theme?

chriscct7:
I had a couple people open support tickets on this, which is why I actually opened the ticket. I'll see if I can dig them out of you want. They were mostly people who thought the code would work for marketify and such since they're paid but link off the main site. But I think the better thing might be to clarify "sold on our site" on the end of that sentence

pippinsplugins:
@chriscct7 Do you remember if they were confused because the themes were free or they were confused because they couldn't use the code on a bundle?

chriscct7:
Its not that they were confused with our themes, what happened was people were seeing 20% off themes, they'd click on the link to go to our site and on our site we list Marketify (for example) which is not free and they were going to ThemeForest from our site and trying to apply the coupon. And obviously we get the bundles issue every week as well. Adding `sold on our site (excludes bundles)` would solve all of these tickets we frequently get

pippinsplugins:
It doesn't exclude all bundles (gateway bundles not excluded).

I'm re-considering my objection from earlier.

chriscct7:
Oh really? I've always thought all bundles were excluded

pippinsplugins:
Nope, just Core and Marketplace.

Issue #3522, opened 2015-06-23
Add parameter to disable pagination for [download]


kiarashi:
If possible, please add an extra parameter inside the download shortcode that will either display the pagination or not. E.g. [download pagination=yes/no/false/true]

cklosowski:
I think this could be a useful addition. i'll slate it for a future release.

pippinsplugins:
I've updated the title to reflect the change.

cklosowski:
This was completed in #3894 with EDD 2.5.7

PR #3526, opened 2015-06-24
Parameter added to disable pagination for [download] #3522

PareshRadadiya:
Added parameter inside the download shortcode that will either display the pagination or not. E.g. [download pagination=yes/no]

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -273,7 +273,7 @@ function edd_downloads_query( $atts, $content = null ) {
 		'orderby'          => 'post_date',
 		'order'            => 'DESC',
 		'ids'              => '',
-		'pagination'       => 'yes'
+		'pagination'       => 'true'
 	), $atts, 'downloads' );
 
 	$query = array(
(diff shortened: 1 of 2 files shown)


pippinsplugins:
For #3522

spencerfinnell:
Should be true/false rather than yes/no?

PareshRadadiya:
An **excerpt**  parameter accept "yes" or "no". The  **thumbnails** parameter requires "true" or "false" :confused:

pippinsplugins:
Ha, oops. Yay us for being inconsistent.

I'd prefer to stick with `true/false`

PareshRadadiya:
Okay, I've updated parameter to `true/false`

includes/shortcodes.php
```diff
@@ -272,20 +272,29 @@ function edd_downloads_query( $atts, $content = null ) {
...
 		'order'          => $atts['order']
 	);
 
-	if ( $query['posts_per_page'] < -1 ) {
-		$query['posts_per_page'] = abs( $query['posts_per_page'] );
+	if(  'true' === $atts['pagination']  ) {
+
+		$query['posts_per_page'] = (int) $atts['number'];
+
+		if ( $query['posts_per_page'] < -1 ) {
```
cklosowski: `$query['posts_per_page'] < -1` should be `$query['posts_per_page'] < 0`. By being `< -1` would allow a value of `-1`.

cklosowski:
See my one comment, other than that, looks good to me

PR #3550, opened 2015-07-02
[l10n] > First update from TX for 2.4.1

fxbenard:
First pull from TX
At least all .mo are based on the latest pot and French is 100% completed


pippinsplugins:
Thanks so much!

PR #3567, opened 2015-07-05
fixed edd_listen_for_failed_payments()

felixarntz:
The `edd_get_payment_status()` function only accepts a `WP_Post` object. However the current version of `edd_listen_for_failed_payments()` passes a payment ID (post ID) to the function, causing it to return false instead of the actual payment status so that the payment status is never actually changed to 'failed'.

This commit fixes that behavior.


pippinsplugins:
Good catch!

PR #3574, opened 2015-07-06
Update pot to 2.4.1

michaelbeil:


pippinsplugins:
@michaelbeil Did you submit this (or pull it from) transifex as well?

michaelbeil:
I only ran a `grunt makepot` to update the recent string changes, but I did not pull from transifex. I would pull from tx as well but I don't have permissions.

PR #3583, opened 2015-07-07
Update class-edd-api.php Fix:bug with bbpress profile editing page

wpcar:


cklosowski:
@wpcar, we went with the option to move the necessary 'upgrade check' functions into a front end accessible file, instead of including all our upgrade routine functions whenever the API is called.

See the changes here: #3597

Issue #3569, opened 2015-07-06
Customer search doesn't work when searching by name or email


mindctrl:
Go to Downloads > Customers, and search for a customer by name or email address. No results are returned. The only way it seems to work is by the customer ID #.

[Reported here](https://secure.helpscout.net/conversation/102472477/12174/).

PR #3587, opened 2015-07-08
Customer search doesnt work when searching by name or email. Resolve …

PareshRadadiya:
Search by email and name query prepare fix for  …#3569

### DIFF
--- a/includes/class-edd-db-customers.php
+++ b/includes/class-edd-db-customers.php
@@ -424,18 +424,20 @@ class EDD_DB_Customers extends EDD_DB  {
 		if( ! empty( $args['email'] ) ) {
 
 			if( is_array( $args['email'] ) ) {
-				$emails = "'" . implode( "', '", $args['email'] ) . "'";
+
+				$emails_count = count( $args['email'] );
+				$emails_placeholder = array_fill( 0, $emails_count, '%s' );
+				$emails = implode( ', ',  $emails_placeholder ) ;
+				$where .= $wpdb->prepare( " AND `email` IN( $emails ) ",  $args['email'] );
 			} else {
-				$emails = "'" . $args['email'] . "'";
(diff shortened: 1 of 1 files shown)


includes/class-edd-db-customers.php
```diff
@@ -424,18 +424,20 @@ public function get_customers( $args = array() ) {
 		if( ! empty( $args['email'] ) ) {
 
 			if( is_array( $args['email'] ) ) {
-				$emails = "'" . implode( "', '", $args['email'] ) . "'";
+
+				$emails_count = count( $args['email'] );
+				$emails_placeholder = array_fill( 0, $emails_count, '%s' );
```
cklosowski: We cannot use `0` as the value for the `array_fill`, this was only allowed in PHP 5.6 and greater. Previous versions cannot use `0` as a value for array_fill:

http://php.net/manual/en/function.array-fill.php#refsect1-function.array-fill-changelog
5.6.0	num may now be zero. Previously, num was required to be greater than zero.

includes/class-edd-db-customers.php
```diff
@@ -424,18 +424,20 @@ public function get_customers( $args = array() ) {
...
+				$emails = implode( ', ',  $emails_placeholder ) ;
+				$where .= $wpdb->prepare( " AND `email` IN( $emails ) ",  $args['email'] );
```
cklosowski: Ignore previous comment, misread the argument order.

I'd say just formatting updates and this should be good:
```
$emails_count           = count( $args['email'] );
$emails_placeholder = array_fill( 0, $emails_count, '%s' );
$emails                      = implode( ', ',  $emails_placeholder ) ;
$where                      .= $wpdb->prepare( " AND `email` IN( $emails ) ",  $args['email'] );
```

PR #3596, opened 2015-07-09
Purchase a download through PayPal with no title results in empty car…

### LINKED ISSUE #3481: Trying to purchase a download through PayPal with no title results in empty cart
If you create a product but do NOT give it a title, like this https://cloudup.com/ccBLpddpZa9, and then try to purchase it through PayPal, PayPal will display a "Your cart is empty" error message:

screen shot 2015-06-05 at 1 38 03 pm

We should set the PayPal `item_name` field to the download ID if no title is present.

PareshRadadiya:
…t Resolve #3481

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -564,7 +564,7 @@ function edd_get_cart_item_price_name( $item = array() ) {
  * @param int $item Cart item array
  * @return string item title
  */
-function edd_get_cart_item_title( $item = array() ) {
+function edd_get_cart_item_name( $item = array() ) {
 	$item_title = get_the_title( $item['id'] );
 
 	if( empty( $item_title ) ) {
@@ -576,7 +576,7 @@ function edd_get_cart_item_title( $item = array() ) {
 		$item['name'] .= ' - ' . edd_get_cart_item_price_name( $item );
 	}
 
-	return apply_filters( 'edd_get_cart_item_title', $item_title, $item['id'], $item );
+	return apply_filters( 'edd_get_cart_item_name', $item_title, $item['id'], $item );
 }
 
 /**
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -558,6 +558,28 @@ function edd_get_cart_item_price_name( $item = array() ) {
 }
 
 /**
+ * Get cart item title
+ *
+ * @since 2.4.2
+ * @param int $item Cart item array
+ * @return string item title
+ */
+function edd_get_cart_item_title( $item = array() ) {
+	$item_title = get_the_title( $item['id'] );
+
+	if( empty( $item_title ) ) {
+		$item_title = $item['id'];
+	}
+
+	if ( edd_has_variable_prices( $item['id'] ) && edd_get_cart_item_price_id( $item ) !== false ) {
+
+		$item['name'] .= ' - ' . edd_get_cart_item_price_name( $item );
+	}
+
+	return apply_filters( 'edd_get_cart_item_title', $item_title, $item['id'], $item );
+}
+
+/**
  * Cart Subtotal
  *
  * Shows the subtotal for the shopping cart (no taxes)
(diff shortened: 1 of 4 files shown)


includes/cart/functions.php
```diff
@@ -558,6 +558,28 @@ function edd_get_cart_item_price_name( $item = array() ) {
...
+ * @return string item title
+ */
+function edd_get_cart_item_title( $item = array() ) {
+	$item_title = get_the_title( $item['id'] );
+
+	if( empty( $item_title ) ) {
+		$item_title = $item['id'];
+	}
+
+	if ( edd_has_variable_prices( $item['id'] ) && edd_get_cart_item_price_id( $item ) !== false ) {
```
cklosowski: Can we write that 2nd condition as `false !== edd_get_cart_item_price_id( $item )`

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions

cklosowski:
looks good other than my note above. There is some confusion in that the array key is `name` and the function is `title` but I think that's avoidable as this new function also calls a function containing `cart item name`, so it's probably better to differentiate.

PareshRadadiya:
@cklosowski Should we change function name `edd_get_cart_item_title` to `edd_get_cart_item_name` ?

cklosowski:
That might be a good idea, since it will cover two things:
1) The array element is `name` not `title`
2) we call `edd_get_cart_item_price_name` as well, so it keeps that consistent.

Let's go with `edd_get_cart_item_name` as the new function name. When it's done let me know and I'll put it on my radar to get some unit tests build for it.

PareshRadadiya:
@cklosowski  It's done !

templates/shortcode-receipt.php
```diff
@@ -148,7 +148,7 @@
 						?>
 
 						<div class="edd_purchase_receipt_product_name">
-							<?php echo esc_html( $item['name'] ); ?>
+							<?php echo esc_html( edd_get_cart_item_name( $item ) ); ?>
 							<?php if( ! is_null( $price_id ) ) : ?>
 							<span class="edd_purchase_receipt_price_name">&nbsp;&ndash;&nbsp;<?php echo edd_get_price_option_name( $item['id'], $price_id, $payment->ID ); ?></span>
```
pippinsplugins: We need to not change this part since the the price ID is added below with it's own HTML.

pippinsplugins:
@cklosowski @PareshRadadiya see my comment above

templates/shortcode-receipt.php
PareshRadadiya: @pippinsplugins
Consider doing something this way. We can remove this 
```php
<?php if( ! is_null( $price_id ) ) : ?>
	<span class="edd_purchase_receipt_price_name">&nbsp;&ndash;&nbsp;<?php echo edd_get_price_option_name( $item['id'], $price_id, $payment->ID ); ?></span>
<?php endif; ?>
```
So it would be
```php
<div class="edd_purchase_receipt_product_name">
	<?php echo esc_html( edd_get_cart_item_name( $item ) ); ?>
</div>
```

pippinsplugins: @PareshRadadiya The problem with that is it changes the markup of the checkout, which could break users styles. While it is nicer using a simple function, we can't do that for backwards compatibility's sake.

PareshRadadiya: @pippinsplugins This would be okay.
```php
<div class="edd_purchase_receipt_product_name">
	<?php
	if( empty( $item['name'] ) ):
		echo $item['id'];
	else:
		echo esc_html( $item['name'] );
	endif;
	?>
	<?php if( ! is_null( $price_id ) ) : ?>
	<span class="edd_purchase_receipt_price_name">&nbsp;&ndash;&nbsp;<?php echo edd_get_price_option_name( $item['id'], $price_id, $payment->ID ); ?></span>
	<?php endif; ?>
</div>
```

pippinsplugins: There's no reason `$item['name']` would be empty.

pippinsplugins:
Looks good, will test further.

Issue #3561, opened 2015-07-03
Improve Batch Exporter Extensibility


cklosowski:
There are a few changes we need to make to the batch exporter to allow it to be super extensible by other plugins.

Mainly adding some hooks to allow including specific extensions of the `EDD_Batch_Export` class and set properties specific to the type of export, instead of a general collection of properties.

cklosowski:
This is needed to release an update for the Commissions extension, fixing a major issue with the generation of large mass payment files.

cklosowski:
@pippinsplugins can you take a look at this issue branch and see if this works for you. The core reports still export, working on getting the Commissions one working with it now.

pippinsplugins:
See my inline comments.

pippinsplugins:
This great, let's get a PR and get it merged.

PR #3599, opened 2015-07-10
Improve batch processing to be extendable 

cklosowski:
#3561


pippinsplugins:
@cklosowski Could each of the file includes be in a single function using a switch statement instead of a separate function for each export type? A switch statement will be better for future expansions.

cklosowski:
@pippinsplugins we could, but I think it makes it slightly more convoluted for people who want to either filter our default ones, or add their own ( by using our example ).

pippinsplugins:
Are you thinking that by having them separate, we could allow people to replace our export classes?

cklosowski:
Yes, it makes it so they can remove an action, and add their own to do an include for a specific type.

cklosowski:
Or an extension builder can.

pippinsplugins:
@cklosowski let's move `do_action( 'edd_batch_export_class_include', $form['edd-export-class'] );` to just after the nonce verification to ensure nothing can be executed during the process before we've verified it's legit.

Issue #3483, opened 2015-06-06
Admin notice for opt in mentions newsletter subscription


brashrebel:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/admin/tracking.php#L246

There is a bit of a discrepancy regarding the 20% discount for opting in offer. The admin notice above states:

`Allow Easy Digital Downloads to track plugin usage? Opt-in to tracking and our newsletter and immediately be emailed a 20% discount to the shop for <a href="https://easydigitaldownloads.com/extensions" target="_blank">Extensions and Themes</a>. No sensitive data is tracked.`

whereas the general settings state:

`Allow Easy Digital Downloads to anonymously track how this plugin is used and help us make the plugin better. Opt-in and receive a 20% discount code for any purchase from the Easy Digital Downloads store. Your discount code will be emailed to you.`

I would recommend aligning these two messages with each other better. Perhaps just remove the newsletter reference from the admin notice. If leaving it in, it should probably link to https://easydigitaldownloads.com/#pmc_mailchimp or something. Also, the discount code comes without necessitating a newsletter subscription currently.

pippinsplugins:
They should be the same. The first one is correct because they do get subscribed automatically.

cklosowski:
@sunnyratilal I'm deleting the branch you pushed up as it's got a lot of extra commits. Got a new one coming up now.

cklosowski:
How do we feel about this wording @pippinsplugins and @brashrebel 
`Opt-in to tracking and our newsletter and immediately be emailed a 20% discount to the EDD shop, valid for our library of extensions.`...the `library of extensions` becomes the link in this case.

pippinsplugins:
:+1:

PR #3602, opened 2015-07-10
Update tracking message to be consistent

cklosowski:
#3483


pippinsplugins:
Looks good

cklosowski:
Actually, while we're at it, let's get some UTM tracking on those links :)

Issue #3580, opened 2015-07-07
URLs to remove Fees and Items can be caught by caching


jdevalk:
It generates URLs with `p` or `page_id` instead of normal URLs. Should be easy enough to fix :)

pippinsplugins:
@jdevalk Why is that a problem? Is the `p` or `page_id` incorrect?

jdevalk:
No, the caching exclusion doesn't work because the URL doesn't contain `/checkout/`, which it would otherwise...

pippinsplugins:
Ah! that makes sense.

We'll get it fixed!

PR #3603, opened 2015-07-10
Remove item/fee URLs honor checkout and avoid cache 

cklosowski:
#3580

### DIFF
--- a/includes/cart/actions.php
+++ b/includes/cart/actions.php
@@ -87,7 +87,7 @@ add_action( 'edd_add_to_cart', 'edd_process_add_to_cart' );
 function edd_process_remove_from_cart( $data ) {
 	$cart_key = absint( $_GET['cart_item'] );
 	edd_remove_from_cart( $cart_key );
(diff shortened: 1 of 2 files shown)


includes/cart/actions.php
```diff
@@ -101,7 +101,7 @@ function edd_process_remove_from_cart( $data ) {
 function edd_process_remove_fee_from_cart( $data ) {
 	$fee = sanitize_text_field( $data['fee'] );
 	EDD()->fees->remove_fee( $fee );
-	wp_redirect( remove_query_arg( array( 'edd_action', 'fee' ) ) ); edd_die();
+	wp_redirect( remove_query_arg( array( 'edd_action', 'fee', 'nocache' ) ) ); edd_die();
```
pippinsplugins: Why do you remove nocache here? Seems like we'd want it to stay

cklosowski: B/C this is the redirect `after` the removal happens. we've already removed, so get that off otherwise it removes the item then redirects to the cart with only the `nocache` query string, we've already removed the `fee` and `action` items after it's removed.

Issue #3537, opened 2015-06-30
edd_microdata_wrapper_open function running in wrong place.


danieliser:
This one is easy to duplicate. Fire up a fresh install with the 2015 theme and create a new download. When you view the download the schema wrapper is added to the default "Recent Posts" widget body like so.

	<aside id="recent-posts-2" class="widget widget_recent_entries">
		<h2 class="widget-title">Recent Posts</h2>
		<ul>
			<span itemtype="http://schema.org/Product" itemscope="">
				<li>
					<a href="http://edd.dev/2015/06/30/hello-world/">Hello world!</a>
				</li>
			</span>
		</ul>
	</aside>

I have corrected this using the first parameter of that hook $query. It now looks like this and wraps only the main loop content not the sidebar.

	function edd_microdata_wrapper_open( $query ) {

		static $microdata_open = NULL;

		if( ! edd_add_schema_microdata() || true === $microdata_open || ! is_object( $query ) ) {
			return;
		}

		if ( $query && ! empty( $query->query['post_type'] ) && $query->query['post_type'] == 'download' && is_singular( 'download' ) && $query->is_main_query() ) {
			$microdata_open = true;
			echo '<span itemscope itemtype="http://schema.org/Product">';
		}

	}

Look forward to any feedback on this solution.

PR #3610, opened 2015-07-13
Fix schema when sidebar is called first

cklosowski:
#3537


pippinsplugins:
Looks good. Will merge one travis passes.

PR #3614, opened 2015-07-13
Signup is a noun.  Sign up is a verb.

JustinSainton:


cklosowski:
Good catch. I was curious about it...did you know there is an `english.stackexchange.com`? Our language is complicated enough to need crowdsourced grammatical resources.

http://english.stackexchange.com/questions/71930/sign-up-vs-signup

PR #3618, opened 2015-07-16
Add the product id in the sales api.

wolfkang:
Add the product id in the sales api to fetch the product information after calling the sales api.


pippinsplugins:
`$item['id']` should be correct here. Why are you proposing changing it to `$item_id`?

chriscct7:
@pippinsplugins $item_id is actually more correct. See about 10 lines above. We do a check in case the $item variable isn't an array but just a number.

cklosowski:
Yes, `$item_id` is actually more correct due to this line:
`$item_id  = isset( $item['id']    ) ? $item['id']    : $item;`

cklosowski:
Thanks @wolfkang

PR #3619, opened 2015-07-17
Get all the purchase of the logged-in user.

wolfkang:


pippinsplugins:
@wolfkang Could you please explain what the problem is that this fixes?

wolfkang:
@pippinsplugins Sorry for the insufficient explanation. I wrote about this in wordpress.org.
https://wordpress.org/support/topic/i-cant-see-my-purchase-history-after-i-changed-the-email-address?replies=3
When I buy an item as a logged-in user and then change my email, I cannot see the previous purchase item in the purchase history.
This is because the EDD_Custmer instance is created by one email address in the purchase history.
So I modified the code to load the purchase items by the given user id.
Please check it out.

pippinsplugins:
@wolfkang The change actually goes against what we're trying to do.

Originally, EDD used user_ids AND email addresses to associate payment records with customer accounts. This become problematic because it created inconsistencies. Some customers were linked by user ID and some were linked by email.

A few versions back we introduced our customers database table and began making the move away from ever using the user_id to look up purchases.

We won't go back to using user IDs to look up the purchases. Instead we need to identify why changing the email on the account did not properly update the customer and payment records for the new email.

Was the email changed from the Profile or by entering a new email on checkout?

wolfkang:
@pippinsplugins Thanks for reply. You know EDD has an email field in the checkout fields. So users write any email  address, which could be different from his profile address. Also users can change his email address in the profile page and this is very normal in web sites.
When a not-logged-in user bought someting, his email would be a key in the database. But a logged-in user bought something, the user_id should be a key. So the purchase history should be fetched by the logged-in user's user_id, not by his email address.
Please reconsider about the logged-in user's purchase history.

pippinsplugins:
When users change their email in the profile, we automatically update all of their purchase records.

If a user uses a different email during checkout, we cannot associate it with the user's previous customer records. Customer's are identified by email address, so that cannot be done. Perhaps in the future if we introduce aliasing for customer records.

pippinsplugins:
Closing this as invalid now. If anyone disagrees, feel free to re-open and explain your opinion.

PR #3622, opened 2015-07-18
Implement exists by any field.

chriscct7:
Sometimes, you don't know the email for a customer, but you do know, for example, their user_id


pippinsplugins:
@chriscct7 Looks good but I'd like to see the column whitelisted. If field passed doesn't match up with a column, default to `email`.

cklosowski:
@pippinsplugins I'm not sure on 'defaulting' to email. I'd rather return false if they give us a field we're not expecting. It's the proper response for their query while defaulting to `email` if they provide a bad field is masking the error request

PR #3624, opened 2015-07-19
Link the User ID to the Edit Profile page

zackkatz:
I always want to see the user profile, not just the customer details.
Alternately, other user details like the username would be helpful.

### DIFF
--- a/includes/admin/customers/customers.php
+++ b/includes/admin/customers/customers.php
@@ -283,7 +283,7 @@ function edd_customers_view( $customer ) {
 
 					<span class="customer-user-id info-item editable">
 						<?php _e( 'User ID', 'edd' ); ?>:&nbsp;
-						<?php if( intval( $customer->user_id ) > 0 ) : die(); ?>
+						<?php if( intval( $customer->user_id ) > 0 ) : ?>
 							<span data-key="user_id"><a href="<?php echo admin_url( 'user-edit.php?user_id=' . $customer->user_id ); ?>"><?php echo $customer->user_id; ?></a></span>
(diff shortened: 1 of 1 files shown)


zackkatz:
My current flow: 

* Go to the customer
* Copy the User ID
* Go to the Users page
* Click on any user
* Change the ID parameter in the URL
* Access the user profile

I do this a *lot*, and having this link would fix it.  The CSS should probably be updated, but I didn't want to mess with your process on how to name/implement.

shazahm1:
+1

includes/admin/customers/customers.php
```diff
@@ -283,8 +283,8 @@ function edd_customers_view( $customer ) {
 
 					<span class="customer-user-id info-item editable">
 						<?php _e( 'User ID', 'edd' ); ?>:&nbsp;
-						<?php if( intval( $customer->user_id ) > 0 ) : ?>
-							<span data-key="user_id"><?php echo $customer->user_id; ?></span>
+						<?php if( intval( $customer->user_id ) > 0 ) : die(); ?>
```
pippinsplugins: Dying here is not right. We still need it to continue with the rest of the page.

pippinsplugins: @zackkatz

zackkatz: Shoot, sorry @pippinsplugins, that was silly testing.

pippinsplugins: Could you remove it so we can merge this in?

zackkatz:
@pippinsplugins done.

Issue #3636, opened 2015-07-23
Bugs/Fixes to Earnings by Category admin report


MythThrazz:
Hello,

I'm lead Wordpress Developer at CreativeMinds, we're using the Easy Digital Downloads to sell our plugins. We were happy that you've added the "Earnings by Category" report, but quickly we've found out that it's missing the date filters which were our core use case. 

So we've decided to add it, thinking about maybe creating an add-on of it, but quickly we've discovered that the report is using the old (soon to be deprecated as the comment says) function for stats instead of the EDD_Stats class.After using this class and the date selector from the Earnings report we've managed to quickly add the filter to the report, so no need for an add-on :)
 
Later during the testing, we've found out that there's a bug in the EDD_Stats class, which caused the transactions containing the same download_id more than once being multiplied (Select query was missing "DISTINCT").

cklosowski:
Would you mind submitting your changes as a Pull Request to the branch `issue/3636`? Then I can pull them down easily and test them.

MythThrazz:
Done.

cklosowski:
@MythThrazz made a couple changes here, fixed up some UX issues with the control order and whitespace. Things look like they are working alright for you on this PR?

cklosowski:
@pippinsplugins @mindctrl Could you guys poke around on this and see if the changes that @MythThrazz and I made are working correctly for you?

MythThrazz:
@cklosowski I was on a short vacation and didn't have time version after merge yet. I'll try to find time when I dig myself out of the pile of tasks.

cklosowski:
@MythThrazz no worries! thanks for getting back to us. I didn't change much, just some general whitespace cleanup and restructured the order of the controls to follow our standards (was a bug on our end not giving you a way to put them in the correct order)

mindctrl:
@cklosowski still testing this in `issue/3636`? It's totally broken for me.

-----------

![image](https://cloud.githubusercontent.com/assets/1254870/11131664/c5d3f68c-8959-11e5-8cd5-d2b98b191444.png)

Getting JS errors in the console. `Uncaught TypeError: Cannot read property 'grid' of null in jquery.flot.pie.js?ver=4.3.1:661`

cklosowski:
Yeah `issue/3636` should be the one, let me bring it up to speed with `release/2.5`. That could be b/c of 0 sales...do you have something with sales?

mindctrl:
[...]

PR #3637, opened 2015-07-23
Code for the Issue #3636

MythThrazz:
#3636

- Displaying the daterange filters on the Earnings by Category report
- Adding the daterange support for the class outputting the report
- Fixing the bug in the EDD_Payment_Stats class

### DIFF
--- a/includes/admin/reporting/class-categories-reports-table.php
+++ b/includes/admin/reporting/class-categories-reports-table.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Gateways Reports Table Class
  *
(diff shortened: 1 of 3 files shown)


cklosowski:
Looks like the transients have been removed. This shouldn't happen. We need to cache the result set. This would probably mean making a hash of the arguments to get a string unique to that request.

pippinsplugins:
@MythThrazz Did you see Chris's comment about the transients?

MythThrazz:
Hi, 
Sorry for late reply, been busy. I'm working on some more fixes to this part, so I'll fix the transient and the naming convention and commit again.

cklosowski:
Two issues left on my end:

There are a number of changes to the whitespace in this PR that don't fall in line with our core codebase...and it's against master.

Could you please bring your branch of this up to date with our `master` and re-submit the PR to `issue/3636`. I'm getting a bunch of notices and warnings with it currently, and while the overall idea of it seems right, I want it in a branch so I can merge your changes, and update where I see necessary.

This is a great 90% of the way there PR though. thanks @MythThrazz

Issue #3626, opened 2015-07-21
Logic seems backwards in edd_store_discount


danieliser:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/discount-functions.php#L255

Shouldn't that be ??

if ( ! empty( $discount_id ) && edd_discount_exists( $discount_id )  ) {

danieliser:
Doing it the way it is now causes an extra lookup before even checking that the ID is empty. if statements process from left to right. so if the first check fails it automatically skips the rest. Saves a query if the id is empty.

pippinsplugins:
`if ( ! empty( $discount_id ) && edd_discount_exists( $discount_id ) ) {` would be better :)

danieliser:
:+1:

PR #3642, opened 2015-07-24
Reverse logic for edd_store_discount() to make it faster when creatin…

pippinsplugins:
…g new discounts. #3626


cklosowski:
Looks good.

Issue #3670, opened 2015-08-04
Add HUF to list of currencies not supporting decimals


mindctrl:
[Reference](https://developer.paypal.com/docs/classic/api/currency_codes/).

PR #3671, opened 2015-08-04
No decimals for HUF currency

cklosowski:
#3670


pippinsplugins:
How about adding some unit tests for these?

cklosowski:
Was already doing that, while you were commenting :P
Yup

pippinsplugins:
Besties!

PR #3681, opened 2015-08-05

### LINKED ISSUE #2794: Remove download attribute on file download links

chriscct7:
YAR of + resolves #2794 and #2798

Refs:
#2393
#2794
#2795
#2798
#3091
#3255
#3512


pippinsplugins:
I've just spent some more time testing this. It works really well with one exception: it breaks any plugin that uses the `edd_process_verified_download` action to output an error message [EDD Recurring does this](https://github.com/easydigitaldownloads/edd-recurring/blob/master/edd-recurring.php#L200).

It breaks because the `download="Filename"` attribute forces the browser to download a file. When taht happens we get a `.html` file in the download dialog:

screen shot 2015-09-03 at 2 35 29 pm

Once the download is started, it shows a Forbidden error:
screen shot 2015-09-03 at 2 36 14 pm

While I fully understand that extensions should probably not be using `edd_process_verified_download` to show an error message, it's the only hook available (or at least was).

For this reason, I vote we `wontfix` this.

Objections? Rebuttals?

Issue #3658, opened 2015-07-29
Inserting invalid string on start and/or end time on purchase payment page leads to fatal error


chriscct7:
?post_type=download&page=edd-payment-history

To replicate: Paste (not try to type) in a string into the fields.

Result: 
![screenshot at 01-39-30](https://cloud.githubusercontent.com/assets/1324637/8950678/b861c364-3592-11e5-89bb-258b7e838163.png)

chriscct7:
Moving this to 2.4.3. Just needs a type check on the construct of the library.

pippinsplugins:
:+1:

cklosowski:
As @chriscct7 and I discussed, it'd be good to get other date formats supported in our date picker implementation, but for now we can rely on `m/d/Y` to get our date format from these selections.

cklosowski:
Ok latest updates have taken care of the date formatting, also, added in some checks on the stats class. checking for wp_error (when the date is a string like `something`) as well as if the strototime doesn't come out anything good.

pippinsplugins:
Changes in https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/a5924eeb2cb0bdae24775dd3d145ffc2e8a2032a have broken today and last month stats.

![screen shot 2015-08-18 at 12 45 29 pm](https://cloud.githubusercontent.com/assets/1034109/9337917/afb1828e-45a7-11e5-94cf-779e395fe202.png)

pippinsplugins:
It's the `strtotime( $this->start_date )` and strtotime( $this->end_date )` that makes it fail.

PR #3682, opened 2015-08-05
Make sure it's a date we can parse

cklosowski:
#3658

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -490,29 +490,15 @@ function edd_count_payments( $args = array() ) {
 
 	// Limit payments count by date
 	if ( ! empty( $args['start-date'] ) ) {
-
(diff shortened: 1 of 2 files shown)


includes/payments/functions.php
```diff
@@ -490,15 +490,29 @@ function edd_count_payments( $args = array() ) {
 
 	// Limit payments count by date
 	if ( ! empty( $args['start-date'] ) ) {
-		$date = new DateTime( $args['start-date'] );
-		$where .= "
-			AND p.post_date >= '" . $date->format( 'Y-m-d' ) . "'";
+
+		// If our date pickers ever support other formats, update this 'format' with the chosen one for the store
+		$date = DateTime::createFromFormat( 'm/d/Y', $args['start-date'] );
```
pippinsplugins: How about a version check. If 5.3+, use createFromFormat(), otherwise use `strtotime()`

PR #3693, opened 2015-08-09
Small stuff

JeroenSormani:
Hey,

Doing a couple smaller things here;
- The checkout was displaying subtotal/totals that were off
- Match the <label for='x'> with the <input id='x'>
- Align some code
- Replace $('body') with $(document.body) (https://github.com/woothemes/woocommerce/pull/7879)


pippinsplugins:
@JeroenSormani Happy to look at each one of these changes but they will have to be done separately. Changing the checkout total vs subtotal is not a small thing so it has to be done by itself and done very carefully.

If you resubmit each of these as a separate PR, we can go from there :)

PR #3695, opened 2015-08-10
Use $download->get_price() function instead of property

JeroenSormani:
Hey,

Just noticed this and made a quick PR inside GH.

Use the sanitized and filtered `$download->get_price()` and `$download->get_prices()` values for the old backward compatible functions instead of the EDD_Download property.


pippinsplugins:
Thanks!

PR #3696, opened 2015-08-10
Fix #3694 Added to cart message on click of checkout

### LINKED ISSUE #3694: Unused output in `edd_get_purchase_link()`
I just stumbled across something weird:

In [this line](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/template-functions.php#L185) the function `printf()` is called, however the second and third parameter are never used (since there is no placeholder for them in the first parameter) which makes no sense to me. Not sure if this is a bug or just old code.

I encountered it since a client asked me to replace the "Checkout" text on buttons with a "Go to Checkout" text, so I browsed around and found this. For the actual use case it would be great if we were able to filter the "Checkout" text.

PatelUtkarsh:
Fix #3694

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -183,7 +183,7 @@ function edd_get_purchase_link( $args = array() ) {
 				<span class="edd-cart-ajax-alert">
 					<span class="edd-cart-added-alert" style="display: none;">
 						<?php printf(
-								'<i class="edd-icon-ok"></i> ' . __( 'Added to cart', 'edd' ),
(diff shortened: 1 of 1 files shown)


PatelUtkarsh:
Should we also add filter for `checkout` and `added to cart` ? @pippinsplugins

pippinsplugins:
That looks correct.

No, it should not have a filter.

I'll test and merge if everything works as expected.

PatelUtkarsh:
No lonnger required as line is changed with echo statement.

PR #3702, opened 2015-08-14
Prefill Billing State / Province and Update user First and Last name on checkout

vanbo:
Prefill Billing State / Province - The State / Province for countries that the it is written instead of selected is currently not pre-filled and always stays empty as opposing to all other billing fields. So I thought it will be a good idea to just pre-fill it, too.

Update user First and Last name on checkout - the user first and last name are not updated, when the user is logged in already. But there can be cases where the user was created without the First and Last name (from place other than the checkout page). In this case the First and Last name will never be updated. Checks can also be added, to make sure that the First and Last name changed before they are updated. Let me know what you think. Thanks


pippinsplugins:
@vanbo Looks good! We'll consider it further for 2.5.

includes/process-purchase.php
```diff
@@ -1064,3 +1066,28 @@ function edd_process_straight_to_gateway( $data ) {
...
+ *
+ * @return int|WP_Error
+ */
+function edd_update_user( $user_data = array() ) {
+	// Bail, if we don't have a user to update
+	if ( ! isset( $user_data['user_id'] ) ) {
+		return false;
+	}
+
+	$user_args = array(
```
chriscct7: This would be better as:
$user_args = array( );

if ( isset( $user_data['user_first'] )  ){
     $user_args['first_name'] = $user_data['user_first']; 
}
// same thing for last name here

if ( !empty( $user_args ) ){
  // add id to array
  // update user here
} 

That way the user only gets updated if and only if they have new information coming in (we don't want to lose old data), and if we only have a first name only update that, not last_name if we don't have that.

pippinsplugins:
Only thing I don't like about this is the `edd_update_user()` function. It definitely makes sense to have a helper function, but looking at the function, I would expect it to support all standard arguments of wp_update_user(), which it doesn't in the current form.

How about for now, since the user is only updated in one instance, we just call wp_update_user() inline. We can abstract it more in the future if it needs re-used.

pippinsplugins:
@vanbo any thoughts on my comments above?

vanbo:
@pippinsplugins It definitely makes sense to either use more specific function or just use the code inline. I do see how `wp_update_user()` might confuse people that it does more then it actually does.

I'll add it inline for now and we can move it, if it needs to.

pippinsplugins:
Sounds good!

vanbo:
Moved the update inline in the `edd_get_purchase_form_user` function

pippinsplugins:
Great! We'll re-review for 2.5.5 or shortly after.

cklosowski:
I agree with pre-filling the state/province fields, but I don't agree with the user first and last names should update the user's first and last name. Mainly for the concept of the payment not being specifically for a single 'person', meaning I can make a purchase as a logged in user, but have an email address an name for say a client, or a department head.

This way when I have a receipt or invoice for the purchase, it's for the person who the purchase was authorized by, not the user account, therefore updating the user to use this information wouldn't be ideal.

@pippinsplugins @vanbo I'd be curious your opinions on this.

cklosowski:
Punting to 2.5.8 as we're trying to get 2.5.7 out shortly.

pippinsplugins:
After reviewing again, I agree, we shouldn't update the user name here. We can store it when creating an account during purchase but not when making a new purchase.

cklosowski:
After looking back over this discussion, we're not ready to update any user information based off change during a purchase. If we come up with another perspective on this, since we've introduced a full Customer Object since this was originally done, let's revisit it in an issue first for discussion.

PR #3706, opened 2015-08-14
Fix (sub)total amount attributes

JeroenSormani:
Splitted PR from https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3693

### DIFF
--- a/includes/checkout/template.php
+++ b/includes/checkout/template.php
@@ -760,7 +760,7 @@ function edd_checkout_final_total() {
 ?>
 <p id="edd_final_total_wrap">
 	<strong><?php _e( 'Purchase Total:', 'edd' ); ?></strong>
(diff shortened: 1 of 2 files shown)


pippinsplugins:
@JeroenSormani why is the subtotal incorrect?

JeroenSormani:
@pippinsplugins 
The attribute says `data-total` while the function used is `edd_get_cart_subtotal()`.
There's also a attribute `data-subtotal` which also has the `edd_get_cart_subtotal()` function (this one is correct).

Vice versa for the other file.

pippinsplugins:
Huh, that's "great".

Let's definitely get that confusion fixed. I'm not going to merge this yet as we need to be super careful with it, but I'm going to milestone it for 2.6.

JeroenSormani:
Cool. I think I did check when I made those changes, and I couldn't find those attributes used in EDD Core.

cklosowski:
Going to fix up the merge conflicts and get this directly into our `release/2.6` branch. Using #4697 instead keeps commit history in place.

Issue #3740, opened 2015-08-26
Use is_ssl() instead of $_SERVER["HTTPS"] check


mindctrl:
On some systems the $_SERVER["HTTPS"] variable doesn't get set, and results in some weird cross domain issues or 'bad requests'. It also apparently works better behind proxies ([reference](https://github.com/pippinsplugins/Restrict-Content-Pro/pull/214)).

We use it in `edd_get_current_page_url()`, and a couple other places, I think.

pippinsplugins:
:+1:

cklosowski:
I fear we may have made this obsolete with the recent release in 2.4.4 in issue #3769. We'll just want to be sure that we're still happy with `edd_get_current_page_url()` when this merge is all said and done.

mindctrl:
@cklosowski didn't you merge this in already?

cklosowski:
Yes I did, then I merged in master, which has the release from #3769, which further altered this function. I'm reviewing the merge right now, to make sure we're not seeing any artifacts of a wonky merge.

cklosowski:
yeah merge looks good, i'm closing this one out as the generation of the URLs with these functions has been tested well in the 2.4.4 and 2.4.5 releases.

PR #3744, opened 2015-08-28
Switch from $_SERVER["HTTPS"] check to is_ssl(). #3740.

mindctrl:
is_ssl() handles the same check and an additional port check for when servers fail to set the HTTPS var. This removes the check from edd_enforced_ssl_redirect_handler() since it's already checked with is_ssl above, and it swaps the check in edd_get_current_page_url() to is_ssl().

### DIFF
--- a/includes/checkout/functions.php
+++ b/includes/checkout/functions.php
@@ -323,10 +323,6 @@ function edd_enforced_ssl_redirect_handler() {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


pippinsplugins:
#3740

cklosowski:
I'm going to merge this in, it's needed before I start on #3769

PR #3746, opened 2015-08-28
.htaccess added filter for file formats

somewebmedia:
Video file formats should be allowed as well, in case a theme needs those files for showcase

### DIFF
--- a/includes/admin/upload-functions.php
+++ b/includes/admin/upload-functions.php
@@ -150,7 +150,7 @@ function edd_get_htaccess_rules( $method = false ) {
 		case 'direct' :
 		default :
 			// Prevent directory browsing and direct access to all files, except images (they must be allowed for featured images / thumbnails)
-			$allowed_filetypes = apply_filters( 'edd_protected_directory_allowed_filetypes', array( 'jpg', 'png', 'gif' ) );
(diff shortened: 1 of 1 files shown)


chriscct7:
Strongly against this PR on the basis that many people have uploaded full videos that they sell via EDD that they are expecting to be protected. This would open those files to be directly hotlinked. If anything, I'm for removing all of the exceptions and then use a bulk upgrade routine to move a copy of previously whitelisted ones to the main uploads folder, and then stop whitelisting any extensions

somewebmedia:
@chriscct7 Ok, I've removed video types, but there should be some filter for the people who want better control of this, so my proposal is 'edd_protected_directory_allowed_filetypes' filter.

somewebmedia:
In that case even .mp3 and .ogg should be removed from the default list.

chriscct7:
Whatever was already allowed should remain allowed for now. We'll consider removing those exceptions on a different issue.

somewebmedia:
@chriscct7 Done. And please consider this filter for filetypes, I encountered a need for video files to be visible as a showcase, but there is no easy workaround.

chriscct7:
Looks good

pippinsplugins:
I like the idea of a filter here.

If @cklosowski agrees as well, we'll merge.

cklosowski:
Filter here is a fantastic idea. It cant hurt and offers people to make changes they deem necessary for their own sites. big :+1: from me.

pippinsplugins:
Thanks @somewebmedia !

Issue #3758, opened 2015-09-02
Add $item to edd_receipt_show_download_files filter


pippinsplugins:
The `edd_receipt_show_download_files` allows us to determine if the download files should be shown on the receipt page.

At the moment there is no way to know the price ID of the item being checked. We can add `$item` as an additional parameter to the function/filter to allow extensions to check the price ID too.

pippinsplugins:
Required for https://github.com/easydigitaldownloads/edd-recurring/issues/62

PR #3759, opened 2015-09-02
Add  to edd_receipt_show_download_files. #3758

pippinsplugins:
#3758


includes/checkout/template.php
```diff
@@ -905,6 +905,6 @@ function edd_filter_success_page_content( $content ) {
  * @since 1.8.6
  * @return boolean
 */
```
cklosowski: Can we add the args into the PHPDoc while we are in here updating it?

pippinsplugins: Done

PR #3764, opened 2015-09-04
Update class-edd-html-elements.php

wolfkang:
When I have more than 30 downloads, I cannot see all the downloads in the bundled downloads selectbox. The number of posts should be changed from 30 to -1 to fetch all.

### DIFF
--- a/includes/class-edd-html-elements.php
+++ b/includes/class-edd-html-elements.php
@@ -38,7 +38,7 @@ class EDD_HTML_Elements {
 			'multiple'    => false,
 			'selected'    => 0,
 			'chosen'      => false,
-			'number'      => 30,
+			'number'      => -1,
 			'bundles'     => true,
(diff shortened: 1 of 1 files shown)


chriscct7:
-1 is a bad idea. On a site with 10,000 downloads, doing that will crash the site. The select boxes for downloads are chosen enabled. By default 30 downloads are populated. As you type the name of the bundle you are looking for, the results will automatically refresh on the fly, and it will find the bundle for you.

wolfkang:
Thanks for reply. As I type the name of a download, I get no results for the match. There's no http request for searching. Please check it out.

pippinsplugins:
Are you entering 4+ characters? Search only happens once at least 4 characters have been entered.

wolfkang:
Thanks, @pippinsplugins I didn't know that.

Issue #3770, opened 2015-09-04
Better object handling with API Customer endpoint


cklosowski:
Before using the `$user_data` in the `class-edd-api:917` we never actually verify we got an object back from the `get_userdata()` call. This results in notices:
```
Notice:  Trying to get property of non-object in /Users/cklosowski/WordPress/src/wp-content/plugins/easy-digital-downloads/includes/api/class-edd-api.php on line 926

Notice:  Trying to get property of non-object in /Users/cklosowski/WordPress/src/wp-content/plugins/easy-digital-downloads/includes/api/class-edd-api.php on line 927
```

pippinsplugins:
Good catch

PR #3772, opened 2015-09-05
Account for -1 as user_id on customer

cklosowski:
#3770

### DIFF
--- a/includes/api/class-edd-api.php
+++ b/includes/api/class-edd-api.php
@@ -914,7 +914,7 @@ class EDD_API {
 				$customers['customers'][$customer_count]['info']['last_name']    = $last_name;
 				$customers['customers'][$customer_count]['info']['email']        = $customer_obj->email;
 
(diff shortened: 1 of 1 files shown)


includes/api/class-edd-api.php
```diff
@@ -914,7 +914,7 @@ public function get_customers( $customer = null ) {
 				$customers['customers'][$customer_count]['info']['last_name']    = $last_name;
 				$customers['customers'][$customer_count]['info']['email']        = $customer_obj->email;
 
-				if ( ! empty( $customer_obj->user_id ) ) {
+				if ( $customer_obj->user_id > 0 ) {
```
pippinsplugins: Maybe we should do ! empty() and greater than. Trying to think if there's ever a case where it wouldn't be set.

cklosowski: Yeah I'm good with that, changing it and merging.

PR #3773, opened 2015-09-06
First pass at plugins_loaded transition

ghost1227:
This is a quick run-through of the cleanest way I can think of to migrate EDD to load on `plugins_loaded` without breaking anything. Short version:

Anything that is required during the installation process is moved to a new class `EDD_Loader`. This class is instantiated during the `register_activation_hook` function and in the main `Easy_Digital_Downloads` class. At present, the loader doesn't do much but instantiate the required classes (`EDD_Roles()`, `EDD_API()`, `EDD_Session()` and `EDD_DB_Customers()`), and define the plugin constants.

During the instantiation of the main `Easy_Digital_Downloads` class, the already instantiated classes are passed to the EDD vars rather than re-instantiating them. This allows backward compatibility with anything currently referencing them through `EDD()->whatever->method()`.

The activation hook has been moved from the install.php file into the main plugin file, and now calls a new function, `edd_load_installer` which instantiates the main loader, then runs the existing `edd_install` function.

In my tests, it works flawlessly.

This is a pass at resolving the issue brought up in #2189.

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -48,6 +48,14 @@ final class Easy_Digital_Downloads {
 	private static $instance;
 
 	/**
+	 * EDD Loader Object
+	 *
+	 * @var object
+	 * @since 2.5
+	 */
+	public $loader;
+
+	/**
 	 * EDD Roles Object
 	 *
 	 * @var object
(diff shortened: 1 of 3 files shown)


ghost1227:
Additionally, I'm loading EDD with a priority of 9. Since `plugins_loaded` runs at 10 by default, this allows the extensions which check for EDD on `plugins_loaded` to still work as expected.

I have not updated the unit tests yet, so the build does fail at the moment.

Issue #3661, opened 2015-07-30
Download Details widget doesn't display unless settings have been modified


mindctrl:
When adding the Download Details widget to a sidebar, if you don't change any of the settings and leave everything at defaults, it doesn't display. You must change at least one setting for it to show. I think we just need to set a default instance.

pippinsplugins:
:+1:

cklosowski:
@mindctrl this works fine for me...maybe the fix from dan in #3778 took care of this, but by default, we don't show up unless a download ID is specified, or the current item being viewed is a download (since the default selection is 'current'). Can you test again and close if this is not needed now?

pippinsplugins:
This works for me as well. Closing as fixed. Re-open if you disagree.

PR #3778, opened 2015-09-07
Default to "current" if download_id is not set

ghost1227:
Issue #3661


pippinsplugins:
@ghost1227 Unit tests need updated for this.

cklosowski:
@ghost1227 I think the unit test you removed should still be in there. For the best code coverage we should have a positive and negative test. So Visit `/`, make sure it returns nothing, visit the `download_id`, make sure there is content. A better positive test would look for the download's name in the content with an `$this->assertContains()` as well. Just be more thorough.

cklosowski:
Other than that, love the idea and update.

Issue #3781, opened 2015-09-08
Improve the check for receipt visibility


mindctrl:
There are cases where our receipt visibility check fails. I've opened some tickets in other repos, but ultimately this is an EDD core issue.

References:
https://github.com/easydigitaldownloads/edd-free-downloads/issues/18

https://github.com/easydigitaldownloads/edd-auto-register/issues/8

What I've found is that the receipt isn't visible on the purchase confirmation screen if a purchase is attached to a user account when purchased while logged out. This happens with Auto Register and with Free Downloads. Customers come to the site, don't log in, and purchase things with the email address of their existing user account.

The part of our check that fails is the `$customer_id` check here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.4.3/includes/shortcodes.php#L653

I have a proposal for fixing this in an incoming PR. Definitely needs more eyes on it.

pippinsplugins:
@mindctrl I would LOVE to see this improved and the proposal looks good. I left a few comments on the PR for you.

pippinsplugins:
Looking really good though I have found a bug.

1. Make a purchase as a guest
2. Make a new purchase as a guest but create an account during checkout
3. Visit purchase history and click View Order Details on the purchase made in step 1. It cannot be viewed.

pippinsplugins:
Wait, maybe ignore me. Might have confused myself. Testing again.

pippinsplugins:
All clear!

pippinsplugins:
This looks perfect.

@mindctrl Could you give this one more test now that @cklosowski has added his changes as well?

Note: I've merged `master` into this so you will see the account verification stuff too.

SDavisMedia:
I was able to replicate the issue on master testing with Free Downloads and then confirm that it works properly on EDD `issue/3781`. :+1:

pippinsplugins:
Merging.

PR #3782, opened 2015-09-08
Fix receipt visibility issue under certain conditions. 

mindctrl:
#3781

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -795,3 +795,36 @@ if ( ! function_exists( 'getallheaders' ) ) :
 	}
 
 endif;
+
+/**
+ * Determines the receipt visibility status
+ *
+ * @return bool Whether the receipt is visible or not.
+ */
+function edd_user_can_view_receipt( $payment_key ) {
+
+	global $edd_receipt_args;
+
+	$edd_receipt_args['id'] = edd_get_purchase_id_by_key( $payment_key );
+
+	$customer_id = edd_get_payment_user_id( $edd_receipt_args['id'] );
+
+	$payment_meta = edd_get_payment_meta( $edd_receipt_args['id'] );
+
+	if ( is_user_logged_in() && $customer_id === get_current_user_id() ) {
+		return true;
+	}
+
+	if ( current_user_can( 'view_shop_sensitive_data' ) ) {
+		return true;
+	}
+
+	$session = edd_get_purchase_session();
+	if ( ! empty( $session ) && ! is_user_logged_in() ) {
+		if ( $session['user_email'] === $payment_meta['user_info']['email'] ) {
+			return true;
+		}
+	}
+
+	return false;
+}
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -813,16 +813,18 @@ function edd_can_view_receipt( $payment_key = '' ) {
 
 	$edd_receipt_args['id'] = edd_get_purchase_id_by_key( $payment_key );
 
-	$user_id = edd_get_payment_user_id( $edd_receipt_args['id'] );
+	$user_id = (int) edd_get_payment_user_id( $edd_receipt_args['id'] );
 
 	$payment_meta = edd_get_payment_meta( $edd_receipt_args['id'] );
 
-	if ( is_user_logged_in() && $user_id == get_current_user_id() ) {
-		$return = true;
-	}
-
-	if ( current_user_can( 'view_shop_sensitive_data' ) ) {
-		$return = true;
+	if ( is_user_logged_in() ) {
+		if ( $user_id === (int) get_current_user_id() ) {
+			$return = true;
+		} elseif ( wp_get_current_user()->user_email === $payment_meta['user_info']['email'] ) {
+			$return = true;
+		} elseif ( current_user_can( 'view_shop_sensitive_data' ) ) {
+			$return = true;
+		}
 	}
 
 	$session = edd_get_purchase_session();
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -820,7 +820,7 @@ function edd_can_view_receipt( $payment_key = '' ) {
 	if ( is_user_logged_in() ) {
 		if ( $user_id === (int) get_current_user_id() ) {
 			$return = true;
-		} elseif ( wp_get_current_user()->user_email === $payment_meta['user_info']['email'] ) {
+		} elseif ( wp_get_current_user()->user_email === edd_get_payment_user_email() ) {
 			$return = true;
 		} elseif ( current_user_can( 'view_shop_sensitive_data' ) ) {
 			$return = true;
@@ -829,7 +829,7 @@ function edd_can_view_receipt( $payment_key = '' ) {
 
 	$session = edd_get_purchase_session();
 	if ( ! empty( $session ) && ! is_user_logged_in() ) {
-		if ( $session['user_email'] === $payment_meta['user_info']['email'] ) {
+		if ( $session['purchase_key'] === $payment_meta['key'] ) {
 			$return = true;
 		}
 	}
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -801,30 +801,36 @@ endif;
  *
  * @return bool Whether the receipt is visible or not.
  */
-function edd_user_can_view_receipt( $payment_key ) {
+function edd_can_view_receipt( $payment_key = '' ) {
+
+	$return = false;
+
+	if ( empty( $payment_key ) ) {
+		return $return;
+	}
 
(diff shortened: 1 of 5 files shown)


chriscct7:
This might have breaking implications with extensions relying on receipts (FES pre-2.3  for view orders for example)

mindctrl:
@chriscct7 not seeing any issues viewing orders on FES 2.2.17. If you can provide more insight on that it would be helpful.

chriscct7:
In FES 2.2.17, the vendor view order page is made by showing the receipt. This itself probably doesn't affect FES because we use the filter right below the change to override whatever EDD says, but it could affect other extensions that use the receipt shortcode

mindctrl:
Thanks @chriscct7. Confirmed this works with PDF Invoices and EDD Invoices. Will check for other extensions.

includes/misc-functions.php
```diff
@@ -795,3 +795,36 @@ function getallheaders() {
 	}
 
 endif;
+
+/**
+ * Determines the receipt visibility status
+ *
+ * @return bool Whether the receipt is visible or not.
+ */
+function edd_user_can_view_receipt( $payment_key ) {
```
pippinsplugins: Can we change this to `edd_can_view_receipt( $payment_key = '' )`?

- Not everyone viewing a receipt is technically a `user`
- Reads better semantically
- Sets a default value

includes/misc-functions.php
```diff
@@ -795,3 +795,36 @@ function getallheaders() {
...
+
+	global $edd_receipt_args;
+
```
pippinsplugins: If `$payment_key` is empty, this should bail early.

includes/misc-functions.php
```diff
@@ -795,3 +795,36 @@ function getallheaders() {
...
+	}
+
+	$session = edd_get_purchase_session();
+	if ( ! empty( $session ) && ! is_user_logged_in() ) {
+		if ( $session['user_email'] === $payment_meta['user_info']['email'] ) {
+			return true;
+		}
+	}
+
+	return false;
```
pippinsplugins: This should have a filter here so that it can be changed by extensions. Return statements above should be changed to a variable definition so there is only one return statement.

includes/misc-functions.php
```diff
@@ -795,3 +795,36 @@ function getallheaders() {
...
+ *
+ * @return bool Whether the receipt is visible or not.
+ */
+function edd_user_can_view_receipt( $payment_key ) {
+
+	global $edd_receipt_args;
+
+	$edd_receipt_args['id'] = edd_get_purchase_id_by_key( $payment_key );
+
+	$customer_id = edd_get_payment_user_id( $edd_receipt_args['id'] );
```
pippinsplugins: This should be `$user_id` since the customer ID is something different.

includes/misc-functions.php
```diff
@@ -795,3 +795,42 @@ function getallheaders() {
...
+
+	global $edd_receipt_args;
+
+	$edd_receipt_args['id'] = edd_get_purchase_id_by_key( $payment_key );
+
+	$user_id = edd_get_payment_user_id( $edd_receipt_args['id'] );
+
+	$payment_meta = edd_get_payment_meta( $edd_receipt_args['id'] );
+
+	if ( is_user_logged_in() && $user_id == get_current_user_id() ) {
```
pippinsplugins: Let's make this a strict comparison with `===` and also add `(int)` to both.

pippinsplugins: Both being `$user_id` and `get_current_user_id()`

includes/misc-functions.php
```diff
@@ -795,3 +795,42 @@ function getallheaders() {
...
+		$return = true;
+	}
+
+	if ( current_user_can( 'view_shop_sensitive_data' ) ) {
+		$return = true;
+	}
+
+	$session = edd_get_purchase_session();
+	if ( ! empty( $session ) && ! is_user_logged_in() ) {
```
pippinsplugins: This should also have a condition for when the user ID on the payment doesn't match the the logged-in user's ID but the email does. This would happen where a payment was purchased while logged out (so no ID recorded) but with same email.

includes/misc-functions.php
```diff
@@ -795,3 +795,44 @@ function getallheaders() {
...
+		return $return;
+	}
+
+	global $edd_receipt_args;
+
+	$edd_receipt_args['id'] = edd_get_purchase_id_by_key( $payment_key );
+
+	$user_id = (int) edd_get_payment_user_id( $edd_receipt_args['id'] );
+
+	$payment_meta = edd_get_payment_meta( $edd_receipt_args['id'] );
```
pippinsplugins: Let's replace this with `edd_get_payment_user_email()`. The one in user_info might get deprecated at some point.

includes/misc-functions.php
```diff
@@ -795,3 +795,44 @@ function getallheaders() {
...
+		} elseif ( wp_get_current_user()->user_email === $payment_meta['user_info']['email'] ) {
+			$return = true;
+		} elseif ( current_user_can( 'view_shop_sensitive_data' ) ) {
+			$return = true;
+		}
+	}
+
+	$session = edd_get_purchase_session();
+	if ( ! empty( $session ) && ! is_user_logged_in() ) {
+		if ( $session['user_email'] === $payment_meta['user_info']['email'] ) {
```
pippinsplugins: I'd prefer that we compare payment IDs here instead of emails. If the purchase session is set, we don't need to validate the email.

includes/payments/functions.php
```diff
@@ -1063,9 +1077,36 @@ function edd_get_payment_user_email( $payment_id ) {
...
+	// check the customer record first
+	$customer_id = edd_get_payment_customer_id( $payment_id );
+	$customer    = new EDD_Customer( $customer_id );
+
+	if ( ! empty( $customer->user_id ) && $customer->user_id > 0 ) {
+		$user_id = $customer->user_id;
+	}
+
+	// check the payment meta if we're still not finding a user with the customer record
+	if ( ! empty( $user_id ) && $user_id > 0 ) {
```
pippinsplugins: No need for `$user_id > 0` check here since it's only getting set to something other than zero if the ID found on line 1083 s greater than 0.

includes/shortcodes.php
```diff
@@ -666,11 +666,27 @@ function edd_receipt_shortcode( $atts, $content = null ) {
...
-		return $edd_receipt_args['error'];
+	if ( ! isset( $payment_key ) ) {
+		return '<p class="edd-alert edd-alert-error">' . $edd_receipt_args['error'] . '</p>';
+	}
+
+	$payment_id    = edd_get_purchase_id_by_key( $payment_key );
+	$user_can_view = edd_can_view_receipt( $payment_key );
+
+	// Key was provided, but user is logged out. Offer them the ability to login and view the receipt
+	if ( ! $user_can_view && ! empty( $payment_key ) && ! is_user_logged_in() && ! edd_is_guest_payment( $payment_id ) ) {
```
mindctrl: @cklosowski do we want the edd_user_can_view_receipt filter above this just in case?

mindctrl:
@cklosowski @pippinsplugins something added since my last commit broke EDD Invoices.

pippinsplugins:
@mindctrl broke in what way?

mindctrl:
@pippinsplugins oops. `You do not have permission to view this invoice.`

pippinsplugins:
@mindctrl EDD Invoices or PDF Invoices?

mindctrl:
EDD Invoices. PDF Invoices works fine.

pippinsplugins:
That text string doesn't exist in EDD Invoices . . ..

mindctrl:
Line 211 in the main plugin file.

pippinsplugins:
Oh, nevermind, mine was outdated.

pippinsplugins:
@mindctrl What is the `edd_get_payment_user_id()` function returning on line 210?

pippinsplugins:
Ah it's because we're now returning `0` instead of `-1` for a guest. While `-1` sucks a lot, we'll need to leave it as `-1` so as to not break things.

mindctrl:
yup, 0.

pippinsplugins:
@mindctrl Can you try again now?

mindctrl:
@pippinsplugins still not working, but it looks like it's because the customer record isn't getting attached to the user account. 

image

mindctrl:
Something here is also breaking the "View Licenses" link output by Software Licensing. It shows no output. Seems to break in this commit. abd1fdded6d2a3ca8008f16b27e8dc67a05c9712, so maybe unrelated to this ticket.

pippinsplugins:
As soon as tests pass, we can merge this.

PR #3787, opened 2015-09-09
Fix #3473: Discount codes: excluded products count towards minimum am…

### LINKED ISSUE #3473: Discount codes: excluded products count towards minimum amount
Excluded products currently count towards the minimum amount required to use a discount code. We need to account for that and make sure they do not.

PareshRadadiya:
…ount

#3473

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -606,10 +606,50 @@ function edd_cart_subtotal() {
  * @return float Total amount before taxes
  */
 function edd_get_cart_subtotal() {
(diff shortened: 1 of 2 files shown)


cklosowski:
Works great, going to see if I can get some unit tests built around this as well. I think we should update the messaging in the future (since we're string locked for 2.5) to mention that "minimum of $20 of qualifying items not met" when there is a product exclusion. But this is a good fix to get in, regardless of the string being reported.

Issue #3769, opened 2015-09-04
edd_get_current_page_url() returns primary site URL in multisite with sub domains


pippinsplugins:
I don't know if this is always the case, but I have at least one reported case of the `edd_get_current_page_url()` returning the primary site URL instead of the current site when in a multisite environment running subdomains.

For example, site is running at `store.domain.com`. On checkout, instead of returning `store.domain.com`, it is returning just `domain.com`.

This was causing the remove from cart links to not work.

See [ticket](https://wordpress.org/support/topic/remove-link-in-checkout-missing-subdomain) on WordPress.org for more info.

cklosowski:
@pippinsplugins why aren't we using something like `site_url()` for the base? The real problem on subdomain multi-site installs is that the `$_SERVER['SERVER_NAME']` variable is pulled from the vhost configuration, and many times, that is simply the primary domain name, as the subdomains are wildcards, there is not a config file for every subdomain.

cklosowski:
The detail is that on things like Apache `$_SERVER['SERVER_NAME']` pulls from the `ServerName` value in the VHOST config.

pippinsplugins:
I honestly don't remember why `site_url()` wasn't used since the beginng.

pippinsplugins:
We also have issues with the `edd_get_current_page_url()` function returning a port number, such as in this ticket: https://secure.helpscout.net/conversation/118695079/16322/

pippinsplugins:
@cklosowski Just had another report of an issue with this (see ticket above). Let's get this fixed up and then shipped ASAP.

chriscct7:
So there's a couple more bugs here (most of these are edd_get_checkout_uri() vs edd_get_current_page_url()) that need to be fixed (migrating over from the HelpScout ticket): 

1. When edd_get_current_page_url() calls home_url on the frontpage it should return right after that. Instead it continues on, and would therefore end up with a final url like: http://www.example.com://www.example.com

2. Despite the fact these functions are used to retrieve urls interchangeable on pages that need cache bursting (ex: the remove action link if/elses between the two), nocache is not available on edd_get_current_page_url() when it should be

3. Again despite the fact these functions are used interchangeably, they are significantly different in how they create a link. This is a problem.

[...]

PR #3790, opened 2015-09-09
Make url detection more universal and reliable

cklosowski:
#3769

### DIFF
--- a/includes/checkout/functions.php
+++ b/includes/checkout/functions.php
@@ -131,8 +131,9 @@ function edd_get_checkout_uri( $args = array() ) {
 		$uri = preg_replace( '/^http:/', 'https:', $uri );
 	}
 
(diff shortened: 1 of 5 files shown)


includes/checkout/functions.php
```diff
@@ -131,8 +131,9 @@ function edd_get_checkout_uri( $args = array() ) {
 		$uri = preg_replace( '/^http:/', 'https:', $uri );
 	}
 
-	if ( edd_get_option( 'no_cache_checkout', false ) && edd_is_caching_plugin_active() )
-		$uri = add_query_arg( 'nocache', 'true', $uri );
+	if ( edd_get_option( 'no_cache_checkout', false ) ) {
```
chriscct7: shouldn't the cache busting function do this check? That way extensions could just call it and if its on use it

cklosowski: No b/c this is a specific check for the checkout url.

PR #3797, opened 2015-09-13
Add page title

### LINKED ISSUE #3795: Settings, Reports, Tools admin page need headings
Just noticed that the Settings, Reports, and Tools admin pages do not have headings. I believe they should. On some screen sizes, you wouldn't even know where you are in the dashboard unless you were already familiar with the UI.

screen shot 2015-09-13 at 2 28 35 am

ghost1227:
Fixes #3795


chriscct7:
a11y page titles should be h2, not h1

Issue #3571, opened 2015-07-06
Change "User" heading to "Customer" in the Payment History screen


mindctrl:
Now that we Customers, let's change the User column heading to Customer to clarify it.

pippinsplugins:
I've actually done more than just update the column name. I've also updated the column values to pull the name from the customer record instead of the user record.

pippinsplugins:
@mindctrl could you give this some testing this week?

mindctrl:
Looks good here.

PR #3799, opened 2015-09-14
Better Customer Links for the Payments Table

chriscct7:
Right now the links here just bring you to a filtered view of the customers payments. 99% of the time, when on the page, I sort of expect this link to bring you to the customers's profile, which after all contains all of their orders anyways.

image

This pr fixes it so it does that
Makes for a much better user experience and lets you more easily cross reference data for a customer

### DIFF
--- a/includes/admin/payments/class-payments-table.php
+++ b/includes/admin/payments/class-payments-table.php
@@ -355,7 +355,7 @@ class EDD_Payment_History_Table extends WP_List_Table {
 		$customer_id = edd_get_payment_customer_id( $payment->ID );
 		$customer    = new EDD_Customer( $customer_id );
 
-		$value = '<a href="' . esc_url( add_query_arg( array( 'user' => urlencode( $customer->email ), 'paged' => false ) ) ) . '">' . $customer->name . '</a>';
+		$value = '<a href="' . esc_url( admin_url( "edit.php?post_type=download&page=edd-customers&view=overview&id=$customer_id" ) ) . '">' . $customer->name . '</a>';
 		return apply_filters( 'edd_payments_table_column', $value, $payment->ID, 'user' );
 	}
 


pippinsplugins:
I'm not opposed to this at all. I quite like it really.

Before we had a customer view, it as important to support filtering by the customer, but I'm not sure that's necessary any more.

cklosowski:
We'll need to do backwards compatibility on this, and if the EDD site hasn't run the upgrade script to include customers, we'll have to maintain current behavior.

cklosowski:
but other than that concern, a bit :+1: from me

chriscct7:
If they haven't run the routine, then the existing links probably won't work either since they are built using the EDD_Customer class

mindctrl:
If back compat is going to complicate things, can we just show a linked notice in the column telling people to run the upgrade routine we've already asked them to run (and they clearly ignored)?

cklosowski:
@chriscct7 we should fix that then :P

chriscct7:
I think we can just remove the customer column if the routine hasn't been run yet in terms of handling it

cklosowski:
@mindctrl I don't think that's the solution. I think we should make the first and last name from teh purchase show (if they exist) just not link.

pippinsplugins:
If the upgrade hasn't run, I vote we just fallback to the current link.

chriscct7:
@pippinsplugins the current link doesn't work if the routine hasn't run ;) See https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/3799#issuecomment-140909019

pippinsplugins:
@chriscct7 Yes it does. It uses the email in payment meta, which has always been thre.

cklosowski:
@pippinsplugins @chriscct7, how about Chris just makes sure it works...if it already does, less work for him and more Mtn Dew time. If it doesn't, he can make it work again

pippinsplugins:
lol I like that method

chriscct7:
@pippinsplugins no it doesn't because you removed that: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/264690b1592b66ee33e684eca0bc0de40395f0de :P

I'll make it work :). I like Mountain Dew.

chriscct7:
Related: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3571

pippinsplugins:
I vote we go ahead and just merge this as is.

pippinsplugins:
Actually we'll fall back to a link that links to payment history search based on email: https://cloudup.com/cZUkvA6k3BU

Issue #3800, opened 2015-09-16
Mixed-content requests breaking checkout ajax


pippinsplugins:
Since 2.4.4 and the changes made in #3769, the ajax requests on checkout are broken if a site uses HTTPS on the checkout but HTTP on admin.

I've confirmed this on two sites now.

pippinsplugins:
If checkout is loaded on HTTPS, the ajax URL on checkout must also be HTTPS.

On the site that I just saw this one, the checkout screen was HTTPS buy `Enforce SSL on Checkout` was _not_ enabled. I believe that's key to this.

pippinsplugins:
Note: quick fix for anyone seeing this is to either enable `Enforce SSL on Checkout` or make HTTPS site wide.

PR #3801, opened 2015-09-16
Fixing AJAX URL in some cases of Checkout page

cklosowski:
Addresses #3800

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -432,8 +432,10 @@ function edd_get_current_page_url( $nocache = false ) {
 
 	if ( is_front_page() ) {
 		$uri = home_url();
+	} elseif ( edd_is_checkout() ) {
+		$uri = edd_get_checkout_uri();
 	} else {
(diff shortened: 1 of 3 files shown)


includes/misc-functions.php
```diff
@@ -455,8 +457,13 @@ function edd_get_current_page_url( $nocache = false ) {
...
 function edd_add_cache_busting( $url = '' ) {
 
-	if ( edd_is_caching_plugin_active() ) {
-		$url = add_query_arg( 'nocache', 'true', $url );
+	$no_cache_checkout = edd_get_option( 'no_cache_checkout', false );
+	$current_url       = urldecode( $url );
+
+	if ( edd_is_caching_plugin_active() || ( edd_is_checkout() && $no_cache_checkout ) ) {
+		if ( false === strpos( $current_url, 'nocache=true' ) ) {
+			$url = add_query_arg( 'nocache', 'true', $url );
```
pippinsplugins: Why the check? add_query_arg() tests it already. Seems redundant.

cklosowski: fair enough...i'll remove and recheck, it was there for when we were getting a duplicate call.

Issue #3817, opened 2015-09-18
Redirect loop when verifying user if no purchase history page is set in options


thomasgriffin:
I don't know exactly when this got implemented, or how exactly users are "verified" - is this a setting I can turn off?? - but if you do not have a page set for your purchase history in the EDD settings, clicking on the user verification link causes an infinite redirect loop.

It also sends out a zillion "Verify your account" emails, none of which have the proper URL because of no purchase history page being set. Very frustrating for the user, and very frustrated that this was something that appeared to be turned on by default.

Can someone explain how this works?

pippinsplugins:
@thomasgriffin I'm going to ping you on slack to go over the fine details of this. Will explain there.

Is there a reason you don't have a purchase history page set?

PR #3818, opened 2015-09-18
Fix redirect loop when no Purchase History page is set

cklosowski:
#3817


cklosowski:
Travis CI is having issues pulling the WP Core unit test. I've tested this using PHP 5.2 and 5.6 locally. Going to merge and push 2.4.6

PR #3840, opened 2015-09-27
Only make payments for products who are published

chriscct7:
Right now, since there isn't a check for it, running the create payments cli command can make sales of products that aren't published yet. Found while testing 2.5


chriscct7:
Nevermind, further testing reveals that this wasn't the case, which makes sense since by default get_posts doesn't default post_status to containing trash

Issue #3365, opened 2015-04-25
Download History in API 


chriscct7:
Right now we let people export download history as a CSV.
Let's let people access that via the api as well

pippinsplugins:
@ghost1227 This is concerning file downloads, not sales. Sales are already accessible in the API.

ghost1227:
and your point?

pippinsplugins:
The pull request was for sales, not file download logs.

ghost1227:
Bet me

ghost1227:
oh wait... i think i completely misunderstood the point of this issue

ghost1227:
this is supposed to be the data from `edit.php?post_type=download&page=edd-reports&tab=logs` isn't it...

pippinsplugins:
That was my point :)

Yes, this issue is related to pulling the FILE download history via the API.

ghost1227:
Well dammit

ghost1227:
@pippinsplugins resubmitted as requested

pippinsplugins:
In `issue/3365` for testing.

pippinsplugins:
Working perfectly. Merged into release/2.5 and closing.

pippinsplugins:
After a  quick call with @topher1kenobe, we feel the endpoint should be changed from `downloads` to `download-logs`. We use `download` frequently to refer to a product so having `/downloads/` creates confusion and is unclear whether it refers to download products or file downloads.

The other option I like is `/edd-api/file-downloads/`

Opinions @easydigitaldownloads/core-devs ?

chriscct7:
+50k

pippinsplugins:
@chriscct7 Which of the two do you prefer?

spencerfinnell:
Haven't ever looked at the API so not sure if this type of endpoint would even work, but what about `/logs/downloads/`, which would allow expansion for the other available logs.

pippinsplugins:
@spencerfinnell At the moment, no, that wouldn't work but I can see that being a great idea for a v2 or v3.

chriscct7:
I personally strongly prefer Spencers but since that is impossible for the current generation we can at least make the endpoint look close to that so my preference is logs-downloads so that when we do move to it its not such a huge jump

chriscct7:
Then when we do update all we're doing is changing a - to a /

cklosowski:
[...]

PR #3845, opened 2015-09-28
Issue/3365

pippinsplugins:
#3365


includes/api/class-edd-api.php
```diff
@@ -1492,6 +1501,86 @@ public function get_discounts( $discount = null ) {
...
+		$paged    = $this->get_paged();
+		$per_page = $this->per_page();
+		$offset   = $per_page * ( $paged - 1 );
+		$count    = 0;
+
+		if( $customer ) {
+			if( is_numeric( $customer ) ) {
+				$meta_query[] = array(
+					'key'    => '_edd_log_user_id',
+					'value'  => $customer
```
cklosowski: You are using `$customer` but the meta query is for `user`. We need to stop making this confusion. If it's `customer`, use customer ids, if it's `user`, use user ids. But this is just itching to be confused and giving wrong information.

pippinsplugins: Agreed.

PR #3866, opened 2015-10-02
Remove unnecessary code

### LINKED ISSUE #3865: Unnecessary code
As far as I can see, these 2 lines do nothing:

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/template-functions.php#L187-L188

When adding a download to the cart, the "Added to cart" message along with the checkmark is the only thing shown. I have a feeling that the anchor link is supposed to wrap around the "Added to cart" text. I think we should just remove it since the button turns into "checkout".

sumobi:
Fixes #3865


cklosowski:
looks fine, merging.

Issue #3492, opened 2015-06-10
Add primary columns to WP_List_Tables


pippinsplugins:
WP core is working on responsive list tables: https://core.trac.wordpress.org/ticket/32395#comment:25

It should work out of the box but it is better if we define a [primary column](https://core.trac.wordpress.org/ticket/25408).

pippinsplugins:
@mindctrl or @topher1kenobe one of you want to tackle this? I can help if you're unsure what to do from the ticket above.

mindctrl:
I'll take it.

pippinsplugins:
:+1:

pippinsplugins:
@mindctrl Think you could hit this up next week?

mindctrl:
@pippinsplugins :+1:

PR #3868, opened 2015-10-02
Add default primary column name to list table classes. #3492.

mindctrl:
#3492


includes/admin/payments/class-payments-table.php
```diff
@@ -247,6 +247,18 @@ public function get_sortable_columns() {
...
 	/**
+	 * Gets the name of the default primary column.
+	 *
+	 * @since 2.5
+	 * @access protected
+	 *
+	 * @return string Name of the default primary column.
+	 */
+	protected function get_default_primary_column_name() {
+		return 'ID';
```
pippinsplugins: I think this should be the email column. ID isn't really important

includes/admin/reporting/class-sales-logs-list-table.php
```diff
@@ -52,6 +52,18 @@ public function __construct() {
...
 	/**
+	 * Gets the name of the default primary column.
+	 *
+	 * @since 2.5
+	 * @access protected
+	 *
+	 * @return string Name of the default primary column.
+	 */
+	protected function get_default_primary_column_name() {
+		return 'ID';
```
pippinsplugins: This should be the download name I think.

mindctrl: @pippinsplugins I agree. I originally tried that but it looks like this:

image

Same for some of the other ones. Thoughts about that? Seems like any 'primary' column that isn't the first one shows up weird like that. Am I missing something?

mindctrl: ^ my comment was meant for the payments table file. But same thing applies here.

pippinsplugins: That looks like it has two primary columns. When collapsed it should only have one column. That's weird.

includes/admin/payments/class-payments-table.php
cklosowski: I think I disagree here. Payments are made in chronological order, and setting the email as the primary column doesn't achieve us much, as it's not something you'd really sort on either. The ID is the unique value in the payments world, not the email.

cklosowski:
Shouldn't we be using `get_primary_column_name` not `get_default_primary_column_name`? Must like sortable columns, the items in the extended class shouldn't use the `default` getter methods.

cklosowski:
Just realized this was to be into 2.5, going to close this and repoen it for 2.5

PR #3879, opened 2015-10-06
Fix broken test (due to typo being fixed)

JeroenSormani:
Hi!

Long time since my last commit, have been quite busy! Trying to pick it back up..

Noticed the tests were failling locally and with other PRs, probably due to a typo that has been fixed.
This fixes the broken unit test.

PR #3880, opened 2015-10-06
Fix test warnings, removing useless 'use \CLASS'

JeroenSormani:
This may be related to my specific PHPUnit version, as I didn't see it on any of the TravisCI tests so far, but it seems like a improvements.

I was getting these warnings:
image

Which is fixed in this PR. 

Importing classes from the global namespace without any alias doesn't have any purpose as far as I know :smile_cat:

### DIFF
--- a/tests/tests-api.php
+++ b/tests/tests-api.php
@@ -1,11 +1,8 @@
 <?php
 
-use \EDD_Roles;
-
 /**
  * @group edd_api
  */
-
 class Tests_API extends WP_UnitTestCase {
 	protected $_rewrite = null;
 
--- a/tests/tests-emails.php
+++ b/tests/tests-emails.php
@@ -1,7 +1,5 @@
 <?php
 
-use \EDD_Email_Template_Tags;
-
 /**
  * @group edd_emails
  */
--- a/tests/tests-payments.php
+++ b/tests/tests-payments.php
@@ -1,6 +1,5 @@
 <?php
 
-use \EDD_Payments_Query;
 /**
  * @group edd_payments
  */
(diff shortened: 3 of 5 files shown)


pippinsplugins:
This goes hand in hand with #3879. Once both merged, it should start passing.

pippinsplugins:
@JeroenSormani I may be wrong, but I believe the `use` statement is required for older PHP Versions.

JeroenSormani:
@pippinsplugins Hmm, that would be a first time hearing for me. 
AFAIK the 'use' is used by namespaces, which were added in 5.3+ (though there is another use case with `function( $y ) use ( $x ) [..]`, which isn't relevant here).

So that made me think this code would be incompatible with PHP 5.3-, which isn't such a big deal as it are unit tests and not the core code and TravisCI isn't doing 5.2 tests.

Do you have any info about `use` for older PHP versions?

Thanks!

pippinsplugins:
Let's ping @cklosowski and @chriscct7. Any thoughts?

chriscct7:
So what happened here is we used to namespace all of our unit test files with EDD_Unit_Tests at the top, so once you do that you have to also then use the classes involved inside those files.

About a year after we did this, Pippin and I decided to move to a system based on WP CLI's unit tests which use the WordPress core tests. Thus we dropped the custom namespace we were doing at the top of all the files, but in the process, the `use` was never removed.

Addition of use: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/553a0babbc2ef7cb6704a2d499c6a3b540e413a0

1 year later
Removal of namespaces: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/eef27c63806c1cc5d0139e37b9d7229c2ce173f4

You should be perfectly fine to remove them

pippinsplugins:
Ah! Thank you for that @chriscct7, I had completely forgotten we had done that.

Issue #3883, opened 2015-10-06
Record error message in log when wp_mail() fails 


pippinsplugins:
We depend on `wp_mail()` a lot. It has a nice feature that causes it to return `false` if the email fails to send.

To help diagnose and identify customer cases where email receipts are not being sent, we should log (and possibly display) an error message when an email fails to send.

We don't be able to identify why the email failed to send, but at least we will not if it fails.

pippinsplugins:
I'm proposing that we use a general `edd-errors.log` file for storing log entries like this. We can quickly access it and identify if there are problems.

When emails fail to send, I'm logging the following:
```
Email from Easy Digital Downloads failed to send.
Send time: October 6 2015 11:28:30
To: {to address here}
Subject: {subject here}
Message: {Full email message here}
```

pippinsplugins:
Reminder that we need to update our emails not sending doc when this is released: https://trello.com/c/Qar0dvTE/105-update-emails-not-sent-doc-with-error-logging

PR #3884, opened 2015-10-06
Log email send failures in log file. #3883

pippinsplugins:
#3883

### DIFF
--- a/includes/emails/class-edd-emails.php
+++ b/includes/emails/class-edd-emails.php
@@ -295,6 +295,19 @@ class EDD_Emails {
 
 		$sent = wp_mail( $to, $subject, $message, $this->get_headers(), $attachments );
 
+		if( ! $sent ) {
+
+			$log_message = sprintf(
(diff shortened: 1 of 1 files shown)


cklosowski:
There are a few cases where sending with stuff like mandrill may not be caught since they don't return `false`, but since core returns `false` we can just go with that.

WPMandrill looks like this:
```
try {
	$sent = wpMandrill::mail( $to, $subject, $message, $headers, $attachments );

	if (    is_wp_error($sent)
	        || !isset($sent[0]['status'])
	        || ($sent[0]['status'] != 'sent' && $sent[0]['status'] != 'queued') ) {

        return wpMandrill::wp_mail_native( $to, $subject, $message, $headers, $attachments );
    }

	return true;
} catch ( Exception $e ) {
	return wpMandrill::wp_mail_native( $to, $subject, $message, $headers, $attachments );
}
```

Do we care?

pippinsplugins:
I don't think we care. If it returns false, it failed, at least as far as I can tell.

Issue #3902, opened 2015-10-12
Missing Personal Info fields when using "Login Form Only" with guest checkouts enabled


mindctrl:
When EDD is configured to:
- allow guest checkouts
- show only the login form on the checkout screen

the `Personal Info` section doesn't show on the checkout screen. This effectively makes it unusable unless you do have a login. 

##### EDD Config
![image](https://cloud.githubusercontent.com/assets/1254870/10433404/a1d278fe-70df-11e5-9078-f918ce185328.png)

##### Checkout
![image](https://cloud.githubusercontent.com/assets/1254870/10433428/c8ffefa6-70df-11e5-8a5f-f97fe3a575ad.png)

When configured in this way, the `Personal Info` fields should be displayed so that customers who do not have a login can still purchase. Like this:

![image](https://cloud.githubusercontent.com/assets/1254870/10433475/25bb7f8a-70e0-11e5-80ca-025742178b4f.png)

pippinsplugins:
Replicated.

mindctrl:
This works now. Opened https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3905 for UI work related to this for 2.6 milestone.

SDavisMedia:
Yes, all good here. :+1:

pippinsplugins:
Great, will let @cklosowski review and merge.

cklosowski:
Looks good to me, verified and tested it out.

PR #3903, opened 2015-10-12
Issue/3902

pippinsplugins:
#3902


cklosowski:
merging, failures are due to another issue we're working through.

Issue #3911, opened 2015-10-14
Payment Method fields are not hidden when a 100% discount code is applied


mindctrl:
When a 100% discount code is applied with ajax enabled, we hide the credit card and billing fields, but not the payment method fields. If you refresh the page, the payment methods are properly hidden.

pippinsplugins:
Works perfectly.

PR #3912, opened 2015-10-14
Hide the payment method box when 100% discount applied. 

mindctrl:
#3911


pippinsplugins:
@mindctrl Can you please re-open against `release/2.5`?

PR #3923, opened 2015-10-15
Invalid HTML markup fixes for edd_receipt

### LINKED ISSUE #3922: edd_render_receipt_in_browser() outputs invalid markup
Hey Guys,

Just noticed that the HTML output from the `edd_render_receipt_in_browser()` function here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/emails/template.php#L239 is not valid.

1. Title tag should be wrapped in the head per W3 standards: "Every HTML document must have a TITLE element in the HEAD section." http://www.w3.org/TR/html4/struct/global.html#edef-TITLE

2. `wp_head()` is used without a 

3. the HTML tag incorrectly closes before the body tag begins - "After document type declaration, the remainder of an HTML document is contained by the HTML element." http://www.w3.org/TR/html4/struct/global.html#edef-HTML

10-15-2015 4-38-49 pm

DevinWalker:
Fixes #3922

### DIFF
--- a/includes/emails/template.php
+++ b/includes/emails/template.php
@@ -246,10 +246,11 @@ function edd_render_receipt_in_browser() {
 ?>
 <!DOCTYPE html>
 <html lang="en">
-	<title><?php _e( 'Receipt', 'easy-digital-downloads' ); ?></title>
-	<meta charset="utf-8" />
-	<?php wp_head(); ?>
-</html>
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@DevinWalker Could you resubmit to `release/2.5`?

PR #3926, opened 2015-10-17
Escape downloads notes and option name in receipt

christ0ph3r:


christ0ph3r:
Actually this will break p tags if someone has line breaks in the download notes... I did not know that wpautop created multiple p tags and didn't test with multiple lines till now.  Not sure how to handle line 158.

christ0ph3r:
used wp_kses_post instead to handle multiple p tags.  Feedback?

templates/shortcode-receipt.php
```diff
@@ -150,12 +150,12 @@
 						<div class="edd_purchase_receipt_product_name">
 							<?php echo esc_html( $item['name'] ); ?>
 							<?php if( ! is_null( $price_id ) ) : ?>
-							<span class="edd_purchase_receipt_price_name">&nbsp;&ndash;&nbsp;<?php echo edd_get_price_option_name( $item['id'], $price_id, $payment->ID ); ?></span>
+							<span class="edd_purchase_receipt_price_name">&nbsp;&ndash;&nbsp;<?php echo esc_attr( edd_get_price_option_name( $item['id'], $price_id, $payment->ID ) ); ?></span>
```
pippinsplugins: This is not an attribute so does not need escaped in the same way.

templates/shortcode-receipt.php
```diff
@@ -150,12 +150,12 @@
...
 							<?php endif; ?>
 						</div>
 
 						<?php if ( $edd_receipt_args['notes'] ) : ?>
-							<div class="edd_purchase_receipt_product_notes"><?php echo wpautop( edd_get_product_notes( $item['id'] ) ); ?></div>
+							<div class="edd_purchase_receipt_product_notes"><?php echo wp_kses_post(wpautop( edd_get_product_notes( $item['id'] ) ) ); ?></div>
```
pippinsplugins: wp_kses_post() is for sanitizing input, not escaping output.

templates/shortcode-receipt.php
christ0ph3r: Your right.  Well how should it be escaped?

pippinsplugins: `esc_html()` would be appropriate except that breaks custom code where people have added HTML to the names.

It could use striptags or similar but would need to allow standard HTML

christ0ph3r: Hey thanks for the help.  I have this pretty much done just got 2 questions.

1.  What tags you think should be allowed?  strip_tags function can take a 2nd parameter for allowed tags.  

2.  Should I submit a new PR?  I feel like I made this one messy.

templates/shortcode-receipt.php
pippinsplugins: The global allowedtags should work: https://core.trac.wordpress.org/browser/tags/4.3.1/src/wp-includes/kses.php#L49

christ0ph3r: How do you use that variable?  I don't understand how to do that.  I tried this and tried with making the variable global at the top of the page but had no luck. 

```
<?php echo strip_tags( edd_get_product_notes( $item['id'] ), $allowedposttags ); ?>
```

pippinsplugins: Add `global $allowedposttags;`

christ0ph3r: Hey Pippin I am struggling with this.  I tried the following and got Parse error: syntax error, unexpected 'global' (T_GLOBAL) in /var/www/html/wp-content/plugins/easy-digital-downloads/templates/shortcode-receipt.php on line 158 .  I tried a lot of other ways and could not get it to work so I must be messing up somewhere.

```
<?php echo strip_tags( edd_get_product_notes( $item['id'] ), global $allowedposttag; ); ?>
```

mindctrl: @christ0ph3r you can add the $allowedposttags at the top with the other global we already have.

```
global $edd_receipt_args, $allowedposttags;
```

Then just use `$allowedposttags` as the second param in `strip_tags()`.

`<?php echo strip_tags( edd_get_product_notes( $item['id'] ), $allowedposttags; ); ?>`

christ0ph3r: @mindctrl thanks.   I realized my last post I was trying with $allowedposttag and not $allowedposttags on accident.  I went ahead and tried your suggestions.

I now have 

line 5
```
 global $edd_receipt_args, $allowedposttags;
```

and 

line 158
```
<?php echo strip_tags( edd_get_product_notes( $item['id'] ), $allowedposttags ); ?>
```

I had to take out the semi colon right after $allowedposttags since it threw syntax error.   I get the following error now.

Notice: Array to string conversion in /var/www/html/wp-content/plugins/easy-digital-downloads/templates/shortcode-receipt.php on line 158

mindctrl: @christ0ph3r Ah yes, my bad on the semi-colon. So `strip_tags` requires the second parameter be a string.

Pretty sure you *can* use `wp_kses_post()` for escaping output: https://vip.wordpress.com/documentation/best-practices/security/validating-sanitizing-escaping/#escaping-securing-output

https://developer.wordpress.org/plugins/security/securing-output/

Though there are possible performance concerns with it. I'll let @pippinsplugins or @cklosowski chime in on the best way to handle this.

cklosowski: I see no issue with using `wp_keses_post` here. 

> There has historically been a perception that `wp_kses()` is slow. While it is a bit slower than the other escaping functions, the difference is minimal and does not have as much of an impact as most slow queries or uncached functions would.

I say we move forward with this... I'm going to close this PR, but branch it into an issue branch, which will keep the commit history in tact and get it milestoned.

PR #3928, opened 2015-10-19
Change capability for the Tools page

grappler:
The reason for this change is that the tools page is hidden when the constant `DISALLOW_FILE_MODS` is defined. This is a bigger issue for me as the fraud monitor settings are in the tools page. 

There does not seem to be any reason to limit the access to those who can install plugins.


pippinsplugins:
I agree with this change.

retlehs:
there's still `install_plugins` used for the 'Easy Digital Download Extensions' page. since that page doesn't actually install plugins via the wordpress admin, shouldn't it be changed as well?

fwiw there's also EDD extensions that are using `install_plugins` in a way that is wrong, such as the product updates plugin.

pippinsplugins:
The extensions page could go either way really. On one hand, it makes sense to only show them to users that have the ability to install plugins. On the other hand, it makes sense that store administrators should be able to see the extensions even if they can't install the plugins as they could then report to store admins the ones they want.

retlehs:
on a bedrock-like setup (`DISALLOW_FILE_MODS` enabled on all environments except dev) you'd only be able to see/reach the extensions list if you were on the development environment. since it's usually devs on the dev environment, store admins are currently missing out on quickly checking out some EDD extensions that they'd want to have installed

this is not really a big deal for the extensions page specifically. i mentioned the EDD product updates plugin — currently unusable on bedrock-like production environments because of `install_plugins`. i'm not sure what other EDD plugins (if any) do the same thing. should i just create a support ticket for that plugin?

pippinsplugins:
Let's change it :)

Note: add-ons are managed separately from core so this change will reflect core only. For add-ons, please do create a ticket and it'll get resolved.

PR #3929, opened 2015-10-19
edd_receipt now uses filter_var with FILTER_VALIDATE_BOOLEAN to deter…

### LINKED ISSUE #3919: [edd_receipt] shortcode options not working as expected
**Scenario:** A user doesn't want to display a receipt item like the date so they use the shortcode like this: `[edd_receipt price="false" date="false"]`

**Problem:** EDD isn't taking into account true/false values as strings rather than booleans and not respecting them:

10-14-2015 4-21-53 pm

**Solution:** `filter_var( $edd_receipt_args['date'], FILTER_VALIDATE_BOOLEAN )`

Please verify so I can submit a PR

DevinWalker:
…mine boolean option

Returns TRUE for "1", "true", "on" and "yes"
Returns FALSE for "0", "false", "off" and "no"
Returns NULL otherwise.

Fixes #3919

### DIFF
--- a/templates/shortcode-receipt.php
+++ b/templates/shortcode-receipt.php
@@ -26,7 +26,7 @@ $status    = edd_get_payment_status( $payment, true );
 	<thead>
 		<?php do_action( 'edd_payment_receipt_before', $payment, $edd_receipt_args ); ?>
 
-		<?php if ( $edd_receipt_args['payment_id'] ) : ?>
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@DevinWalker could you resubmit to `release/2.5`?

pippinsplugins:
@DevinWalker could you resubmit to `release/2.5`?

PR #3938, opened 2015-10-23
Fix "Remove from Cart" ajax request when there are multiple purchase forms on same page

maxrossi:
Fix "Remove from Cart" ajax request: purchase forms on the page can be more than one


pippinsplugins:
Looks right to me.

We'll test further and merge if everything checks out.

Thanks!

maxrossi:
OK thank you!

PR #3940, opened 2015-10-23
filter_var and FILTER_VALIDATE_BOOLEAN for shortcode conditional checks

### LINKED ISSUE #3919: [edd_receipt] shortcode options not working as expected

DevinWalker:
Resubmitted to release/2.5 - Fixes #3919


pippinsplugins:
Looks good to me.

Issue #3927, opened 2015-10-19
cart widget breaks Add to Cart button functionality


SDavisMedia:
Visit a page that has an Add to Cart button on it for a product. Make sure the cart widget is NOT showing on that page anywhere. Click the Add to Cart button as many times as you can before it turns into the Checkout button (yes... literally click it fast several times). When you go to checkout, there should only be one item in the cart.

Do the same test again but this time, do it from a page that has the cart widget displaying somewhere on the screen. It's adding the product to the cart one time for every time the Add to Cart button was clicked. 

Let me know if you guys are getting the same behavior.

SDavisMedia:
Though I can't replicate this myself, the user is reporting that even if the cart widget is not present, he is still able to add the item multiple times. So the fix should cover all angles.

mindctrl:
I was just able to replicate this. Here's a screencast: http://cl.ly/1a3p2I2p3x2F

I was also able to replicate it without the widget.

ZackWard:
It looks like this is happening because the add to cart button makes an ajax request for each click until the button is actually replaced with the "checkout" button. A user can add duplicates of the item each time they click the button until the ajax call actually returns from the server. I've added an ajax lock object that will cause the request to the server not to be made if there is already a current request pending.

pippinsplugins:
@zackkatz I believe you are correct. The change proposal in #3953 looks like it should work fine, though I think simply disabling the input field would be simpler and would prevent any `on('click')` event from firing as well, instead of just the ajax.

Thoughts?

zackkatz:
I think Pippin meant to ping @ZackWard

pippinsplugins:
Whoops!

ZackWard:
@pippinsplugins You're right, that would be much simpler. I'll work on an updated PR.

pippinsplugins:
Changes in #3959 look like they should work fine.

ZackWard:
@pippinsplugins Should I modify the minified js or will that be taken care of automatically by the build process?

pippinsplugins:
@ZackWard we do that pre-release.

cklosowski:
Thanks @ZackWard. I've merged this into master, will get it wrapped back up into 2.5 shortly.

PR #3953, opened 2015-10-29
Prevent rapid addition of items via ajax - Issue #3927

ZackWard:
A user may click the "add to cart" button for a download multiple times while the ajax request to the server is made. These changes introduce an ajax lock object that will prevent this behavior.


ZackWard:
This addresses issue #3927

ZackWard:
Closing PR, better solution found.

PR #3959, opened 2015-10-29
Disable add to cart button during ajax request

ZackWard:
Disable the add to cart button during ajax requests to prevent items from being added multiple times due to user rapidly clicking the button (issue #3927). The add to cart button is re-enabled when the ajax request is complete, so that it can be used again in the event that the user removes the item from the cart and then needs to re-add it.


cklosowski:
:+1:

PR #3967, opened 2015-10-30
Add unused ticksize_num graph option

stephywells:
$graph = new EDD_Graph( $data );
$graph->set( 'ticksize_num', 5 );

This still results in in a ticksize of 1. This patch uses the ticksize_num option instead of always using 1.


pippinsplugins:
Looks good to me. We can merge once 2.5 is closer to release. Thanks @stephywells !

stephywells:
I added one more small change to this, so the PayPal transaction link will go directly to the transaction.

pippinsplugins:
@stephywells While the transaction ID link is great, that's unrelated to this PR.

I'm going to go ahead and merge it but please ensure that the only changes in a PR are for the issue at hand.

Thanks!

stephywells:
Oops. Didn't mean for that one to be in the same PR. Sorry about that.

Issue #3803, opened 2015-09-16
Symlink file download option is broken


pippinsplugins:
In 1.5 (#786) we introduced an option to create symbolic links during the file download process.

This feature was designed to only be used with the `Forced` download method as a way to make large file downloads more reliable.

In #1160 we changed the `Symlink File Downloads` option to only work with the `Redirect` download method since it was no longer compatible with the improved `.htaccess` rules we introduced.

In #3459 it was discovered (and fixed) that the JS toggle on the `Symlink File Downloads` option did not properly display when the `Download Method` was set to `Redirect`.

Today we discovered that even though the symlink option is only available with using the `Redirect` download method, the feature itself only runs when using the `Forced` download method.

We need to update our logic to properly support symlinks on the `Redirect` download method.

In `process-download.php`, we run this:

```php
if( $direct ) {
	edd_deliver_download( $file_path );
} else {
	// The file supplied does not have a discoverable absolute path
	header( "Location: " . $requested_file );
}
```
If `Forced` ($direct) is the download method, we pass the file to `edd_deliver_download()`. This function is responsible for creating / redirecting to the symlink:

```php
/**
 * Deliver the download file
 *
 * If enabled, the file is symlinked to better support large file downloads
 *
 * @access   public
 * @param    string    file
 * @return   void
 */
function edd_deliver_download( $file = '' ) {
	$symlink = edd_get_option( 'symlink_file_downloads', false );
	$symlink = (bool) apply_filters( 'edd_symlink_file_downloads', $symlink );

	/*
	 * If symlinks are enabled, a link to the file will be created
	 * This symlink is used to hide the true location of the file, even when the file URL is revealed
	 * The symlink is deleted after it is used
	 */

	if( $symlink && function_exists( 'symlink' ) ) {

		// Generate a symbolic link
		$ext       = edd_get_file_extension( $file );
		$parts     = explode( '.', $file );
		$name      = basename( $parts[0] );
		$md5       = md5( $file );
		$file_name = $name . '_' . substr( $md5, 0, -15 ) . '.' . $ext;
		$path      = edd_get_symlink_dir() . '/' . $file_name;
		$url       = edd_get_symlink_url() . '/' . $file_name;

		// Set a transient to ensure this symlink is not deleted before it can be used
		set_transient( md5( $file_name ), '1', 30 );

		// Schedule deletion of the symlink
[...]

PR #3975, opened 2015-11-03
Make symlink file downloads work with redirect download method

pippinsplugins:
#3803


cklosowski:
Seems to be working great for me. Merging into `release/2.5`

Issue #3985, opened 2015-11-04
Add CSV Product Export


pippinsplugins:
CSV product export is currently handled by EDD CSV Manager, which we all know has some pain points. We should make use of our new bulk processing API and add this to core.

I've set this for 2.5 simply because we've had 5+ tickets in the last month about CSV Manager not working. I'm going to write this now and get it in.

pippinsplugins:
I have this nearly complete. There are just a few column values that are not being populated properly.

cklosowski:
Merged into `release/2.5` for further testing, seems to be working great alongside #3633

pippinsplugins:
I'd like for @SDavisMedia and/or @mindctrl to give this a quick look before we close it out.

mindctrl:
@pippinsplugins found an issue here where columns get shifted, possibly due to quotes/shortcodes in post_content. Reproduced on two sites.

Steps to reproduce:
- Create new download
- Insert an audio playlist in it. Example: [playlist ids="44884,44865"]
- Publish post
- Export CSV

The CSV columns get broken:
![image](https://cloud.githubusercontent.com/assets/1254870/10983443/7b173f4c-83e0-11e5-9c52-07158260a7e0.png)

Results in all following data getting shifted over to the right one column.

mindctrl:
Actual shortcode from previous screenshot was `[playlist ids="276,278"]`

mindctrl:
More data:

This:
`This is a "test" dude. "What we have here," moo.`

Results in:
![image](https://cloud.githubusercontent.com/assets/1254870/10983515/e4fd31fa-83e0-11e5-9e27-6fc663311802.png)

pippinsplugins:
@mindctrl try again now on `issue/3985`

pippinsplugins:
@cklosowski Could you test this once more as well?

pippinsplugins:
Once 2.5 goes out, we need to remember to update `EDD CSV Manager` to remove the product export from there.

pippinsplugins:
This has all been merged into `release/2.5` but still needs further testing to confirm everything is finalized.

cklosowski:
Looks good to me, the one thing I see missing is the bundle data coming in. Is that something we're not supporting, since it would need to be configured after the other products are imported, and their IDs might be different?

pippinsplugins:
Maybe the export could contain the product names that are bundled?

cklosowski:
[...]

Issue #3633, opened 2015-07-22
Add recount / reset earnings tools to core


pippinsplugins:
Let's merge the [recount earnings extension](https://github.com/easydigitaldownloads/EDD-Recount-Earnings) into core and improve it.

- Use bulk processing
- Option to recount earnings / sales on products
- Option to reset earnings / sales on products (should NOT delete payments)
- Option to recount store-wide earnings (or maybe this is part of recounting products)

cklosowski:
I'm going to pull down what you have and finish this up.

pippinsplugins:
:+1:

cklosowski:
Ok, so this is my initial work on it. Currently it'll do recount the store earnings, recount an individual download's earnings & sales, recount ALL downloads earnings & sales w/ store earnings update, and then a reset earnings and sales.

*NOTE:* Currently this _does not_ delete the sales logs, so the reset leaves those present. I'm doing this for the mean time, to allow the recounts to be much easier to test. I've still go to work out the performance side of deleting logs, etc.

cklosowski:
Some UX for you:
![screen shot 2015-09-22 at 11 14 05 am](https://cloud.githubusercontent.com/assets/1390124/10027332/3946419a-611b-11e5-9dd7-d7d308e0e8eb.png)
![screen shot 2015-09-22 at 11 09 14 am](https://cloud.githubusercontent.com/assets/1390124/10027334/3965cc5e-611b-11e5-8e5d-913455fd0cef.png)
![screen shot 2015-09-22 at 11 09 08 am](https://cloud.githubusercontent.com/assets/1390124/10027336/39663d74-611b-11e5-94c6-bf37ac451e7b.png)
![screen shot 2015-09-22 at 11 14 08 am](https://cloud.githubusercontent.com/assets/1390124/10027337/3966a17e-611b-11e5-9dce-09e28e5717a3.png)
![screen shot 2015-09-22 at 11 14 16 am](https://cloud.githubusercontent.com/assets/1390124/10027333/3961e1a2-611b-11e5-94f3-acc121ad9bc4.png)
![screen shot 2015-09-18 at 4 15 05 pm](https://cloud.githubusercontent.com/assets/1390124/10027335/3966200a-611b-11e5-93fe-2041ca364d4f.png)

cklosowski:
[...]

PR #3986, opened 2015-11-05
Add batch export UI / class for products - Issue/3985

pippinsplugins:
#3985


cklosowski:
Works great, merging so we can get it tested with the updates in #3633

Issue #3710, opened 2015-08-14
Multiple API pings per page load


stephywells:
When I view the plugins page, wp_remote_post is triggered twice for each add-on I have installed.
[14-Aug-2015 19:25:08 UTC] https://easydigitaldownloads.com/edd-sl-api/ Content Restriction 
[14-Aug-2015 19:25:09 UTC] https://easydigitaldownloads.com/edd-sl-api/ Manual Purchases 
[14-Aug-2015 19:25:09 UTC] https://easydigitaldownloads.com/edd-sl-api/ PDF Invoices 
[14-Aug-2015 19:25:09 UTC] https://easydigitaldownloads.com/edd-sl-api/ Software Licensing 
[14-Aug-2015 19:25:16 UTC] https://easydigitaldownloads.com/edd-sl-api/ Content Restriction 
[14-Aug-2015 19:25:17 UTC] https://easydigitaldownloads.com/edd-sl-api/ Manual Purchases 
[14-Aug-2015 19:25:17 UTC] https://easydigitaldownloads.com/edd-sl-api/ PDF Invoices 
[14-Aug-2015 19:25:18 UTC] https://easydigitaldownloads.com/edd-sl-api/ Software Licensing 

This is adding a lot of time to the page load. I'm planning on adding more add-ons as well so this will add even more time to my page load.

I'm launching a new plugin and am hoping to use EDD to do it. Before I spend time on adding transients to save the API responses in the EDD_SL_Plugin_Updater class, I want to make sure that's a direction your team is okay with going. I've done this in Formidable in order to reduce the load on our API site as well. Let me know if I can add these transients in. Thanks!

pippinsplugins:
Are you seeing it on every page load or just one time (with follow up loads pulling from cache)?

stephywells:
I'm seeing it on every page load. Is there already some caching in place?

pippinsplugins:
We hook into WP core's transients API.

Do you have `set_site_transient( 'update_plugins', null )` uncommented at the top of the plugin updater class? That would cause it to run on every page load.

stephywells:
No, that line is commented, and I've disabled all other plugins. The transient is getting set twice on the page load, so the remote post is also being triggered twice. In Formidable, I'm saving the response from wp_remote_post into a different transient to prevent any of the add-ons from checking in more than once every 12 or 24 hours. This looks like it would be a lot of strain on the api site.

[...]

PR #3991, opened 2015-11-05
Fix Multiple API pings per page load

stephywells:
My server has been getting bombarded since releasing a version with EDD licensing support. I've included this change in the version of this class packaged with the plugin, but I'd love for it to be in the core. Thank you!
Fix https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3710

### DIFF
--- a/includes/EDD_SL_Plugin_Updater.php
+++ b/includes/EDD_SL_Plugin_Updater.php
@@ -295,7 +295,7 @@ class EDD_SL_Plugin_Updater {
             'url'        => home_url()
         );
 
-	$cache_key = md5( 'edd_plugin_' . sanitize_key(  $api_params['license'] . $api_params['item_name'] . $api_params['item_id'] ) . '_' . $api_params['edd_action'] );
+	$cache_key = md5( 'edd_plugin_' . sanitize_key(  $api_params['license'] ) . '_' . $api_params['edd_action'] );
 	$cached_response = get_transient( $cache_key );
 	if ( $cached_response ) {
 		// if this has been checked within 24 hours, don't check again
--- a/includes/EDD_SL_Plugin_Updater.php
+++ b/includes/EDD_SL_Plugin_Updater.php
@@ -310,16 +310,11 @@ class EDD_SL_Plugin_Updater {
 
         if ( $request && isset( $request->sections ) ) {
             $request->sections = maybe_unserialize( $request->sections );
+            set_transient( $cache_key, $request, DAY_IN_SECONDS );
         } else {
             $request = false;
+            set_transient( $cache_key, 0, DAY_IN_SECONDS );
         }
-
-		if ( $request === false ) {
-			// avoid saving false in the transient to prevent repeat checks
-			set_transient( $cache_key, 0, DAY_IN_SECONDS );
-		} else {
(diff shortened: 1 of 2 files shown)


includes/EDD_SL_Plugin_Updater.php
```diff
@@ -295,6 +295,13 @@ private function api_request( $_action, $_data ) {
             'url'        => home_url()
         );
 
+	$cache_key = md5( 'edd_plugin_' . sanitize_key( $api_params['item_name'] . $api_params['item_id'] ) . '_' . $api_params['edd_action'] );
```
pippinsplugins: `item_name` and `item_id` are not always included. How about passing in the license key instead of item_id and item_name.

stephywells: That's updated now. Thank you!

pippinsplugins:
Looks great.

@cklosowski any objections to this?

stephywells:
Pagely is reporting some sites that are pinging our site with every single page load. We really need this change in place. Our site has been suffering from a lot of lag since we switched over. Please merge this?

pippinsplugins:
It's tagged for 2.5. Can you apply the merge to your site manually until
released?

I'm still baffled why your site sees the problems but we don't, nor does
anyone else.

We are also on pagely

stephywells:
Thank you! I've made this change in our version as well, so it should help with our next release this week. Pagely came to us to report problems, and they are saying they see others with the same problems with EDD. Maybe it's the quantity of running sites right off the bat, so it's easier to see the sudden change?

pippinsplugins:
That could be it. We've never received another report of it.

The changes are good regardless.

includes/EDD_SL_Plugin_Updater.php
```diff
@@ -307,6 +314,8 @@ private function api_request( $_action, $_data ) {
             $request = false;
         }
 
+	set_transient( $cache_key, $request, 60*60*24 );
```
cklosowski: Can we use `DAY_IN_SECONDS` instead of `60*60*24`? It's just easier to read on a quick pass through.

includes/EDD_SL_Plugin_Updater.php
```diff
@@ -295,6 +295,13 @@ private function api_request( $_action, $_data ) {
             'url'        => home_url()
         );
 
+	$cache_key = md5( 'edd_plugin_' . sanitize_key(  $api_params['license'] . $this->version ) . '_' . $api_params['edd_action'] );
```
cklosowski: Instead of having `edd_plugin_` in the cache key, I'd prefer we move it to outside of that, so when we have issues with transients not clearing (we've been seeing reports of this on some hosts) it's easier for us to identify EDD based transients.

cklosowski:
Looks like a good change, left a few comments. Let me know if you are good with those @pippinsplugins @stephywells

includes/EDD_SL_Plugin_Updater.php
cklosowski: On second thought, talked to @pippinsplugins about this and I think we're prefer to match Core's cache and make it `HOUR_IN_SECONDS * 12`.

stephywells:
I'm made those changes. Let me know if you see anything else.

PR #4004, opened 2015-11-10
Make all products show up on add discount page

jdevalk:
The `number` parameter was missing, causing products not to show up for us.


pippinsplugins:
@jdevalk We can't do this because it wouldn't scale on large sites. We use Chosen with the ajax search for this instead. Simply enter the first four letters of the product name and a search will be performed.

PR #4007, opened 2015-11-11
Multiple reverse proxy IPs in $_SERVER['HTTP_X_FORWARDED_FOR']

geoffdutton:
In situations where there are multiple reverse proxies, there will be multiple IPs stored in $_SERVER['HTTP_X_FORWARDED_FOR'].

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -169,7 +169,9 @@ function edd_get_ip() {
 		$ip = $_SERVER['HTTP_CLIENT_IP'];
 	} elseif ( ! empty( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
 		//to check ip is pass from proxy
-		$ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
+		// can include more than 1 ip, first is the public one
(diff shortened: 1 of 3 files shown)


includes/misc-functions.php
```diff
@@ -169,7 +169,9 @@ function edd_get_ip() {
 		$ip = $_SERVER['HTTP_CLIENT_IP'];
 	} elseif ( ! empty( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
 		//to check ip is pass from proxy
-		$ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
+		// can include more than 1 ip, first is the public one
+		$ip = explode(',',$_SERVER['HTTP_X_FORWARDED_FOR']);
+		$ip = trim($ip[0]);
```
pippinsplugins: If we're detecting multiple IPs, it makes sense that we also store each of the IPs.

cklosowski: @pippinsplugins actually it doesn't. Multiple reverse proxy IPs will likely be inaccessible to the public and actually are likely to be private IP delegations in the private space. The only real IP that we should need to concern ourselves with is the first one `$ip[0]` since it is the first one before being tunneling through an internal only environment and is available via the internet.

Issue #3826, opened 2015-09-23
Quantity not added with variable prices and multi-purchase mode on when ajax is disabled


wpweb101:
Hi

There is still problem in some other cases.

I have disabled AJAX from the global setting page.

See this Screenshot : http://awesomescreenshot.com/06958vl1af

Now on front-side, In Variable Download with multiple purchase option, I selected two different quantiites as per screen shot below :

http://awesomescreenshot.com/0da59cgu2c

But now when I click on add to cart button, it only add 1 product into the cart of both Variable Download Options.

See this Screenshot : http://awesomescreenshot.com/0e559chk28

pippinsplugins:
Duplicate: of https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3804

pippinsplugins:
Re-opening as we've now split the original issue into two separate bugs.

cklosowski:
@wpweb101 @pippinsplugins this should be ready for some testing.

pippinsplugins:
@cklosowski This isn't working for me. I'm testing with Vendd currently.

pippinsplugins:
It works properly when adding the first price option in the list is added; it does not work with any other price options.

cklosowski:
Working fine on Vendd for me. Will retest though. I was able to add any and all price options.

pippinsplugins:
I also tested on Twenty Fifteen and saw the same behavior.

cklosowski:
Here's what I'm seeing:
![screen shot 2015-11-14 at 8 31 47 pm](https://cloud.githubusercontent.com/assets/1390124/11167046/d570625e-8b0e-11e5-81fc-fb09161142e8.png)


![screen shot 2015-11-14 at 8 31 58 pm](https://cloud.githubusercontent.com/assets/1390124/11167044/c933d6a6-8b0e-11e5-9879-455b41d6a34d.png)

cklosowski:
ok I see what you mean. working on it.

cklosowski:
@pippinsplugins give that a shot.

pippinsplugins:
@cklosowski There we go!

PR #4011, opened 2015-11-11
Allow multi select quantities with variable pricing

cklosowski:
#3826


pippinsplugins:
You did this much better and more elegant than my futile attempt a few weeks ago

Looks good at first glance but I haven't tested it yet.

cklosowski:
@pippinsplugins good deal. I built in a unit test as well, that should help us out with making sure it doesn't break in future ones.

PR #4014, opened 2015-11-12
Make state field select instead of text if possible

brashrebel:
If applicable, this gets the states for the given country and makes the state field on the profile editor a select field instead of just plain text.

### DIFF
--- a/templates/shortcode-profile-editor.php
+++ b/templates/shortcode-profile-editor.php
@@ -10,6 +10,8 @@ if ( is_user_logged_in() ):
 	$last_name    = get_user_meta( $user_id, 'last_name', true );
 	$display_name = $current_user->display_name;
 	$address      = edd_get_customer_address( $user_id );
+	$states       = edd_get_shop_states( $address['country'] );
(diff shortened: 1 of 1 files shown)


pippinsplugins:
Looks good to me at first glance.

sdmtt:
I tested this and it works fine, but the field should be updated when the country is changed (`admin-ajax.php?action=edd_get_shop_states`).

pippinsplugins:
@brashrebel Would you be up for updating for the comment above?

brashrebel:
@pippinsplugins yep! I'll get on that soon.

brashrebel:
@pippinsplugins I took a stab at this and was unsuccessful. Think you can point me in the right direction? I have very little ajax experience. I know that in the admin and on the checkout screen there are .js files which include functions for this but there's nothing on the profile editor page to use. Just need a little guidance.

pippinsplugins:
@brashrebel I'm not 100% what is needed here. Will review more after 2.5 is out.

cklosowski:
Pulling this into a new branch and PR to merge directly into `release/2.6` and make any necessary updates.

Issue #4015, opened 2015-11-12
Error while saving Issue


christ0ph3r:
I recently started noticing error while saving issues on one of my sites.  Today I started debugging and EDD seems to be the issue.   I tested on another host, digital ocean with only EDD and 2015 theme.  I just spun up a brand new droplet with wordpress and issue persists still.   It happens when saving pages, posts, and custom post types.

```
### Begin System Info ###

-- Site Info

Site URL:                 http://104.236.186.247
Home URL:                 http://104.236.186.247
Multisite:                No

-- Hosting Provider

Host:                     DBH: localhost, SRV: 104.236.186.247
                 
-- WordPress Configuration

Version:                  4.3.1
Language:                 en_US
Permalink Structure:      /%year%/%monthnum%/%day%/%postname%/
Active Theme:             Twenty Fifteen 1.3
Show On Front:            posts
Remote Post:              wp_remote_post() works
Table Prefix:             Length: 3   Status: Acceptable
WP_DEBUG:                 Disabled
Memory Limit:             40M
Registered Post Stati:    publish, future, draft, pending, private, trash, auto-draft, inherit, refunded, failed, revoked, abandoned, active, inactive

-- EDD Configuration

Version:                  2.4.9
Upgraded From:            None
Test Mode:                Disabled
Ajax:                     Enabled
Guest Checkout:           Enabled
Symlinks:                 Disabled
Download Method:          Direct
Currency Code:            USD
Currency Position:        before
Decimal Separator:        .
Thousands Separator:      ,
Upgrades Completed:       upgrade_payment_taxes,upgrade_customer_payments_association,upgrade_user_api_keys,remove_refunded_sale_logs

-- EDD Page Configuration

Checkout:                 Valid
Checkout Page:            http://104.236.186.247/checkout/
Success Page:             http://104.236.186.247/checkout/purchase-confirmation/
Failure Page:             http://104.236.186.247/checkout/transaction-failed/
Downloads Slug:           /downloads

-- EDD Gateway Configuration

Enabled Gateways:         None

-- EDD Tax Configuration

Taxes:                    Disabled
Tax Rate:                 0
Display On Checkout:      Not Displayed
Prices Include Tax:       No

-- Must-Use Plugins


-- WordPress Active Plugins

Easy Digital Downloads: 2.4.9

-- WordPress Inactive Plugins

Akismet: 3.1.5
Hello Dolly: 1.6

-- Webserver Configuration

[...]

PR #4016, opened 2015-11-13
Only load inline-edit-post when not viewing a single item

cklosowski:
#4015

### DIFF
--- a/includes/scripts.php
+++ b/includes/scripts.php
@@ -172,13 +172,8 @@ function edd_load_admin_scripts( $hook ) {
 	wp_enqueue_script( 'jquery-chosen' );
 
 	$admin_deps = array();
-	if ( empty( $_GET['action'] ) && ( isset( $_GET['post_type'] ) && 'download' === $_GET['post_type'] ) ) {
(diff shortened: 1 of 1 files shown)


includes/scripts.php
```diff
@@ -171,7 +171,19 @@ function edd_load_admin_scripts( $hook ) {
 	wp_register_script( 'jquery-chosen', $js_dir . 'chosen.jquery' . $suffix . '.js', array( 'jquery' ), EDD_VERSION );
 	wp_enqueue_script( 'jquery-chosen' );
 
-	wp_register_script( 'edd-admin-scripts', $js_dir . 'admin-scripts' . $suffix . '.js', array( 'jquery', 'inline-edit-post' ), EDD_VERSION, false );
+	$admin_deps = array();
```
pippinsplugins: @cklosowski This should use our `edd_is_admin_page()` function.

cklosowski:
@pippinsplugins updated.

includes/scripts.php
cklosowski: See below, the update

pippinsplugins: Looks good!

Issue #4017, opened 2015-11-14
Deleting a non-complete payment updates the customer stats


stephywells:
1. Find an abandoned payment and view the user details.
2. Delete the abandoned payment
3. The customer purchase count and amount is now wrong.

cklosowski:
Confirmed this happens on 'Pending' payments as well.

PR #4018, opened 2015-11-14
only update customer stats when payment being deleted is completed 

cklosowski:
#4017

### DIFF
--- a/includes/class-edd-customer.php
+++ b/includes/class-edd-customer.php
@@ -333,6 +333,10 @@ class EDD_Customer {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/class-edd-customer.php
```diff
@@ -333,6 +333,10 @@ public function remove_payment( $payment_id = 0, $update_stats = true ) {
 			return false;
 		}
 
+		if ( 'publish' !== edd_get_payment_status( $payment_id, false ) ) {
```
pippinsplugins: `revoked` should be used here since that is a completed payment (not reversed) that has simply been disabled.

cklosowski:
:+1: looks good to me. will merge when CI is done.

PR #4029, opened 2015-11-18
Unit tests for issue/3473

cklosowski:
#3473

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -653,7 +653,7 @@ function edd_is_discount_maxed_out( $code_id = null ) {
  * @return bool $return
  */
 function edd_discount_is_min_met( $code_id = null ) {
-	$discount = edd_get_discount(  $code_id );
+	$discount = edd_get_discount( $code_id );
 	$return   = false;
 
 	if ( $discount ) {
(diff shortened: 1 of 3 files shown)


includes/discount-functions.php
```diff
@@ -929,7 +929,7 @@ function edd_get_discount_id_by_code( $code ) {
  * @return string $discounted_price Amount after discount
  */
 function edd_get_discounted_amount( $code, $base_price ) {
-	$amount      = 0;
+	$amount      = $base_price;
```
pippinsplugins: This part doesn't make sense to me. Can you explain the logic here?

cklosowski: id you pass an invalid discount, the price after discount was 0

cklosowski: For the notes (discussion happened on Slack), the function is returning the `$base_price` minus the `$rate` of the discount, so the returned value isn't the amount this item has been discounted, but the amount of the item AFTER discounts have been applied, so setting the amount of `0` initially means that if the `$code` passed in is not a valid discount, then the item price after discounts would be returned `0`, instead of the `$base_price`.

PR #4034, opened 2015-11-23
Moved click handler on purchase submission to a named function.

collizo4sky:
I am working on a payment gateway and it is important i temporarily off (.off) the click handler that is run when the purchase submit button is clicked and then on (.on)  and trigger it after a our custom JS validation is complete.

I really need this commit merged to proceed with my work.


collizo4sky:
Simply ignore this PR. i was able to fix the issue in a more graceful way.

PR #4036, opened 2015-11-23
Utilize manage_options instead of install_plugins

etcook:
Many WordPress stacks use DISALLOW_FILE_MODS in production because they follow a proper build routine. As a result, checking for install_plugins on these installations will return false even if the user is an administrator.


chriscct7:
Related #3084 and #3039

Original reason for doing it this way here:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/1291#issuecomment-20003914

Maybe instead we can change install_plugins to manage_shop_settings and then show a notice on those pages that DISALLOW_FILE_MODS is on

etcook:
@chriscct7 that actually makes sense - one of those edge case considerations with no really good solution

pippinsplugins:
Resolved in #3928

PR #4037, opened 2015-11-24
Set in_array check to strict

grappler:
I wanted to add another select field and came across a problem that the placeholder text was being added as a selected item once I selected an option.

This was not happening in the bundle option as the values were numbers but in my case I needed to values to be text.
image
image

The following will return `1`
```php
$array[0] = 'abc';
echo in_array( 0, $array );
```
Where as the following will not return a result.
```php
$array[0] = '123';
echo in_array( 0, $array );
```

By setting the check to strict the result will be the same for both situations.

P.s. You may want to think about making all of the checks strict.


pippinsplugins:
A strict check makes sense to me.

PR #4038, opened 2015-11-25
Pass selected country to the edd_shop_states filter

sdmtt:
Makes it possible to add custom states without modifying core.
```php
function custom_edd_states($states, $country) {
    if(empty($states) && $country === 'AR') {
        $states['BA'] = 'Buenos Aires';
        $states['EA'] = 'Mendoza';
    }

    return $states;
}
add_filter('edd_shop_states', 'custom_edd_states', 10, 2);
```


pippinsplugins:
This looks good to me.

@cklosowski Any objections to including this ins 2.5?

PR #4039, opened 2015-11-25
We should allow for developers to run actions on a deleted option

chriscct7:


pippinsplugins:
What about instead of passing the `$did_update`, we pass the old value.

Thoughts?

chriscct7:
I think we should pass $did_update because in this context its more important than the old value, but we could add a second param for $old_value for those who want it

pippinsplugins:
I was thinking `old_value` would be nice for times someone wants to run actions conditionally based on what the value was.

chriscct7:
This is no longer required for the initial purpose I needed it for. Feel free to modify at will :)

pippinsplugins:
Since not needed I'm going to punt to 2.5.4 just to help get 2.5 finished faster.

PR #4047, opened 2015-12-01
Lowercase "e" for the schedule_events method in EDD_Cron class

DevinWalker:
I'm not sure why it was capitalized but I standardized it to match other methods in the class


cklosowski:
:+1: Thanks @DevinWalker

Issue #2667, opened 2014-09-29
Introduce EDD_Payment class


pippinsplugins:
Creating payments and retrieving payment data can be obtusely difficult, especially creating payments like the Manual Purchases extension does.

We should really introduce an `EDD_Payment` class that takes care of this.

For example, to create a new payment, it might look like this:
```php
$payment = new EDD_Payment;
$payment->add_download( $id, $price_id );
$payment->add_fee( $label, $amount );
$payment->add_discount( $code );
$payment->set_total( $amount );
$payment->set_tax( $amount );
$payment->complete();
```
All meta data should be able to be set in this way.

Retrieving an existing payment would be as simple as:
```php
$payment = new EDD_Payment( $id );
$total = $payment->total;
$downloads = $payment->downloads;
$tax = $payment->tax;
```
Updating an existing payment:
```php
$payment = new EDD_Payment( $id );
$payment->set_total( $new_total );
$payment->add_note( 'Total manually changed because why not?' );
```

ghost1227:
um... no...

pippinsplugins:
Can you push it up? The commit links above all 404

danieliser:
Any help needed on this?

cklosowski:
Not yet. We're waiting on some work that's been done, then it's on my list to knock out for 2.3

cklosowski:
@ghost1227 If you can, see what happened to your commits for this Would like to get things moving, If you can make it a PR or something that'd be great.

pippinsplugins:
@ghost1227 If you can't find them, we'll rewrite it.

ghost1227:
It's always been there... and I've linked it to you at least once in the last few weeks in hall. It's at https://github.com/easydigitaldownloads/Easy-Digital-Downloads/tree/feature/payments-class

pippinsplugins:
Unfortunately we're running out of time for this so we're going to punt it to 2.4.

ghost1227:
@pippinsplugins Anything that you see that needs to be fixed?

pippinsplugins:
Since we have punted to 2.4, I haven't looked at it in depth yet. Will
after 2.3 is out.

pippinsplugins:
`edd_insert_payment()` should function as a wrapper for this.

pippinsplugins:
2.4 beta is Wednesday. We'll have to punt this to 2.5.

cklosowski:
Blocker for #2714

[...]

PR #4071, opened 2015-12-10
Fixing fees added via old filter and items now added with per-item price not total amount

cklosowski:
#2667

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -626,6 +626,7 @@ final class EDD_Payment {
 		$new_download = array(
 			'id'       => $download->ID,
 			'quantity' => $quantity,
+			'fees'     => $args['fees'],
 		);
 
 		$default_options = array(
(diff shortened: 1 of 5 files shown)


pippinsplugins:
@cklosowski Please add `@since` tags to all properties.

includes/payments/class-edd-payment.php
```diff
@@ -33,39 +33,208 @@
...
+
+	/**
+	 * Array of global fees for this payment
+	 * @var array
+	 */
+	protected $fees = array();
+
+	/**
+	 * The sum of the fee amounts
+	 * @var integer
```
pippinsplugins: This is a float.

includes/payments/class-edd-payment.php
```diff
@@ -33,39 +33,208 @@
...
+	 * The Unique Payment Key
+	 * @var string
+	 */
+	protected $key = '';
+
+	/**
+	 * The total amount the payment is for
+	 * Includes items, taxes, fees, and discounts
+	 *
+	 * @var integer
```
pippinsplugins: This is a float

includes/payments/class-edd-payment.php
```diff
@@ -591,7 +787,7 @@ public function add_download( $download_id = 0, $args = array(), $options = arra
 		$defaults = array(
 			'quantity'    => 1,
 			'price_id'    => false,
-			'amount'      => false,
+			'item_price'  => false,
```
pippinsplugins: `item_price` should probably default to either `0.00` or `''`

pippinsplugins:
Fees look really good. Just one thing: it doesn't look like the fees handling in `add_download()` ensures the `download_id` is set on the fee.

If a fee is supplied in the `$args` array of `add_download()`, we should force it to include `download_id` and set that to the ID passed to `add_download()`

cklosowski:
:+1: will get those done quick, should be easy

cklosowski:
@pippinsplugins Ok these are ready for your look again.

Issue #3894, opened 2015-10-08
Option to Disable Pagination in the Shortcode


spencerfinnell:
Right now you can pass `pagination` as `false`, but that will still show the actual pagination links (since its not disabling pagination, it's disabling the query limit). What if you want to only display 5 downloads but no pagination, and can't pass the individual IDs? 

`[downloads pagination=false number=10]`

Should not show the pagination links below results.

`[downloads number=10]`

Should.

pippinsplugins:
@spencerfinnell I believe this is duplicate of #3522. Is that not working properly?

spencerfinnell:
That's disabling pagination by fetching all results. I want to fetch a set/limited number and not output the HTML pagination links.

spencerfinnell:
Basically the issue is you can't combine the number and pagination attributes. Setting pagination to false overrides it and doesn't affect the pagination output below if.

pippinsplugins:
Ah, makes sense.

PR #4085, opened 2015-12-15
Allow the use of `pagination` and `number` shortcode attributes toget…

spencerfinnell:
…her. See #3894


pippinsplugins:
@spencerfinnell I updated it to use `FILTER_VALIDATE_BOOLEAN` with `filter_Var()`.

PR #4088, opened 2015-12-17
Add support for translation packs

grappler:
I have developed a plugin for WordPress developers who sell themes and plugins with EDD and use the Software Licensing plugin.

This code is needed for the EDD plugins that hook into this class and wish to ship translations to their customers.


cklosowski:
I'm closing this out, we've actually got an open issue in the Software Licensing extension to add support for this, and EDD core simply uses the updated version of the plugin updater from there. Since SL is the source of truth for the updater class, we'll just that as the canonical point of the enhancement.

PR #4090, opened 2015-12-19
Add filter for $edd_receipt_args

grappler:
I would like to remove the discount from the purchase confirmation page. I could copy the template to my child theme and make the changes there. I don't really want to do this as this template is quite large and I will miss out of changes in future update.

With the filter I could remove the discount code and then the check below will return false and not show the discount field.

### DIFF
--- a/templates/shortcode-receipt.php
+++ b/templates/shortcode-receipt.php
@@ -4,7 +4,8 @@
  */
 global $edd_receipt_args;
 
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@grappler I'm not sure this is where we want a filter.

You can remove the discount with the shortcode:

```[edd_receipt discount="false"]```

grappler:
Oh, that woks too. :smile:

PR #4093, opened 2015-12-22
Always show reports and logs navigation

### LINKED ISSUE #4092: Always show bulk actions when used for navigation
In multiple reports and all logs in EDD, `WP_List_Table` is used and the `bulk_actions()` method is utilized as a convenient way to output the select field for navigating to other views in the right place. However, bulk actions only display if there are items to show in the table which makes sense when the bulk actions are actually actions but not when they are navigational as they are in EDD.

I'm making a pull request right now with a simple solution that will ensure these navigation items always show.

brashrebel:
Currently the method for switching between different reports and logs views in EDD is to choose a view from the select field in the bulk actions part of `WP_List_Table`. The problem with this is that bulk actions only display if there are items to show in the table which results in users losing their navigation if they select a view that contains no entries.

This solution ensures that the navigation will always display by simply setting the `has_items()` method to `true` no matter what on the tables that require bulk actions regardless of whether items exist.

Resolves #4092

### DIFF
--- a/includes/admin/reporting/class-api-requests-logs-list-table.php
+++ b/includes/admin/reporting/class-api-requests-logs-list-table.php
@@ -302,4 +302,15 @@ class EDD_API_Request_Log_Table extends WP_List_Table {
 			)
 		);
 	}
+
+	/**
+	 * Since our "bulk actions" are navigational, we want them to always show, not just when there's items
+	 *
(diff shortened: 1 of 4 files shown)


pippinsplugins:
@brashrebel Could you resubmit this against `release/2.5`?

Issue #3788, opened 2015-09-09
window.location.reload() not executed when removing a 100% discount


felixarntz:
I encountered an issue where the `window.location.reload()` that is supposed to happen when removing a 100% discount does not get executed.

See [this line](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/assets/js/edd-checkout-global.js#L283): The problem happens when using a different decimal separator other than the dot (for example in German it would be `0,00`) - this setting is not honored here. An alternative solution, bypassing the need to honor the setting would be to check the `data-total` attribute of the node instead of the node text since this is always a non-formatted number.

pippinsplugins:
Thanks for the report! We'll get it fixed!

ghost1227:
@pippinsplugins Problem with using the `data-total` attribute is that at present it doesn't seem to take discounts into account. In other words, if I have a $1.00 item in the cart, and add a 100% discount, the `data-total` attribute still shows as `1`.

pippinsplugins:
That should be fixed then :)

pippinsplugins:
Let's get this in for 2.5.8

cklosowski:
👍

pippinsplugins:
PR works perfectly.

PR #4102, opened 2015-12-31
fixed JS issue where decimal separator has not been considered

felixarntz:
see #3788: If a different decimal separator than `.` is used, the `window.location.reload()` will not be reached even if a 100% discount has been removed. This commit fixes it.


felixarntz:
btw the fix is only in the non-minified file. Not sure if that's sufficient - I wasn't sure how assets are minified in the EDD workflow.

PR #4106, opened 2016-01-01
Replace $edd_options[ $key ] with edd_get_option()

### LINKED ISSUE #4104: Replace $edd_options[ $key ] with edd_get_option()
I hooked into `'edd_get_option_' . $key` to change the page id depending on the language used in Polylang.

Everything works fine in the front end but in the backend it does not work. The reason is that `global $edd_options;` is still being used in the callbacks.

To fix this we can replace `global $edd_options;` with `$edd_options = edd_get_option( $args['id'] );`and replace `$edd_options[ $args['id'] ]` with `$edd_options` in `/includes/admin/settings/register-settings.php`

Let me know what you think and I can create a PR.

grappler:
Fixes #4104

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -19,6 +19,7 @@ if ( !defined( 'ABSPATH' ) ) exit;
  * Looks to see if the specified setting exists, returns default if not
  *
  * @since 1.8.4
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@grappler This won't merge once `release/2.5` is merged into `master`. Could you resubmit later this week?

PR #4116, opened 2016-01-04
convert @nodes to attributes

gordielachance:
Hi, I don't know if was handling a badly formatted json, but I had problems with attributes that were recognized as nodes.
This commit checks if the node name starts with "@", and makes an attribute of it if needed.


cklosowski:
This is a 3rd party library we're bundling, we should not be making direct changes to it. I'll open an issue to see about getting the most recent version of the `array2xml.php` library included.

PR #4139, opened 2016-01-08
Fix test constants plugin folder path check

chriscct7:
On a Windows environment this unit test will fail. 

This is purely a problem with phpunit and not an actual bug in EDD.

What happens is Windows sees the EDD constant declaration and ends the path with a "/" whereas the unit test declaration path ends in "\". This is simply a result of PHPUnit's path being called in a folder above the file versus the original file. In EDD since the constant is declared once, as it's a constant, this isn't a problem. The path removes the last character of each path, allowing the test to pass on Windows.

### DIFF
--- a/tests/tests-easy-digital-downloads.php
+++ b/tests/tests-easy-digital-downloads.php
@@ -27,7 +27,9 @@ class Tests_EDD extends WP_UnitTestCase {
 
 		// Plugin Folder Path
 		$path = str_replace( 'tests/', '', plugin_dir_path( __FILE__ ) );
-		$this->assertSame( EDD_PLUGIN_DIR, $path );
+		$path = substr( $path, 0, -1 );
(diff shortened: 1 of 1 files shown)


tests/tests-easy-digital-downloads.php
```diff
@@ -27,7 +27,9 @@ public function test_constants() {
 
 		// Plugin Folder Path
 		$path = str_replace( 'tests/', '', plugin_dir_path( __FILE__ ) );
```
pippinsplugins: What about using `DIRECTORY_SEPARATOR` instead of `/`. That would make the `substr()` below unnecessary.

chriscct7: It's not the / on the end of tests, its actually the / that ends the path
Example:
..../wp-content/plugins/Easy-Digital-Downloads/
vs 
..../wp-content/plugins/Easy-Digital-Downloads\

PR #4144, opened 2016-01-08
composer: phpunit to latest of 3.7 branch

Ramoonus:


chriscct7:
There's a reason we set it to use this specific version but I can't remember why. Will try to remember

pippinsplugins:
@cklosowski Do you remember any reason?

Ramoonus:
not sure why you are still on 3.x at all when v4 is stable

Ramoonus:
https://travis-ci.org/easydigitaldownloads/Easy-Digital-Downloads/jobs/101123201
line 162: 162PHPUnit 4.5.0 by Sebastian Bergmann and contributors.

Ramoonus:
your composer and Travis are independent of eachother

chriscct7:
Yes, Travis CI will use the 4.5 or 5.0 branch respectively. We do not use Composer to install PHPUnit on Travis, because the composer installed version of PHPUnit requires that the boostrap file in the phpunit.xml file be pointed at the composer vendor folder, wheras we need that to be pointed at tests/boostrap.php.

The reason we require that specific version of PHPUnit in our composer file is the WordPress Unit Tests library and subsequently EDD's testing suite is designed to be compatible with a pear, phar or composer installation of PHPUnit. As 3.7.31 was the last issue of pear PHPUnit, requiring this specific version in the composer file ensures that regardless of someone is running the EDD unit test suite via PEAR or via a phar or via Composer, they will all run using the same PHPUnit installed. In fact, EDD's travis suite used to install PHPUnit (before Travis made it default installed) using PEAR. I believe this was the reason it was done, although there could have been additionally bug in a newer version (another 3.7 branch that caused us to do this).

cklosowski:
As per this notice on the PHPUnit GitHub pages, where it's talking about the PEAR deprecation:

> Update: On December 1, 2014 we released PHPUnit 3.7.36 and PHPUnit 4.0.18 as PEAR packages. These PEAR packages do not install a working version of PHPUnit. They merely install a script that prints an error message.

I'm going to merge this, since PEAR was the hold out reason for installing a specific version. Let's merge this in, and then we can look at upping the version at a later date if necessary.

Issue #4167, opened 2016-01-12
Payment date not saving when changed in admin


pippinsplugins:
When updating a payment record in the admin, the date cannot be changed. After updating the payment, the date reverts back to what it was.

PR #4172, opened 2016-01-12
Allow updating the date on EDD_Payment objects

cklosowski:
#4167

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -770,6 +770,15 @@ final class EDD_Payment {
 						$this->update_meta( '_edd_payment_number', $this->number );
 						break;
 
+					case 'date':
+						$args = array(
+							'ID'        => $this->ID,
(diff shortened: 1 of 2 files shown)


includes/payments/class-edd-payment.php
```diff
@@ -770,6 +770,15 @@ public function save() {
 						$this->update_meta( '_edd_payment_number', $this->number );
 						break;
 
+					case 'date':
+						$args = array(
+							'ID'        => $this->ID,
+							'post_date' => $this->date,
```
pippinsplugins: Does this also need to set the post_date_gmt or does that happen automatically?

cklosowski: We didn't in the previous update payment action:
```
	// Update main payment record
	$updated = wp_update_post( array(
		'ID'        => $payment_id,
		'post_date' => $date
	) );
```

Issue #4187, opened 2016-01-13
The View Payment Screen needs to ensure an option for payment status is present


chriscct7:
With the recent WP Parsidate plugin conflict, a new issue was exposed in the view payment screen.

A user using the plugin had a customer succesfully place an order. As a result of the aforementioned bug, the payment had a future date. 

When a payment object is made for an order, the payment status key is set to the value of the post status. However, on the view payment page, the dropdown attempts to match that key versus the keys of the edd_get_payment_statuses() array. However because of the Parisdate bug, since the payment has a status of 'future' it will make order appear to be "pending" even though the order was really "future" because the key wasn't found and thus the select shows the first result as selected even if its not (in this case Pending is the first in the array). Further, because the user didn't realize the post was a future post, they didn't realize they needed to adjust the date (which if you looked at it was a future date/time), and instead repetitively tried to save the order as completed, which didn't do anything since the post would still be a future post to WordPress.
In fact, we're going to see this issue on a lot of customers sites over the next few days who had the Parsidate plugin conflict, as people look at their orders.

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/admin/payments/view-order-details.php#L74 just needs to have a 

```
if ( ! array_key_exists( $payment->status, edd_get_payment_status_keys() ) ){
  //html for the option here
}
```

That way if the key doesn't exist in the array, the current payment status will still be selected

Related: #4185, #4180

pippinsplugins:
Seems like a good idea to me. Helps diagnose problems like this a LOT more easily.

cklosowski:
This broke master...the values are all numerical, `Publish` has replaced `Complete`, and it's using the `post_status` value as the Display value:
![screen shot 2016-01-18 at 2 44 10 pm](https://cloud.githubusercontent.com/assets/1390124/12403563/f621edc2-bdf1-11e5-8bd1-b308268276c5.png)

![screen shot 2016-01-18 at 2 44 01 pm](https://cloud.githubusercontent.com/assets/1390124/12403566/fa94a9a8-bdf1-11e5-981d-d036d4195348.png)

pippinsplugins:
PR reverted.

cklosowski:
#4209 was the PR to revert the original fix.

pippinsplugins:
Punting to 2.5.8. Not a priority for this.

cklosowski:
[...]

PR #4188, opened 2016-01-14
If key not in payment status array then still show the correct payment status selected

chriscct7:
Just wrote this on a phone (untested). Solves #4187.


chriscct7:
Unrelated Travis failures. Looks like things Chris K was working on

cklosowski:
Can you pull in the latest from master as a merge so this has passing tests? Just to be sure.

chriscct7:
They're passing now

PR #4202, opened 2016-01-16
fix `hide_on_checkout` bug in `edd_cart_widget`

### LINKED ISSUE #4199: Cart Widget doesn't show on checkout page when configured to do so
Customer reported and confirmed locally.

`$instance['hide_on_checkout']` is always set [here](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.5.6/includes/widgets.php#L44), it's just empty.

felixarntz:
fixes #4199


includes/widgets.php
```diff
@@ -41,7 +41,7 @@ function __construct() {
 	/** @see WP_Widget::widget */
 	function widget( $args, $instance ) {
 
-		if ( isset( $instance['hide_on_checkout'] ) && edd_is_checkout() ) {
+		if ( isset( $instance['hide_on_checkout'] ) && $instance['hide_on_checkout'] && edd_is_checkout() ) {
```
pippinsplugins: This could be simplified to just `if ( ! empty( $instance['hide_on_checkout'] ) `

cklosowski:
Thanks @felixarntz

Issue #4205, opened 2016-01-18
Auto flush permalinks if we detect a 404 error on a post type or taxonomy page


pippinsplugins:
404 errors on post type and taxonomy pages are very common. The fix is usually to just refresh the permalinks. Today it occurred to me that we can actually programmatically fix these 404 errors immediately when they happen, potentially saving us a large number of support tickets.

My proposal is that we detect when a 404 error is thrown and then do a soft flush on permalinks if the 404 was on an EDD post type or taxonomy page.

SDavisMedia:
This is dope. To test, I need a way to screw up the permalinks in a way that this will fix it on the fly. Any tips?

pippinsplugins:
Comment out the flush rewrite in install.php, deactivate Easy Digitial
Downloads, refresh perms links, activate edd, go to category page. It
should auto redirect and fix itself.

SDavisMedia:
Okay `issue/4205` is working for me but I want to make sure another step I added isn't the case.

I did everything mentioned above but after refreshing permalinks, I also deleted all transients. With those steps, on `master`, I could trigger the 404 as expected. On `issue/4205`, it worked (no 404) as expected.

So as long as deleting all transients doesn't bork the test, this is working.

pippinsplugins:
Deleting transients will not bork the test :)

cklosowski:
@SDavisMedia nope no borking there, it just simply let you test multiple times without having to wait 12 hours. Thanks.

pippinsplugins:
:D

PR #4206, opened 2016-01-18
Introduce edd_refresh_permalinks_on_bad_404() - #4205

pippinsplugins:
Introduce edd_refresh_permalinks_on_bad_404() to auto flush permalinks when a 404 error is thrown on an EDD post type or taxonomy page. #4205


cklosowski:
Since @SDavisMedia and I are good with this, I'm going to merge into master.

PR #4221, opened 2016-01-20
Added option to show image in details widget

deviodigital:
In the Download Details widget, there was not a way to show the image of the product that's being sold, which ends up just showing the title of the product, a large buy button and the categories/tags.

This commit will update the widget to give the user an option to show the featured image that's set with their product, making the widget more visually appealing and let shoppers know more about what they may be purchasing when clicking the buy button.

### DIFF
--- a/includes/widgets.php
+++ b/includes/widgets.php
@@ -249,8 +249,8 @@ class EDD_Product_Details_Widget extends WP_Widget {
 		echo $download_title;
 		
 		if($image) {
-			if(wp_get_attachment_url( get_post_thumbnail_id($download_id) ) !='') {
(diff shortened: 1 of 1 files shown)


cklosowski:
Looks like this is causing the unit tests for the widget to fail, can you please update to fix that @deviodigital?

deviodigital:
yeah, I'll look the code over tomorrow afternoon and get it fixed up

cklosowski:
@deviodigital if you could submit this against the `issue/4239` branch instead of `master`, i'll go ahead and get it cleaned up for the unit tests before we get it into master.

Issue #4223, opened 2016-01-21
Add a method in EDD_Payment for making as refunded


cklosowski:
We currently have a method of marking a payment as complete, but the other action that has the most hook points is a 'refund'.

We need to add a refund_payment method in `EDD_Payment` to allow for extensions to easily hook when a refund is being processed.

pippinsplugins:
@cklosowski I get `refund failed` when trying to refund a payment now: https://cloudup.com/cRcPQPzwpNh

pippinsplugins:
Ok actually looks like the refund processed but it showed the wrong admin notice.

cklosowski:
I got the proper notice in admin, I'll try again.

pippinsplugins:
Just tried it again. It refunded fine but I get an undefined index now:

Notice: Undefined variable: message in /Users/pippinwilliamson/Sites/trunk/wp-content/plugins/edd-recurring/includes/admin/class-admin-notices.php on line 53

cklosowski:
@pippinsplugins this was a bug in recurring, I found it and fixed it in `issue/203` of recurring

pippinsplugins:
:+1:

pippinsplugins:
Heh, major bug.

Refunding a payment causes the sale count on each item in the payment to decrease by `3`, not `1`.

pippinsplugins:
Hmm, maybe not all the time. It definitely happened on one particular sale but I just tried again and it didn't do it.

pippinsplugins:
OK all is good. I forgot that I had had quantities enabled and then later disabled them.

pippinsplugins:
@cklosowski just one item I noticed: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/4228#discussion_r51338394

cklosowski:
@mindctrl this could use some testing, if you have some bandwidth. Pretty major addition to the platform.

cklosowski:
@pippinsplugins your comment here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/4223#issuecomment-177048269), was probably caused by that bug I found in the `edd_undo_purchase` where, things were looping through each download in the payment, and then the function was doing the same, but for a specific download ID...should be resolved now.

pippinsplugins:
@cklosowski Agreed.

mindctrl:
[...]

PR #4228, opened 2016-01-22
Move refund into the EDD_Payment and allow filtering on the status ch…

cklosowski:
…anges #4223


includes/payments/actions.php
```diff
@@ -153,53 +153,11 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
...
-		)
-	);
+	$payment = new EDD_Payment( $payment_id );
+	$payment->refund();
 
-	// Clear the This Month earnings (this_monththis_month is NOT a typo)
-	delete_transient( md5( 'edd_earnings_this_monththis_month' ) );
 }
-add_action( 'edd_update_payment_status', 'edd_undo_purchase_on_refund', 100, 3 );
+// add_action( 'edd_update_payment_status', 'edd_undo_purchase_on_refund', 100, 3 );
```
pippinsplugins: This wasn't supposed to be commented out.

cklosowski: Actually it needs to be removed. We have a clause in the changing of the payment status that fires specific changes, that were on this hook prior. The hook still runs, but we don't use it anymore.

includes/payments/functions.php
```diff
@@ -223,15 +223,9 @@ function edd_delete_purchase( $payment_id = 0, $update_customer = true, $delete_
...
-
-	if ( is_array( $downloads ) ) {
-		// Update sale counts and earnings for all purchased products
-		foreach ( $downloads as $download ) {
-			edd_undo_purchase( $download['id'], $payment_id );
-		}
-	}
 
+	// Update sale counts and earnings for all purchased products
+	edd_undo_purchase( false, $payment_id );
```
pippinsplugins: First parameter is expected to be the download ID. Since we determined that this function was a bit broken, we should mark this as deprecated and force it to false inside the function.

cklosowski: Ok I'll put a notice for deprecation in the function itself if `$download_id` is anything but empty.

includes/misc-functions.php
```diff
@@ -520,6 +519,47 @@ function _edd_deprecated_function( $function, $version, $replacement = null, $ba
...
+ *   trigger or false to not trigger error.
+ *
+ * @param string  $argument    The arguemnt that is being deprecated
+ * @param string  $function    The function that was called
+ * @param string  $version     The version of WordPress that deprecated the function
+ * @param string  $replacement Optional. The function that should have been called
+ * @param array   $backtrace   Optional. Contains stack backtrace of deprecated function
+ */
+function _edd_deprected_argument( $argument, $function, $version, $replacement = null, $backtrace = null ) {
+	do_action( 'edd_deprecated_argument_run', $argument, $function, $replacement, $version );
```
pippinsplugins: Let's please swap the order of `$replacement` and `$version` so they match the order in the function definition.

includes/payments/actions.php
```diff
@@ -152,55 +152,9 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
  * @return void
  */
 function edd_undo_purchase_on_refund( $payment_id, $new_status, $old_status ) {
```
pippinsplugins: Let's go ahead and deprecate this entire function.

includes/misc-functions.php
cklosowski: This actually follows the same usage that the `_edd_deprecated_function` call is using, the order is `$version, $replacement` in the function arguments, then `$replacement, $version` in the filter. I vote we keep it similar to avoid confusion.

cklosowski:
```
function _edd_deprecated_function( $function, $version, $replacement = null, $backtrace = null ) {
	do_action( 'edd_deprecated_function_run', $function, $replacement, $version );
```

pippinsplugins: Fair enough.

PR #4234, opened 2016-01-23
Replace $edd_options[ $key ] with edd_get_option()

### LINKED ISSUE #4104: Replace $edd_options[ $key ] with edd_get_option()

grappler:
Remove global $edd_options;
Fixes #4104

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -19,6 +19,7 @@ if ( !defined( 'ABSPATH' ) ) exit;
  * Looks to see if the specified setting exists, returns default if not
  *
  * @since 1.8.4
+ * @global $edd_options Array of all the EDD Options
  * @return mixed
  */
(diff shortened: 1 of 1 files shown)


cklosowski:
merged in https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/2ad8530a5f517c2ea922ab0db3fa69b2b1dda8b0

PR #4249, opened 2016-01-27
Make edd_view_customers_role filter work.

jdevalk:
The `edd_view_customers_role` filter is called in the render function but not on the `add_submenu_page` and thus you can never get there, even if you filter it.


pippinsplugins:
Looks good to me. Merging.

Issue #4242, opened 2016-01-26
Allow access to .jpeg files in .htaccess rules


mindctrl:
We allow access to `.jpg` files in the protected directory rules, but not .jpeg files. [Reference](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.5.6/includes/admin/upload-functions.php#L153). I've seen this cause problems several times for folks uploading `.jpeg` files.

Let's add `.jpeg` to the allowed filetypes.

pippinsplugins:
:+1:

PR #4260, opened 2016-02-01
Add .jpeg to protected directory allowed types. #4242.

mindctrl:
#4242


pippinsplugins:
Will merge once unit tests pass.

Issue #3981, opened 2015-11-04
Download links break if campaign variables are added to them


pippinsplugins:
I just discovered that file download links break (by giving a permissions error) if UTM tracking parameters are added to them.

We should look at allowing UTM variables on download URLs since that will be a common occurrence.

We could probably resolve it by simply stripping the UTM variables off the URL before we validate the token.

pippinsplugins:
The Better Google Analytics plugin automatically adds the UTM links to outgoing emails: https://wordpress.org/plugins/better-analytics/

cklosowski:
@pippinsplugins I would propose we strip any query string items with `utm_` in there. Might be even better to set an array of whitelisted query string parameters for download links, and then strip them based off that whitelist. Make it filterable as well so things can be added to them.

pippinsplugins:
@cklosowski How about we strip _all_ query args that are no intentionally added by EDD.

cklosowski:
@Pippinsplugins yeah let's just strip all that aren't ours.

pippinsplugins:
:+1:

mindctrl:
Related: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/2597

pippinsplugins:
Time for testing!

PR #4274, opened 2016-02-09
Automatically remove non-permitted query args from the URL - #3981

pippinsplugins:
#3981


includes/download-functions.php
```diff
@@ -1190,6 +1190,29 @@ function edd_validate_url_token( $url = '' ) {
 
 		wp_parse_str( $parts['query'], $query_args );
 
+		// These are the only URL parameters that are allowed to affect the token validation
+		$allowed = array(
```
cklosowski: We should throw a filter on here allowing added whitelist items.

PR #4280, opened 2016-02-09
Made Sales Log currency-aware

daigo75:
Target
--
Ensuring that all the rows in the Sales Log show the currency currency symbol.

Changes
--
* Added currency to the data returned by the `edd_get_payment_meta_cart_details()` function.
* Tweaked the Sales Log to use the currency associated with each item.

Why
--
The Sales Log, and practically every other log, calls the `edd_currency_filter()` function without passing a currency. This is imprecise, as the currency associated with a payment may or may not be the "global" one (this is true in multi-currency environements, but also if the admin changes the base currency). Since the log export classes are pretty much hard-coded, it's quite difficult to intercept the call to `edd_currency_filter()` and "sneak in" the correct currency. It's much simpler to do that within the core, as it takes few lines of code, as opposed as multiple "workaround" filters.

The solution is, simply, to use the currency associated with each order, and pass it to the `edd_currency_filter()` function. Since this information is attached to a payment, it's not included by `edd_get_payment_meta_cart_details()`, which just returns the cart items. I tweaked that function as well, so that the currency is attached to each item (it's the same for all of them, anyway), and it can be picked up by the report.

Remarks
--
The solution works correctly both in single and multi-currency environments (the payment currency is a core attribute, which is always returned by `EDD_Payment->currency`. In other words, it doesn't depend on any specific multi-currency implementation.

### DIFF
--- a/includes/admin/reporting/class-sales-logs-list-table.php
+++ b/includes/admin/reporting/class-sales-logs-list-table.php
@@ -76,6 +76,9 @@ class EDD_Sales_Log_Table extends WP_List_Table {
 	 */
 	public function column_default( $item, $column_name ) {
 		$return = '';
+		// Use the currency associated with the payment from which each item is
+		// extracted
+		$currency = $item['currency'];
 
 		switch ( $column_name ){
 			case 'download' :
@@ -88,11 +91,11 @@ class EDD_Sales_Log_Table extends WP_List_Table {
 				break;
 
 			case 'item_price' :
-				$return = edd_currency_filter( edd_format_amount( $item['item_price'] ) );
+				$return = edd_currency_filter( edd_format_amount( $item['item_price'] ), $currency );
 				break;
 
 			case 'amount' :
-				$return = edd_currency_filter( edd_format_amount( $item['amount'] / $item['quantity'] ) );
+				$return = edd_currency_filter( edd_format_amount( $item['amount'] / $item['quantity'] ), $currency );
 				break;
 
 			case 'payment_id' :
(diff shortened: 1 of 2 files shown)


pippinsplugins:
This looks good to me. It needs some minor clean up but we can take care of that after merge.

@cklosowski agreed?

daigo75:
I tried to stick to the original coding standards as much as possible, I might have missed something here and there (I'm in fully "*patching mode*" these days, EDD 2.5 caused quite a stir).

cklosowski:
Merging, I'll clean up the comments after merge.

pippinsplugins:
@daigo75 Thank you for the fix and, as always, the indepth feedback. Note, if you haven't already, please consider subscribing to the EDD dev blog: https://easydigitaldownloads.com/development - EDD 2.5 was in beta for more than a month specifically so we could try and catch items like this before the release. We announce beta versions on the dev blog well in advance to other developers a good opportunity to test with their plugins and custom integrations.

Thanks again!

daigo75:
Thanks for the heads up. In my case, a month is not much, mainly because my other products (the ones for WooCommerce) absorb most of my time. I only had a couple of days to test EDD 2.5 and I missed several changes that caused side effects (now I should have addressed them all). Having WC 2.5 released to close to EDD 2.5 didn't help either, unfortunately. 

Recently I started reorganising my business so that I can spread the workload more evenly. The developers blog will be a useful resource. :smile:

pippinsplugins:
A month beta period is at least enough time to identify if there are problems, even if it takes us longer to fix them :)

daigo75:
From a calendar perspective, yes. :smile:

PR #4287, opened 2016-02-11
Change type to wordpress-plugin and add ...

michaelbeil:
composer/installers to require.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,11 +6,11 @@
     "homepage": "https://easydigitaldownloads.com/",
     "license": "GPL-2.0+",
     "support": {
-   	  "issues": "https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues"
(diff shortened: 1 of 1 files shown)


michaelbeil:
#4287

composer.json
```diff
@@ -1,12 +1,16 @@
 {
     "name": "easydigitaldownloads/Easy-Digital-Downloads",
-    "type": "project",
+    "type": "wordpress-plugin",
     "description": "The world's easiest way to sell digital downloads through WordPress for free",
     "keywords": ["edd", "eddwp", "easy digital downloads"],
     "homepage": "https://easydigitaldownloads.com/",
     "license": "GPL-2.0+",
+    "support": {
```
pippinsplugins: This should be our support page

michaelbeil: Roger that.

Issue #4291, opened 2016-02-12
New User Notification Email URL Not Linked


DevinWalker:
It would be nice to have this URL clickable / linked:

![2-12-2016 3-42-38 pm](https://cloud.githubusercontent.com/assets/1571635/13023701/2bcfdc4c-d1a0-11e5-888c-3ea02c06a1bf.png)

pippinsplugins:
I agree, let's link it. I see you committed a fix. Could you drop that in a PR?

pippinsplugins:
Nevermind, you already did :)

PR #4292, opened 2016-02-13
Issue/4291

DevinWalker:
#4291


pippinsplugins:
Looks good to me.

cklosowski:
I think this should be `log in`. `login` references a username or identifier. `log in` references the action of logging in. Is that correct @mindctrl?

pippinsplugins:
That is correct.

michaelbeil:
Yes @cklosowski: http://public.wsu.edu/~brians/errors/login.html.

cklosowski:
Going to merge and then fix that line.

PR #4295, opened 2016-02-13
Add download name helper function.

pderksen:
Not a must-have, but would help the readability of some of my custom EDD filters.

### DIFF
--- a/includes/class-edd-download.php
+++ b/includes/class-edd-download.php
@@ -245,6 +245,16 @@ class EDD_Download {
 	}
 
 	/**
+	 * Retrieve the download name
+	 *
+	 * @since
+	 * @return string
+	 */
+	public function get_name() {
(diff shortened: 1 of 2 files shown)


JeroenSormani:
I'd suggest moving this to the `EDD_Download` class, the loose functions aren't actively developed/added anymore if I'm correct. 

May not be a bad idea to have a specific method for that as `$download->__get( 'post_title' );` kinda looks rubbish

cklosowski:
I vote for in `EDD_Download` as well.

pippinsplugins:
I as well.

pderksen:
That work?

pippinsplugins:
I like it.

@cklosowski ?

cklosowski:
Yep! Merging.

PR #4310, opened 2016-02-17
Add missing <p> wrapping payment status drop-down.

pderksen:
Also affected "Refund Charge in Stripe" checkbox placement.

image


pippinsplugins:
Ah! So THAT's what caused the Stripe checkbox to move. Been wondering what that was :)

pderksen:
@pippinsplugins Yeah I was looked in edd-stripe first hence the recent PR over there. :)

Issue #3177, opened 2015-02-21
Bug on widgets.php


itibook:
Hello,

I spent a few hours on this one, unsure if this is the wanted behaviour or not, but here's my findings...

My widgets.php was breaking right after the Download Detail widget. As soon as I unregistered the widget everything went back to normal. I also had issue on the customise.php that was not working.

In /includes/widgets.php at line 302 there's the args array for the get_posts function. 

		$args = array( 
	            'post_type'      => 'download', 
	            'posts_per_page' => -1, 
	            'post_status'    => 'publish', 
	        );
$downloads = get_posts( $args );

I am not sure if the -1 for the post_per_page is the meant to be that way, but in my case, where I have thousands of downloads it seems that it had a negative effect (sometimes also had memory issues popping up as errors).

as soon as I changed the -1 to a normal number (e.g. 20) the issue disappeared.

could you please confirm if this is an issue only affecting installation with large amount of downloads or if the -1 is indeed a bad practice and should be changed to something more sensible. I am not sure if there are other implications by changing the -1 (I did not see any, but who knows..)

ciao

Luca

danieliser:
Hey Luca, Your correct in the assumption its bad for sites with lots of downloads. The alternative unfortunately will require quite a bit of work to set up and would be an ajax search input similar to how tags and discounts work.

itibook:
Hi Daniel,

I am not really clear how the widget works and how it impacts other areas. any idea where changing the -1  to let's say 20 could have a negative impact?

From what I see I only use it in the sidebar for a single download, so I am not quite clear why it has to pick up all the other downloads in the dropdown menu, but maybe I am missing the point. Basically I don't see the functional need for that.

What about creating two version of widget? one that only uses the current download and the other one that has the full drop down menu...

or there could be an setting in EDD to limit the amount posts (i.e. the -1 parameter) to show in the dropdown menu

any suggestion on how I could best work around this issue will be appreciated. Not really a developer, so any hint/tip is gold :-)

ciao

Luca

mindctrl:
Luca,

If we change the -1 to 20 as you suggest, you would only see 20 items in the dropdown when adding the widget to a sidebar area, leaving you no way to select the vast majority of your downloads. As Daniel noted, to better accommodate sites with a lot of downloads, that section probably needs to be reworked to allow for searching from the dropdown.

danieliser:
[...]

PR #4322, opened 2016-02-19
Widgets admin fix for huge sites

jasongill:
By default, the "Download Details" widget form in the admin area attempts to load every download item into a drop-down. On sites with tens or hundreds of thousands of downloads, this causes wp-admin/widgets.php to break. This proposed change limits the drop-down to 1000 items, as opposed to -1 (no limit).

### DIFF
--- a/includes/widgets.php
+++ b/includes/widgets.php
@@ -313,7 +313,7 @@ class EDD_Product_Details_Widget extends WP_Widget {
 		 <?php
 			$args = array(
 				'post_type'      => 'download',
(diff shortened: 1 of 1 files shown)


mindctrl:
@jasongill thanks for this, but see https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3177.

jasongill:
Ah yes - sorry about that, didn't see the prior PR for the same issue.

PR #4324, opened 2016-02-19
Only show hyphen if item has variable pricing

### LINKED ISSUE #4323: Payment record: Harden price ID check
We need to add an `edd_has_variable_prices` check to the price ID here: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/admin/payments/view-order-details.php#L301-L302. Right now a hypehn displays on the payment record screen without this check: http://beil.io/1eHLs .
 
To reproduce on 2.5.9:
1. Create a product
2. Buy the product 
3. View payment record (see http://beil.io/1eHLs) 

So, we can replace: 

```php
if ( isset( $price_id ) ) {
    echo ' - ' . edd_get_price_option_name( $item_id, $price_id, $payment_id );
}
```
with:
```php
if ( edd_has_variable_prices( $item_id ) && isset( $price_id ) ) {
    echo ' - ' . edd_get_price_option_name( $item_id, $price_id, $payment_id );
}
```

This fixes the trailing `-` in the payment record (view order details) screen: http://beil.io/bI91.

Related: #4303.

michaelbeil:
Fixes #4323.

### DIFF
--- a/includes/admin/payments/view-order-details.php
+++ b/includes/admin/payments/view-order-details.php
@@ -298,7 +298,7 @@ $currency_code  = $payment->currency;
 														if ( isset( $cart_items[ $key ]['item_number'] ) && isset( $cart_items[ $key ]['item_number']['options'] ) ) {
 															$price_options = $cart_items[ $key ]['item_number']['options'];
 
-															if ( isset( $price_id ) ) {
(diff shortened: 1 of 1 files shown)


pippinsplugins:
Looks good to me

michaelbeil:
Solid.

PR #4330, opened 2016-02-22
Bug in date_end filter in payment history

rene-hermenau:
Wrong end_date condition in payment history. 
As a result the payment history is not showing all purchase made in the selected time frame.

Examples

Alt text

Alt text


includes/admin/reporting/export/class-batch-export-payments.php
```diff
@@ -96,7 +96,7 @@ public function get_data() {
 			$args['date_query'] = array(
 				array(
 					'after'     => date( 'Y-n-d H:i:s', strtotime( $this->start ) ),
```
cklosowski: Same as stated in #4332, should we set the `after` to be `00:00:00`

cklosowski:
closing to use #4404

PR #4331, opened 2016-02-22
Wrong end date in payment export function leads to missing payments

rene-hermenau:
End time must be 23:59:59 when the data filter is used or data result gets empty for end day in the time frame. 

Example: 

Filtering data from 10/01/2016 to 10/02/2016 does not contain purchases from 10/02/2016 without that fix. (This should be save as long as it is never be tried to filter by time)

Alt text


rene-hermenau:
closed. error in it

PR #4332, opened 2016-02-22
Wrong end date in payment export function - missing payments in csv

rene-hermenau:
End time must be 23:59:59 when the data filter is used or data result gets empty for end day in the time frame. 

Example: 

Filtering data from 10/01/2016 to 10/02/2016 does not contain purchases from 10/02/2016 without that fix. (This should be save as long as it is never be tried to filter by time)

Alt text


rene-hermenau:
@cklosowski agree. Do not foresee any drawbacks with doing that. Moreover, it makes filtering code more robust. However, could be worth to write a unit test for this.

cklosowski:
@rene-hermenau Yes, currently we don't have a unit test for the batch processing system. We'll need to get one of those spun up somehow. I'll go ahead and work these PRs into an issue branch and work from there.

cklosowski:
closing to use #4404

PR #4351, opened 2016-02-25
setup_fees_total

### LINKED ISSUE #4350: Payment object doesn't setup fees_total
I'm adding downloads to a payment after purchase (to handle bundles easily), this is done with `$payment->add_download()` after `edd_insert_payment`. Anyway, I found out that modifying the payment will cause the total to be recalculated incorrectly based on the fees_total not being setup during instantiation.

```
		$this->fees_total      = $this->setup_fees_total();
```

```


	/**
	 * Setup the payment fees
	 *
	 * @since  XXX
	 * @return float The fees total for the payment
	 */
	private function setup_fees_total() {
		$fees_total = (float) 0.00;

		$payment_fees = isset( $this->payment_meta['fees'] ) ? $this->payment_meta['fees'] : array();
		if ( ! empty( $payment_fees ) ) {
			foreach ( $payment_fees as $fee ) {
				$fees_total += (float) $fee['amount'];
			}
		}

		return $fees_total;

	}
```

dancameron:
Fixes #4350

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -435,6 +435,7 @@ final class EDD_Payment {
 		// Currency Based
 		$this->total           = $this->setup_total();
 		$this->tax             = $this->setup_tax();
+		$this->fees_total      = $this->setup_fees_total();
 		$this->subtotal        = $this->setup_subtotal();
 		$this->currency        = $this->setup_currency();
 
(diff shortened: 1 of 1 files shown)


cklosowski:
I'm closing this in favor of #4388, It has your original commits @dancameron just has unit tests and a since tab on the method..

Issue #4319, opened 2016-02-18
undefined index download_method


SDavisMedia:
```
Notice: Undefined index: download_method in /Applications/MAMP/htdocs/localhost/edd/wp-content/plugins/easy-digital-downloads/includes/admin/settings/register-settings.php on line 999

Notice: Undefined index: download_method in /Applications/MAMP/htdocs/localhost/edd/wp-content/plugins/easy-digital-downloads/includes/admin/settings/register-settings.php on line 1001
```

Happened when I attempted to enable the SKU field. So I didn't actually change the download method... just resaved it.

PR #4355, opened 2016-02-25
Remove check for download method in misc accounting sanitization call…

cklosowski:
…back. #4319

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -996,11 +996,6 @@ function edd_settings_sanitize_misc_accounting( $input ) {
 		return $input;
 	}
 
-	if( edd_get_file_download_method() != $input['download_method'] || ! edd_htaccess_exists() ) {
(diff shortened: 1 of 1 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -996,11 +996,6 @@ function edd_settings_sanitize_misc_accounting( $input ) {
 		return $input;
 	}
 
-	if( edd_get_file_download_method() != $input['download_method'] || ! edd_htaccess_exists() ) {
```
pippinsplugins: We still need this to fire somewhere.

cklosowski: It does fire, in the proper section https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/admin/settings/register-settings.php#L967

pippinsplugins: :+1:

Issue #4343, opened 2016-02-23
In variable pricing table remove repeatable field should be a button, not a link


samikeijonen:
In general links are for links when you actually go somewhere, buttons are for actions. Here is the [reference](https://make.wordpress.org/core/handbook/best-practices/coding-standards/accessibility-coding-standards/#semantics-for-controls).

![remove_repetable_field](https://cloud.githubusercontent.com/assets/1820415/13252918/d73b3e9c-da41-11e5-88a0-8055ac6a12c5.png)

Code for remove "button" looks like this at the moment.

```
<td>
	<a href="#" class="edd_remove_repeatable" data-type="price" style="background: url(<?php echo admin_url('/images/xit.gif'); ?>) no-repeat;">&times;</a>
</td>
```

Screen readers see that as "times" and it's not clear what that actually do. I suggest that we change it to button like this:

```
<td>
	<button class="edd_remove_repeatable" data-type="price" style="background: url(<?php echo admin_url('/images/xit.gif'); ?>) no-repeat;"><span class="screen-reader-text"><?php printf( __( 'Remove price option %s', 'easy-digital-downloads' ), $key ); ?></span><span aria-hidden="true">&times;</span></button>
</td>
```

This would need a small CSS change/addition like this:

```
.edd_remove_repeatable {
    border: none;
    padding: 0;
}
```

samikeijonen:
This is not related but `edd-admin.css` could be coded using [CSS standard](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/) :)

samikeijonen:
The same should be done in Download Files metabox for remove "buttons".

pippinsplugins:
:+1:

pippinsplugins:
PR works great, merged into `release/2.6`.

PR #4363, opened 2016-02-26
Use buttons in remove repeatable fields

samikeijonen:
In general links are for links when you actually go somewhere, buttons are for actions. This PR replaces links with buttons in remove repeatable fields. There is more details in issue #4343.


samikeijonen:
In the bundle we have this button:

```
<button class="edd_remove_repeatable" data-type="file" style="background: url(<?php echo admin_url('/images/xit.gif'); ?>) no-repeat;"><span class="screen-reader-text"><?php printf( __( 'Remove bundle option %s', 'easy-digital-downloads' ), $key ); ?></span><span aria-hidden="true">&times;</span></button>
```

Is it possible to get bundle download name instead of just `$key`?

pippinsplugins:
That might be possible but will likely require more javascript.

samikeijonen:
Okay. No worries, it might be totally overkill in this situation. At least it's not the same text for all buttons.

pippinsplugins:
Works great, merging into `release/2.6`.

Issue #4345, opened 2016-02-23
Losing focus when removing price option


samikeijonen:
When you add new price option I really like that you set the focus on Option Name field.

![set_focus](https://cloud.githubusercontent.com/assets/1820415/13257913/86132bde-da59-11e5-9d29-8e59cd5e7122.png)

But when you remove the price option you lose the focus. It should at least stay inside Download price metabox. Probably setting the focus in previous focusable element would be preferable.

If you remove the first price option I'm not sure where the focus should go, still probably just previous focusable element.

samikeijonen:
Or we could set the focus always to the next element.

```
focusable  = row.next( 'tr' ).find( 'select, input, textarea, button, a' ).filter( ':visible' ),
firstFocusable = focusable.eq(0);

// If count > 1
firstFocusable.focus();

```

pippinsplugins:
I vote we set focus to the previous price option's name, or the next price option when removing the first one in the lst.

samikeijonen:
I might need help on this one. I'm not sure how do I conditionally check if I'm about to remove the first one on the list.

samikeijonen:
Okay, I tried something in PR #4366. It's working but let me know if it can be done better.

pippinsplugins:
Excellent, thanks! We have all of this a11y items milestoned for 2.6 so we'll review them as we get closer.

pippinsplugins:
PR works great for me. I'm manually merging into `release/2.6`.

PR #4366, opened 2016-02-26
Set focus when removing repeatable field

samikeijonen:
When you remove repeatable field focus get lost. This PR sets the focus on next or previous element.

More details are in issue #4345.


pippinsplugins:
Manually merged into release/2.6 so closing.

PR #4367, opened 2016-02-26
Add filter for user AJAX search

atimmer:
After #4249 the page to view customers is accessible. However the AJAX search however is broken. This PR fixes that.


pippinsplugins:
Looks good to me.

cklosowski:
Yes I'm good with this change, thanks @atimmer

Issue #4052, opened 2015-12-03
Sale logs not deleted when payment changed from complete to pending


pippinsplugins:
If a completed payment is changed back to `pending`, the sale log entries and sale counts / earnings for products are not decremented.

When changing a complete payment to pending, the following should happen:

1. Status changed to pending
2. Delete sale logs
3. Decrement sale count on products in payment
4. Decrease earnings on products in payment
5. Delete `_edd_completed_date` meta key/value.

SDavisMedia:
@cklosowski Replicated on `2.3.9` and then tested in `master`. Verified that changing the payment record from complete to pending did the following:

1. deleted the sales log
2. decremented the product sales count in all locations
3. decreased the product earnings in all locations
4. removed the `_edd_completed_date` meta value from the payment record

All good. :+1:

pippinsplugins:
:+1:

PR #4368, opened 2016-02-26
Fix Complete->Pending transition to alter stats and logs 

cklosowski:
#4052

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -1620,8 +1620,6 @@ final class EDD_Payment {
 			return;
 		}
 
-		do_action( 'edd_pre_pending_payment', $this );
-
 		$decrease_store_earnings = apply_filters( 'edd_decrease_store_earnings_on_pending', true, $this );
 		$decrease_customer_value = apply_filters( 'edd_decrease_customer_value_on_pending', true, $this );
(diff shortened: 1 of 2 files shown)


includes/payments/class-edd-payment.php
```diff
@@ -1568,11 +1569,81 @@ private function process_refund() {
...
+		}
+
+		// Allow extensions to filter for their own payment types, Example: Recurring Payments
+		$process_pending = apply_filters( 'edd_should_process_pending', $process_pending, $this );
+
+		if ( false === $process_pending ) {
+			return;
+		}
+
+		do_action( 'edd_pre_pending_payment', $this );
```
pippinsplugins: I'm not sure this action name makes sense unless this method also fires when a pending payment is initially created.

cklosowski: It does not, any suggestions...something like `edd_pre_to_pending_payment`?

pippinsplugins: How about no actions aside from the status status change ones?

cklosowski: Works for me.

cklosowski:
@pippinsplugins take a look at that?

pippinsplugins:
Looks good to me. Get some PHPDocs on there and we should be good to merge.

cklosowski:
sounds good!

cklosowski:
Merging since PHPdocs are updated.

Issue #4211, opened 2016-01-18
Extensions tab with no legacy extensions looks off


cklosowski:
When you are viewing the new Extensions tab, with only extensions that support subsections it looks a little weird to have a main:

![screen shot 2016-01-18 at 3 56 17 pm](https://cloud.githubusercontent.com/assets/1390124/12404815/3641d098-bdfc-11e5-8c60-e9bc360bbf1b.png)

We should look at hiding that save button or simlar if we're on a main without any settings.

pippinsplugins:
How about defaulting to the first sub section and not showing `Main`?

cklosowski:
@pippinsplugins We can look into that as a solution, I'll have to dig into the subsection code to see where we want to handle that.

cklosowski:
@mindctrl are we good here? Can I merge this and close it out?

mindctrl:
@cklosowski looks great here.

PR #4369, opened 2016-02-26
Remove the main section if there are no settings registered and default

cklosowski:
...to the first non-empty section #4211

### DIFF
--- a/includes/admin/settings/display-settings.php
+++ b/includes/admin/settings/display-settings.php
@@ -34,6 +34,22 @@ function edd_options_page() {
 
 	$registered_sections = edd_get_settings_tab_sections( $active_tab );
 	$section             = isset( $_GET['section'] ) && ! empty( $registered_sections ) && array_key_exists( $_GET['section'], $registered_sections ) ? $_GET['section'] : $key;
+
+	// Unset 'main' if it's empty and default to the first non-empty if it's the chosen section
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@mindctrl could you give this a test?

mindctrl:
@pippinsplugins @cklosowski this causes extensions that don't support subsections yet to not show at all. In the screenshot below, Mailchimp 2.5.5 (pre-subsections version) is activated but the settings aren't available.

image

cklosowski:
@mindctrl i'll reverify, I thought I had one there that wasn't configured yet. I'll update when I find the issue.

cklosowski:
@mindctrl if you have ONLY extensions that were not updated yet...(acquisition survey) does it show that set of settings, but no subsections?

mindctrl:
@cklosowski yup, that works. I see the settings like this:

image

JeroenSormani:
A FYI, this caused a issue in two of my plugins. I was loading a settings view (via the hook callback field type) using `require_once`, which resulted in it being loaded in the check if it had any settings, and not when actually displaying the settings.

I already fixed it by changing it to just `require` (probably better too), but thought it could be a learning moment for the future for both :smile:

pippinsplugins:
@JeroenSormani Thanks for the headsup! We'll watch out for similar cases.

PR #4370, opened 2016-02-27
Issue/4359

### LINKED ISSUE #4359: {receipt_link} markup stripped in plain text emails
When you select "No template, plain text only" and have `{receipt_link}` all you get is the non-clickable text. We should be checking for this within `edd_email_tag_receipt_link` and output URL if plain text so email clients can hyperlink it.

Reported via Give here: https://github.com/WordImpress/Give/issues/384

Fixed here: https://github.com/WordImpress/Give/pull/495

LMK if I can submit a PR if you agree.

DevinWalker:
Fixes #4359


DevinWalker:
Err... this doesn't look right.

Issue #4376, opened 2016-02-29
Change message on account page for password


akyusa01:
Original Ticket: https://secure.helpscout.net/conversation/175925713/28854/

Message shows "Please note after changing your password, you must log back in." which is misleading as the user is still logged in.

![account local testing 2016-02-29 11-37-08](https://cloud.githubusercontent.com/assets/9019502/13401103/d014dc7c-ded8-11e5-983b-5013c3b657d5.png)

michaelbeil:
I believe when a user changes the password, he/she is automatically logged out.

akyusa01:
Hey man :)  I tested on my local with the latest version and didn't get logged out.  But as soon as the profile editor loads, you have that message there whether you are there to change your password or now.

pippinsplugins:
Let's get it fixed in 2.6

pippinsplugins:
There is actually no requirement that the user has to log back in, so I vote we remove the message instead of forcing them to log back in.

michaelbeil:
👍

PR #4380, opened 2016-03-01
Logs user out if they're changing their password #4376

sunnyratilal:
The user is now automatically logged out if they change their password from the Profile Editor and will be redirected back to the back where they will be presented with a log in form.

#4376


pippinsplugins:
Closing in favor of https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4463

Issue #4384, opened 2016-03-02
Fatal Error on adding multi purchase price id to cart when it already exists


cklosowski:
`Fatal error: Unsupported operand types in /easy-digital-downloads/includes/cart/functions.php on line 220`

cklosowski:
To test this easily do the following:
1) Disable AJAX
2) Create variable price item with mult-price purchase on
3) Visit the download and add one price point to the cart
4) After reload is complete, add another price point to the cart
5) No fatal error (hopefully?)

mintplugins:
Success. No fatal error 
<img width="758" alt="screen shot 2016-03-09 at 2 26 38 pm" src="https://cloud.githubusercontent.com/assets/1638128/13647638/00004b84-e603-11e5-8d7d-abea0e7d2af3.png">

PR #4385, opened 2016-03-02
Fix quantity opperand issue with multi-price 

cklosowski:
#4384


cklosowski:
@pippinsplugins you good with this one? @mintplugins has tested it out and verified it doesn't throw the fatal anymore.

PR #4390, opened 2016-03-07
Deprecate edd_get_success_page_url()

### LINKED ISSUE #4389: EDD has 2 (almost identical) functions for returning the success page URL
It turns out that the new version of EDD Recurring prevents a conditional redirect from working in EDD Conditional Success Redirects. This is because EDD Recurring uses `edd_get_success_page_url()` to get the URL of the success page. In the CSR plugin it uses the `edd_get_success_page_uri` filter available within the `edd_get_success_page_uri()` function.

These two functions appear almost identical, they're just written slightly differently. Both return the success page URL such as `http://site.com/checkout/purchase-confirmation/`.

`edd_get_success_page_uri()`
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/checkout/functions.php#L63

`edd_get_success_page_url()`
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/checkout/functions.php#L179

For now, I'm going to add the extra filter to CSR to fix the redirect, but we should probably deprecate one or the other. EDD core doesn't use `edd_get_success_page_url` anywhere, perhaps we should deprecate that one and make it return `edd_get_success_page_uri`.

sumobi:
Fix #4389

### DIFF
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -544,3 +544,16 @@ function edd_verify_download_link( $download_id = 0, $key = '', $email = '', $ex
 	// Payment not verified
 	return false;
 }
+
+/**
+ * Get Success Page URL
+ *
+ * @param string $query_string
+ * @access      public
+ * @since       1.0
+ * @deprecated  2.6
+ * @return      string
+*/
+function edd_get_success_page_url( $query_string = null ) {
+	return apply_filters( 'edd_success_page_url', edd_get_success_page_uri( $query_string ) );
+}
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -551,7 +551,7 @@ function edd_verify_download_link( $download_id = 0, $key = '', $email = '', $ex
  * @param string $query_string
  * @access      public
  * @since       1.0
- * @deprecated  2.6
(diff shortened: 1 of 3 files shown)


sumobi:
PR keeps the `edd_success_page_url` filter intact. Also adds `$query_string` to the `edd_get_success_page_uri` function in case any plugin is using it. I've done some testing and all seems fine to me.

includes/deprecated-functions.php
```diff
@@ -543,4 +543,17 @@ function edd_verify_download_link( $download_id = 0, $key = '', $email = '', $ex
...
+/**
+ * Get Success Page URL
+ *
+ * @param string $query_string
+ * @access      public
+ * @since       1.0
+ * @deprecated  2.6
+ * @return      string
+*/
+function edd_get_success_page_url( $query_string = null ) {
```
pippinsplugins: Let's add a deprecated notice to it as well.

sumobi:
@pippinsplugins ahh :)

pippinsplugins:
Looks good.

@cklosowski Ready to merge if you approve.

cklosowski:
This is failing builds currently...I think we need to get the `release/2.6` branch updated with master. I'll go ahead and merge this...and then get 2.6 passing again.

pippinsplugins:
:+1:

Issue #4377, opened 2016-02-29
Safari Auto-fill on checkout form


akyusa01:
Original ticket: https://secure.helpscout.net/conversation/176246649/28915/?folderId=750133

I was able to recreate this on my local dev site.  When you use the Safari browser and auto-fill to complete the form, the billing country model window states 'USA' but when you click the auto-fill, it changes it to Virgin Islands (USA).  Does not do this in Chrome or FF, only Safari.

![screen shot 2016-02-29 at 12 59 55 pm](https://cloud.githubusercontent.com/assets/9019502/13403728/7141e116-dee4-11e5-88e8-e0ac3ce9c08b.png)
![screen shot 2016-02-29 at 12 59 47 pm](https://cloud.githubusercontent.com/assets/9019502/13403729/7142da44-dee4-11e5-97c5-55c0b7d066ac.png)

pippinsplugins:
So weird . . . Not sure how we can fix this but let's look into it.

chriscct7:
If you add a hidden iframe inside the <form> element it should fix the issue.

chriscct7:
@akyusa01 does https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/4393 fix the issue?

akyusa01:
I just re-tested, set it up on a live server.  Still the same, still bringing over Virgin Islands.  I updated the includes/checkout/template.php first before trying.  Adding the 2 lines and removing the 1 from #4393.

sunnyratilal:
The easiest way would be to just add `autocomplete="off"` to the Country attribute but that's an inconvenience then for some customers.

Also, Safari doesn't seem to autofill the country for me even when I change my autofill data to US.

pippinsplugins:
@chriscct7 I don't really care for that approach.

I don't care for the `autocomplete="off"` approach either. Let's see what else we can come up with?

pippinsplugins:
Suggestion fro, user:

>>>I wonder if you tried simply removing the "USA" portion from the entry: "Virgin Islands (USA)"?

pippinsplugins:
I can no longer replicate this.

Could someone else please try?

michaelbeil:
I cannot replicate this either.

sunnyratilal:
Ditto

pippinsplugins:
Closing as invalid then.

PR #4393, opened 2016-03-08
Add a hidden iframe handler to the checkout form. See #4377

chriscct7:
#4377


pippinsplugins:
#4377 was punted so closing.

Issue #4115, opened 2016-01-04
"No checkout page has been configured" can't be properly dismissed


mindctrl:
Customer reported that the notice can't be dismissed when a checkout page hasn't been configured. Looks like the conditional [here](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.4.10/includes/admin/class-edd-notices.php#L44) is wrong.

pippinsplugins:
Let's get this fixed in 2.5.5

pippinsplugins:
Moving to 2.5.8

PR #4395, opened 2016-03-08
Fix logic for dismissing checkout page missing 

cklosowski:
#4115


pippinsplugins:
Once this passes Travis, we can merge.

PR #4402, opened 2016-03-11
Allow payments via PayPal to be refunded in the WordPress admin.

thomasgriffin:
Merry Christmas from myself and all the other devs at Awesome Motive. :-)

See issue #4403 for more info.

This code allows you to refund payments made with PayPal Standard or PayPal Express right inside of the WordPress admin, just like how Stripe refunds work. It works with both sandbox and live transactions. You simply need to add your live/test PayPal API credentials in the PayPal Standard settings area.

Hooks and filters are added throughout, and the checking process is very thorough. We have been using this in production on OptinMonster, Soliloquy and Envira for 10+ months, so it is battle tested and works well. Refund IDs are stored as a note once it is confirmed that the transaction has been successfully refunded. Should an error occur during the process, we print out a note and let the user know exactly what went wrong and guide them to refund in the PayPal interface.

One thing to note: **this does work with sandbox transactions, but you do need to have a sandbox account created with the same email as your live account, and you should use those credentials in the admin.** From that point, it doesn't matter which sandbox account you use to test purchases - they will be refunded properly.

Let me know if you have any questions. :-)

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -653,3 +653,206 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
 
 }
 add_filter( 'edd_payment_details_transaction_id-paypal', 'edd_paypal_link_transaction_id', 10, 2 );
+
+/**
+ * Shows checkbox to automatically refund payments made in PayPal.
+ *
+ * @access public
+ * @since  2.6.0
+ *
+ * @param int $payment_id The current payment ID.
+ * @return void
+ */
+function edd_paypal_refund_admin_js( $payment_id = 0 ) {
+	// If not the proper gateway, return early.
+	if ( 'paypal' !== edd_get_payment_gateway( $payment_id ) && 'paypalexpress' !== edd_get_payment_gateway( $payment_id ) ) {
+		return;
+	}
+
+	// If our credentials are not set, return early.
+	$key = get_post_meta( $payment_id, '_edd_payment_mode', true );
+	if ( empty( edd_get_option( 'paypal_' . $key . '_api_username' ) ) || empty( edd_get_option( 'paypal_' . $key . '_api_password' ) ) || empty( edd_get_option( 'paypal_' . $key . '_api_signature' ) ) ) {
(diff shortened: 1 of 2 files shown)


cklosowski:
<3 This is awesome. Can't way to test it out.

includes/gateways/paypal-standard.php
```diff
@@ -653,3 +653,206 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
...
+function edd_refund_paypal_purchase( $payment_id = 0 ) {
+	// Verify the payment ID is a valid integer.
+	$payment_id = absint( $payment_id );
+	if ( ! $payment_id ) {
+		echo '<pre>' . var_export( __( 'Payment could not be refunded because the payment ID was invalid.', 'easy-digital-downloads' ), true ) . '</pre>';
+		die;
+	}
+
+	// Grab the transaction ID for the payment. Try notes first since transaction IDs weren't stored in transaction ID field originally.
+	$notes = edd_get_payment_notes( $payment_id );
```
pippinsplugins: Let's retrieve it from the transaction ID helper function first. That function actually looks through notes as well, so probably don't need to do both.

includes/gateways/paypal-standard.php
```diff
@@ -653,3 +653,206 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
...
+		'api_signature' => edd_get_option( 'paypal_' . $key . '_api_signature' )
+	);
+	$credentials = apply_filters( 'edd_refund_paypal_api_credentials', $credentials, $payment_id );
+
+	// Prepare body of the refund request.
+	$body = array(
+		'USER' 			=> $credentials['api_username'],
+		'PWD'  			=> $credentials['api_password'],
+		'SIGNATURE' 	=> $credentials['api_signature'],
+		'VERSION' 		=> '94',
```
pippinsplugins: Let's bump this to `124`, the latest version.

pippinsplugins:
I've manually merged this into `issue/4403` for further development and tweaks.

PR #4411, opened 2016-03-14
Fix lack of JS support when ajax is turned off

chriscct7:
When ajax is turned off, we give a back button from step 2 (checkout fields) to step 1 (gateway selection). Unfortunately, apparently you can't use Javascript to do the redirect when Javascript is disabled. Also the lack of query param stripping means that extensions that depend on edd_get_chosen_gateway() to work will find that there's a mismatch between what that function says the selected gateway is and what the radio buttons have selected (because those fallback to edd_get_default_gateway()). Crossref: https://github.com/chriscct7/edd-gateway-fees/issues/16 1 and 2


pippinsplugins:
Looks good to me. We can merge once unit tests pass.

Issue #3186, opened 2015-02-23
Proposal: Remove Price Inclusive of Tax Option


chriscct7:
This ticket, proposes the idea of removing the price inclusive of tax option but _without_ removing any of the existing functionality we have in EDD core (it may take a couple read through for this to be understood)

People use the "enter price inclusive of tax" option for 2 main reasons:
- They are required to show the price inclusive of tax, by law, on the product page
- It's easier to setup a product so the total price at checkout will be a specific number (say 19.99)
- Let's people charge the same total price everywhere in the world regardless of the tax amount

The main problems with this option are:
- It causes alot of bugs with the tax api
- Its the root cause of a rounding issue #3104 
- It makes for alot of complications inside EDD core when figuring out the item price
- It forces us to back calculate the product price and the amount of tax inside of it on the checkout page
  - It causes alot of complexity for all of our extensions

What if we could provide all the benefits of this option, without providing the option? 
This ticket, proposes the idea of removing the price inclusive of tax option.

Here's how it would work:
- On the edit product page, we'd add a new column next to the price called "Price after tax", where we'd use a bit of JS to automatically calculate the price after tax by adding the price to the left + the tax for the product based on the base country. This lets people instantly see how much their product will cost after checkout. 
- We continue using the "display product price inclusive of tax" on the product page, but we'd simply the logic greatly, to simply just $product_price_to_show = $product_price + $product_price \* $tax_rate;
- On checkout, since all product prices are entered exclusive of tax, our logic here is greatly reduced, because we don't need to back calculate anything, which solves the problem completely about taxes after discount calculations
- We'd use an update routine to set all currently set prices inclusive of taxes and back calculate and save the amounts exclusive of taxes

This lets people get all of the legal and convenience benefits of entering the product prices inclusive of tax on the backend without actually offering that option. It would take the entire tax API and simplify it so greatly, @pippinsplugins might actually start liking that API

Removing this option and providing complete backwards compatibility in terms of functionality and code really isn't as hard as it sounds.

We could rip out so much code (thousands of lines of pure tax calculation torture), headaches, heartaches, and needless complexity while providing everything we provide right now in terms of functionality. 

The end result? Putting the Easy back in Easy Digital Downloads.

chriscct7:
[...]

Issue #2347, opened 2014-06-04
Incorrect taxes for taxes after discounts


pippinsplugins:
I just received this report from a user but have not verified it yet:

The Taxes tab under settings has a checkbox to apply taxes after discount, but if you check it, taxes are applied before discount. Little confusing. 
The text reads:

> Check this if you would like taxes calculated after discounts. By default taxes are calculated before discounts are applied.

But if that's actually what you want, you have to uncheck that option.

Issue #1744, opened 2013-11-22
Tax system rewrite


pippinsplugins:
This is just a wrapper issue for the entire tax system rewrite.

This includes: #1739,  #1676, #1674, , #1513, #1511, #1505, and #1490

PR #4413, opened 2016-03-14
Allow for the removal of tax from edd_get_cart_item_price

chriscct7:
For simplicity, we'll have a cart with 1 product, with an entered cost of $100. 
We've set the global tax rate to 10%, and have an active discount fee of 10%

As in the rest of EDD, we should always calculate tax after discounts.

When item price excludes tax
Total should be ($100 - 10%) * 1.10 = $99
of which the tax amount is ($100 - 10%) *.10 = $9
of which the product amount is ($100 - 10%) * 1 = $90

This part works fine.

Also calculating without taxes works fine. 

So why does this code only not work when prices are entered inclusive of taxes?

When item price includes tax
Raw product amount: ( 100 / 1.10 (this step makes the price exclusive of 10% tax) ) = $90.91 ******
Final raw product amount (post-discount) should be $90.91 - 10% = $81.82
Final product amount should be $81.82 * 1.1 = $90.00
Final tax amount should be $81.82 * .1 = $8.18

Total should be $90.

However, the problem isn't that fee based discount isn't calculating correctly. As a matter of fact, when I substitute the get cart item price function calls for (float) 90.91, this is exactly what I get.

Note also, EDD's discounts should always match the fee based discount. So why does EDD handle both inclusive and exclusive fine, but the fee based discount cannot? 

It all boils down to one function, which at some point has changed. In fact, the docbloc on this function is completely misleading:

If you get the price of an item in the cart using edd_get_cart_item_price. As per the docbloc __this excludes tax always__. Well, unless of course it doesn't at all. In fact, this function doesn't take any sort of taxes into account. So when it claims to always return the price exclusive of tax, this only works for when an EDD site is set to exclude tax because the price entered excludes tax. If you call this function, when you have edd set to include tax, the result is "100.00". This is wrong, because the *actual* product price, exclusive of tax is $90.91. This is why when I substitute the calls to this function for the "90.91" it works.

Moreover, this function has a display function, called edd_cart_item_price(). This function is used to display the price "exclusive of tax". However, as you can see by comparing the functions, the latter actually works as expected. It does take into account that prices can be entered inclusive of taxes, and if that's the case, remove the tax amount. And only then does it format the price (which is unfortunate because otherwise we could just use this function).

So basically, what needs to happen here is EDD needs to correct edd_get_cart_item_price to work like it's supposed to, the way it's described, and the way the display equivalent of it correctly works.

For comparison purposes,
here is the display function, which works: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/f2f1f7db7322f5abe2d5900be0c40895978aed21/includes/cart/functions.php#L393

and here is the non-display function, which doesn't: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/f2f1f7db7322f5abe2d5900be0c40895978aed21/includes/cart/functions.php#L444

In this PR, I've adjusted the former function to have an extra parameter to allow for the removal of the raw tax amount.

While it's unideal to add another parameter to support something that the function claims to do by default, looking through EDD core, unwinding this mess might be more than anyone has bargained for.

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -406,7 +406,8 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
 
 		if( edd_prices_show_tax_on_checkout() && ! edd_prices_include_tax() ) {
 			$price += edd_get_cart_item_tax( $item_id, $options, $price );
-		} else if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
+		} 
+		if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
 			$price -= edd_get_cart_item_tax( $item_id, $options, $price );
 		}
 
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -405,13 +405,9 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
 	if ( ! edd_is_free_download( $item_id, $price_id ) && ! edd_download_is_tax_exclusive( $item_id ) ) {
 
 		if( edd_prices_show_tax_on_checkout() && ! edd_prices_include_tax() ) {
-
 			$price += edd_get_cart_item_tax( $item_id, $options, $price );
-
-		} if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
-
+		} else if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
 			$price -= edd_get_cart_item_tax( $item_id, $options, $price );
-
 		}
 
 		if( edd_display_tax_rate() ) {
@@ -445,9 +441,10 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
  * @since 1.0
  * @param int   $download_id Download ID number
  * @param array $options Optional parameters, used for defining variable prices
+ * @param bool  $$remove_tax_from_inclusive Remove the tax amount from tax inclusive priced products.
  * @return float|bool Price for this item
  */
-function edd_get_cart_item_price( $download_id = 0, $options = array() ) {
+function edd_get_cart_item_price( $download_id = 0, $options = array(), $remove_tax_from_inclusive = false ) {
 
 	$price = 0;
 	$variable_prices = edd_has_variable_prices( $download_id );
@@ -476,6 +473,11 @@ function edd_get_cart_item_price( $download_id = 0, $options = array() ) {
 		// Get the standard Download price if not using variable prices
 		$price = edd_get_download_price( $download_id );
 	}
+	
+	if ( $remove_tax_from_inclusive && edd_prices_include_tax() ) {
+
+		$price -= edd_get_cart_item_tax( $item_id, $options, $price );
+	}	
 
 	return apply_filters( 'edd_cart_item_price', $price, $download_id, $options );
 }
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -480,7 +480,7 @@ function edd_get_cart_item_price( $download_id = 0, $options = array(), $remove_
 	
 	if ( $remove_tax_from_inclusive && edd_prices_include_tax() ) {
 
-		$price -= edd_get_cart_item_tax( $item_id, $options, $price );
+		$price -= edd_get_cart_item_tax( $download_id, $options, $price );
 	}	
 
 	return apply_filters( 'edd_cart_item_price', $price, $download_id, $options );
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -405,13 +405,13 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
 	if ( ! edd_is_free_download( $item_id, $price_id ) && ! edd_download_is_tax_exclusive( $item_id ) ) {
 
 		if( edd_prices_show_tax_on_checkout() && ! edd_prices_include_tax() ) {
-			
+
 			$price += edd_get_cart_item_tax( $item_id, $options, $price );
-			
+
 		} if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
-			
+
 			$price -= edd_get_cart_item_tax( $item_id, $options, $price );
-			
+
 		}
 
 		if( edd_display_tax_rate() ) {
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -406,8 +406,7 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
 
 		if( edd_prices_show_tax_on_checkout() && ! edd_prices_include_tax() ) {
 			$price += edd_get_cart_item_tax( $item_id, $options, $price );
-		} 
-		if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
+		} if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
 			$price -= edd_get_cart_item_tax( $item_id, $options, $price );
 		}
 
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -445,7 +445,7 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
  * @since 1.0
  * @param int   $download_id Download ID number
  * @param array $options Optional parameters, used for defining variable prices
- * @param bool  $$remove_tax_from_inclusive Remove the tax amount from tax inclusive priced products.
+ * @param bool  $remove_tax_from_inclusive Remove the tax amount from tax inclusive priced products.
  * @return float|bool Price for this item
  */
 function edd_get_cart_item_price( $download_id = 0, $options = array(), $remove_tax_from_inclusive = false ) {
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -405,9 +405,13 @@ function edd_cart_item_price( $item_id = 0, $options = array() ) {
 	if ( ! edd_is_free_download( $item_id, $price_id ) && ! edd_download_is_tax_exclusive( $item_id ) ) {
 
 		if( edd_prices_show_tax_on_checkout() && ! edd_prices_include_tax() ) {
+			
 			$price += edd_get_cart_item_tax( $item_id, $options, $price );
+			
 		} if( ! edd_prices_show_tax_on_checkout() && edd_prices_include_tax() ) {
+			
 			$price -= edd_get_cart_item_tax( $item_id, $options, $price );
+			
 		}
 
 		if( edd_display_tax_rate() ) {


pippinsplugins:
No. If you need the item without tax, you can use `edd_get_cart_item_tax()`

chriscct7:
What? edd_get_cart_item_tax gets the amount of tax in a product price. That isn't what's required. What is required is the price of a product always exclusive of tax, something that this function claims do do quite explicitly in the docbloc: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/263c60f9d57a61ff795b9d81c51b2cd418e8ad0c/includes/cart/functions.php#L440

```
 * Get Cart Item Price
 *
 * Gets the price of the cart item. Always exclusive of taxes
 *
 * Do not use this for getting the final price (with taxes and discounts) of an item.
 * Use edd_get_cart_item_final_price()
```

The fact that when price includes tax returns a product price inclusive of tax when the function clearly is not only stated to do the opposite, but also previously did the opposite, is what the pr fixes in a backwards compatible way

chriscct7:
Actually not only did the function do this before, the third parameter, which was actually removed (non-backwards compatibily) was literally this exact thing:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/1744

Here's the commit that deprecates that param (I can find the one that removed it without backwards compat if you really want):
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/c74a79b621589b13b3db9fb384c5e6a70ce88932

pippinsplugins:
`edd_get_cart_item_price()` does not include any tax on it.

pippinsplugins:
Actually that's not what you want. Hang on.

chriscct7:
And we also have the note where this docbloc description is also made as a comment: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/463dc851e45ebb1e307e43fe6caf699741cd61d7#diff-d3d6b11bcd1da4cb4dd855a546f8a7e1R371

which I might point out you literally wrote this is how this function is designed to work, and that's how extensions have used it

Later you fixed it so that you wouldn't need to pass a param (a backwards compatibility breaking but minor change given the function at this point still does the same thing): https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/c74a79b621589b13b3db9fb384c5e6a70ce88932

Then you made it better: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/f0599939d074b2d2e4109332dc21e4d63c0c891e/includes/cart/functions.php#L346

Then you finally removed all of it: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/117a2a34bcb0e0bcd892aa655f959aa7c35e54ed

chriscct7:
^That's how the function went from working for inclusive of tax to not working for inclusive of tax.

pippinsplugins:
That docbloc is not wrong or misleading. It's exactly correct. `edd_get_cart_item_price()` does nothing but retrieve the amount set in the backend for the product. It does not take taxes or discounts into account. Adding an option to make it exclude taxes will not happen.

pippinsplugins:
Ok I see where the disconnect is. Sorry, forgot the `inclusive of tax` portion. Thinking.

chriscct7:
Pippin, it doesn't anymore because you broke backwards compatibility. 

This is the commit in which you added that docbloc:
image

chriscct7:
All that this does is re-introduce the parameter that at one point you removed which used to literally do this same thing.

pippinsplugins:
@SDavisMedia could you test this along side https://github.com/chriscct7/edd-discounts/issues/92 and let me know if it resolves it 100% for you?

chriscct7:
In an ideal world, we'd go back and make the function automatically remove the inclusive of tax part like it used to do. In that case, we might even still re-introduce the third parameter, and make it do nothing so at least we don't have people getting strict errors for using a third parameter that was yanked out.

However, this function is used all over the place. Its going to require a lot of time to do that ^, so in the meantime this pr re-adds the third parameter, and makes it do exactly what it did before, remove the inclusive of taxes portion of a product price. This serves as an excellent stopgap until we can test and untangle the rest and then deprecate this function.

pippinsplugins:
I'm trying to remember why we removed it in the first place. If I recall it had to do with it causing some real cluster fucks that resulted in a _lot_ of miscalculations.

chriscct7:
We removed the functionality because of the discounts mess. https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/2347. As a result, you then decided to move the removal of inclusive tax to the display function so the checkout didn't break: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/commit/19ca686b32359effffc969357a8eaf65738bbe82

That's why the display function correctly has the right no-tax value and why the underlying code one does not.

chriscct7:
Again, to unwind this and fix it to how it worked before will probably involve a ton of testing. In the meantime though re-adding the third parameter, and making it do what it did before is a great stop gap measure and we can always deprecate it again later. This solves both PHP strict notices for too many parameters and this issue in one simple change.

cklosowski:
Removing the milestone, there's no way we can test this throughly enough for the `2.5.10` release. Just can't happen that quickly. I'd be happy to look into it for `2.6` provided we get an open issue for it, so we can make it apparent it's coming to developers that are relying on this info.

chriscct7:
Let me summarize the history of this function so as to further explain what the problem is, how it got here, how this pr fixes it (without the need for any extensive testing), and what should be done in the future:

This isn't difficult:
1. edd_get_cart_item_price is introduced. It has 3 params for the download id, variable prices, and whether or not to remove the inclusive of tax portion from a product price.
2. Pippin decides that the function, correctly, as documented should not require the third parameter, that the function should automatically always remove the inclusive of tax portion of a product price, and deprecates the third parameter.
3. Pippin removes the third parameter (bc breaking, no actual reason for it)
4. Issue arises when discounts after/before taxes comes up. Instead of fixing the function the inclusive of tax removal portion is removed from the function. The function now no longer does what it says it does or is supposed to do (major bc breaking).
5. Issue arises that as a result of 4, the checkout now displays the wrong price. This is fixed by moving what used to auto remove the inclusive of tax code into the display function. This is why the display function handles the math correctly but the function its derived from does not.
6. Discounts Pro relies on how the function originally was supposed to work. It now finds that product prices are inclusive of tax, leading to miscalculated discounts.
7. PR is opened to readd the third parameter. As a reminder this is a __default off__ parameter. Which by definition means it cannot break anything. It solves 2 issues: that strict errors are thrown when a plugin calls the function with the third param, and the inclusive of tax by letting plugins remove tax like they used to
8. In the future, we can fix it to auto correct. This is the part that will require testing, and its not what is in the pr. At that time the third  param will be deprecated, like it was before.

SDavisMedia:
@pippinsplugins @chriscct7 @cklosowski EDD `feature/edd_get_cart_item_price_inclusive_of_tax` (with a manually bumped version number to meet DP requirements) and DP `issue/92` fixes the issue for me between these two plugins. Tested everything outlined here: https://github.com/chriscct7/edd-discounts/issues/92

cklosowski:
@SDavisMedia Do you have any of the Tax extensions installed? Just curious, most of us do'nt.

SDavisMedia:
@cklosowski no not at all.

pippinsplugins:
@SDavisMedia Good, nice to confirm it resolves the issue at hand. Now we just need to determine beyond any shadow of a doubt that we're not going to cause a problem if we merge this.

chriscct7:
One of the commonly brought up "worries" of this change is it's impact on extensions like EU VAT. So I spent the last 3 hours refreshing the checkout and product pages over and over and changing the settings and digging through their code. Here's the results:

There are 3 uses of the third param in EDD EU VAT. 

The first is in the PDF template for PDF Invoices, which actually on current master displays the incorrect result. The pr actually fixes it. If you turn off WP_DEBUG to silence the strict errors it throws, the code is supposed to output the prices of the products without inclusion tax. However, because of the lack of the third param, currently the PDF invoice incorrectly shows the price included of tax, and then shows the same tax in the line item for tax. Thus if you actually try to add the numbers together, you'd find the result of the summation of the product line items and the tax line is greater than the provided total (since the summation of the product lines is greater than the subtotal). When using the provided pr, the result is correct.

Before:
image

After:
image

The second and third are on the product page for the product price and the custom cart template.

In both cases, regardless of whether you use the PR or master, the EU VAT labels picked are wrong. For example using master:
image
On a $100 item, price entered inclusive of tax, the base price of the item should be $90.91 with $9.09 of tax

Despite this, the label chosen by EU VAT to represent this is:
image

It should read $90.91- Exclusive of 10% VAT

However, regardless of whether you use the PR or not, the end result is that the calculation results in an exactly identical result. The reason for this is that EU VAT has chosen to write their plugin in such a way it avoids the issue completely. This requires (by a rough search of their files) 28 version comparisons. That's how much of a mess this simple removal has caused. We can fix all of that, with 7 lines of code. Once we fix it, they over time can bump their min edd version requirement and eventually require this version or higher. That would allow them, from a quick look to eliminate a few hundred lines of code that they'll no longer need, since they won't have to hope and pray (to no fault of their own) that between EDD and their overrides that the inclusion tax is accounted for correctly. It'd make a lot of their functions quite a bit more simple.

In the end, this pr has no negative impact on the extension (they will still need to fix their label bug), while actually fixing some bugs. I did find quite a few bugs testing their extension. If someone knows the repo for it, I'd be happy to help report and fix some of these.

Across every single extension we have a downloaded copy of for the marketplace, there are only 3 calls of this function with 3 parameters. And we just covered all of them. Because they are all in EU VAT.

Now I know what you're thinking, oh but there must be people doing this elsewhere. That's why I took the time to one at a time manually page through 151 pages of GitHub code search results for "edd_get_cart_item_price". And if you ignore EU VAT, old copies of EDD, the only other plugin that uses it, in all of GitHub is.........the crowdfunding plugin by Appthemer (not the main astoundify one, the really tiny one). But wait, there's more. If you activate aforementioned plugin and dive into their codebase, you'll find, yet again (and sorta unsurprisingly given the plugin ceased development soon after or before the bc breaks) that the plugin is expecting the third param to work the same way it does in the pr. In fact, the comparison they use for the custom price check calculation (the 1 place they use it with 3 params) doesn't work properly without it.

Then there's https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3186#issuecomment-76911549.

Regardless of this, there are absolute certainties:
- This restores functionality previously in EDD. 
- Previously there was 3 parameters in the function. While you may not care that this has the possibility to throw tons of strict errors, the only argument against this pr that have been heard is that it could break plugins that have a third parameter. This is a pretty terrible argument, for a simple reason: why would a plugin developer, go out of his way and not only call a function with a third parameter that does not exist, and moreover turn it default off to on, unless they wanted something to happen? That's like calling wp_mail( $a, $b, $c, $d, $e, $f, $g, $h, $i, $j, $k) and hoping WP core never adds any parameters. 
- It is not a guess, nor a wait and see, it is an absolute certainty that the result of a series of 3 backward compatibility breaking changes has broken several extensions. It has broken Discount Pro, it has broken EU VAT, and it has broken AppThemer's Crowdfunding plugin. Basically, it's, as expected, broken everyone who uses the third param, because everyone's code expects it to work.

pippinsplugins:
@bseddon Would you be able to take a quick look to confirm that this change won't hurt anything on your end?

pippinsplugins:
@chriscct7 Excellent work. If we can get a quick confirmation from Bill that everything lines up with what you've found, we'll merge.

bseddon:
I'm tied up for another couple of hour but will then attempt to take a look.

pippinsplugins:
Thanks!

chriscct7:
@bseddon if you have a private repo somewhere, if I could get access to it that would be great. I found some bugs during the testing to report.

bseddon:
I don't have an accessible copy.  As you mention in the thread above, the function 'edd_get_cart_item_price' is called using the third parameter.  However, the discussion above does not apply to the VAT plugin because I don't support prices including taxes as the philosophy (for want of a better term) does apply in a VAT context.  The philosophy of this EDD option is that the price entered is the price to be charged.  That may work in a scenario where there is one rate of tax but not when there are 28.

Suppose that I have a product which cost $10.  My VAT rate in the UK is 20% so the inclusive price is $12.  If I sell to someone who is also in the UK or is in France where the rate is also 20% then everything works.  But suppose I sell this product to Hattie from Denmark where VAT is 25%.  Because the price is fixed and the amount of tax to be accounted for is now $2.40 not $2 I'm out of pocket.

However, it gets much worse for someone buying who is VAT exempt - an EU business with a valid VAT number or a non-EU citizen.  In these scenarios the effective VAT rate is zero.  I'm fat, dumb and happy because I'm collecting all of the $12.  However, the buyers are fuming because they are still paying $12.

bseddon:
Also, I think this behavior is illegal for EU businesses.  When prices are entered including taxes we are advertising a price that includes tax but then, potentially, 'charging' and keeping the tax.  I think the tax authorities of EU member states will see this as problematic and perceive it as a ruse to overcharge the customer - that's their job.

bseddon:
I've downloaded the pull request and applied it to my test site.  Everything seems to work OK.  Given my comment above, I did not enable the option to enter prices including VAT.

pippinsplugins:
Thanks @bseddon.

pippinsplugins:
Merging.

PR #4416, opened 2016-03-18
i18n updates

ramiy:
Fixing different translation string to make it easier to translate the plugin.

Fox example, replacing `<code>` tags with `%s` placeholders, moving the code out of the translation string. Replacing difficult to understand strings like `%shere%s` with intuitive strings like `<a href="%s">here</a>`, and more...

We've done the same in WordPress core in 4.1 - 4.5, same standards should be implemented in EDD.


JJJ:
This PR has gone quite stale in 2 years.

* Some things are fixed or non-existent in 3.0
* Placeholders may not be ideal in all locations

Going to close this PR, since it's unlikely to see a refresh anytime soon. If someone wants to prioritize, please feel free to do so.

Issue #4429, opened 2016-03-24
Add way to blacklist URLs from triggering a session start


pippinsplugins:
Sometimes there are pages where we never want a session to start (think high traffic landing page). To aide in this, we should add a way to blacklist URLs from starting sessions.

This will be a simple, filtered array of URLs / slugs. If we detect we're on a black listed URL, we prevent the session from starting.

Related: #4428 and #4284

pippinsplugins:
@cklosowski Any further feedback here or are you fine with merging?

cklosowski:
Let me run one more test @pippinsplugins. I'll merge if all is well.

pippinsplugins:
:+1:

cklosowski:
Working much better. Merging.

cklosowski:
@pippinsplugins quick thing...`@since 2.5.11` please.

Issue #4428, opened 2016-03-24
Never start sessions on RSS


pippinsplugins:
It was brought to my attention today that we are allowing sessions to be started on RSS feeds. That's just silly and causes some serious performance problems on sites that have very active RSS feeds (such as our own easydigitaldownloads.com).

We need to prevent sessions from ever being started from an RSS feed.

pippinsplugins:
Related: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/4284

PR #4430, opened 2016-03-24
Introduce should_start_session(). Issues 4428 and 4429

pippinsplugins:
Includes edd_session_start_uri_blacklist filter and edd_start_session filter to provide ways to disable session starts. #4429 #4428

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -75,6 +75,10 @@ class EDD_Session {
 
 		} else {
 
+			if( ! $this->should_start_session() ) {
+				return;
+			}
+
 			// Use WP_Session (default)
 
 			if ( ! defined( 'WP_SESSION_COOKIE' ) ) {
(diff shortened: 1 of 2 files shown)


includes/class-edd-session.php
```diff
@@ -300,13 +304,56 @@ public function use_cart_cookie() {
...
+				'feed/rss2',
+				'feed/rdf',
+				'feed/atom',
+				'comments/feed/'
+			) );
+
+			$uri = ltrim( $_SERVER[ 'REQUEST_URI' ], '/' );
+			$uri = untrailingslashit( $uri );
+
+			if( in_array( $uri, $blacklist ) ) {
```
cklosowski: This breaks for domains like `http://local.dev/edd/feed/` where you have WordPress in a subdirectory like multisite or even just a sub directory base install since it's looking for the request URI to be just `feed` not `edd/feed` which is the result of the `ltrim` and `untrailingslashit`

cklosowski:
left a comment, also, if we could get a unit test in place for these since it's a public function we could easily test it out to verify that URLs will return true or false when expected.

includes/class-edd-session.php
pippinsplugins: Ok, so if it's multisite, we should also grab the site's path and include that. Agreed?

PR #4431, opened 2016-03-26
Remove old WordPress version compatibility. EDD doesn't support WP < 4.0

chriscct7:
Removes back compat that's wrapped in wp_version checks, for versions of WP that EDD doesn't support.


pippinsplugins:
:+1:

PR #4434, opened 2016-03-28
Let extensions validate fields after a user has been logged in

chriscct7:
Right now, a couple of our extensions add extra validation to the checkout form.

The problem is that if they rely on knowing the correct user_id, right now, the hook they have to use is before EDD logs in a user if they've used the registration or login forms. To accurately log these users in, we need to add a new hook to let the extensions do this.

I've submitted this to master, as right now, a couple of our extensions have bugs that mean the validation will fail even for valid users, and the sooner we can get this hook out, the better, so the extensions can immediately update and fix this problem.

Additionally, there appears to be a related bug in core in that EDD is not correctly storing the user_id of the user placing an order if the user uses the login/registration forms while checking out (for example with guest checkout disabled). Haven't had time to look into this yet, but it's possible that either EDD isn't saving the correct user_id in the payment post meta, or the customer table isn't getting the user_id, or both.

Working on finding it

### DIFF
--- a/includes/process-purchase.php
+++ b/includes/process-purchase.php
@@ -46,6 +46,9 @@ function edd_process_purchase_form() {
 
 	// Validate the user
 	$user = edd_get_purchase_form_user( $valid_data );
+	
+	// Let extensions validate fields after user is logged in if user has used login/registration form
+	do_action( 'edd_checkout_user_error_checks', $user, $valid_data, $_POST );
 
 	if ( false === $valid_data || edd_get_errors() || ! $user ) {
 		if ( $is_ajax ) {


pippinsplugins:
Why not just move `edd_checkout_error_checks` down below `$user = edd_get_purchase_form_user( $valid_data );`? I don't see any reason that wouldn't work.

chriscct7:
Because for plugins that hook into that, it would mean that their error messages are output if there's an empty cart.

chriscct7:
To avoid that, extensions using the new hook must use `! edd_get_cart_contents() && ! edd_cart_has_fees()` to not output an edd error if the cart is empty

pippinsplugins:
@chriscct7 can you show me an example of how you're hooking into it?

chriscct7:
@pippinsplugins to the existing hook? https://github.com/chriscct7/edd-checkout-fields/blob/master/classes/frontend/class-checkout.php#L46

But the problem is that the validation runs before the user is logged in if logging in/registering on the form because that's done here https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/bug/user-post-login-error-action/includes/process-purchase.php#L48 and the extensions all hook in here https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/bug/user-post-login-error-action/includes/process-purchase.php#L37

chriscct7:
Right now, extensions don't need to check for empty cart before outputting the errors because the hook only runs if the cart isn't empty. The new hook will have to make extensions count for that because its outside the logic for if cart is empty, because it has to be after the login function call

chriscct7:
Though frankly, a better option might be to 
- move the login thing above the existing hook extensions use. There's no point in validating if the user logged in or registered or inserting their account if the cart is empty, as they shouldn't be able to register on an empty cart when using that specific function.

OR

- return right away when the cart is found to be empty

pippinsplugins:
I prefer `return right away when the cart is found to be empty`

cklosowski:
Punting for response to feedback.

chriscct7:
`return right away when the cart is found to be empty` as it turns out that's what EDD already does. As a result, pr is ready for merge

Issue #4443, opened 2016-03-31
In Customer details page icons are not focused


samikeijonen:
In Customer details page there are dashicons with links (Profile, Notes, Tools, Delete). I'd add hover and focus styles for those links.

Also HTML semantic feels weird for me. Now the icons list are like this:

```
<ul>
<a><li><span class="dashicon"></span></li></a>
<a><li><span class="dashicon"></span></li></a>
<a><li><span class="dashicon"></span></li></a>
</ul>
```

I'd change it to something like this:

```
<ul>
<li><a><span class="dashicon" aria-hidden="true"></span><span class="screen-reader-text">Info text</span></a></li>
<li><a><span class="dashicon" aria-hidden="true"></span><span class="screen-reader-text">Info text</span></a></li>
<li><a><span class="dashicon" aria-hidden="true"></span><span class="screen-reader-text">Info text</span></a></li>
</ul>
```

samikeijonen:
Also when you click any of the icons, you lose the focus. Focus could switch to element `edd-item-card-wrapper` or the first focusable element inside it.

pippinsplugins:
Thanks @samikeijonen! Milestoned for 2.6

samikeijonen:
I'll create PR as soon as I have time :)

pippinsplugins:
Great! Please submit against the `release/2.6` branch.

samikeijonen:
Hmmm I think I broke the internet:) In PR #4445 I tried to fix this but it seems I got all the balls rolling.

pippinsplugins:
New one looks good!

pippinsplugins:
Looks good, merged!

PR #4447, opened 2016-03-31
Set focus on icons in Customer details page

samikeijonen:
Also change HTML semantics, now links are inside LI element.


samikeijonen:
Here is a fix for issue #4443.

PR #4451, opened 2016-04-02
Fix couple things in docblocks

JeroenSormani:
Some typo's / added descriptions on some params/return / other fixes


includes/payments/functions.php
```diff
@@ -1474,7 +1477,7 @@ function edd_get_payment_notes( $payment_id = 0, $search = '' ) {
 	}
 
 	remove_action( 'pre_get_comments', 'edd_hide_payment_notes', 10 );
-	remove_filter( 'comments_clauses', 'edd_hide_payment_notes_pre_41', 10, 2 );
```
chriscct7: why was the 2 removed?

chriscct7:
@JeroenSormani looks good. One little issue to fix (left a note)

includes/payments/functions.php
JeroenSormani: `remove_filter` doesn't have a 4th parameter ;-)

chriscct7: oh lol, I didn't look at the remove part. Its all good

PR #4452, opened 2016-04-02
Small tweaks in EDD_Download

JeroenSormani:
Mostly replacing a function call which is a workaround to call a property, to just directly call the given property.

### DIFF
--- a/includes/class-edd-download.php
+++ b/includes/class-edd-download.php
@@ -611,7 +611,7 @@ class EDD_Download {
 	public function increase_sales( $quantity = 1 ) {
 
 		$quantity    = absint( $quantity );
-		$total_sales = $this->sales + $quantity;
+		$total_sales = $this->get_sales() + $quantity;
 
 		if ( $this->update_meta( '_edd_download_sales', $total_sales ) ) {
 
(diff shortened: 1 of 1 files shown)


chriscct7:
This actually breaks quite a bit of things.

For example $this->sales is not populated unless get_sales is called, and by default that's not run in the constructor.

See https://travis-ci.org/easydigitaldownloads/Easy-Digital-Downloads/jobs/120330303

JeroenSormani:
Hmm, interesting to see.. 

Say this function `edd_get_download_sales_stats` basically does this: `EDD_Download( $download_id )->sales`
While me doing `$this->sales` doesn't work..?

Are `$this` called properties not going through the magic `__get` and from outside are, or is there some other difference that I'm missing here?
(I never use magic methods, so I'm new on that area ;-)

cklosowski:
Magic __get functions do not work in the $this reference. Look up magic
methods. They only get run on "inaccessible" properties. So externally they
are run. But internally they are not.
-- 
Chris Klosowski

cklosowski:
Ran into the same thing writing EDD_Payment. That's where I learned that
little "trick" ;-)
-- 
Chris Klosowski

JeroenSormani:
Thanks for the crash course @cklosowski! 

I've already made some edits so its calling the methods instead of the properties. Locally its passing all tests :+1: 

This also is sort-of the reason I've always avoided magic methods, they do some 'magic' behind the scenes (/aren't dead obvious), which IMO is not a good thing in programming. But glad I learned this :smile:

Issue #4459, opened 2016-04-05
Plugin update POST firing twice on "Dashboard" > "Updates"


mintplugins:
Plugins built using the EDD_SL_Plugin_Updater class "Phone Home" twice when on the "Dashboard" > "Updates" page or the "Plugins" page. This may be un-necessary. 

The "plugins_api" hook is what's causing it to "phone home" twice on plugins and updates pages. If we cache the result of that for 24 hours in a transient (but not the result that comes through "pre_set_site_transient_update_plugins"), then "Dashboard" > "Updates" page will check for updates every time but only once every 24 hours on the "Plugins" page - and in turn only once every 24 hours will it check twice on the "Dashboard" > "Updates" page as well.

mintplugins:
PR: ​https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/4460

pippinsplugins:
Let's get this in. Left a few small tweaks on the PR.

PR #4460, opened 2016-04-05
Store api_request in a transient for 24 hours.

mintplugins:
Plugins built using the EDD_SL_Plugin_Updater class "Phone Home" twice when on the "Dashboard" > "Updates" page or the "Plugins" page. This may be un-necessary.

The "plugins_api" hook is what's causing it to "phone home" twice on plugins and updates pages. If we cache the result of that for 24 hours in a transient (but not the result that comes through "pre_set_site_transient_update_plugins"), then "Dashboard" > "Updates" page will check for updates every time but only once every 24 hours on the "Plugins" page - and in turn only once every 24 hours will it check twice on the "Dashboard" > "Updates" page as well.

store-api_request-in-a-transient-for-24-hours

### DIFF
--- a/includes/EDD_SL_Plugin_Updater.php
+++ b/includes/EDD_SL_Plugin_Updater.php
@@ -241,11 +241,22 @@ class EDD_SL_Plugin_Updater {
 				'reviews' => false
 			)
 		);
+		
+		//Get the transient where we store the api request for this plugin for 24 hours
+		$edd_api_request_transient = get_site_transient( 'edd_api_request_' . $this->slug );
+			
+		//If we have no transient-saved value, run the API, set a fresh transient with the API value, and return that value too right now.
+		if ( empty( $edd_api_request_transient ) ){ 
 
-		$api_response = $this->api_request( 'plugin_information', $to_send );
-
-		if ( false !== $api_response ) {
-			$_data = $api_response;
+			$api_response = $this->api_request( 'plugin_information', $to_send );
+			
+			//Expires in 1 day (86400 seconds)
+			set_site_transient( 'edd_api_request_' . $this->slug, $api_response, 86400 );
+			
+			if ( false !== $api_response ) {
+				$_data = $api_response;
+			}
+			
 		}
 
 		return $_data;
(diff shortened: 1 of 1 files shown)


includes/EDD_SL_Plugin_Updater.php
```diff
@@ -79,35 +79,47 @@ public function init() {
...
-
-				if( version_compare( $this->version, $version_info->new_version, '<' ) ) {
-
-					$_transient_data->response[ $this->name ] = $version_info;
-
+		
+		global $edd_update_plugins_flag;
+	
+		//If flag is true, this is our second go around with pre_set_site_transient_update_plugins
+		if ( $edd_update_plugins_flag ){
```
pippinsplugins: Let's adjust this so that it returns early if `! $edd_update_plugins_flag` instead of putting everything below inside of the conditional. Gives us less levels of indention that way.

includes/EDD_SL_Plugin_Updater.php
```diff
@@ -79,35 +79,47 @@ public function init() {
...
-				$_transient_data->checked[ $this->name ] = $this->version;
-
+	
 			}
-
+		
+			$edd_update_plugins_flag = false;
 		}
-
+		else{
```
pippinsplugins: Should be `} else {` to be consistent with code formatting standards.

includes/EDD_SL_Plugin_Updater.php
```diff
@@ -241,11 +253,22 @@ public function plugins_api_filter( $_data, $_action = '', $_args = null ) {
...
+		//Get the transient where we store the api request for this plugin for 24 hours
+		$edd_api_request_transient = get_site_transient( 'edd_api_request_' . $this->slug );
+			
+		//If we have no transient-saved value, run the API, set a fresh transient with the API value, and return that value too right now.
+		if ( empty( $edd_api_request_transient ) ){ 
+
+			$api_response = $this->api_request( 'plugin_information', $to_send );
+			
+			//Expires in 1 day (86400 seconds)
+			set_site_transient( 'edd_api_request_' . $this->slug, $api_response, 86400 );
```
pippinsplugins: To make it easier to grok, change `86400` to `DAY_IN_SECONDS`

mintplugins:
did it :)

cklosowski:
For #4459 (linking)

cklosowski:
this is passing and we've improved it with a hash instead of the slug (since it's possible for the slug to be over the limit of the transient name after adding the prefix). Merging.

Issue #4064, opened 2015-12-09
Customize product preview with oEmbed in 4.4


pippinsplugins:
WordPress 4.4 supports oEmbed for post objects. Let's expand it to include support for embedding products on other sites along with their prices and purchase buttons.

https://make.wordpress.org/core/2015/10/28/new-embeds-feature-in-wordpress-4-4/

cklosowski:
👍 great idea.

leewillis77:
Came here to raise this (+1)

pippinsplugins:
:)

BurlesonBrad:
@pippinsplugins @cklosowski Something in my cranium wants to use this as a way to satisfy:
Always Be Learning
Contribute
Challenge My Self
Now, being a very slow learning, time constrained, dad of three, thick headed, and perseverantly stubborn kinda guy, I ask for a bit of help from support & 'Phil' gave me a few generalized links.  But, would either of you two coding super rocket ninja geniuses have any other helpful 'breadcrumbs' that would help me learn, and contribute to EDD getting their products to work w/ oEmbed?

pippinsplugins:
Happy to help though I haven't even look into the API to see how it works yet.

pippinsplugins:
Initial pass at this:
![screen shot 2016-04-05 at 9 38 57 pm](https://cloud.githubusercontent.com/assets/1034109/14304247/d80bcebc-fb76-11e5-995d-f8f78d98d1eb.png)
I've opted to only add the price and not purchase buttons to keep it nice and simple for now.

pippinsplugins:
@easydigitaldownloads/core-devs anyone else have any feedback on this?

BurlesonBrad:
Sorry.  gimme a sec.....

BurlesonBrad:
Ok.... Instead of Site Icon & Site Name in the lower left, I think it should *indeed* need to have some type of CTA ~*as in a button of some sort* 

Existing oEmbeds have the featured image & the excerpt ~ I think those would be critical.
Then afterwards, instead of the ```site_icon``` & ```blog_name```, shouldn't the priority of *"digital-real-estate"* be some type of **actionable** item? *(not just the site name?)*

....just my two cents.  And, **_yes,_** you might need a refund :laughing:

spencerfinnell:
Was going to say the same -- a purchase link/button should be able to be added right? A direct link to the checkout on single priced items, and a more details button for variable pricing?

[...]

PR #4465, opened 2016-04-06
Introduce edd_add_oembed_price(). #4064

pippinsplugins:
#4064

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -925,3 +925,26 @@ function edd_add_download_post_classes( $classes, $class = '', $post_id = false 
 	return $classes;
 }
 add_filter( 'post_class', 'edd_add_download_post_classes', 20, 3 );
+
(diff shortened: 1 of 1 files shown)


includes/template-functions.php
```diff
@@ -925,3 +925,26 @@ function edd_add_download_post_classes( $classes, $class = '', $post_id = false
...
+ * @since 2.6
+ * @return void
+ */
+function edd_add_oembed_price() {
+
+	if( 'download' !== get_post_type( get_the_ID() ) ) {
+		return;
+	}
+
+	$show = ! get_post_meta( get_the_ID(), '_edd_hide_purchase_link', true );
```
cklosowski: We need to filter the `$show` variable so that extensions like EDD Hide Downloads can tell the embed not to work if the download is supposed to be hidden.

PR #4466, opened 2016-04-06
Added docblock for edd_post_insert_discount action

JeffMatson:
Was already digging through to get more information on this hook.  Figured I would document it while I was here.  Not familiar with your doc standards, so feel free to do whatever with it.


pippinsplugins:
Wonderful.

PR #4472, opened 2016-04-06
Fix return value in docblock for edd_store_discount()

leewillis77:
Hi;

Minor doc fix for edd_store_discount() which returns a post ID, not bool.


pippinsplugins:
Thanks!

PR #4478, opened 2016-04-07
action for users to include their own advanced filters on the payments table

norcross:
currently there is no way to include your own advanced filters (my use case is a filter for showing payments by a particular gateway). this allows that.


cklosowski:
@pippinsplugins I see no reason this shouldnt' just go into `2.5.11`

cklosowski:
Merging.

PR #4490, opened 2016-04-14
Added filters to process the fees on add and retrieval

daigo75:
The filters allow any 3rd party to process the fees as they added and retrieved.

Notes
--
1. There are two filters in EDD_Fees::add_fee():
* `edd_fees_add_fees` allows to process a single fee being added to the list of fees. This is useful when one only wants to alter a specific fee as it's added. It's sufficient to check the fee type to determine if it needs to be processed.
* `edd_fees_set_fees` allows to process all the fees that are being associated to a session. This is slightly different from the first filter, and it can't be used easily for the same purpose. To alter a single fee, it would be necessary to loop through all of them, find the appropriate fee and, if it wasn't already processed, update it.

2. The filter names follow the `<scope>_<action>` structure. They can be changed, if necessary.


includes/class-edd-fees.php
```diff
@@ -97,7 +97,10 @@ public function add_fee( $args = array() ) {
 		$args['amount'] = edd_sanitize_amount( $args['amount'] );
 
 		// Set the fee
-		$fees[ $key ] = $args;
+		$fees[ $key ] = apply_filters( 'edd_fees_add_fees', $args, $this );
```
pippinsplugins: This should be `fee` not `fees`.

daigo75: Yes, I probably copy/pasted the filter names and forgot to adjust it. Would you like me to change it?

pippinsplugins: If you could, yes please :)

daigo75: [Done already, this morning](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4490/commits/d0d83153fa9e45820857e95a67d5f547908e0aed). :+1:

pippinsplugins:
I've manually merged this into our `release/2.6` branch. Thanks!

pippinsplugins:
https://github.com/easydigitaldownloads/easy-digital-downloads/commit/0754f8ccedef3270766e753953700e7585386af0

Issue #4329, opened 2016-02-22
filesize() being run on external files in process-download.php


pippinsplugins:
I've seen this error twice today:

PHP Warning:  filesize(): stat failed for https://github.com/{redacted}/natural-lite/archive/master.zip in /wp-content/plugins/easy-digital-downloads/includes/process-download.php on line 199

If a file is external, we should never run `filesize()` on it.

mintplugins:
This is the same issue as this: 
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/4321

I had that exact error message when we were initially discussing this issue:
https://eddwp.slack.com/archives/edd-support/p1455901084001739

chriscct7:
The solution to all of these issues seems to be exactly what me and john talked about last week, which is if we implement a simple if not path or URL in WP content then redirect method, then this ^ would be bypassed as would the other nginx issue. As opposed to trying to hunt down one or more issues without fixing the one we already know exists, we're (based on all the theories I've seen thus far) probably just baindaiding over the real issue, which is that (not path && not in WP content) URLs are being sent to direct.

chriscct7:
Related #4321

pippinsplugins:
@chriscct7 I agree.

mintplugins:
@chriscct7 @pippinsplugins Preventing the issue further up the line seems to make the most sense to me as well.

cklosowski:
This is being done in #4328, and is in PR #4335

pippinsplugins:
I've merged this into `2.5.10-beta` and am sending it to a few users.

mintplugins:
I have a ticket where this is still happening for relative links but not affecting downloads. However it is still causing an error in the server logs:

https://secure.helpscout.net/conversation/191893627/32117/?folderId=749231

cklosowski:
Relative links are local so should have their file size determined.

On Sunday, April 17, 2016, Phil Johnston <notifications@github.com> wrote:

> Reopened #4329
> <https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4329>
> .
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub
> <https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4329#event-630525382>
>


-- 
Sent from Gmail Mobile

mintplugins:
[...]

PR #4502, opened 2016-04-18
Suppress errors on filesize() call 

cklosowski:
#4329


pippinsplugins:
Merge once builds pass.

PR #4527, opened 2016-04-30
fix the payment icon name.

suifengtec:
fix the payment icon name.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -1266,7 +1266,9 @@ function edd_payment_icons_callback( $args ) {
 
 				} else {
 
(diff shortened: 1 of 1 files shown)


suifengtec:
@cklosowski  the '$option' may be '支付宝'  or other non-English string, but I think a filter hook 'edd_accepted_payment_支付宝_image' is not a good idea.

cklosowski:
I don't think this is a 'bad' change, however we can't simply remove that filter because people maybe be depending on it. If someone has a current gateway they've written, and are expecting us to use the `$option` not the `$key` this will break.

I agree that with non-ascii characters this could be a problem, but we cannot break backwards compatibility. I'm going to break this off into an issue branch, keeping your commits in tact so I can work up a fully backwards compatible platform.

Issue #4311, opened 2016-02-17
Introduce customer meta table


pippinsplugins:
We really need to introduce a customer meta table that allows us to store various (arbitrary?) pieces of information for customer records.

This can be based off of how we did it for AffiliateWP: https://github.com/AffiliateWP/AffiliateWP/issues/537

One example of something that relies on a customer meta table: storing the customer ID from Stripe that is associated with the EDD customer.

cklosowski:
:heart:

crsmoore:
You already know how much I would love this one. :+1:

cklosowski:
Going to make a checklist of things I See (feel free to alter if needed)
- [ ] Update `uninstall.php` with any tables we need to remove, including the new `customermeta` table.

Ok only one thing I saw so far. Looking good. I say let's get the uninstall stuff written and merge this in so we can start knocking out some of the dependent issue, and further refine it after we use it a bit.

pippinsplugins:
Merged into `release/2.6`. Closing this out. We will re-open if needed as we work with it in related issues.

SDavisMedia:
Instructions from @pippinsplugins to reopen this and report that when switching to branch `issue/3755` (support additional emails), I'm facing this output on the custom management table: http://glui.me/?i=9zgggk8hurvmmnz/2016-05-25_at_11.44_AM_(1).png/

pippinsplugins:
@cklosowski I think we need to set up an automatic upgrade system that triggers anytime the `EDD_VERSION` constant is changed, similar to how we did it in AffiliateWP: https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/admin/class-upgrades.php

That way we can make database changes without requiring users click a link to run it.

cklosowski:
So like we did with Recurring Payments as well.

pippinsplugins:
@cklosowski take a look at #4616

cklosowski:
@pippinsplugins manually reset my `edd_version` option, delete the customermeta table and then checked out this PR. Upon a refresh, table was created and the version was updated.

:+1:

cklosowski:
We will just need to make sure when we release the beta that we update the `EDD_VERSION` constant to be `2.6-beta1` so that the routines trigger.

PR #4535, opened 2016-05-04
Introduce customer meta table and EDD_Customer meta methods. #4311

pippinsplugins:
#4311

### DIFF
--- a/includes/class-edd-db-customer-meta.php
+++ b/includes/class-edd-db-customer-meta.php
@@ -48,12 +48,14 @@ class EDD_DB_Customer_Meta extends EDD_DB {
 	/**
 	 * Retrieve customer meta field for a customer.
 	 *
+	 * For internal use only. Use EDD_Customer->get_meta() for public usage.
+	 *
 	 * @param   int    $customer_id   Customer ID.
 	 * @param   string $meta_key      The meta key to retrieve.
 	 * @param   bool   $single        Whether to return a single value.
 	 * @return  mixed                 Will be an array if $single is false. Will be value of meta data field if $single is true.
 	 *
-	 * @access  public
+	 * @access  private
 	 * @since   2.6
 	 */
 	public function get_meta( $customer_id = 0, $meta_key = '', $single = false ) {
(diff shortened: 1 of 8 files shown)


includes/class-edd-db-customer-meta.php
```diff
@@ -0,0 +1,144 @@
...
+	 *
+	 * @param   int    $customer_id   Customer ID.
+	 * @param   string $meta_key      The meta key to retrieve.
+	 * @param   bool   $single        Whether to return a single value.
+	 * @return  mixed                 Will be an array if $single is false. Will be value of meta data field if $single is true.
+	 *
+	 * @access  public
+	 * @since   2.6
+	 */
+	public function get_meta( $customer_id = 0, $meta_key = '', $single = false ) {
```
cklosowski: Should we make this `protected` so that the only way to get customer meta data is to call `EDD_Customer::get_meta()`? If we make them both public we could end up with two ways to get the same data.

pippinsplugins: Yes, I think that's a good idea.

pippinsplugins: Wait no, we can't do that because then the `EDD_Customer` class can't call those since it's a separate class.

pippinsplugins: Let's add `@access private` so at least it's apparent they are for internal use only.

includes/class-edd-db-customer-meta.php
```diff
@@ -0,0 +1,152 @@
...
+	 * @access  public
+	 * @since   2.6
+	*/
+	public function create_table() {
+
+		require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
+
+		$sql = "CREATE TABLE {$this->table_name} (
+			meta_id bigint(20) NOT NULL AUTO_INCREMENT,
+			customer_id bigint(20) NOT NULL DEFAULT '0',
```
cklosowski: If we're not allowing `NULL`, we shouldn't allow `0` to be the `customer_id`. Setting `0` to the default means it's possible to create a `customer meta` entry tied to no customer. We could add a 'CHECK` into the creation that requires that `customer_id` be greater than `0`:
`CONSTRAINT check_customer_id CHECK (customer_id > 0)`

With that constraint in place, when data is inserted, it verifies the `customer_id` value provided is greater than `0` or else it fails the insert.

pippinsplugins: Good idea. Mind ading it in?

cklosowski: So after a little more reading, while an accepted item in the `create` it's not applied, just parsed. We will just have to handle this within the classes, which I'll push up here shortly.

Issue #4506, opened 2016-04-19
Flat rate discounts can apply > than cart total


chriscct7:
If you make a flat rate discount for $80 and apply it to a cart of $50, it correctly gives the order total remaining at checkout as $0, incorrectly lists the amount of discount applied at $80 instead of $50 and then shows the order total was -$30 on the purchase history shortcode

When a flat rate discount is applied, it should only apply up to the value of the cart at the time, re-evaluating whenever the taxes, cart total/quantity/items change, or when fees reapply

pippinsplugins:
@chriscct7 Could you test this PR?

mintplugins:
Tested https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4545 and it fixes the issue here.

pippinsplugins:
Great, merging.

PR #4545, opened 2016-05-06
Set discounted_price to 0 if discounted amount is less than 0. - #4506

pippinsplugins:
#4506


mintplugins:
Tested and works properly. Using a flat rate discount code of $1000 off when buying a $100 product leaves the payment at $0 (as opposed to previously when it set the payment to -$900).

Issue #4435, opened 2016-03-28
Creating a payment sometimes doesn't save the correct user_id or customer_id


chriscct7:
Related to #4434, there are cases where EDD isn't saving the correct user_id or customer_id (not sure which one yet, possibly both, probably user_id only). 

You can replicate this several different ways. For example:
- Disable guest checkout, set allow registration to true
- Buy a product while logged out, using the registration form to create a new user that the payment should be assigned to.

Or even easier:
- Install manual payments
- Create a new user manually in wp-admin
- Insert a payment manually for the user by using their email address

The result is the payment is not linked to the customer

This bug is impacting several of our extensions that rely on EDD to correctly identify the customer using the payment's user_id.

Briefly reading through the functions that process a payment, we switch between $payment_data['user_info']['user_id'] and $payment_data['user_info']['id'] to represent the customer's user_id all over the place (starts as the former, and by the time it goes into the payment class it should, and appears to be in the latter). My guess is somewhere along the lines, this is what causes the problem.

chriscct7:
Another report: https://secure.helpscout.net/conversation/185898165/30859/?folderId=367285

chriscct7:
Possibly related: https://secure.helpscout.net/conversation/185825189/30846/?folderId=180499

pippinsplugins:
The manual payments scenario can be ignored. I've already fixed that in Manual Payments 2.0 (in development still).

pippinsplugins:
@chriscct7 I just tested this on `master` and it worked properly.

![screen shot 2016-03-28 at 1 17 32 pm](https://cloud.githubusercontent.com/assets/1034109/14085785/8119883c-f4e7-11e5-8882-f3ca28b717f5.png)
![screen shot 2016-03-28 at 1 17 43 pm](https://cloud.githubusercontent.com/assets/1034109/14085784/81192982-f4e7-11e5-8aca-5575b453790c.png)
```php
<?php
$p = new EDD_Payment( 3771 );
var_dump( $p->user_id );
```
`string(2) "79"`

pippinsplugins:
@chriscct7 Can you try and replicate this again?

@cklosowski could you also try please?

cklosowski:
Testing now.

cklosowski:
Things connected just fine for me. No issues with the first scenario:
1) Disabled Guest Checkout
2) Made a purchase registering on the checkout
3) User ID is assigned to the payment

[...]

PR #4553, opened 2016-05-09
Pull user ID from customer record if _edd_payment_user_id is empty - Issue/4435

pippinsplugins:
#4435


mintplugins:
Tested this but users who signed-up upon checkout still have no meta saved through CFM. However, the correct user/customer ID is attached to both the payment and the customer.

cklosowski:
@pippinsplugins looks like this didn't fix the CFM stuff...are we expecting it to?

pippinsplugins:
@cklosowski CFM stuff was found to be something else. This is still relevant though, just in a different way.

Issue #4555, opened 2016-05-09
Standardize the display of file names with edd_get_file_name()


cklosowski:
We're using two different methods of getting the `file name` on the receipt shortcode.

On non-bundle products:
`<?php echo edd_get_file_name( $file ); ?>`

On bundle products:
`<?php echo esc_html( $file['name'] ); ?>`

Because of this, extensions like Git Updater cause bundles to not show download links, since there is no name to display. Git Updater has an open issue for this, however I think we need to fix it in EDD core before we investigate more in extensions. Using the helper function runs a fallback in place for when the supplied name is empty.

cklosowski:
Git Updater reference: https://github.com/wpninjas/edd-git-download-updater/issues/19

cklosowski:
As a note, if we do this in the template file we should be making a post on the Dev Blog about this change so that people who've customized the file can have some reference to what changed.

PR #4556, opened 2016-05-09
Standardize the method of getting file names

cklosowski:
This swaps usage of `$file['name']` for `edd_get_file_name( $file )`, which means we're getting file names in our templates the same way everywhere.

Addresses #4555


cklosowski:
This fixes the output for the purchase receipt, download history, as well as emails.

Issue #3572, opened 2015-07-06
Error in structured data


ankitpokhrel:
Looks like there is a problem with the structured data edd generates. Google says _The property price is not recognised by Google for an object of type SoftwareApplication._

![EDD Structured Data](http://ankitpokhrel.com.np/img/edd-structured-data.png)

The price should go inside itemprop offers
```
<div itemprop="offers" itemscope itemtype="http://schema.org/Offer">
    <span itemprop="price">$25.00</span>
    <meta itemprop="priceCurrency" content="USD" />
    <link itemprop="availability" href="http://schema.org/InStock" />In stock
</div>
```
Also a recommended field _operatingSystem_ is missing.

pippinsplugins:
Thanks for the headsup!

Related: #3355 and #3537

pippinsplugins:
@ankitpokhrel what version of EDD are you using? We already have it setup like that: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/master/includes/template-functions.php#L254

ankitpokhrel:
@pippinsplugins I am using v2.4.2. Still getting the same error in [structured data test tool](https://developers.google.com/structured-data/testing-tool/) for [this download](http://ankitpokhrel.com.np/blog/downloads/dynamic-featured-image-pro/) which has variable pricing enabled. `edd_add_schema_microdata()` seems to be false so the itemprop attribute is missing. Or is there anything in settings to enable schema microdata :smile:?

pippinsplugins:
@ankitpokhrel on a variable priced download or single price?

ankitpokhrel:
@pippinsplugins only on a variable priced download. No problem with the single priced download.

pippinsplugins:
The `name` item appears to be missing too: https://wordpress.org/support/topic/add-schema-name-to-download-items

pippinsplugins:
Another related PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4567

pippinsplugins:
Merged #4567 into `release/2.6`

Issue #3355, opened 2015-04-20
Schema markup missing "priceCurrency"


SDavisMedia:
It looks like we're adding `itemprop="price"` which is correct. But we're also putting the currency symbol in there. Instead, the `priceCurrency` property should be used in ISO 4217 format to describe the `price` property.

Validators don't seem to be enforcing this, though, which is weird. Not sure what to make of that considering `priceCurrency` is required.

SDavisMedia:
Looks like we may also need to look into the `AggregateOffer` property when variable pricing is used. Right now we still use the Price property but AggregateOffer appears to be designed for price variations and has its own set of required properties, like `lowPrice`.

cklosowski:
Starting to look into this one. We're going to have to think carefully about it, as it's going to have to cause us to rewrite quite a bit of the purchase button code.

SDavisMedia:
It also looks like we're not including `itemprop="price"` at all for non-variable pricing items. Hopefully I'm tripping. Can you confirm that, @cklosowski?

cklosowski:
@SDavisMedia yeah i think you are right. but there's a lot here to get it right. I'm moving to 2.4. With our past rush jobs on Schema, I want to make sure we get this right. I'm no expert on schema so if someone wants to jump on this I'd be happy to have the help

SDavisMedia:
I'm with you 100%. It's sort of new to me too. Reading docs on it while implementing it is the name of the game for me. Good call pushing it back. We need to get it right because what we learn here will come in handy for the EDD Reviews rewrite. I'll help get it in there too.

cklosowski:
@SDavisMedia thanks, pushed, and we'll regroup once we've had some time to do research on how to do this right.

pippinsplugins:
Punting to 2.5 due to lack of time to do adequate research.

pippinsplugins:
Correction: 2.4.3

cklosowski:
:+1:

pippinsplugins:
Needs to go to 2.5.

pippinsplugins:
Another related PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4567

pippinsplugins:
Merged #4567 into `release/2.6`

PR #4566, opened 2016-05-12
Fix price metadata by providing separate price and priceCurrency

leewillis77:
Currently EDD outputs the following schema.org metadata:

```html
<span class="edd_price_option_price" itemprop="price">$17.00</span>
```

This is incorrect according to https://schema.org/price, which states

> Use the priceCurrency property (with ISO 4217 codes e.g. "USD") instead of including ambiguous symbols such as '$' in the value.

The attached PR removes the itemprop="price" from the human-readable price, and adds two hidden meta values, e.g. 

```html
<span class="edd_price_option_price">$17.00</span>
<meta itemprop="price" content="17.00">
<meta itemprop="priceCurrency" content="USD">
```

This removes warnings when pages are run through Google's structured data testing tool (https://developers.google.com/structured-data/testing-tool/)

Before:
https://www.dropbox.com/s/rf2k8s5miiflq94/Screenshot%202016-05-12%2010.31.24.png?dl=0

After:
https://www.dropbox.com/s/hberjk7l8airt6r/Screenshot%202016-05-12%2010.34.18.png?dl=0


pippinsplugins:
This should fix both #3355 and #3572

pippinsplugins:
I've manually merged this into release/2.6. Thanks!

Issue #2963, opened 2015-01-15
nostates


bseddon:
Updated to 2.2.3 yesterday.  Now when I selected a country without any states I see 'nostates' shown.  While this is correct, it is inelegant.

![image](https://cloud.githubusercontent.com/assets/1221824/5756713/5fc5cede-9caa-11e4-9d57-696faee6250e.png)

Could the field just not be hidden?

But I have a more fundamental concern about the change made in this release. I tried to raise my concern (in a not very articulate way) by asking if there was a specification for this change in the discussion thread.

By prescribing countries and states you have removed the flexibility to cope with the real world.  Now we have to fit in a world that matches with your world view.  While the UK does not have 'states' we do have regions (England, Scotland, Wales and Northern Ireland) and counties.  Counties in particular are a normal part of a UK mailing address.  I live in the county of SURREY in the UK and my proper address includes the county in which I reside.  It is

16 Lynton Road
New Malden
SURREY
KT3 5EE
United Kingdom

This is not unusual and most countries in Europe have some kind of regional component, above a city, to their address.  The state field would have often been used to collect this additional address element in an unstructured way.  Now that is not possible.

I think there has been an attempt to model the changes after Woo Commerce.  While Woo Commerce does not show a county field for the UK it does at least hide the redundant field.

Is this a killer?  No, but presenting 'nostates' makes it look like it was conceived during amateur hour.  I would like to see the site owner given the option to present an unstructured, non-validated field in which a buyer is able to enter this additional address information.  It would be great if the name for the address element could be named.  'State' works in some countries, 'County' would work for the UK. 'Department' for France and so on.  At the very least it would be good to be able to show something other than 'nostates'

cklosowski:
Can you try with no other plugins active? Seems like you're having a conflict somewhere...because I'm getting the intended results:

![image](https://cloud.githubusercontent.com/assets/1390124/5757799/b84290b0-9c8c-11e4-9be0-2601e4abe6f2.png)

bseddon:
I apologize I'm wrong.  In fact I was seeing the problem in 2.2.2 which emits this when the country is changed:

includes\ajax-functions.php:            $response = 'nostates';

I need to do a better job of remembering which test site I'm using

cklosowski:
No worries! Thanks for double checking.

PR #4569, opened 2016-05-12
Trim states response

jamesckemp:
I was having an issue (as described here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/2963) where the states field was just showing `nostates` instead of the field.

Turns out the response from the ajax was: `"        nostates"` (without quotes, but with spaces infront). this meant the conditional was not met. Added $.trim() to fix the issue and the field shows as it should.

Not sure why there were spaces before the response, but this solution works well and should avoid it in future.

### DIFF
--- a/assets/js/edd-checkout-global.js
+++ b/assets/js/edd-checkout-global.js
@@ -98,7 +98,7 @@ window.EDD_Checkout = (function($) {
 					withCredentials: true
 				},
 				success: function (response) {
-					if( 'nostates' == response ) {
+					if( 'nostates' == $.trim(response) ) {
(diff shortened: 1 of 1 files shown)


jamesckemp:
On further review, looks like the `EDD Quaderno` plugin is adding this additional space to the response.

pippinsplugins:
Good catch! We'll get that included in 2.6.

cklosowski:
Might be something outputting some pre-output buffer spaces/line endings. Trimming makes perfect sense here. Thanks @jamesckemp

jamesckemp:
No problem, happy to see it merged.

PR #4573, opened 2016-05-13
Improved EDD_Customer::increase_value() and decrease_value()

daigo75:
* Changed `edd_customer_pre_increase_value` and `edd_customer_pre_decrease_value `actions to filters, to allow altering the value dynamically. This will make it easier to ensure that the increase/decrease is in the correct currency.
* Added the `EDD_Customer` instance to the filter. Needed to be able to alter the `EDD_Customer::$purchase_value` attached to the instance.

Personal suggestion
--
Unless there is a specific reason not to do so, it's better to pass the instance of an object rather than its ID. If one needs the ID, he can get it from the instance, but if one needs the actual instance (i.e. the "live" instance, not a local one), it's not possible to get it from the ID.


daigo75:
Fixed previous commit. I didn't consider that one can't just change actions to filter, as the latter require a return value, and existing "consumers" of the original action would not return one.

pippinsplugins:
This looks good to me.

@cklosowski any objections to including this in 2.6?

cklosowski:
This looks good, going to spin up an issue, assign it to 2.6 and merge this PR into that issue branch.

cklosowski:
Closing to use #4614 to push this into `release/2.6`, this isn't ready for `master` just yet.

Issue #4574, opened 2016-05-13
Functions related to "total earnings" can't take the currency into account


daigo75:
Today I noticed on a customer's site (multi-currency) that the `edd_earnings_total` were increasing and decreasing in an unexpected way. After an investigation, I realised that the `edd_earnings_total` value is manipulated by functions that are not currency-aware, and also not context-ware.

Examples
1. `edd_get_total_earnings()`
This function fetches the total of each payment, adds it to a total, disregarding the currency, and stores the result to `edd_earnings_total`. 100 USD + 100 GBP + 100 EUR become "300".
2. `EDD_Tools_Recount_Store_Earnings::get_data()`
Same as n.1.
3. `edd_increase_total_earnings()` and `edd_decrease_total_earnings()`
These functions are invoked by passing a raw numeric value, without currency or context information. Such information also cannot be retrieved, as these functions can be invoked in several different cases.
## Proposed solution

Since the total earnings are expected to be in base currency, a filter should be added to process any value that is going to be used to alter them, passing as much context information as possible.
For example, for a payment total, the call should pass the payment currency, so that any filter would be able to determine which conversion should be performed.

pippinsplugins:
I've closed the https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4575 PR as the proposed solution there won't work for performance reasons. We'll need an alternate proposal in order to do this.

cklosowski:
With 3.0 we have currency aware database queries and conversions with the 'rate' column.

PR #4575, opened 2016-05-13
Improvement - Total earnings filters

daigo75:
[Ref. issue 4574](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4574)

Added several filters to allow the manipulation of totals used to increase and decrease earning totals. 
## Notes
- The filter names may need to be changed, I just needed to put a name to illustrate the concept.
- Since I'm not entirely familiar with EDD stats, more filters may be needed elsewhere.


includes/payments/functions.php
```diff
@@ -840,7 +840,10 @@ function edd_get_total_earnings() {
 
 				if( ! empty( $payments ) ) {
 					$payments = implode( ',', $payments );
-					$total += $wpdb->get_var( "SELECT SUM(meta_value) FROM $wpdb->postmeta WHERE meta_key = '_edd_payment_total' AND post_id IN({$payments})" );
+
+					foreach( $payments as $payment_id ) {
```
pippinsplugins: We can't do a foreach here. It's simply too expensive of a query.

daigo75: True, but you can't do a `SUM()` either, as each of the totals has to be processed individually.

pippinsplugins: Perhaps the better solution would be to just return the amount through a filter that you can then use to return a custom amount?

daigo75: That would make things even worse: you would calculate the total, just to throw it away and have the whole loop done anyway.

pippinsplugins: This loop only ever runs once: when the total earnings option isn't present.

I think you could safely ignore this completely.

daigo75: I wish I could, but there's a chance that the totals calculated so far have always been wrong (at least, in EDD 2.5). If you have an order for 100 USD and one for 100 EUR, the logic calculates the total earnings as 100 + 100, which is incorrect. Having the possibility to recalculate them from scratch would allow to rectify this, to a certain extent.

daigo75: In any case, even a single filter for the total would be better than nothing. :smile:

cklosowski: Yes I agree, we cannot do this foreach. This is a performance breaker.

Issue #4519, opened 2016-04-27
Add SSL notice in PayPal settings for IPN verification


pippinsplugins:
To prepare for the SSL requirement for IPN verification coming from PayPal in September, 2016, we should add notice text to the IPN verification settings in the PayPal gateway settings:
![screen shot 2016-04-27 at 3 44 11 pm](https://cloud.githubusercontent.com/assets/1034109/14867607/ea4067c2-0c8e-11e6-88e9-18ca3bee5789.png)

cklosowski:
@pippinsplugins is this the matter of detecting if SSL is disabled, show a notice that IPN Verification requires SSL?

pippinsplugins:
Yes.

cklosowski:
@easydigitaldownloads/core-devs Thoughts on this notice when the `is_ssl` check fails (site isn't configured for SSL)
![screen shot 2016-05-16 at 10 34 51 am](https://cloud.githubusercontent.com/assets/1390124/15297962/d8fe5cfe-1b51-11e6-8b1f-8a721bef98d2.png)

SDavisMedia:
I don't see anything wrong with that.

mindctrl:
@cklosowski Maybe link to our docs? http://docs.easydigitaldownloads.com/article/1199-do-i-need-ssl or http://docs.easydigitaldownloads.com/article/994-how-to-set-up-ssl or something like that?

cklosowski:
@mindctrl Yeah I'll do that too

pippinsplugins:
I don't think the message is very clear.

It needs to convey several pieces of information:

1. SSL is only required for IPN verification.
2. IPN Verification !== IPN processing for marking payments as complete.

Can we update both the non-SSL notice and the description of the setting to better explain exactly what it does?

spencerfinnell:
Agreed with @pippinsplugins -- the notice should also only display if there is no SSL and the setting is unchecked. 

Is it feasible to invert the setting as well? I know I always harp on this but for some reason reversed check boxes kill me. I think it would make creating an easy to understand description easier as well.

spencerfinnell:
Doesn't look like it without an upgrade routine after looking at how checkboxes are stored. Nevermind!

pippinsplugins:
How's this? Note: I have not inverted the settings.
[...]

PR #4579, opened 2016-05-16
Move PayPal settings to own registration and add non-SSL notice for I…

cklosowski:
…PN #4519


includes/gateways/paypal-standard.php
```diff
@@ -22,6 +22,59 @@
  */
 add_action( 'edd_paypal_cc_form', '__return_false' );
 
+
```
pippinsplugins: Please go ahead and add PHPdoc

cklosowski: Doh, forgot one, on it.

Issue #3157, opened 2015-02-17
Allow Customers to use dates in API


cklosowski:
As requested in https://easydigitaldownloads.com/support/topic/need-api-features-adding-to-edd/

Would be nice to have the `customers` data in the API to use our date ranges to search by the `date_created`.

pippinsplugins:
:+1:

chriscct7:
Dependent on #3337

chriscct7:
(for backwards compatibility reasons)

pippinsplugins:
Punting this to 2.5

cklosowski:
:+1:

cklosowski:
@pippinsplugins since we don't have the week ranges currently in the API, is this good to go? We can look at mirroring the Reporting date ranges in #2462 possibly?

pippinsplugins:
Let's go ahead and mirror them.

cklosowski:
@pippinsplugins adding `this_week` and `last_week` is a far larger project than adding dates to the customers endpoint. I propose we merge customers endpoint with the existing date ranges, and then when we come in to revamp the dates part of the API we rely on the stats class dates parsing as it's far more flexible. No sense in mirroring them in this branch just to have them redone in another.

pippinsplugins:
That's fine by me.

pippinsplugins:
Just added a fix that allows customers to be shown even if no date is specified. IT was returning empty results if no date was specified.

PR #4588, opened 2016-05-18
Add date queries to the Customers endpoint

cklosowski:
#3157 #4602


includes/api/class-edd-api-v2.php
```diff
@@ -148,4 +148,152 @@ public function get_products( $args = array() ) {
...
+			}
+
+			if ( ! empty( $args['enddate'] ) ) {
+				$date_range['end'] = $dates['year_end'] . sprintf('%02d', $dates['m_end'] )   . $dates['day_end'];
+			}
+
+			$args['date'] = $date_range;
+
+		} else {
+			if( $args['date'] == 'this_quarter' || $args['date'] == 'last_quarter'  ) {
```
pippinsplugins: Can we add the other pre-defined date ranges too please? Today, yesterday, this week, last week.

cklosowski: I need to add `*_week`, but `yesterday/today` are supported already. They fallback to the last case here on line 231

cklosowski: @pippinsplugins I don't see `this_week` or `last_week` currently supported in the `get_dates` method of the existing API.

pippinsplugins: Huh, they're in the reports UI but not in the REST API. Oops :P

PR #4595, opened 2016-05-18
Discontinue use of global in settings sanitization Props @grappler

cklosowski:
Thanks @grappler for your help getting this ready for #4104. This should do the trick, just going to let some CI pass and do one more pass over the testing.


includes/admin/settings/register-settings.php
```diff
@@ -1249,19 +1249,19 @@ function edd_header_callback( $args ) {
...
+	$edd_option = edd_get_option( $args['id'] );
 
 	if ( isset( $args['faux'] ) && true === $args['faux'] ) {
 		$name = '';
 	} else {
 		$name = 'name="edd_settings[' . edd_sanitize_key( $args['id'] ) . ']"';
 	}
 
-	$checked = isset( $edd_options[ $args['id'] ] ) ? checked( 1, $edd_options[ $args['id'] ], false ) : '';
+	$checked = isset( $edd_option ) ? checked( 1, $edd_option, false ) : '';
```
pippinsplugins: This isn't right. `isset()` will always be true here. While it may work fine, let's update this to properly check with `empty()`.

cklosowski: Using `empty` here completely breaks it actually. A checkbox will fail `isset` if unchecked. If it's not checked, it's not part of the `$_POST`

pippinsplugins: But `isset()` will never return false here. `$edd_option` will _always_ be set.

cklosowski: @pippinsplugins unless of course you meant an `! empty` check. which is possible.

pippinsplugins: I just meant we should use `empty()` instead of `isset()`. I wasn't proposing changing `isset( $edd_option )` to `empty( $edd_option )` verbatim.

cklosowski:
@pippinsplugins updated to use the empty check instead, working fine now.

Issue #3338, opened 2015-04-08
Better Coupon Endpoint


chriscct7:
Once https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3337 is done, we should look at improving the coupon endpoint, for v2.

Right now, the sales API tells you the coupons used on a sale, but not how much they impacted the total of the sale. This is making things like writing a report that shows you which discount codes are used on higher value purchases, or most often, or lead to better returns, etc pretty difficult (since you then have to call the api to get the rules for a discount code and then the extending program has to then manually write its own rules for calculating the discount codes for figuring out how much each code took off).

Like if I know code xxxx is used and code yyyy is used on a sale, I have to right now call the coupons endpoint for each to retrieve the coupon rules, and then I have to calculate how much each would affect the sale I'm looking at manually (using my own custom implemented logic).

That gets really expensive and unnecessarily complicated when there's alot of coupons. It'd be nice, since EDD knows this information, if it would tell you how much each coupon took off.

In v1, coupons are shown like this on the sales endpoint:
```
'coupons' => array(
     'coupon1',
     'coupon2'
);
```

In v2, we'd do:
```
'coupons' => array(
     'coupon1' => 8.3,
     'coupon2' => 1.2
);
```

pippinsplugins:
Makes sense to me.

chriscct7:
Dependent on #3337

pippinsplugins:
API versioning is going into 2.4 but I'm going to propose that we wait to introduce this until 2.5 so that we can make other API changes at the same time. Better to have one new version instead of two or three in a short time.

Any objections?

pippinsplugins:
We'll need to store the discount amounts for each discount code in payment meta data to make this possible.

pippinsplugins:
Or we could just show the amount set on the discount code, opposed to the actual value of the discount on the cart.

cklosowski:
Couple issues with the latter of those @pippinsplugins is that the discount could have been removed from the system or modified, so looking up it's discounted amount (from the discount post type) might not be possible or accurate.

I'm not sure that's a case to completely go with future payments having this data only though. I vote we use it, and just account for discount posts types that have been removed.

pippinsplugins:
The approach taken in the PR won't work. See https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4620#discussion_r64749463.

pippinsplugins:
[...]

Issue #3113, opened 2015-02-14
We should add unit tests for the API


chriscct7:
Normally I'd advocate against opening tickets regarding code coverage, but the EDD API class which people are starting to use and rely on has only 24% coverage, so low that we're starting to get reports of issues as people and extensions start using it more frequently: https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/2979. For being the main API, we should try to get it to at least 80% for 2.4.

pippinsplugins:
Agreed.

ghost1227:
Can do!

chriscct7:
Assigning to Dan per above. Might take back towards the end if it needs help

cklosowski:
While `release/2.6` is going through beta I'm going to try and get some more unit tests for the API in place, specifically with the differences between `v1` and `v2` (introduced in 2.6)

cklosowski:
First pass at getting our API able to be unit tested. Now that I have it working locally, will see if Travis likes it, and keep working from here to see if we can get some more coverage on it prior to 2.6 release.

cklosowski:
Not quite sure why, this passes locally 100% but on Travis any tests that require the API Authentication fail. I'd say it's related to `$wp_query` not getting correctly passed, however, other parts of the `$wp_query` variable end in the right place. I'll come back to this a bit later.

pippinsplugins:
Could you try using the go_to() helper instead of manually setting query
vars?

cklosowski:
@pippinsplugins I'll give that a shot next.

cklosowski:
@pippinsplugins `go_to()` doesn't work because it's outside the context of the unit test's instantiation of the `API`. It basically just returns data like the API was never called.

pippinsplugins:
Boo

cklosowski:
Punting this from 2.6...I need to build a local docker container of the Travis Environment so I can mess with this to figure out why it'll pass locally, and not on Travis.

PR #4620, opened 2016-05-26
Change  to  and add amount discounted on payment for get_recent_sales…

cklosowski:
… #3338

### DIFF
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -358,27 +358,6 @@ class EDD_API_V2 extends EDD_API_V1 {
 				$sales['sales'][ $i ]['ID']             = $payment->number;
 				$sales['sales'][ $i ]['transaction_id'] = $payment->transaction_id;
 				$sales['sales'][ $i ]['key']            = $payment->key;
-
-				$discounts       = ! empty( $payment->discounts ) ? explode( ',', $payment->discounts ) : array();
-				$discount_values = array();
-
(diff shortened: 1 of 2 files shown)


includes/api/class-edd-api-v2.php
```diff
@@ -317,4 +317,111 @@ public function get_customers( $args = array() ) {
...
+
+					// Clean up the discount code for use in the JSON
+					$code  = trim( $discount );
+					$total = $payment->subtotal;
+
+					if ( edd_use_taxes() && edd_prices_include_tax() ) {
+						$total += $payment->tax;
+					}
+
+					$discount_values[ $code ] = (float) $total - ( edd_get_discounted_amount( $code, $total ) );
```
pippinsplugins: This won't always be correct since some discounts are limited to specific items.

Example, if a discount is limited to item A but an order has item A and B, using the subtotal of the order here will give incorrect amounts.

cklosowski: Good catch. So we'll have to iterate over each cart item, determine the discount each discount applies to an item, and then total it that way?

cklosowski:
@pippinsplugins Check these changes over?

pippinsplugins:
That looks like it should work. Can we get some unit tests added to help verify?

cklosowski:
@pippinsplugins right now our API isn't unit testable.  Which is why #3113 is still open. We need to spend some deep time getting a way to do this

Issue #4632, opened 2016-05-30
Taxes should never be included on negative cart fees


pippinsplugins:
Negative tax fees (aka discounts) are currently included in tax calculations. This can result in incorrect tax calculations and, in some cases, even negative taxes.

To help illustrate the issue, look at this:
![screen shot 2016-05-30 at 10 25 16 am](https://cloud.githubusercontent.com/assets/1034109/15653185/f1efe840-2650-11e6-983b-6ce3111195d8.png)

- 20% tax rate
- `Pro Package` has taxes disabled on the product
- Tax is negative since it is being calculated at 20% of of `-135.00`.

Two things we should do:

1. Never allow taxes to go negative
2. Do not include negative fees in tax calculations

cklosowski:
Seems to work great for me. Tested with a single product that met the requirements. Then tested with an item that should have a tax calculation.

Merging.

cklosowski:
Leaving `Needs Testing` on for now.

mintplugins:
![should-these-say-0](https://cloud.githubusercontent.com/assets/1638128/15718059/9983cb42-27f7-11e6-9540-456a43e7ddcf.jpg)

mintplugins:
When doing another test, I'm getting the wrong taxation calculation. If the Discount is 50% and the cart quantity is used, the tax amount is calculated before the discount is applied. See screenshot:
![tax-wrong](https://cloud.githubusercontent.com/assets/1638128/15722249/c71c9918-280a-11e6-99b3-b2c13383eb3e.jpg)

pippinsplugins:
Subtotal is pre discounts, so it is the item total only.

Taxes are supposed to always be after discounts, so that looks like a bug we need to address.

mintplugins:
I know this probably goes against how the EDD fee system works - but is it possible to have fees/discounts not show up as cart items? I feel like they make more sense after the subtotal if the subtotal is pre-discounts/fees.

pippinsplugins:
Perhaps in the future but not right now.

pippinsplugins:
Ticket from a customer experiencing this issue with Recurring Payments: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4632

mintplugins:
The above issue with incorrect tax amounts has its own issue here now: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4658

pippinsplugins:
[...]

PR #4633, opened 2016-05-30
Do not allow taxes to go negative or apply to negative fees - Issue/4632

pippinsplugins:
#4632


glondon:
I'm OK with negative taxes.... Just kidding...  Good catch on this one..

PR #4643, opened 2016-06-01
Only return first IP address fetched from $_SERVER

### LINKED ISSUE #4642: Possible for edd_get_ip() to return a CSV of IP addresses
`edd_get_ip()` can use the `$_SERVER['HTTP_X_FORWARDED_FOR']` value as the IP address. This can cause problems because `$_SERVER['HTTP_X_FORWARDED_FOR']` [allows returning a comma-separated value](https://en.wikipedia.org/wiki/X-Forwarded-For#Format), not just a single IP.

This causes problems with payment gateways that verify the IP addresses, like PayPal, which threw this error message: "Invalid IP: The IP Address provided is invalid."

I'll submit a patch in a few minutes that fixes the problem.

zackkatz:
Fixes #4642

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -173,7 +173,11 @@ function edd_get_ip() {
 	} elseif( ! empty( $_SERVER['REMOTE_ADDR'] ) ) {
 		$ip = $_SERVER['REMOTE_ADDR'];
 	}
-	return apply_filters( 'edd_get_ip', $ip );
+	
+	// Fix potential CSV returned from $_SERVER variables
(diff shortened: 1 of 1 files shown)


zackkatz:
Note: I have no idea what changed on my server. I use CloudFlare, so that may be it, but it seems like it just happened out of the blue. So, I don't know how to reproduce.

cklosowski:
Going to close this PR, and get a new one setup for a few changes. your commit history will stay in tact @zackkatz. Going to make a few small modifications and add a unit test around it.

PR #4647, opened 2016-06-02
Fix role creation check

### LINKED ISSUE #4646: Roles are being initialized on every admin request
The function [edd_install_roles_on_network() in includes/install.php](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/install.php#L309-L335) creates user roles and permissions if missing. This line
```
if( ! in_array( 'shop_manager', $wp_roles->roles ) ) {
```
always evaluates to false and thus always triggers role creation on every admin page load because `$wp_roles->roles` is now a multi-dimensional array and so we need to update this to check for array keys. 

Steps to re-create:

1. Install WP 4.5.2
2. Install Easy Digital Downloads and activate
3. Set a debug breakpoint inside `edd_install_roles_on_network`
4. Load Dashboard pages and see that the function keeps re-creating roles. 

I will open a pull request momentarily with a fix for this issue... this problem has been preventing me from being able to customize the capabilities for the EDD roles.

bcole808:
resolves #4646

### DIFF
--- a/includes/install.php
+++ b/includes/install.php
@@ -322,7 +322,7 @@ function edd_install_roles_on_network() {
 		return;
 	}
 
-	if( ! in_array( 'shop_manager', $wp_roles->roles ) ) {
+	if( ! array_key_exists( 'shop_manager', $wp_roles->roles ) ) {
 
 		// Create EDD shop roles
 		$roles = new EDD_Roles;


pippinsplugins:
I've manually merged this into `release/2.6`. Thanks!

Issue #4658, opened 2016-06-07
Taxes should be affected by negative fees.


mintplugins:
If you have a negative fee on the cart, it will still charge tax as if no negative fee was added. 

![100 -off-discounttax-issue](https://cloud.githubusercontent.com/assets/1638128/15841850/3bea7966-2c22-11e6-9d52-6a148dea54d1.jpg)

mintplugins:
Related: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4632

mintplugins:
This affects any function that uses negative fees including:

- Discounts Pro
- Points and Rewards

pippinsplugins:
I'm pretty sure this is resolved in the same way as #4632

mintplugins:
That's definitely possible. Just to clarify, I am running this test on the 2.6 branch and it is still occurring. However, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4632 is resolved.

pippinsplugins:
Yeah, will still likely require a small change.

pippinsplugins:
@mintplugins mind giving that PR a try? It's definitely not working 100% now but I _think_ it's better.

mintplugins:
Testing...

mintplugins:
That definitely fixed the issue for Point and Rewards. However, Discount Pro still seems to be affected

<img width="887" alt="screen shot 2016-06-07 at 9 39 43 pm" src="https://cloud.githubusercontent.com/assets/1638128/15880131/756eab90-2cf8-11e6-8817-593e780fbc1e.png">

chriscct7:
That screenshot looks fine to me, what's wrong with it

mintplugins:
@chriscct7 it's using the "Points and Rewards" extension - not Discounts pro. Here's one with a Discounts Pro discount:

<img width="863" alt="screen shot 2016-06-07 at 11 21 29 pm" src="https://cloud.githubusercontent.com/assets/1638128/15881778/a5f09d92-2d06-11e6-9d89-8691b5d5c866.png">

mintplugins:
Okay so doing further testing, it's not a Points and Rewards VS Discounts Pro issue. The issue is whether the discount cover 100% of the original price. If the discount iis for 100% it is fine. If it is not for 100% the issue remains.

chriscct7:
What's your tax rate set to?

pippinsplugins:
[...]

PR #4659, opened 2016-06-07
Prevent fees from affecting taxes if subtotal is 0 or less. #4658

pippinsplugins:
#4658


includes/cart/functions.php
```diff
@@ -782,7 +791,19 @@ function edd_get_cart_fees( $type = 'all', $download_id = 0, $price_id = NULL )
  * @return float Total Cart Fees
  */
 function edd_get_cart_fee_total() {
-	return EDD()->fees->total();
+	$fees = EDD()->fees->get_fees( 'all' );
+
+	$fee_total = 0.00;
+	foreach ( $fees as $fee ) {
+		if ( ! empty( $fee['download_id'] ) && ( $fee['amount'] <= 0 || $fee['no_tax'] === true ) ) {
```
pippinsplugins: What's the reasoning behind excluding `no_tax` fees?

cklosowski: Might have been an oversight, I'll test it out again, there is a scenario I was accounting for on fees with a download ID and that shouldn't affect taxes. I'll check it again.

cklosowski:
@pippinsplugins removed that check, it was left over from some testing I was doing. @mintplugins @pippinsplugins ready for a bunch of testing!

Issue #4669, opened 2016-06-09
New filter hook for 'edd_price_options' classes


mintplugins:
I am proposing a new filter hook which would allow the adding/removing of css classes for the variable price div (edd_price_options).

I think if we set it up as an array it makes it easy to add/remove classes and then implode them upon output. 

I am thinking we call the hook: "edd_price_options_classes" and pass the classes array and the download_id to the filter hook.

Here is a screenshot of the proposed code change:
<img width="862" alt="screen shot 2016-06-09 at 10 08 13 am" src="https://cloud.githubusercontent.com/assets/1638128/15932882/705c8378-2e2b-11e6-939a-6985b4b3b55d.png">

Any thoughts or changes?

shazahm1:
$css_classes should be sanitized before being output using [sanitize_html_class()](https://codex.wordpress.org/Function_Reference/sanitize_html_class).

mintplugins:
@shazahm1 great catch - thanks!

mintplugins:
PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4670

PR #4670, opened 2016-06-09
Add "edd_price_options_classes" Filter

mintplugins:
This change makes an array for the css classes which will be used for the edd_price_options div. It is filterable using "edd_price_options_classes" and sanitized using sanitize_html_class

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -251,26 +251,21 @@ function edd_purchase_variable_pricing( $download_id = 0, $args = array() ) {
 	$mode   = edd_single_price_option_mode( $download_id ) ? 'multi' : 'single';
 	$schema = edd_add_schema_microdata() ? ' itemprop="offers" itemscope itemtype="http://schema.org/Offer"' : '';
 	
-	//Set the default classes string
-	$css_classes_string = NULL;
-	
-	//Filter the class names for the edd_price_options div
+	// Filter the class names for the edd_price_options div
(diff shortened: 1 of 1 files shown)


mintplugins:
#4669

pippinsplugins:
@mintplugins a couple of nitpicky things:

1. There should always be a space after `//` for comments
2. The default value for the classes should be an empty string instead of null since the datatype we're setting in the filter is a string.
3. Let's please add `esc_attr()` to the class output (around `rtrim()`).

pippinsplugins:
The `foreach( $css_classes_array as $css_class ) {` could also be simplified to this:
```php
$css_classes_string = array_map( 'sanitize_html_class ', implode( $css_classes_array, ' ' ) );
```

mintplugins:
@pippinsplugins Good catches, thanks. I just had to put the implode around the array_map since it must be passed an array and implode creates a string.

`
$css_classes_string = implode( array_map( 'sanitize_html_class', $css_classes_array ), ' ' );
`

shazahm1:
Since `sanitize_html_class()` is being used, there is no need to run `esc_attr()` on the imploded string. Basing this on how core does pretty much the same with `get_body_class()` and `body_class()`.

mintplugins:
@shazahm1 That definitely true. I think we are going to leave it in now since escaping should be done as close to the actual output as possible and future changes to this could possible change how it functions. Having that last esc_attr doesn't necessarily hurt anything so I would argue it's better to have it in there.

Issue #4673, opened 2016-06-11
Allow new public API query modes to be added


ericnicolaas:
Currently, there is no easy way to add a new _public_ API endpoint (i.e. one that doesn't require API keys). I think there are valid reasons for wanting to add new public endpoints (one example below), and it makes sense to add that possibility when we can already declare new endpoints that are not public.

I have a working version of this and will put together a PR. 

**Example where this would be useful:** I want to add a `versions` endpoint that prints out a simple array of all products + their current version. This information is already available in the `products` mode, but by adding a mode that _only_ returns versions, the request speed can be much faster. The `versions` endpoint will then be used to check for updates to our plugins, doing away with the current technique we used (based on the Software Licensing sample plugin). The end result should be less load on our server and faster experiences in wp-admin for users who have multiple plugins installed.

pippinsplugins:
Oh, I love this!

I think we should go ahead and include this in 2.6.

ericnicolaas:
Updated PR with 2.6 for the `@since` tag.

pippinsplugins:
Thanks!

cklosowski:
@easydigitaldownloads/core-devs This is pretty awesome, could you guys hit this up and just test the current API endpoints to make sure that things are still working as expected?

mintplugins:
Tested all built-in endpoints and they all work properly:

Private (needs API key):
- stats
- sales
- customers
- discounts
- file-download-logs

Public:
- products

When adding the code (https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4674) for the versions endpoint it works publicly as well:

PR #4674, opened 2016-06-11
Added is_public_query() method to EDD_API class

ericnicolaas:
This allows new public query modes to be defined. The filter acts on an array to make implementation very concise. The following both registers a mode and defines it as an endpoint: 
```
function eddvapi_add_versions_query_mode( $modes ) {
    $modes[] = 'versions';
    return $modes;
}

add_filter( 'edd_api_valid_query_modes', 'eddvapi_add_versions_query_mode' );
add_filter( 'edd_api_public_query_modes', 'eddvapi_add_versions_query_mode' ); 
```
The `@since` doc comment hasn't been defined yet, but happy to update that to whatever it should be.

#4673 (cklosowski)

### DIFF
--- a/includes/api/class-edd-api.php
+++ b/includes/api/class-edd-api.php
@@ -311,7 +311,7 @@ class EDD_API {
 		$this->override = false;
 
 		// Make sure we have both user and api key
-		if ( ! empty( $wp_query->query_vars['edd-api'] ) && ( $wp_query->query_vars['edd-api'] != 'products' || ! empty( $wp_query->query_vars['token'] ) ) ) {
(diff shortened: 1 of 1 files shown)


cklosowski:
@ericnicolaas I'm going to close this PR but move your changes into an `issue/4673` branch that will be a PR against `release/2.6` will over easier testing from the team and what not, plus that way we can get it into the Release Candidates when we get to that point :)

cklosowski:
#4681 is now open for `release/2.6` so we can test this out as a team and get it into our documentation

Issue #4641, opened 2016-06-01
Improve inline help throughout the plugin


mindctrl:
One of the best ways we can improve the customer experience and lower the volume of support requests is to make it easier to understand what settings mean and how things work. After chatting with @pippinsplugins we've decided to add some inline tooltips throughout the plugin to make it easier to get quick information about various things in the plugin. We want to make this a priority.

Something like this, which we're starting to implement in RCP. 

![image](https://cloud.githubusercontent.com/assets/1254870/15716645/251206da-27f1-11e6-9554-4c6d2d29ca53.png)

It's pretty quick and easy to add and doesn't require a lot of code. We're using the jQuery TipTip library. Example code required to add the tooltip in the screenshot:

```<span alt="f223" class="rcp-help-tip dashicons dashicons-editor-help" data-tip="<?php _e( '<strong>Subscription level</strong>: a subscription level refers to a membership option. For example, you might have a Gold, Silver, and Bronze membership level. <br/><strong>Access Level</strong>: refers to a tiered system where a member\'s ability to view content is determined by the access level assigned to their account. A member with an access level of 5 can view content assigned to access levels of 5 and lower.', 'rcp' ); ?>"></span> ```

@samikeijonen we'd love your thoughts on accessibility.

cklosowski:
I rather like this idea, however I am concerned about the accessibility.

pippinsplugins:
The method above is identical to how WooCommerce does it. Not saying they've nailed it just because they're Woo, but I think it's likely that it's fine.

samikeijonen:
I'm sure tooltip can be accessible, like in this [example](http://a11y.nicolas-hoffmann.net/dialog-tooltip/). I can ask around more info about it.

cklosowski:
Ok @pippinsplugins and @samikeijonen, @mindctrl.

We should outline a few areas we see the most support requests about clarification. Since this is the initial addition of them, we could do a super sharp focus on the top 10-20 pain points and add them in those first. And then follow up once we get some feedback.

cklosowski:
# Areas to add tooltips for
- [ ] Price assignments for files (Edit/Add Download)
- [ ] Bownload link expiration time (Settings)
- [ ] Button behavior (Settings & Edit/Add Download)
- [ ] File Download Method (Settings)
- [ ] IPN verification (Settings)
- [ ] File Download Limit (Edit/Add Download)
[...]

PR #4694, opened 2016-06-20
Add the edd prefix to the ui-tooltip class ...

michaelbeil:
...  selector.

#4641

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -51,7 +51,7 @@ a.edd-delete:hover {
 	cursor: help;
 }
 
-.ui-tooltip {
+.edd-ui-tooltip {
 	background: #333 !important;
 	border-radius: 3px !important;
 	box-shadow: 1px 1px 2px 1px rgba(214,214,214,0.5);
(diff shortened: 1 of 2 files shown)


pippinsplugins:
This will also need to adjust the classes on the tool tip spans, otherwise it breaks the CSS>

michaelbeil:
I do not see any tooltip spans on release/2.6 within our screens. Am I missing something?

pippinsplugins:
I believe that class gets added automatically. All we do in the HTML to set them up is this:
```
<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help">content here</span>
```

michaelbeil:
I only see the edd-help-tip class being added here: 
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/2.6/includes/admin/payments/view-order-details.php#L81 https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/2.6/includes/admin/payments/view-order-details.php#L180  
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/2.6/includes/admin/payments/view-order-details.php#L237

cklosowski:
@m @pippinsplugins test that out, there is a `jquery ui tooltip` option to define a class for the hover state.

michaelbeil:
This works well.

Issue #4013, opened 2015-11-12
Use <select> instead if <input type="text"> for "state" on profile editor when possible


brashrebel:
Currently there is an inconsistency when it comes to customers declaring their state. On the checkout form, after country has been specified, the state field becomes a select populated with available states if they are available, otherwise it becomes a text input. When using [edd_profile_editor] however, the state field is always a text input.

It would be nice to make this a select field when possible. Displaying a text field instead allows for inconsistencies and typing errors. It also outputs the state code instead of the state name which, according to a customer I've been helping, is not always ideal. Some countries (Japan in this case) do technically have state codes but they are apparently not used much and displaying the state name would be much preferred.

I have a PR coming for consideration.

pippinsplugins:
@brashrebel See https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3982

brashrebel:
@pippinsplugins yeah, I read through that. It seemed like a different issue though. This is just an enhancement to make the profile editor make state a select field _if possible_. I've got it working locally, just need to make a PR. Unless I'm missing something and shouldn't bother.

pippinsplugins:
Oh, sorry, I missed the `profile editor` part :)

:+1:

brashrebel:
Just made a PR.

SDavisMedia:
This isn't ready for testing yet. https://github.com/easydigitaldownloads/Easy-Digital-Downloads/pull/4014

cklosowski:
Pushing to 2.6 to match #4014 since they are related.

cklosowski:
Thought of a little issue here when AJAX is disabled. I'll work up a fix on that.

SDavisMedia:
Only issue I'm having is that with AJAX disabled, [edd_profile_editor] doesn't adjust the state/province field when you switch between counties. Is that the issue you found, @cklosowski.

Everything else seems to be working great.

cklosowski:
@SDavisMedia yes that's the problem. @pippinsplugins and i were just talking about this, and we think we're just going to leave like that. Thoughts? I think we'd like to eventually get rid of the 'Disable AJAX' option, in the future.

SDavisMedia:
I don't think there's anything wrong with that for now. They can save the form with the new country selected and then edit the state/province field once it reloads.

cklosowski:
Yep that's my thought. I'm going to merge this in then.

PR #4698, opened 2016-06-21
Make state field select instead of text if possible

cklosowski:
Replaces #4014 for #4013

### DIFF
--- a/assets/js/edd-ajax.js
+++ b/assets/js/edd-ajax.js
@@ -369,6 +369,69 @@ jQuery(document).ready(function ($) {
 
 	});
 
+	$('body').on('change', '#edd_cc_address input.card_state, #edd_cc_address select, #edd_address_country', update_state_field);
+
+	function update_state_field() {
+
+		var $this = $(this);
+		var $form;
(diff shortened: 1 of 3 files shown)


cklosowski:
@SDavisMedia @brashrebel mind testing out this PR for me?

cklosowski:
You can use branch `brashrebel-issue/4013`

brashrebel:
Just tested and it works for me. Although it seems that this was already working on the checkout screen in `release/2.6`

cklosowski:
Yes, it was working on checkout, but I changed where the JS lives, so it's
best to test them both to make sure I didn't break the existing
functionality.

Chris Klosowski
cklosowski@gmail.com
@cklosowski

cklosowski:
Actually I think there's a problem here when we have AJAX disabled. Previously all the JS to do this was in the `edd-checkout-global.js` which is loaded even if the AJAX is disabled. With the functions to do this existing in the `edd-ajax.js` file, if AJAX is disabled, then these will never load for the checkout.

We'll need to fix that.

PR #4706, opened 2016-06-22
Improvement - Filters for download earnings amounts

daigo75:
The filter have the same purpose of the action added for customers, i.e. to allow processing the earnings before they are updated on the product. This needed in cases like multi-currency shops, where earnings cannot just be added or subtracted, but must be first converted to the base currency.


pippinsplugins:
@daigo75 can you please based your branch off of `release/2.6` and then resubmit so that your PR doesn't include hundreds of other commits?

pippinsplugins:
And please submit to `release/2.6`.

daigo75:
Oh, I thought I had patched the release 2.6, but now I noticed I patched master. Fixing it immediately.

PR #4713, opened 2016-06-22
Adding changelog and updating composer and package JSON

cklosowski:


michaelbeil:
@cklosowski: 

- New: Public API query mode to allow extension developers **the ability** to create API endpoints that do not require API keys.
-  New: **Purchased** Products column in payment export improved for better parsing.   
-  New: (API v2) Customers endpoint support**s** filtering customers by creation date.  
-  New: Add price(s) to oEmbed output for downloads**.** 
-  New: Add **filters** to the add_fee() and get_fees() methods of EDD_Fees class. 
-  New: PayPal **Standard** payments are now able to be refunded within the WordPress admin. 
-  New: Add default redirect option to [edd_login] **and** [edd_register] shortcodes. 
-  New: Add JavaScript trigger **to** edd_load_gateway() for 3rd party hooks.    
-  New: Support additional email address**es** for customers.    
-  Tweak: Update translation .mo files and add development tool for downloading translations from **Language** Packs. 
-  Tweak: Use existing **function** to build {receipt_link} email tag.        
-  Fix: Roles are being **initialized** on every admin request. 
-  Fix: System info **shows** Must-Use plugins when *none are* present. 
-  Fix: Flat rate **discounts** can apply an amount larger than the cart total.
-  Fix: Improve performance of transient deletion when uninstalling Easy **Digital** Downloads.     
-  Fix: (a11y) **Repeatable** field link should be a button.

Issue #4718, opened 2016-06-23
Turn edd_get_payment into a caching layer for EDD_Payment, and allow false on no-payment.


zackkatz:
When passing a `$payment_id` that isn't a valid payment, `EDD_Payment::setup_payment()` returns false before setting up the payment.

`edd_update_payment_status()` assumes that instantiating `EDD_Payment` returns a valid `EDD_Payment` class, and continues with that assumption, calling `EDD_Payment::save()` no matter the return value.

Related to https://trello.com/c/j7LgHA80/436-gravity-forms-checkout-for-easy-digital-downloads#comment-575edeabeb0ad751c227c4e9

pippinsplugins:
Let's definitely do this, but I'm going to milestone it out a bit because there's a very good likelyhood of this causing issues with any one that is doing `if( $payment->ID > 0 )` type checks to determine if a payment was found.

SDavisMedia:
Replicated still being able to create a new payment while passing a payment ID that doesn't exist. @zackkatz's PR fixes that. Also confirmed by running `edd_get_payment_gateway( 999 )` and getting no results (`999` is the non-existent payment ID I tried to create).

pippinsplugins:
New PR from issue branch set up https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4921

cklosowski:
So after some discussion, the expectation here is that if we do something like the following:
```
// Where 123 is not a payment
$payment = new EDD_Payment( 123 );
```
That `$payment` would be `false`. B/C of PHP that's now how object instantiation works, and `__construct` does not return `false`, it returns an empty object. What our goal should be is that `$payment` be equal to `false`...will have to research this a bit.

cklosowski:
Ok, researched this more, and you cannot make a class' construct return `false`. In order to do this we'd need a helper function to instantiate `EDD_Payment` that would return `false` after checking the inner data of the instantiated version.

This is a general principal of the PHP Class system. So we either need to reword this and go deeper or close it since it's not possible in it's current state.

pippinsplugins:
We could have `edd_get_payment()`, similar to `get_post()`

cklosowski:
@pippinsplugins yes, and we could then even go as far as to use the caching method that we're using on the AffiliateWP classes. Would probably actually improve performance quite a bit.

cklosowski:
[...]

PR #4719, opened 2016-06-23
Return false if EDD_Payment::setup_payment() fails

zackkatz:
This could cause problems with all the functions in /payments/functions.php - I started updating them, but it's so many that I figured I'd let @cklosowski and @pippinsplugins think about it first.

Related: #4718

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -312,7 +312,11 @@ class EDD_Payment {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


cklosowski:
I moved this into an issue branch PR...will allow me to get some unit tests in place to make sure it's working as expected. @SDavisMedia tested this out a bit and hasn't seen any issues with the helper function @zackkatz but I'll be sure to merge to the `issue/4718` branch and get some unit tests in place to be positive.

Thanks for the PR.

Issue #4717, opened 2016-06-23
Welcome page not showing updating to 2.6


cklosowski:

pippinsplugins:
Pretty sure we need to move the transient check to `edd_after_install()`.

PR #4724, opened 2016-06-24
Check for activation redirect during upgrade routine on major version…

pippinsplugins:
…s. This ensures that plugin updates get sent to whats new page. #4717


cklosowski:
Verified that manually setting the version to `2.6` redirected to the welcome page and `2.6.1` did not.

PR #4732, opened 2016-06-28
EDD employee user_role must not be able to delete payments

rene-hermenau:
user_roles _shop_accountant_, _shop_worker_ and _shop_vendor_ are all able to delete payments! 
Employees should never be able to delete anything, especially not a important thing like a payment transaction.

Quick fix would be to change the deletion capability to `delete_posts`.

This is the minimum capability a user should be able to delete any payment and this would include all above mentioned user_roles. 

If a store owner needs a user_role with deleting capabilities he still can select the shop_manager or would be able than to create a custom user_role with a third party plugin. 
This is not possible at the moment, so all EDD default users are able to delete payments.

Can we change that urgently?

### DIFF
--- a/includes/admin/payments/actions.php
+++ b/includes/admin/payments/actions.php
@@ -287,7 +287,7 @@ function edd_trigger_purchase_delete( $data ) {
 
 		$payment_id = absint( $data['purchase_id'] );
 
-		if( ! current_user_can( 'delete_shop_payments', $payment_id ) ) {
+		if( ! current_user_can( 'delete_posts', $payment_id ) ) {
 			wp_die( __( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
 		}
 


rene-hermenau:
**Edit:** Have found the capability `delete_shop_payment`. 

This would be much more suitable. Just set `'delete_shop_payment'     => true` in class-edd-roles.php for the _shop_manager_  user_role to be backward compatible. 

Adding this to the other default user_roles _shop_accountant_ and _shop_vendor_  would not change any permission for existing default roles  but than we are able to create custom _user_roles_ with the ability for non deletion. 

Would be much helpful

chriscct7:
It is intended that shop_accountants and shop_managers should have the ability to delete payments, as defined by the definition of their roles in their ability to fully manage payments (which by definition includes the right to delete them). These roles are not equivalent to each other, as the former does not have access to certain parts of EDD (for example settings) but does have the ability to see certain things (for example edd_sensitive_data) that a shop worker cannot.

It could be argued shop workers should not have the right to delete payments, but given they've had this permission for 3 years now, its not something that should probably not be removed. If we remove it, we'd have to deal with the repercussions of this breaking a long time used workflow for countless online stores, and so far, this is the only registered complaint we've received about this functionality, despite the large number of active installs who have had it for many years now.

The shop vendor role, at least in testing it, does not have the ability that I could find to delete payments.

If anything, we should add `delete_shop_payments` to the shop accountants make this permission more explicit. We could move to using that over `edit_shop_payments` for the deletion, but that would be something we'd have to do either over time BC-breaking or meta_cap mapping for a BC option.

It should be noted the shop_manager account already has `delete_shop_payment`.

rene-hermenau:
@chriscct7 The shop vendor is not able to access global payments or customer data etc. so this user role is not a real solution here. I am just looking for a solution to create a user_role like a shop worker for doing some maintenance but without the power of deleting payments. Really wonder that no one else noticed that before. Maybe i just overlook something. ?

> We could move to using that over edit_shop_payments for the deletion, but that would be something we'd have to do either over time BC-breaking or meta_cap mapping for a BC option.
> It should be noted the shop_manager account already has delete_shop_payment.

You are right, i see this now `in get_core_caps()` so chances are much better that this would not break anything;)

Is there anything i can do for the team for a long term test to proof the compatibility?

pippinsplugins:
I agree with @chriscct7 that these roles should have the ability to delete payments.

@rene-hermenau you could create a custom user role and then assign all necessary capabilities to them except `delete_shop_payment`. Wouldn't that solve the issue for you?

rene-hermenau:
Thanks for stopping by @pippinsplugins 

No, thats not working because `edit_shop_payments ` is the capability which allows deletion.

If it was `delete_shop_payment` everything would be fine and we would be able to define a user without deleting capabilities. Changing the capability to `delete_shop_payment` would be the solution!

Alt text

Thats the original code:

```
/**
 * Trigger a Purchase Deletion
 *
 * @since 1.3.4
 * @param $data Arguments passed
 * @return void
 */
function edd_trigger_purchase_delete( $data ) {
	if ( wp_verify_nonce( $data['_wpnonce'], 'edd_payment_nonce' ) ) {

		$payment_id = absint( $data['purchase_id'] );

		if( ! current_user_can( 'edit_shop_payments', $payment_id ) ) {
			wp_die( __( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
		}

		edd_delete_purchase( $payment_id );
		wp_redirect( admin_url( '/edit.php?post_type=download&page=edd-payment-history&edd-message=payment_deleted' ) );
		edd_die();
	}
}
```

pippinsplugins:
Oh! So we need both `edit_shop_payments` and `delete_shop_payments`.

I would be fine with introducing that new capability and then applying it to all three of the roles that currently get `edit_shop_payments`.

chriscct7:
Shop manager already has that capability actually

rene-hermenau:
Shop worker and administrator, too!.
So nothing breaks when we change that capability. We could even change that capabilities on several other places where data can be deleted. (So it would be possible to create a shop user who can edit anything but delete nothing)

However, let's ask heavy guys like @dannyvankooten @jdevalk @zackkatz  if there could be any possible unwanted  effect before we do so.

I am going to open a new new pull request for this than.

jdevalk:
The more specific the better. We have Custom roles anyway.

pippinsplugins:
Even easier than, just need to adjust the capability checked.

cklosowski:
I'm closing this PR, as we did the necessary in #4746

Issue #4735, opened 2016-06-29
Discounts become inactive on upgrade


psahalot:
I upgraded to 2.6 and an active discount got deactivated. Thought, it might be because 2.6 was a major upgrade. 
But, the same thing happened when upgrading to 2.6.1 and 2.6.2

cklosowski:
Did these discounts have expiration dates or usage limits on them?

psahalot:
Yes, the expiration date is set to June 30, 2016. No usage limits.

cklosowski:
What timezone is your site set to?

cklosowski:
Just to clarity. there was a cron added to the daily tasks to look for discounts past their expiration or at their usage limit and mark them as inactive. This was a measure to clean up the discount codes in #3198 because normally codes were never marked as expired until their expiration date was queried.

pippinsplugins:
Another report: https://secure.helpscout.net/conversation/221290632/37130/?folderId=180499

pippinsplugins:
We should add `if( ! empty() )` checks around the SQL queries so they only run if IDs were found.

sunnyratilal:
I think it's an issue with the `meta_query`.  Calling the function on `init` just to see if it runs, it deactivated all discounts I created which were set to expire tomorrow (1st July).

sunnyratilal:
Running an `if( ! empty() )` would still cause the SQL query to run because the IDs of the discounts are being returned from `edd_get_discounts`

pippinsplugins:
@sunnyratilal Pretty sure that's backwards now. The expiration date needs to be in the past.

sunnyratilal:
Yes you're correct - my mistake. Currently incorrect discounts are being pulled from `edd_discount_status_cleanup`.  However, the correct discounts are being returned from `edd_is_discount_expired`.

sunnyratilal:
Think I got it, incorrect date formats are being compared.

The date format stored in `_edd_discount_expiration` is `06/30/2016 10:00:00` whereas the format returned from `current_time( 'mysql' )` is `2016-06-30 18:05:54`.

cklosowski:
This is our fault we store an odd format on the post meta. We should
convert it if possible.

-- 
Sent from Gmail Mobile

sunnyratilal:
I've got a commit ready which works in the mean time, should we add an upgrade routine instead which just converts the date format of all the discounts?

pippinsplugins:
[...]

PR #4745, opened 2016-06-30
Issue/4735

sunnyratilal:
Fix for #4735 

The date format stored in _edd_discount_expiration is 06/30/2016 10:00:00 whereas the format returned from current_time( 'mysql' ) is 2016-06-30 18:05:54 causing the MySQL comparison query to be incorrect hence returning discounts which shouldn't be returned.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1585,7 +1585,7 @@ function edd_discount_status_cleanup() {
 			array(
 				'key'     => '_edd_discount_expiration',
 				'value'   => current_time( 'mysql' ),
-				'compare' => '<',
+				'compare' => '>',
 			),
 		),
 	);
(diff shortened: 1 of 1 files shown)


includes/discount-functions.php
```diff
@@ -1584,7 +1584,7 @@ function edd_discount_status_cleanup() {
 			),
 			array(
 				'key'     => '_edd_discount_expiration',
-				'value'   => current_time( 'mysql' ),
+				'value'   => date( 'm/d/Y H:i:s' ),
```
pippinsplugins: please pass `current_time( 'timestamp' )` in as the second parameter to ensure the timezone is correct.

sunnyratilal: Fixed in a4686cb92564e0764a4e7e2332cf737acd1871e6

Issue #4645, opened 2016-06-01
Address Line 2 set to required after some ajax actions


pippinsplugins:
For some reason, after ajax actions fire, the "Line 2" field of the billing address section is getting `required="required"` appended to it, triggering an HTML5 validation error when left empty.

No where do we mark line 2 as a required field.

It happens randomly and does not appear to be something we can reliably replicate, but it definitely happens. I've seen it on the live EDD site and numerous customer sites.

Latest report: https://secure.helpscout.net/conversation/209878954/35441/?folderId=180499

It appears to only happen after an ajax action results in the fields being reloaded, such as changing gateways.

sunnyratilal:
Tracked it down.

We're adding a `required` attribute to all inputs here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/edd-checkout-global.js#L152

The [`inputs`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/edd-checkout-global.js#L142) array include `#edd_cc_address .edd-input` and the Address Line 2 HTML is `<input type="text" id="card_address_2" name="card_address_2" class="card-address-2 edd-input" placeholder="Address line 2" value="" required="required">` which has a class of `edd-input`.

What would you propose is the best way to go about this?

We could wrap the code like this:
```js
if( ! inputs.hasClass('card-address-2') ) {
    inputs.attr('required','required');
}
$('#edd_cc_fields,#edd_cc_address').slideDown();
```

sunnyratilal:
In hindsight, `hasClass` won't work, we'd need to use `.is()`

pippinsplugins:
OOOH, good catch!

seedprod:
Yep,  saw that code where inputs are set with #edd_cc_address .edd-input and required is applied in edd-checkout-global.js 
This occurs consistently after applying a discount after a gateway is selected.

sunnyratilal:
I'll get a patch written up

stevengliebe:
Thanks for that!

sunnyratilal:
@pippinsplugins should I generate minified scripts too or would that be pre-deployment of 2.6.3?

pippinsplugins:
@sunnyratilal No, we do that pre deployment.

PR #4749, opened 2016-06-30
Ensure that Address Line 2 doesn't have the required attr set after AJAX calls. #4645 

sunnyratilal:
Fix for #4645


pippinsplugins:
Looks good, merge once tests pass.

PR #4754, opened 2016-07-03
Fixes #4753 - Warning: array_key_exists() on activation in WPMS sub-d…

### LINKED ISSUE #4753: Warning: array_key_exists() on activation in WPMS sub-domain
When I activated EDD, having installed a local Git repo at the latest level ( v2.6.3, head 8cb0101 ), I noticed the following warning message.

```
Warning: array_key_exists() expects parameter 2 to be array, boolean given in 
<b>C:\github\easydigitaldownloads\easy-digital-downloads\includes\install.php</b>
on line 335.
```

### Problem explanation
In this instance, the value of wp_roles->roles was false. 
This can occur when no user roles have been defined for the site. 
get_option will return false, but EDD expects an array.

### Workaround
run with WP_DEBUG false, or simply ignore the message the first time you see it.

### Problem recreation

I recreated the problem by removing the roles that EDD creates and deleting the option ( from wp_6_user_roles, 6 being the site ID. ).

An easier way is to activate EDD in another site.

image

As an aside, the screen capture also shows two messages about tracking plugin usage.

bobbingwide:
When no roles are define WP_roles->roles is set to false. 
Adding a test for empty() before ! array_key_exists() avoids the Warning message that may be issued. 

Fix for #4753

### DIFF
--- a/includes/install.php
+++ b/includes/install.php
@@ -331,8 +331,9 @@ function edd_install_roles_on_network() {
 	if( ! is_object( $wp_roles ) ) {
 		return;
 	}
+	
 
-	if( ! array_key_exists( 'shop_manager', $wp_roles->roles ) ) {
+	if( empty( $wp_roles->roles ) || ! array_key_exists( 'shop_manager', $wp_roles->roles ) ) {
 
 		// Create EDD shop roles
 		$roles = new EDD_Roles;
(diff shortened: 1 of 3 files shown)


bobbingwide:
The second commit updates the unit tests. Note: I've not actually run them. I don't know how the tests detect unexpected output such as the Warning message that this fix is intended to avoid.

cklosowski:
@bobbingwide There was a bug in the unit tests...we only wanted to reset the `roles` property of the `$wp_roles` object, not the entire thing, so I've fixed up the unit tests so we don't completely blast away the roles for the remainder of the tests.

cklosowski:
Closing in favor of #4760. It fixes the unit tests and also removes code intended to solve a separate issue. Commit history remains in tact.

PR #4756, opened 2016-07-04
Fixes #4755 - Notice from purchase_link with invalid sku parameter

### LINKED ISSUE #4755: Notice from purchase_link with invalid sku parameter
When attempting to use the purchase_link shortcode with an invalid sku parameter ( e.g. `[purchase_link sku=SKU]` )
the following Notice was displayed.

`Notice: Trying to get property of non-object in C:\apache\htdocs\wp-a2z\wp-content\plugins\easy-digital-downloads\includes\shortcodes.php on line 61`

Note: WP_DEBUG is true

bobbingwide:
Should no longer issue the Notice when WP_DEBUG true and the given sku is invalid. #4755

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -58,7 +58,10 @@ function edd_download_shortcode( $atts, $content = null ) {
 	if( ! empty( $atts['sku'] ) ) {
 
 		$download = edd_get_download_by( 'sku', $atts['sku'] );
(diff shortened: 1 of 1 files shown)


cklosowski:
Thanks @bobbingwide. Moved your code to an issue branch and made #4761 which includes a unit test for your case as well.

Issue #4539, opened 2016-05-04
Deleted Downloads Remain in Cart


spencerfinnell:
If you add some downloads to the cart, delete the downloads (I reset the database) and view the cart again the deleted items still appear in the cart: https://cldup.com/2EpImrOYxE.png

`edd_get_cart_contents()` should filter published downloads only (or at least in the cart template).

cklosowski:
a little Work In Progress, pausing and committing up what I have for now to focus on 2.6.3 release.

pippinsplugins:
@cklosowski see comment on PR

cklosowski:
Fixed.

PR #4772, opened 2016-07-06
Don't allow items in the cart that the user cannot purchase (or are deleted) #4539

cklosowski:
#4539

### DIFF
--- a/includes/download-functions.php
+++ b/includes/download-functions.php
@@ -1240,6 +1240,13 @@ function edd_validate_url_token( $url = '' ) {
 	return apply_filters( 'edd_validate_url_token', $ret, $url, $query_args );
 }
 
+/**
+ * Checks if the current user can purchase the specified download ID
+ *
+ * @since  2.6.4
+ * @param  integer $download_id The Download ID
+ * @return bool                 If the current user can purcahse the download ID
+ */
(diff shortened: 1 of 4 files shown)


includes/download-functions.php
```diff
@@ -1239,3 +1239,25 @@ function edd_validate_url_token( $url = '' ) {
...
 }
+
+/**
+ * Checks if the current user can purchase the specified download ID
+ *
+ * @since  2.6.4
+ * @param  integer $download_id The Download ID
+ * @return bool                 If the current user can purcahse the download ID
+ */
+function edd_user_can_purcahse_download( $download_id = 0 ) {
```
pippinsplugins: Can we please make a method of `EDD_Download` please? Also `purchase` is typoed in function name as `purcahse`

cklosowski:
Updated @pippinsplugins

pippinsplugins:
@cklosowski looks like tests are failing. Let's also please add a test that adds a `__return_false` via the filter.

cklosowski:
Failure is unrelated.

The command "bash bin/install-wp-tests.sh wordpress_test root '' localhost
$WP_VERSION" failed and exited with 35 during .

pippinsplugins:
Ah yep, I didn't look at the actual failure. Let's still please add the additional test.

Issue #4757, opened 2016-07-05
Date strings not properly parsed in product import


topher1kenobe:
If you do a product import with a date format like this: 5/06/16 the import will fail but report success.

topher1kenobe:
I was not able to make this fail on master. I have a theory. At the same time we were working on this, we were also discovering that an invalid status would ALSO fail, and report success.  I'm beginning to think this wasn't an issue at all. Which doesn't mean tightening the logic around the date is bad.

Testing on this branch next.

topher1kenobe:
UPDATE: I was wrong, I *did* make it fail on master. And it's verified fixed on this branch.

cklosowski:
Fixing merge conflicts then will merge the PR.

PR #4774, opened 2016-07-06
Reformat dates on payments and downloads imported for mysql format #4757

cklosowski:
#4757

### DIFF
--- a/includes/admin/import/class-batch-import-downloads.php
+++ b/includes/admin/import/class-batch-import-downloads.php
@@ -106,6 +106,21 @@ class EDD_Batch_Downloads_Import extends EDD_Batch_Import {
 					}
 				}
 
+				// Format the date properly
+				if ( ! empty( $args['post_date'] ) ) {
+
+					$timestamp = strtotime( $args['post_date'] );
+					$date      = date( 'Y-n-d H:i:s', $timestamp );
+
+					// If the date provided results in a date string, use it, or just default to today so it imports
+					if ( ! empty( $date ) ) {
+						$args['post_date'] = $date;
+					} else {
+						$date = '';
+					}
+
+				}
+
 				$download_id = wp_insert_post( $args );
 
 				// setup categories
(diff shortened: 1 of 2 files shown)


includes/admin/import/class-batch-import-downloads.php
```diff
@@ -106,6 +106,21 @@ public function process_step() {
 					}
 				}
 
+				// Format the date properly
+				if ( ! empty( $args['post_date'] ) ) {
+
+					$timestamp = strtotime( $args['post_date'] );
```
pippinsplugins: This needs to pass `current_time( 'timestamp' )` as the second paramter.

cklosowski: No, the 2nd parameter is if you only want the timestamp to be relative to the 2nd time. Supplying it would not give us the proper date.

cklosowski:
```
Parameters ¶

time
A date/time string. Valid formats are explained in Date and Time Formats.

now
The timestamp which is used as a base for the calculation of relative dates.
```

pippinsplugins: That would be correct. We want the timestamp to be consistent with the site's timezone. _Not_ setting it could give us incorrect results.

cklosowski: I'll add it, but i've verified on many timezones that adding this makes no difference in the output:
```
<?php
$timestamp = strtotime( '6/5/2016' );
var_dump($timestamp);
$timestamp2 = strtotime( '6/5/2016', current_time( 'timestamp' ) );
var_dump($timestamp2);

$date = date_i18n( 'Y-n-d H:i:s', $timestamp );
$date2 = date_i18n( 'Y-n-d H:i:s', $timestamp2 );
var_dump($date, $date2);
```
```
/app/public/wp-content/plugins/debug-bar-console/class-debug-bar-console.php(84) : eval()'d code:3:int 1465084800
/app/public/wp-content/plugins/debug-bar-console/class-debug-bar-console.php(84) : eval()'d code:5:int 1465084800
/app/public/wp-content/plugins/debug-bar-console/class-debug-bar-console.php(84) : eval()'d code:9:string '2016-6-05 00:00:00' (length=18)
/app/public/wp-content/plugins/debug-bar-console/class-debug-bar-console.php(84) : eval()'d code:9:string '2016-6-05 00:00:00' (length=18)
```

I did this with timezones at, before and after GMT.

PR #4790, opened 2016-07-11
Add css rule to target the new tel input for the padding declaration

Ismail-elkorchi:
This will give the tel input the same padding as the other fields.


cklosowski:
:+1: Thanks @Ismail-elkorchi

PR #4798, opened 2016-07-15
Restrict get parameters

### LINKED ISSUE #4797: Issue with "Redirect to Checkout" setting.
When "Redirect to Checkout" setting is enabled, parameters in the URL are NOT maintained.
This is important, especially when the links are affiliate, like one example below - 

http://example.com/checkout/?ref=10&edd_action=add_to_cart&download_id=7511&edd_options[price_id]=1&campaign=dynamic

Case 1 - When "Redirect to Checkout" is enabled - 
http://example.com/checkout

Case 2 - When "Redirect to Checkout" is disabled - 
http://example.com/checkout/?sujay=10&campaign=dynamic

Here's a video I quickly recorded if that could help understand the issue better - https://www.youtube.com/watch?v=2wIRwP1Y0ak

This at the least affects to affiliates that might be using "Add to Cart" link to the store.

pratikchaskar:
Fixes #4797

- Modified edd_straight_to_checkout function()
- Checked if URL has get parameters, if so - then stop auto redirect to checkout

### DIFF
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -137,43 +137,35 @@ function edd_process_register_form( $data ) {
 	do_action( 'edd_pre_process_register_form' );
 
 	if( empty( $data['edd_user_login'] ) ) {
-		$empty_username_message = apply_filters( 'empty_username_message', 'Invalid username' );
-		edd_set_error( 'empty_username', __( $empty_username_message, 'easy-digital-downloads' ) );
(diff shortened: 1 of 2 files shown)


pippinsplugins:
The redirect to checkout should not be stopped, instead it should simply re-add any non-EDD parameters to the checkout URL so that they are preserved after the redirection.

pippinsplugins:
Closing in favor of https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4900

PR #4821, opened 2016-07-22
Add $items parameter to the `edd_post_add_to_cart` action

### LINKED ISSUE #4820: Add $items parameter to edd_post_add_to_cart action
There are lots of calculations done in the `edd_add_to_cart()` function to process adding multiple price IDs and different quantities of a download.

These are stored in the `$items` variable, but not passed to the `edd_post_add_to_cart` action. This means additional calculations are necessary if hooking into this action.

zackkatz:
This makes it unnecessary to reproduce the logic inside `edd_add_to_cart()`.

Closes #4820

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -238,7 +238,6 @@ function edd_add_to_cart( $download_id, $options = array() ) {
 	foreach ( $items as &$item ) {
 		$item = apply_filters( 'edd_add_to_cart_item', $item );
 		$to_add = $item;
-		unset( $item );
 
 		if ( ! is_array( $to_add ) )
 			return;
@@ -263,6 +262,8 @@ function edd_add_to_cart( $download_id, $options = array() ) {
 
 		}
 	}
+
+	unset( $item );
 
 	EDD()->session->set( 'edd_cart', $cart );
 
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -233,8 +233,9 @@ function edd_add_to_cart( $download_id, $options = array() ) {
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


includes/cart/functions.php
```diff
@@ -233,8 +233,9 @@ function edd_add_to_cart( $download_id, $options = array() ) {
 		);
 	}
 
-	foreach ( $items as $item ) {
-		$to_add = apply_filters( 'edd_add_to_cart_item', $item );
+	foreach ( $items as &$item ) {
```
cklosowski: Great idea to send the `$item` in as a reference so we're directly modifying it  with the filter. Can we be sure to put an `unset( $item );` call right after the foreach here then? 

They recommend this in the PHP docs:
`Warning: Reference of a $value and the last array element remain even after the foreach loop. It is recommended to destroy it by unset().`

cklosowski: Other than that I think this is good to go.

includes/cart/functions.php
```diff
@@ -233,8 +235,11 @@ function edd_add_to_cart( $download_id, $options = array() ) {
...
+		$item = apply_filters( 'edd_add_to_cart_item', $item );
+		$to_add = $item;
+		unset( $item );
```
cklosowski: Hey @zackkatz I think I might not spoken correctly. the `unset` has to be right after the trailing `}` of the foreach:
```
foreach ( $items as &$item ) {
    // Do the awesome filter
    // Do the rest of the stuff
}
unset( $item );
```

Apparently with the referenced method that variable exists not outside of the foreach, as it doesn't destroy it.

PR #4832, opened 2016-07-26
Allow custom Arguments for Settings API Callbacks

d4mation:
The way that `add_settings_field()` was being called before was very restrictive in what you could and could not pass to EDD Settings API Callbacks.

Now it utilizes `wp_parse_args()` to set the same defaults as before and simply passes `$args` to `add_settings_field()` instead.

This is useful if a User is creating a custom Callback function for a custom Field Type and wants to add more, custom Arguments than what EDD usually supports for its Settings API.

So if a User is adding a custom Field Type to their Settings Tab and wants to have an Argument called `add_item_text` or anything else that they need, they can.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -179,34 +179,34 @@ function edd_register_settings() {
 				if ( empty( $option['id'] ) ) {
 					continue;
 				}
-
-				$name = isset( $option['name'] ) ? $option['name'] : '';
+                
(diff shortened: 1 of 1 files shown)


pippinsplugins:
This looks really nice. Milestoning for 2.6.7. We'll review further and test. Thanks!

michaelbeil:
Issue #4856.

cklosowski:
Clearing the milestone as we've opened an issue to track it, which has the milestone attached.

Issue #4796, opened 2016-07-14
Update MasterCard Logo


chriscct7:
MasterCard has announced they are changing their logo. We need to update the one we include in EDD. The new one looks like:
![image](https://cloud.githubusercontent.com/assets/1324637/16846430/6b0d5942-49ba-11e6-8b7a-d39ba7a68e84.png)

The name "MasterCard" is no longer camelcased, but on websites should be displayed "mastercard."

cklosowski:
@SDavisMedia don't we have an open issue to work on the cc payment icons anyway? #4238?

pippinsplugins:
Is the change already official or planned?

cklosowski:
![screen shot 2016-07-14 at 9 03 01 am](https://cloud.githubusercontent.com/assets/1390124/16846568/c6936284-49a1-11e6-8db9-99e7448056da.png)

It's live

chriscct7:
It's already in effect on most of their properties

SDavisMedia:
Yea I was just gonna say that, @cklosowski. See, I took forever on the other issue for a reason. :)

I'll have to find some good icons with an update MC logo ASAP.

SDavisMedia:
For reference: https://www.mastercardbrandcenter.com/us/getourbrand/index.shtml?pageId=dl_0340&expertVisible=false&staticPage=yes

They need to update their assets.

sunnyratilal:
They are releasing updated brand assets on September 30 so we still got a while.
![image](https://cloud.githubusercontent.com/assets/1525554/16894902/1ffb19cc-4b5d-11e6-97b1-ad166d84fbc6.png)

chriscct7:
Ironically that banner uses "Mastercard" which is not either their old
brand "MasterCard" or their new one "mastercard."

lalunecreative:
I'd be willing to volunteer to create some new icons for you guys. It wouldn't take me long. I'm assuming you want something in the same sizes as before until you can get another solution. They'll be the updated versions of the logos for 2016. Mind if I make them square to get rid of the bleed problem from this ticket? #4238 

Ideally they would be .png files since those look better for retina screens as well.

SDavisMedia:
@lalunecreative I'd love it if you took a stab at it! However, I do think they should remain rounded at the corners. The current icons could very will be rounded and transparent as well. The fact that they are not is just an oversight.

Since you're considering PNG, could you keep them rounded and roughly the same size?

[...]

Issue #4238, opened 2016-01-25
accepted payment method icons need transparent backgrounds


SDavisMedia:
Right now, the .gif files actually appear to be icons with a white background. The icons have rounded corners so the white background pokes through on all four corners. That's no good from a designer's perspective.

Excuse the ugly screenshot but you can see it here in the dashboard settings:

![screen shot 2016-01-25 at 8 48 20 am](https://cloud.githubusercontent.com/assets/2359131/12557598/7b5e8484-c340-11e5-9c22-dd49d62124a6.png)

cklosowski:
:+1: might be nullified once #2717 (and dashicons for that matter) is figured out but this is a good short term solution

cklosowski:
@SDavisMedia can you do this? I don't have a Creative Suite license anymore, just have Illustrator.

SDavisMedia:
Sure thing... anyone know where we got the icons from? They might just need to be replaced unless we have some files somewhere.

cklosowski:
the files are in the repo, https://github.com/easydigitaldownloads/Easy-Digital-Downloads/tree/master/templates/images/icons

SDavisMedia:
@cklosowski Not sure what's important about those files. Editing them to cut the white corners is probably not the best route, if that's what you're implying. Should be a last resort. I'll shoot for a better solution and only do that if it comes down to it.

cklosowski:
Ok perfect. I assumed that would be the easiest method.

pippinsplugins:
Merged into master!

I'm going to leave the `.gif` versions in the plugin just in case anyone is loading them manually.

michaelbeil:
@pippinsplugins What about theses instances within edd.css?

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/templates/edd.css#L409-L419

cklosowski:
Woot! no more iPhone 3GS!

Issue #2717, opened 2014-10-13
Merge a set of payment font icons into the EDD Font, remove current payment gateway images


adampickering:
Here's the project that has already created many payment gateway icons: http://www.orlandotm.com/payment-webfont/

We could merge these into the EDD icon font and create any that need to be created, i'm pretty good with SVG.

Thoughts? @pippinsplugins @spencerfinnell

pippinsplugins:
@ghost1227 Could you please make a pull request for this?

pippinsplugins:
We can't move forward with this anymore, at least not until the iconfont > SVG changes are finished in core.

ghost1227:
Good. Iconfonts ended up being a horrible idea from a UX perspective. Once the SVG changes happen and we can use multi-color graphics, it'll be much more epic!

adampickering:
If you need me to convert everything over to individual SVG images let me know!

pippinsplugins:
:+1:

pippinsplugins:
We can't do anything here until core decides what the heck they're doing with dashicons.

pippinsplugins:
See #4846

PR #4846, opened 2016-08-03
Changing out credit card icons for new ones

lalunecreative:
New credit card icons. Tickets #4796 #4238 #2717


lalunecreative:
I'm not sure what that iphone icon is for. I couldn't find where it's showing up in EDD anywhere, but if I need to update that one as well, let me know. It's also outdated.

michaelbeil:
If you can update the iPhone icon as well, that would be great!

lalunecreative:
iPhone icon. 
 #iphone

michaelbeil:
That looks great! Want to add it to the PR?

chriscct7:
While it's not in line with their logo design, I almost feel like we should increase the size of the "mastercard" text in the mastercard logo so it's readable. Anyone else have thoughts on that?

lalunecreative:
Committed iphone icon. I could personally go either way with the mastercard text. It's going to be a tiny icon either way. Even with the new branding most of the tiny ones used for this only show the circle branding and not the text at all.

michaelbeil:
I think it looks ok, but we could slightly increase the font size, though there is not much room for it as the icon is quite small indeed.

michaelbeil:
The more I look at it, the more it seems to be spot on.

pippinsplugins:
Manually merged!

Issue #4877, opened 2016-08-16
Include zip code when recalculating tax on the front-end


erktime:
Sales tax in the united states often varies by zip code. If we include the zip code when calling the ajax endpoint for recalculating sales tax, custom EDD plugins can provide more accurate tax rates based on country, state and zip code.

I will attach a pull to support this shortly. Its minimal change but offers a lot of flexibility for those that need more precise tax calculations for everywhere in the US.

pippinsplugins:
Thanks @erktime, we'll get it tested.

SDavisMedia:
I don't have anything custom set up to test that this works fully but the ajax call seems to be working just fine. I'm more than willing to test further if anyone has a snippet that could take advantage of this.

erktime:
@SDavisMedia Sorry for the slow response. My github notifications are a mess. If you created a plugin with this code, you could see how this could be leveraged. All this does is return a new tax rate if we have a non $0 balance shopping cart and the customer is in the US and has a zip code we want to alter the tax rate for.



```php
/**
 * Alter the edd tax rate for zip codes we find in the db for US purchasers.
 */
function avaTax_edd_tax_rate_filter($rate, $country = null, $state = null) {
  // Get price of items
  $cart_items = edd_get_cart_contents();
  $total = 0;
  foreach( $cart_items as $key => $item ) {
    $item['quantity'] = edd_item_quantities_enabled() ? absint( $item['quantity'] ) : 1;
    $price_id = isset( $item['options']['price_id'] ) ? $item['options']['price_id'] : NULL;
    $item_price = edd_get_cart_item_price( $item['id'], $item['options'] );
    $discount   = edd_get_cart_item_discount_amount( $item );
    $discount   = apply_filters( 'edd_get_cart_content_details_item_discount_amount', $discount, $item );
    $quantity   = edd_get_cart_item_quantity( $item['id'], $item['options'] );
    $fees       = edd_get_cart_fees( 'fee', $item['id'], $price_id );
    $subtotal   = $item_price * $quantity;
    foreach ( $fees as $fee ) {
      if ( $fee['amount'] < 0 ) {
        $subtotal += $fee['amount'];
      }
    }
    $total = $total + $subtotal - $discount;
  }
  
  if ($total > 0) {
    $user_address = edd_get_customer_address();
    // error_log(print_r($_POST, true));
    if (empty($country)) {
[...]

PR #4878, opened 2016-08-16
Include zip code in front-end tax re calculation. Fix issue 4877

erktime:
US tax districts can vary by zip code, not just state, this commit
allows for plugins to alter the tax based on the user's billing zip
code.

Send zip code in tax recalculate ajax call.

Recalculate tax on zip code change.

This should address issue https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4877


erktime:
Just updated this pull request with a new commit https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4878/commits/6eefceed15daaf307fad20c5f54417bdfd0c9ee1

Noticed there was a bug with browser form autocomplete and simultaneous ajax calls fired to recalculate sales tax.

cklosowski:
PR looks good, I'm going to branch it off, into an issue branch, refresh it with `release/2.7` and submit the PR to the `release/2.7` branch since that's where the issue is set for a milestone. Commit history will be kept in tact @erktime

PR #4879, opened 2016-08-16
Add Placeholder attribute for Text Inputs

d4mation:
Sometimes it is nice to have something for the `placeholder` attribute like desired input examples that you may not want to otherwise put in the Field's Description. 

I needed/wanted it for a project I'm working on, so I made this quick tweak. Just repurposes the `$args['placeholder']` that is also used within Chosen `<select>` field creation since their intent is similar.


d4mation:
The only Travis CI checks that failed were related to WP Latest Version. My only guess as to why would be because WP 4.6 dropped earlier today and it could be messing with Travis CI.

chriscct7:
Once the latest version of WordPress is tagged, it takes a few hours before the SVN for the tagged latest to start including the development tools. These are expected fails over the next hour or so.

Issue #4487, opened 2016-04-12
Offer variation reporting


chriscct7:
We should offer a pie chart for sales and another one for earnings of each variation of a download representative of the whole download. That way store owners could make informed decisions on adjustments on price or discontinuation of  variations

pippinsplugins:
Definitely.

mindctrl:
Closing this in favor of https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4889

Issue #3972, opened 2015-11-02
Further improve Customer Export functionality


cklosowski:
With #3025 we've made it possible to export all customers of a download, or only specific price IDs, however with #3175 we're going to make the multi-select product drop down support price IDs, and we can then make this export even better by allowing export of multiple products, and price IDs combined.

Great idea @brashrebel

chriscct7:
See https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4895

Issue #3405, opened 2015-05-15
Allow discounts to be tied to variable price options.


mindctrl:
We currently can't add discounts for just a single variable option. Probably dependent on https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/3175. [Requested here](https://easydigitaldownloads.com/support/topic/discount-code-for-specific-price-option/#post-523694) and I think a few other times in the past.

cklosowski:
We're going to have to set some limitations. The multi selector doesn't have a great way to achieve this and the performance of looking up all products and their price IDs is a concern.

It may just be that discounts can o my apply to a single price is of a product. But that seems really limited as a feature.

pippinsplugins:
:+1:

mindctrl:
Can we do it like we do on the payment details screen? Look up just the products, then if a variable one is selected, show the options in another dropdown?

cklosowski:
Not in the current method because it's a multi select. Not a single select.

flowdee:
This feature is absolutely required in my opinion!

I want to give out discounts to push high price variants but this is simply not possible right now.

cklosowski:
@flowdee we're not debating if we should or shouldn't do it, just trying to figure out the best way to do it from a UX perspective.

flowdee:
ok fine, then go on :+1:

chriscct7:
See https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4895

cklosowski:
With #5024 merged, we can go forward with this!

pippinsplugins:
#4895 needs refreshed.

chriscct7:
PR was removed from milestone for refreshing, but ticket wasn't updated to reflect this

sunnyratilal:
I'll get it refreshed

collizo4sky:
Any news on this? This is an absolute needed feature for us.

pippinsplugins:
@collizo4sky we do not have an ETA on it. If it's an important feature for you, assisting in testing / updating the associated PR could help facilitate getting this completed.

collizo4sky:
If only this would happen. I will be a happy man.

mihaijoldis:
Another request for the variable price discount codes https://secure.helpscout.net/conversation/1937172486/2026525?folderId=5762562

Issue #3175, opened 2015-02-20
The HTML class's product_dropdown should support variable prices


chriscct7:
We should add a variable key to the product_dropdown so that if variable=true it also shows the individual keys for the variable products, similar to how Discounts Pro shows them on it's dropdown. Maybe even offer the ability to pass in a product ID and if so, show the variable options for that product in the dropdown

pippinsplugins:
:+1:

pippinsplugins:
Punting this to 2.5. It's great but not greatly necessary yet.

cklosowski:
I'm going to punt this to 2.6, to coincide with #3972

chriscct7:
This is also a major blocker for Discounts Pro. It's the only field blocking it from moving to using entirely EDD HTML functions which would solve 90% of the issues it has as an extension.

chriscct7:
This also blocks a pair of FES features that require variable prices in the inputs as well.

chriscct7:
See https://github.com/easydigitaldownloads/easy-digital-downloads/pull/4895

cklosowski:
@easydigitaldownloads/core-devs Hey guys, so we've got quite a few extensions that could really benefit from this addition. What do you guys feel about getting into something prior to 2.7. I know it's a big change, so it _might_ need to go into a major release, but the sooner the better.

SDavisMedia:
I personally want to see it in there as quickly as possible. I've got two users I can update for two different extensions (once they're updated, based on this).

chriscct7:
So what we urgently need for discounts pro is just the html-elements file
changes. The other files changed are to add discounts by variations which
we don't need for Discounts Pro's fixes or for my other extensions. I'd
recommend splitting that file's changes out and doing it in hte next point
release and then we can add discounts by variations as a feature later if
we want

pippinsplugins:
I'm fine with moving it up to pre 2.7.

SDavisMedia:
I can verify that it's implemented like so: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5024#issuecomment-253064753 But there's not much left to test at the moment.

pippinsplugins:
@cklosowski could you please put together a quick sample plugin (or update Discounts Pro for it) to demonstrate it to make it easier to test this?

chriscct7:
Discounts Pro already works using the proposed two dropdowns

cklosowski:
[...]

PR #4895, opened 2016-08-22
Implement variable picker in select field & first pass with variation support for coupons

chriscct7:
This is a first pass of implementing variations in the EDD HTML class product_dropdown (working). It also enables the ability to create discounts that require or limit by product variations (working).

Note, I do not have the bandwidth to go around to other places of EDD that use product_dropdown to enable variation support, like for example in reporting ( #4487 ), so this pr should either be used as a base that is extended to do that or as guidance on how to do this for a completely new pr.

Solves https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3175 and  https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3405

This also makes doing https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3972 pretty trivial (just turn on chosen + variable support)

### DIFF
--- a/includes/admin/discounts/add-discount.php
+++ b/includes/admin/discounts/add-discount.php
@@ -72,8 +72,9 @@ if ( ! defined( 'ABSPATH' ) ) exit;
 							'name'        => 'products[]',
 							'id'          => 'products',
 							'multiple'    => true,
-                            'chosen'      => true,
-                            'placeholder' => sprintf( __( 'Select one or more %s', 'easy-digital-downloads' ), edd_get_label_plural() )
+							'chosen'      => true,
+							'variations'  => true,
+							'placeholder' => sprintf( __( 'Select one or more %s', 'easy-digital-downloads' ), edd_get_label_plural() )
 						) ); ?><br/>
 					</p>
 					<div id="edd-discount-product-conditions" style="display:none;">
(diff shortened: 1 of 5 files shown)


chriscct7:
Note, if this gets merged into 2.7, please let me know as this would allow Discounts Pro to move to using EDD HTML class items for it's admin panel (solving many issues), and we can also use this in FES as well.

chriscct7:
Also, someone should write unit tests for this. I didn't have the time to do so, and it doesn't look like I'll have the time to do so in the near future

SDavisMedia:
So lovely. It would be awesome if we could get this going sooner than later. I have a Discounts Pro ticket where the fix depends on this functionality. 🎉

chriscct7:
Please notify https://github.com/chriscct7/edd-discounts/pull/105 when status changes. Thanks :+1:

pippinsplugins:
This is looking awesome!

One quick thing I noticed. `'data' => array( 'search-type' => 'customer' )` in the `user_dropdown()` method. That should be `user` for the search type and then the search function needs updated to return users instead of customers.

chriscct7:
I would also suggest adding a helper function, like `edd_get_download_from_string` which has 1 parameter:, the string to parse

If the string sent in does not contain a `_`, if it is an integer greater than zero, the returned result should be an array( 'post_id' => {the number}. If the string contains an `_`, if it is in the format [0-9]_[0-9] such as 9_2, then the return should be an array( 'post_id' => {post_id}, 'variation'=> {variation}). If the string is not in the [0-9] (for normal product) or [0-9]_[0-9] (for variation) format, the returned result should be a false.

This would allow for redundant logic of getting the price ID out of a string and seeing what type of download a string is to be removed.

SDavisMedia:
I'm pretty sure Product Updates can benefit greatly from this as well. Right now, it can't send emails to a specific price variation. A customer just asked for the feature and I'm sure it's tied to that. I think Product Updates needs that enhancement for sure!

chriscct7:
Split the functionality out required for minor into https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5024/files

PR #4900, opened 2016-08-25
Preserve get parameters when straight to checkout is selected

### LINKED ISSUE #4797: Issue with "Redirect to Checkout" setting.

Nikschavan:
Fixes #4797 

Removes the edd parameters from the URL and then append the remaining parameters to checkout URL and then redirect.

### DIFF
--- a/includes/cart/actions.php
+++ b/includes/cart/actions.php
@@ -80,7 +80,15 @@ function edd_process_add_to_cart( $data ) {
 	$cart        = edd_add_to_cart( $download_id, $options );
 
 	if ( edd_straight_to_checkout() && ! edd_is_checkout() ) {
-		wp_redirect( edd_get_checkout_uri(), 303 );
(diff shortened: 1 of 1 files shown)


cklosowski:
nice work thanks @Nikschavan

PR #4911, opened 2016-08-30
Send the purchase receipt to the primary customer email address.

### LINKED ISSUE #4896: Resend Receipt button sends receipt to original purchase email
When a user hits the "Resend Receipt" button, the receipt will be successfully sent to the email address that was used during the purchase. However, when the primary address for that user is changed (even after completely removing the original one) the receipt will still be sent to the original address. 

One needs to go into the DB and manually edit the purchase record there to make it send to the new address.

We should allow the receipt and new download links to be sent to the customer's primary address instead.

ericnicolaas:
This fixes #4896. I made the change in the `edd_email_purchase_receipt()` function, instead of changing the email that is returned by `edd_get_payment_user_emai()`. I would expect the latter to always return the email address used when the purchase was made.

### DIFF
--- a/includes/emails/functions.php
+++ b/includes/emails/functions.php
@@ -31,7 +31,8 @@ function edd_email_purchase_receipt( $payment_id, $admin_notice = true ) {
 	$from_email   = edd_get_option( 'from_email', get_bloginfo( 'admin_email' ) );
 	$from_email   = apply_filters( 'edd_purchase_from_address', $from_email, $payment_id, $payment_data );
 
-	$to_email     = edd_get_payment_user_email( $payment_id );
+	$customer     = new EDD_Customer( edd_get_payment_customer_id( $payment_id ) );
+	$to_email     = $customer->email;
 
 	$subject      = edd_get_option( 'purchase_subject', __( 'Purchase Receipt', 'easy-digital-downloads' ) );
 	$subject      = apply_filters( 'edd_purchase_subject', wp_strip_all_tags( $subject ), $payment_id );


cklosowski:
We've had many discussions about this, the problem is, that the purchase receipt should always be sent to the email address used at checkout, not necessarily the customer. This is a situation of when someone purchases on behalf of, or in the situation of an employee who has the purchase receipt sent to the company's accounting department.

Am I remembering our conversations correctly @SDavisMedia and @pippinsplugins ?

ericnicolaas:
@cklosowski That makes sense for the initial receipt. When the receipt is _resent_, I think it should be sent to the primary email address (or there should be an option to choose the email address to send the receipt to). 

The problem I am trying to solve is when a customer purchases but does not receive the purchase receipt email (their email might be incorrect, firewall might be blocking it, etc.). The customer reaches out and asks for the receipt to be sent to their other email address.

My way of solving this was to add their alternative email address, set that as primary and resend the receipt, which I figured would force the email to be sent to the primary address.

If it's not safe to assume that a resend should go to the primary address, I would suggest that the Resend Button be accompanied by a dropdown of the customer's email addresses, so that the correct one can be chosen. This dropdown only needs to be visible for customers with multiple addresses.

Does that sound like an acceptable solution?

pippinsplugins:
A drop down could work well, though that'd have to be limited to the view order details screen (not shown on the list table).

ericnicolaas:
@pippinsplugins That's fine by me. Do you have a specific way you want to build this, or can I update the PR with a revision?

pippinsplugins:
@ericnicolaas I'd think a drop down next to the `Resend Receipt` button would work well.

cklosowski:
I agree here, I think a dropdown would be ideal here since we've introduced that 'multi-email' feature. It could default to the one on the payment (if it's attached to the customer) or default to the primary? Thoughts here @pippinsplugins ?

pippinsplugins:
Default to one on payment followed by primary :+1:

pippinsplugins:
Closing in favor of https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5010

Issue #4629, opened 2016-05-28
Price Options Don't Reappear When Removing Item From Cart


metal450:
EDD 2.5.16, WP 4.5.2, Digital Store (theme) 1.3.6

Upon adding an item to the cart, the price options disappear and the "add to cart" button changes to "checkout" (as desired).  Upon removing it from cart (via the widget), the "checkout" button changes back to "add to cart" (as desired), but the price options do not reappear.  This means that unless the user explicitly refreshes the page, they will not be able to re-add to the cart.

Screen capture: http://screencast.com/t/hiXoIU7VcR

pippinsplugins:
@metal450 I _think_ that is a bug specific to the Digital Store theme. As a quick test, do you get the same behavior with Twenty Sixteen?

metal450:
I don't have TwentySixteen installed (can test tomorrow if necessary?), but it does happen on TwentyTwelve, so it doesn't appear to be just Digital Store...

cklosowski:
this appears to only affect using the shortcode for a purchase link. When viewing the product page itself, everything works as designed. I'll keep digging into this, but on the downoad single page itself the page refreshes, on the homepage with a shortcode, it does it all via ajax.

SDavisMedia:
Replicated and confirmed page refresh/fix in `issue/4629`. :+1:

cklosowski:
@metal450 does this solution meet your requirements?

metal450:
@cklosowski Thanks for the follow-up; I'll report back if I still see this issue after the next chance I get to update WP/EDD/DigitalStore, but due to other commitments it might be some months until I can evaluate, so for now if your testing shows this as resolved, feel free to go ahead and close it :)

Thanks again!

PR #4913, opened 2016-08-30
Force removal redirect when purchase links are present #4629

cklosowski:
#4629

### DIFF
--- a/includes/scripts.php
+++ b/includes/scripts.php
@@ -34,6 +34,11 @@ function edd_load_scripts() {
 		$position = edd_get_item_position_in_cart( $post->ID );
 	}
 
+	$has_purchase_links = false;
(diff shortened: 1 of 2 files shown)


includes/scripts.php
```diff
@@ -34,6 +34,11 @@ function edd_load_scripts() {
 		$position = edd_get_item_position_in_cart( $post->ID );
 	}
 
+	$has_purchase_links = false;
+	if ( has_shortcode( $post->post_content, 'purchase_link' ) || has_shortcode( $post->post_content, 'downloads' ) || is_post_type_archive( 'download' ) ) {
```
pippinsplugins: Shouldn't this include `is_singular( 'download' )`?

cklosowski: No the JS handles that part:
`( parseInt( edd_scripts.position_in_cart, 10 ) === parseInt( item, 10 ) )`.

Basically it forces the page to refresh if download we're viewing has a position in the cart already. The addition of this boolean check simply allows us to add another `edd_scripts` item for when we're not on a download that is currently in the cart.

Issue #4657, opened 2016-06-06
Add edd_get_php_arg_separator_output on System Info


chriscct7:
We should add edd_get_php_arg_separator_output() to the system info page, and detect if it's not the default &

PR #4914, opened 2016-08-31
Add PHP Arg Separator to system info #4657

cklosowski:
Addresses #4657 and adds a unit test for the system info to make sure it nearly completes.


tests/tests-tools.php
```diff
@@ -0,0 +1,75 @@
...
+
+
+/**
+ * @group edd_tools
+ */
+class Tests_Tools extends WP_UnitTestCase {
+
+	public function setUp() {
+		parent::setUp();
+		require_once EDD_PLUGIN_DIR . '/includes/admin/tools.php';
```
chriscct7: This string should not start with a slash.

PR #4926, opened 2016-09-08
Issue with unsetting cart discount when code isn't set

nosegraze:
I encountered a problem when using `edd_unset_cart_discount()` with a discount code that isn't actually set/active. This causes `$key` to be `false` and trying to use `unset` with a `false` array key empties the entire array out (thus removing other discount codes that weren't expected to be removed).

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1140,7 +1140,9 @@ function edd_unset_cart_discount( $code = '' ) {
 
 	if ( $discounts ) {
 		$key = array_search( $code, $discounts );
-		unset( $discounts[ $key ] );
+		if ( false !== $key ) {
+			unset( $discounts[ $key ] );
+		}
 		$discounts = implode( '|', array_values( $discounts ) );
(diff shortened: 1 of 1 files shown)


pippinsplugins:
@nosegraze Thanks for the PR! Could you tell me how to replicate the problem? Is it just a matter of manually setting a fake discount code and then trying to unset it?

nosegraze:
Basically I was using this code to auto apply a discount code behind the scenes: https://gist.github.com/nosegraze/ff650536ee3884400f2ea55d1282da84

But if the conditions weren't met, I was unsetting that discount code, without necessarily checking to see if that discount was already applied (because I didn't think I'd need to).

Then I ran into the issue when someone tried applying a different discount code manually. They enter the real discount, click apply, it gets applied for a second, then the page does a quick update and the discount was suddenly removed (triggered by that last `else` bit in the gist).

Make sense?

pippinsplugins:
Makes perfect sense, thanks! Merged!

PR #4937, opened 2016-09-12
Loop until all discounts have been processed or active one is found

### LINKED ISSUE #4936: `edd_has_active_discounts` returns false with active discounts present
Internally `edd_get_discounts` uses a 30 item limit, which is used in `edd_has_active_discounts` to retrieve discounts to check for active status.

This means that if in the first 30 discounts there is no active one, the status is returned as false.

Will provide a PR with a fix.

moorscode:
Loop through discounts until an active one has been found.
Optimize discount fetching to only retrieve IDs.
Only look for 'active' discounts to begin with.

Fixes #4936

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -65,16 +65,42 @@ function edd_get_discounts( $args = array() ) {
 function edd_has_active_discounts() {
 	$has_active = false;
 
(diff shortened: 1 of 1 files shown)


sunnyratilal:
Updated PR opened at #5197

Issue #4662, opened 2016-06-07
Chosen fields in admin don't recover from no results


Shelob9:
Happens in a few places, for example adding download to bundle.

I add a new download, type in the wrong thing, often times -- not consistently -- I have no way of starting over, have to delete the field and restart.

https://www.dropbox.com/s/z6zs5rbzg728fys/edd-select2.mov?dl=0

pippinsplugins:
I just ran into this today.

Shelob9:
Not sure if its just me, but I think it is worse on the payment admin screen. I'm struggling to figure out the exact behavior to reproduce this.

sunnyratilal:
Looks like the `input` field isn't clearing itself at times. I think it's a bug in Select2.

chriscct7:
We use chosen for that not Select2 don't we?

cklosowski:
Yea the point still stands at times though. Chosen doesn't always recover.

On Monday, July 4, 2016, Chris Christoff <notifications@github.com> wrote:

> We use chosen for that not Select2 don't we?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4662#issuecomment-230366046>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABU2LFMjdLOhmDj7U8l78FBq-j1vcmcQks5qSaODgaJpZM4IwZ5I>
> .
>


-- 
Sent from Gmail Mobile

michaelbeil:
I am seeing this today as well.

michaelbeil:
A customer is also having an issue with this in Manual Puchases: https://secure.helpscout.net/conversation/239907705/40288/?folderId=725711

cklosowski:
i can replicate it on the Download Bundle fields, but not on the `View Order Details` screen. I'll dig into Chosen and see what's going on there.

pippinsplugins:
I see it most often on `Manual Purchases.

cklosowski:
I think it has to do with our repeatable fields with the download selector. The Clone is doing something odd to them.

SDavisMedia:
I can only replicate when adding products to a bundle. Manual Purchases works fine but I can make the bundle test fail every time. 

Unfortunately, `issue/4662` is not changing the behavior for me.

cklosowski:
[...]

PR #4944, opened 2016-09-13
Only reset the results when we have some #4662

cklosowski:
#4662


pippinsplugins:
@cklosowski after 2.6.9, can you please refresh this PR?

PR #4965, opened 2016-09-20
Don't output a label if one isn't set for text and select HTML elements

chriscct7:
Right now is label passed to an HTML element is empty, it still outputs an empty label. This is bad practice, because it causes issues if you're defining a label for that field elsewhere.


pippinsplugins:
All good here.

PR #4966, opened 2016-09-21
Error: Trying to get property of non object on non existing pages

rene-hermenau:


cklosowski:
This PR won't work because of the last part of the conditional:

`is_post_type_archive( 'download' )`

Issue #4969, opened 2016-09-22
CSV Product exporter/importer does not honor file conditions for variable pricing.


mintplugins:
If you import a product through the CSV Product Importer and then purchase it, the deliverables files do not show for the customer. However, if you simply open > save the product, they then show.

Before open > save customer sees:
<img width="906" alt="screen shot 2016-09-22 at 11 34 36 am" src="https://cloud.githubusercontent.com/assets/1638128/18755210/45bd0d32-80b9-11e6-82ca-78055fbfdd4d.png">

After open > save customer sees:
<img width="908" alt="screen shot 2016-09-22 at 11 36 32 am" src="https://cloud.githubusercontent.com/assets/1638128/18755220/4d9b045a-80b9-11e6-8919-5e58e66a7072.png">

Currently, this requires the store owner to open and save every imported product to make sure the deliverable files are available to their customers.

cklosowski:
I've confirmed this only happens on variable priced items, and it's because of the condition attribute. It looks like we're not setting that condition, so the file downloads aren't shown.

cklosowski:
Confirmed, our file fields for the export don't respect or even store the price assignments.

cklosowski:
I think we need to add a condition to the export, this will mean we'll have to do a backwards compatibility that if it's a variable priced item and doesn't contain these, we just assign to all...thoughts @pippinsplugins ?

pippinsplugins:
@cklosowski :+1: for assigning to `All`.

cklosowski:
@mintplugins this is ready to test out if you have a few minutes. The issue title has been updated to reflect the bug in detail. Let me know if you have any quesitons.

mintplugins:
@cklosowski I am still having the exact same issue. I double checked my branch and that I've pulled the latest commits for sure. 

One add thing I noticed is that the file price assignments were originally both "All" before exporting" upon import, the first file is assigned to the second price ID.

<img width="913" alt="screen shot 2016-09-23 at 12 00 15 pm" src="https://cloud.githubusercontent.com/assets/1638128/18792676/6a423c5a-8185-11e6-932d-db7c8894d521.png">

mintplugins:
Whoops - I was wrong about the before settings. Here's Screenshots:

EXPORTED PRODUCT:
[...]

PR #4972, opened 2016-09-22
Add support for variable price file conditions to ex/import #4969

cklosowski:
#4969


cklosowski:
@pippinsplugins ready for a review. As a note, this also fixes a bug where the file downloads array keys were starting at an index of `0` when imported. When we do it via the admin we index the array starting with `1`.

Issue #4890, opened 2016-08-18
If you go to the report for categories on a site with no categories the report picker dissappears


chriscct7:
In EDD's report section, switch the report type from earnings to categories on a site with no categories, and unexpectedly you'll find the report type picker has gone away

cklosowski:
ok @chriscct7 I think this should fix it up for us.

PR #4979, opened 2016-09-23
Load psudo-bulk-actions for reports when no categories exist #4890

cklosowski:
#4890


pippinsplugins:
Look good, merge once tests pass.

Issue #4934, opened 2016-09-10
Countries/States cannot be localized


schlessera:
Currently, the [country and state strings](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/country-functions.php) can not be localized apart from one exception: the [Spanish state strings](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/country-functions.php#L1567-L1621).

All country and state strings should be localizable, to allow the dropdown lists to be properly displayed in different locales.

michaelbeil:
@cklosowski and @pippinsplugins this one could be added to 2.6.7 I think. 

The changes look good.

sunnyratilal:
PR is at #4935

cklosowski:
2.6.7 is pretty locked, being that it's ready for deployment, I'd rather not add more to it. I'll milestone this for 2.6.8. Since it's been running stable overnight on EDD I'd prefer to not add more changes, even minor ones.

michaelbeil:
Fine by me.

cklosowski:
I'm reverting this. We need to come up with a more performant way to do this. We may want to look how WooCommerce is doing it by only loading the country states that are needed, instead of loading ALL of the states for all of the countries into the memory. the WooCommerce team saw a drastic improvement to their performance when doing this:

https://github.com/woothemes/woocommerce/blob/master/includes/class-wc-countries.php#L116

cklosowski:
For now the only way we have to do this without translations is for Store owners to use the filter in place to make the modifications they need to with the filter that's in place:

`return apply_filters( 'edd_shop_states', $states, $country );`

ashleyfae:
See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7798 for list of tickets.

robincornett:
@ashleyfae I would like to maybe start on this, at least just moving the lists to individual files and using `include` to retrieve them, and then see about figuring out more performant things. I feel like getting them set up that way, at a minimum, can at least begin helping us dig through the code? But not sure what milestone to tie it to.

ashleyfae:
I think it's probably better if we hold off on this until 3.1. You can start posting ideas here, but let's hold off on main development for now. You're welcome to assign it to yourself for post-3.0 if you want it though.

PR #4982, opened 2016-09-26
Revert "Make countries translatable #4934"

cklosowski:
Reverts easydigitaldownloads/easy-digital-downloads#4978

This is a performance hit I'm not ready to release yet. We need to do a better job of allowing translations for the states to be loaded dynamically (like they are doing in WooCommerce possibly? https://github.com/woothemes/woocommerce/blob/master/includes/class-wc-countries.php#L116)


cklosowski:
#4934

PR #4995, opened 2016-09-30
edd_get_random_download() now actually returns a value

nosegraze:
I'm assuming this function is indeed supposed to return something? Lol.


nosegraze:
Also while using this I realized that the `post_count` arg in the `edd_get_random_downloads()` query wasn't even working. Switched that to `numberposts`.

pippinsplugins:
Huh . . .I guess no one has actually used this before. LOL

pippinsplugins:
Thanks @nosegraze, we'll get this included in the next point release.

Issue #4987, opened 2016-09-27
record_fees clears fees in the session


ayvazj:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/370d71d60a090ee167acc7ade2ce71114ea53cde/includes/class-edd-fees.php#L336

Calling $payment->save(); calls record_fees which then clears the fees associated with the session.  If the payment fails the fees are lost.

pippinsplugins:
We should only clear the fees if the status is `complete`, `publish`, `failed`, or `revoked` I think.

pippinsplugins:
Ready for testing.

SDavisMedia:
The fix is working for me.

Replicated on `master` by adding a product to the cart, proceeding to checkout, and having a discount (negative fee) automatically applied by RCP Member Discounts ... then refreshing the checkout page after running the following in Debug Bar Console (where `22` is the ID of my product):

```php
$p = new EDD_Payment();
$p->add_download( 22 );
$p->save();
```

I lost the fee after refresh on `master`. The fee stuck after refresh on `issue/4987`. :+1:

PR #5023, opened 2016-10-11
Do not clear fees from session when status of payment is pending. #4987

pippinsplugins:
#4987


cklosowski:
@pippinsplugins just look over this quickly? Fixed it up so our unit tests would pass a bit, plus, just a good check since this data may have a filter run on it where someone could remove an array key (which would break, but it's possible)

pippinsplugins:
Looks good.

PR #5024, opened 2016-10-11
Support variable prices in product dropdown.

chriscct7:
As previously discussed and agreed upon in slack, this pr contains the 1 file of changes from HTML elements from pr #4895 that is the blocker to among other things Discounts Pro being able to fix several bugs, and FES/CFM functionality, as well as other extensions including Product Updates needing this functionality as well. This needs to go out in a point release as soon as possible.


chriscct7:
Note, this needs to get `one quick thing I noticed. 'data' => array( 'search-type' => 'customer' ) in the user_dropdown() method. That should be user for the search type and then the search function needs updated to return users instead of customers.` implemented into it. I won't have time to do that today, but someone else should be able to do that pretty quickly.

cklosowski:
running into a fatal right away:
`( ! ) Parse error: syntax error, unexpected 'elseif' (T_ELSEIF) in /app/public/wp-content/plugins/easy-digital-downloads/includes/class-edd-html-elements.php on line 110`

cklosowski:
This doesn't seem to be working correctly at all. Variations show on first load, but not all of them, and then when searching none of them show up:

screen shot 2016-10-11 at 9 36 41 am

screen shot 2016-10-11 at 9 36 22 am

screen shot 2016-10-11 at 9 36 35 am

chriscct7:
@cklosowski the search needs to be adjusted probably to add variations in

cklosowski:
This will resolve #3175. It does not implement it's usage yet in core, but I did use the discounts add/edit to verify it works. You can use this as the test too by simply adding in `'variations' => true` to any `product_drop_down()` call.

cklosowski:
@chriscct7 this is ready for you to test out with your DP addon. I think it meets all the needs for your release, even includes a couple helper functions to parse the data that's saved as well.

cklosowski:
Clearing milestone since we have #3175 on this milestone as an issue.

includes/class-edd-html-elements.php
```diff
@@ -71,26 +72,103 @@ public function product_dropdown( $args = array() ) {
...
+
 		} elseif ( is_numeric( $args['selected'] ) && $args['selected'] !== 0 ) {
-			if ( ! in_array( $args['selected'], $options ) ) {
-				$options[$args['selected']] = get_the_title( $args['selected'] );
+
+			if ( ! array_key_exists( $args['selected'], $options ) ) {
+
+				if ( strpos( $args['selected'], '_' ) !== false ) {
+
+					$pieces = explode( '_' , $item );
```
pippinsplugins: Reminder on this

includes/class-edd-html-elements.php
```diff
@@ -71,26 +72,103 @@ public function product_dropdown( $args = array() ) {
...
 			foreach( $args['selected'] as $item ) {
-				if( ! in_array( $item, $options ) ) {
-					$options[$item] = get_the_title( $item );
-				}
+
+				if ( ! array_key_exists( $item, $options ) ) {
+
+ 					if ( strpos( $item, '_' ) !== false ) {
+
+ 						$pieces = explode( '_' , $item );
```
pippinsplugins: Can this use our new parsing function please?

includes/class-edd-html-elements.php
cklosowski: Ah, yeah I'll update that quickly.

cklosowski:
@pippinsplugins changes have been made, and tested by @SDavisMedia and I.

pippinsplugins:
Merging!

Issue #5027, opened 2016-10-11
Add framework for the new SL beta support


ghost1227:
With https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/662 we've added beta support to Software Licensing. EDD core needs to be updated to implement handlers for beta support in our own products (and any products using the default licensing class).

ghost1227:
![screen shot 2016-10-11 at 1 30 00 pm](https://cloud.githubusercontent.com/assets/36433/19283277/d9f1e110-8fb6-11e6-8602-d0253e5048ec.PNG)

ghost1227:
Utilizing the framework is a two-part process. First, plugins wishing to utilize betas need to add themselves to the `has_beta_support` array:

```
function add_me_to_betas( $products ) {
    $products['my-slug'] = 'My Product';

    return $products;
}
add_filter( 'edd_has_beta_support', 'add_me_to_betas' );
```

Once added, the item will be added to the beta list on the Tools page. Checked items will be saved under the option `enabled_betas`. To actually enable beta notification, you have to pass an extra argument to `EDD_License` as follows:

```
new EDD_License( __FILE__, 'My Product', 'Product Version', Author', 'API URL', edd_is_beta_support_enabled( 'my-slug' ) );
```

ghost1227:
Forgot to tag this for testing.

pippinsplugins:
>>>Utilizing the framework is a two-part process. First, plugins wishing to utilize betas need to add themselves to the has_beta_support array

We need this to be handled entirely by core.

I would vote that all plugins that utilize our `EDD_License` class automatically get an opt-in option showing, otherwise we have to push an update for every extension before we can release betas for extensions, which is kind of silly.

ghost1227:
I'm not against that... tbh, it never occurred to me.

pippinsplugins:
Let's go ahead and go with that route please.

ghost1227:
@pippinsplugins how's that?

SDavisMedia:
@ghost1227 What exactly am I testing here? Outline testing steps for this please.

ghost1227:
[...]

PR #5028, opened 2016-10-11
Add framework for beta support

ghost1227:
Issue #5027

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -68,7 +68,7 @@ function edd_get_tools_tabs() {
 	$tabs['general']       = __( 'General', 'easy-digital-downloads' );
 	$tabs['api_keys']      = __( 'API Keys', 'easy-digital-downloads' );
 
-	if( count( edd_has_beta_support() ) > 0 ) {
+	if( count( edd_get_beta_enabled_extensions() ) > 0 ) {
 		$tabs['betas'] = __( 'Beta Support', 'easy-digital-downloads' );
 	}
 
@@ -256,7 +256,7 @@ add_action( 'edd_tools_tab_api_keys', 'edd_tools_api_keys_display' );
 /**
  * Display beta opt-ins
  *
- * @since       2.6.9
+ * @since       2.6.11
  * @return      void
  */
 function edd_tools_betas_display() {
@@ -264,7 +264,7 @@ function edd_tools_betas_display() {
 		return;
 	}
 
-	$has_beta      = edd_has_beta_support();
+	$has_beta      = edd_get_beta_enabled_extensions();
 	$enabled_betas = edd_get_option( 'enabled_betas', array() );
 
 	do_action( 'edd_tools_betas_before' );
@@ -303,26 +303,26 @@ add_action( 'edd_tools_tab_betas', 'edd_tools_betas_display' );
 
 
 /**
- * Return an array of all products with beta support
+ * Return an array of all extensions with beta support
  *
- * Products should be added as 'product-slug' => 'Product Name'
+ * Extensions should be added as 'extension-slug' => 'Extension Name'
  *
- * @since       2.6.9
- * @return      array $products The array of products
+ * @since       2.6.11
+ * @return      array $extensions The array of extensions
  */
-function edd_has_beta_support() {
-	return apply_filters( 'edd_has_beta_support', array() );
+function edd_get_beta_enabled_extensions() {
+	return apply_filters( 'edd_beta_enabled_extensions', array() );
 }
 
 
 /**
- * Check if a given product has beta support enabled
+ * Check if a given extensions has beta support enabled
  *
- * @since       2.6.9
- * @param       string $slug The slug of the product to check
+ * @since       2.6.11
+ * @param       string $slug The slug of the extension to check
  * @return      bool True if enabled, false otherwise
  */
-function edd_is_beta_support_enabled( $slug ) {
+function edd_extension_has_beta_support( $slug ) {
 	$enabled_betas = edd_get_option( 'enabled_betas', array() );
 	$return        = false;
 
@@ -337,7 +337,7 @@ function edd_is_beta_support_enabled( $slug ) {
 /**
  * Save enabled betas
  *
- * @since       2.6.9
+ * @since       2.6.11
  * @return      void
  */
 function edd_tools_enabled_betas_save() {
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -67,6 +67,11 @@ function edd_get_tools_tabs() {
 	$tabs                  = array();
 	$tabs['general']       = __( 'General', 'easy-digital-downloads' );
 	$tabs['api_keys']      = __( 'API Keys', 'easy-digital-downloads' );
+
+	if( count( edd_has_beta_support() ) > 0 ) {
+		$tabs['betas'] = __( 'Beta Support', 'easy-digital-downloads' );
+	}
+
 	$tabs['system_info']   = __( 'System Info', 'easy-digital-downloads' );
 	$tabs['import_export'] = __( 'Import/Export', 'easy-digital-downloads' );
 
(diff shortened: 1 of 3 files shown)


includes/admin/tools.php
```diff
@@ -249,6 +254,111 @@ function edd_tools_api_keys_display() {
...
+
+/**
+ * Return an array of all products with beta support
+ *
+ * Products should be added as 'product-slug' => 'Product Name'
+ *
+ * @since       2.6.9
+ * @return      array $products The array of products
+ */
+function edd_has_beta_support() {
```
mindctrl: I think this function name should be something like `edd_get_beta_enabled_products()` or something like that. It more accurately describes what it does.

includes/admin/tools.php
```diff
@@ -249,6 +254,111 @@ function edd_tools_api_keys_display() {
...
+}
+add_action( 'edd_tools_tab_betas', 'edd_tools_betas_display' );
+
+
+/**
+ * Return an array of all products with beta support
+ *
+ * Products should be added as 'product-slug' => 'Product Name'
+ *
+ * @since       2.6.9
```
mindctrl: The @since version is wrong here and on the other functions.

includes/admin/tools.php
```diff
@@ -249,6 +254,111 @@ function edd_tools_api_keys_display() {
...
+
+
+/**
+ * Check if a given product has beta support enabled
+ *
+ * @since       2.6.9
+ * @param       string $slug The slug of the product to check
+ * @return      bool True if enabled, false otherwise
+ */
+function edd_is_beta_support_enabled( $slug ) {
```
mindctrl: This function name sounds like it's checking if the overall feature is enabled instead of checking for a particular product. I think it should be something like `edd_is_download_beta_enabled()` or something similar.

includes/admin/tools.php
pippinsplugins: How about `edd_get_beta_enabled_extensions()` to help make it more clear this is for add-on plugins and not store products.

includes/admin/tools.php
pippinsplugins: Good idea. I like `edd_extension_has_beta_support( $slug )`

includes/admin/tools.php
cklosowski: Agree with this.

includes/admin/tools.php
cklosowski: Yeah, we failed to get it in `2.6.9`, Just needs a quick update to `2.6.11`

includes/class-edd-license-handler.php
```diff
@@ -120,6 +118,8 @@ private function hooks() {
 
 		add_action( 'in_plugin_update_message-' . plugin_basename( $this->file ), array( $this, 'plugin_row_license_missing' ), 10, 2 );
 
+		// Register plugins for beta support
+		add_filter( 'edd_has_beta_support', array( $this, 'register_beta_support' ) );
```
cklosowski: Actually this is the same filter that exists in the proposed changes to `edd_has_beta_support()` function, I would think since this is an array of extensions that have support it should read something like `edd_beta_enabled_extensions`

includes/admin/tools.php
```diff
@@ -249,6 +254,111 @@ function edd_tools_api_keys_display() {
...
+	if( ! wp_verify_nonce( $_POST['edd_save_betas_nonce'], 'edd_save_betas_nonce' ) ) {
+		return;
+	}
+
+	if( ! current_user_can( 'manage_shop_settings' ) ) {
+		return;
+	}
+
+	if( ! empty( $_POST['enabled_betas'] ) ) {
+		edd_update_option( 'enabled_betas', $_POST['enabled_betas'] );
```
pippinsplugins: This should be sanitized before being saved.

includes/class-edd-license-handler.php
```diff
@@ -127,11 +129,13 @@ private function hooks() {
...
 	 */
 	public function auto_updater() {
+		$betas = edd_get_option( 'enabled_betas', array() );
 
 		$args = array(
 			'version'   => $this->version,
 			'license'   => $this->license,
-			'author'    => $this->author
+			'author'    => $this->author,
+			'beta'      => ( array_key_exists( $this->item_shortname, $betas ) ? true : false )
```
pippinsplugins: This should use our `edd_extension_has_beta_support()` helper function

includes/admin/tools.php
ghost1227: Given that the `enabled_betas` is an array, I'm not entirely sure how best to sanitize it. Can you provide some guidance?

pippinsplugins: Use `array_map`

ghost1227: I've never really had a good grasp of `array_map`... guess it's time to do some studying!

pippinsplugins: Allows you to run each value in the array through a call back function. Very useful.

ghost1227: Ok so I get what it's doing, but is there an inbuilt way of enforcing boolean or do we need to write a new callback function?

includes/class-edd-license-handler.php
cklosowski: Fixed.

includes/admin/tools.php
cklosowski: There are a couple challenges actually. The first being that we would need to build a callback function since there isn't an existing function that allows us to sanitize that type of data easily. The best is a `filter_var` that we can use `filter_var($string, FILTER_VALIDATE_BOOLEAN);`

The 2nd challenge is there doesn't appear to be a `value` on the input:
`									<input type="checkbox" name="enabled_betas[<?php echo esc_attr( $slug ); ?>]" id="enabled_betas[<?php echo esc_attr( $slug ); ?>]"<?php echo $checked; ?> />`

Because of that we're relying on the HTML weirdness that is, a checkbox exists in the `$_POST` with or without it. We should probably add in a `value` attribute of `1`, and then write a specific callback function that uses the above `filter_var` to give us the proper data for the save.

cklosowski: Pushed a change for this.

cklosowski:
Ok, @pippinsplugins you're review is still pending. Want anymore changes on this?

includes/admin/tools.php
```diff
@@ -67,6 +67,11 @@ function edd_get_tools_tabs() {
 	$tabs                  = array();
 	$tabs['general']       = __( 'General', 'easy-digital-downloads' );
 	$tabs['api_keys']      = __( 'API Keys', 'easy-digital-downloads' );
+
+	if( count( edd_get_beta_enabled_extensions() ) > 0 ) {
+		$tabs['betas'] = __( 'Beta Support', 'easy-digital-downloads' );
```
pippinsplugins: I think this should be named `Beta Versions`

includes/admin/tools.php
```diff
@@ -249,6 +254,124 @@ function edd_tools_api_keys_display() {
...
+	}
+
+	$has_beta      = edd_get_beta_enabled_extensions();
+	$enabled_betas = edd_get_option( 'enabled_betas', array() );
+
+	do_action( 'edd_tools_betas_before' );
+	?>
+
+	<div class="postbox edd-beta-support">
+		<h3><span><?php _e( 'Enable Beta Support', 'easy-digital-downloads' ); ?></span></h3>
```
pippinsplugins: This should be `Enable Beta Versions`

includes/admin/tools.php
```diff
@@ -249,6 +254,124 @@ function edd_tools_api_keys_display() {
...
+		<div class="inside">
+			<p><?php _e( 'Checking any of the below checkboxes will opt you in to receive pre-release update notifications. You can opt-out at any time. Pre-release updates do not install automatically, you will still have the opportunity to ignore update notifications.', 'easy-digital-downloads' ); ?></p>
+			<form method="post" action="<?php echo admin_url( 'edit.php?post_type=download&page=edd-tools&tab=betas' ); ?>">
+				<table class="form-table edd-beta-support">
+					<tbody>
+						<?php foreach( $has_beta as $slug => $product ) : ?>
+							<tr>
+								<?php $checked = ( array_key_exists( $slug, $enabled_betas ) ? ' checked="checked"' : '' ); ?>
```
pippinsplugins: This should use the helper function to determine if beta support is enabled

includes/class-edd-license-handler.php
```diff
@@ -462,6 +466,20 @@ public function plugin_row_license_missing( $plugin_data, $version_info ) {
 		}
 
 	}
+
+	/**
+	 * Adds this plugin to the beta page
+	 *
+	 * @access  public
+	 * @param   array $products
+	 * @since   2.6.10
```
pippinsplugins: Version number is wrong here.

PR #3052, opened 2015-02-03
Dashicon as EDD media button

corvannoorloos:
I originally just wanted to update the EDD media button to make use of the WP Dashicon font, but also came across some inline styles and spaces used as indentation.

I hope it's ok I've added these in the same request.

38-new-post

38-new-post-mobile

I've tested the Add New Post, Add New Download and Downloads screen running WordPress 3.7 and 3.8 and 4.2 on Windows 7, Ubuntu 14.04.1 and iOS 8.

PR #5032, opened 2016-10-12
Move Download CPT icon to registration arguments

nosegraze:
I think this is better because then it's easier and more intuitive to change the menu icon.

And should #3052 be revisited? It seems like the effort to move to SVG has been basically dead for over a year unless there's more discussion I haven't found.

### DIFF
--- a/includes/post-types.php
+++ b/includes/post-types.php
@@ -58,6 +58,7 @@ function edd_setup_edd_post_types() {
 		'show_ui'            => true,
 		'show_in_menu'       => true,
 		'query_var'          => true,
+		'menu_icon'          => 'dashicons-download',
 		'rewrite'            => $rewrite,
 		'capability_type'    => 'product',
 		'map_meta_cap'       => true,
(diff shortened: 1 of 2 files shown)


michaelbeil:
Could you share a screenshot of the changes you are proposing?

nosegraze:
Well visually it wouldn't look any different so I'm not sure what you want me to screenshot.

Basically right now the admin menu icon (the download icon) is being added via CSS in the header. This makes it kind of a pain to change. I'm suggesting removing that CSS and utilizing the `menu_item` argument in the post type registration instead: https://codex.wordpress.org/Function_Reference/register_post_type#menu_icon Putting the dashicon name there allows it to be changed via the CTP arguments filter.

michaelbeil:
This would be a good addition. Gonna make an issue for it so we can track it accordingly.

michaelbeil:
Regarding the SVG issue, why don't you bring that question up in that issue?

nosegraze:
Will do!

cklosowski:
Lining issue with PR #5033

Issue #5006, opened 2016-10-04
Exported CSVs show "Complete" status as "publish"


mintplugins:
When exporting a CSV of Payments, the "status" column will show "publish" for posts. When exporting, the verbiage is to export "Complete" payments. Should we show "Complete" as the status instead of "publish"?

<img width="150" alt="screen shot 2016-10-04 at 2 23 27 pm" src="https://cloud.githubusercontent.com/assets/1638128/19087168/18da04e8-8a3f-11e6-9322-c51d45a2867f.png">

jaredatch:
Seemingly at the same time this was introduced, Date is no longer properly exported. All Date values are empty.

pippinsplugins:
Dates have been fixed in #4994

jaredatch:
Whoops, sorry about that!

cklosowski:
There's two sides of this.
1) Do we make it human readable and make it `complete` into which, the importer will simply change it to `publish`.
2) Do we keep it `publish` so that it's matching whatever the end result of the import is going to be.

Thoughts @easydigitaldownloads/core-devs

pippinsplugins:
I vote for `complete`

michaelbeil:
complete seems better.

sanketio:
@pippinsplugins 

I have sent PR ( https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5042 ). Let me know if there is any change in that.

pippinsplugins:
PR looks fine to me.

sanketio:
Great. :)

SDavisMedia:
Exporting on the PR branch producing "complete" for all the payment statuses is importing just fine for me on `master`. :+1:

pippinsplugins:
:+1:

SDavisMedia:
I'm not sure what else this needs.

pippinsplugins:
@SDavisMedia nothing else :)

PR #5042, opened 2016-10-13
Issue #5006

sanketio:
Status change from `publish` to `complete` when exporting payment history


includes/admin/reporting/export/class-batch-export-payments.php
```diff
@@ -224,7 +224,7 @@ public function get_data() {
 					'currency'     => $payment->currency,
 					'ip'           => $payment->ip,
 					'mode'         => $payment->get_meta( '_edd_payment_mode', true ),
-					'status'       => $payment->status
+					'status'       => ( 'publish' === $payment->status ) ? __( 'complete', 'easy-digital-downloads' ) : $payment->status
```
pippinsplugins: I don't believe this should be translatable. Allowing it to be translated will mess with the ability to import it.

sanketio: So should I remove translatable text and instead use direct `complete`?

michaelbeil:
Issue #5006.

PR #5043, opened 2016-10-13
Cannot rely on strict checks #3175

cklosowski:
#3175


cklosowski:
@pippinsplugins in working with Discounts pro 1.5, I found we cannot strict type check this.

Issue #4964, opened 2016-09-20
Allow exporting a download history of a specific product


cklosowski:
It's currently not possible to export the file download history of a specific download. We need to support this.

sanketio:
Hi @cklosowski 

I have sent a PR ( https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5045 ). Please let me know if you need any change in that.

cklosowski:
Working up a new PR, I've merged the current work into `issue/4964` to add a few enhancements.

cklosowski:
Ok, looking at the changes needed in order to support the download ID within the file name, we'll have to rewrite the export classes a bit. I'm going to make a future enhancement for that.

mintplugins:
@cklosowski Working properly for me in my tests. Should we include the payment ID attached to the file download in the CSV as well?

SDavisMedia:
Working well on my end. :+1:

cklosowski:
@mintplugins see my above comment. To do it is going to take an extensive rewrite, I've opened an issue to make the file names exported to be more flexible #5097

mintplugins:
@cklosowski It's above my pay grade :P But I'm confused as to why the filename needs to change for us to include a payment ID column in the exported CSV.

cklosowski:
Oh sorry, I misread your comment @mintplugins. Forgive me as I was mobile at the time. I'd say let's open another ticket and milestone that for 2.8 for now. Since it's out of the scope of the current request for this issue.

PR #5045, opened 2016-10-14
Issue #4964

sanketio:
- Option to choose product in exporting a download history


michaelbeil:
Issue https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4964.

cklosowski:
Going to pull this into an issue branch, I think we should add the download ID into the file name we're saving as well. I'll merge this into an issue branch and add that feature. This will keep the commit history in tact for you @sanketio

Issue #5019, opened 2016-10-10
JS for variable pricing option needs to be refined


chriscct7:
Still investigating a different bug related to this, where the JS was doing the reverse for `Enable variable prices` on a new download screen. However, regardless of whether that bug becomes replicatable or not, it showed a flaw in how the JS checks before it does the action. The JS should not have permitted for that behavior to be reversed. If checking that box it should open if and only if not open and unchecking close if and only if not closed.

Bug still in testing showing this reversal: https://www.dropbox.com/s/k28sx6260u6hsp0/2016-10-10_11-27-13.MP4?dl=0

chriscct7:
To replicate the bug, open the new download screen and the second the dom starts registering, immediately tick on the variable pricing box while it's still loading. Wait a few seconds for the page to actually finish, and then use the checkbox. And that's how it gets reversed.

chriscct7:
This needs to be tested on other pages we use JS based toggle style events like the payment page, for the same type of thing

cklosowski:
@chriscct7 so the issue is checking the box and DOM completion can get the toggle code into a mixed up state? If that's the case we can probably just focus on not using `toggle` and being specific about our cases.

chriscct7:
Yep. There's s demo of this on the attached video

michaelbeil:
I could not replicate the issue that @chriscct7 mentioned, but the functionality that @cklosowski introduced in #5061 works great.

chriscct7:
In order to replicate you have to move quickly and do it before the page finishes loading. Adding a plugin that adds more JS like Yoast SEO or something like that might help make it easier to replicate. I know @SDavisMedia replicated it in Slack

SDavisMedia:
I can definitely replicate this. Just gotta be fast. I like @cklosowski's solution.

michaelbeil:
That explains it since my screen is too short to get down to the meta box fast enough 😄  as I do not have my monitors set up yet.

SDavisMedia:
Easily replicated again on `master`. Confirmed fixed on `issue/5019`. :+1:

PR #5061, opened 2016-10-19
Change jQuery toggle() to show/hide to account for DOM parsing lag #5019

cklosowski:
#5019

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -213,7 +213,13 @@ jQuery(document).ready(function ($) {
 
 		prices : function() {
 			$( document.body ).on( 'change', '#edd_variable_pricing', function(e) {
-				$( '.edd_pricing_fields,.edd_repeatable_table .pricing' ).toggle();
+				var checked = $(this).is(':checked');
+				var target  = $( '.edd_pricing_fields,.edd_repeatable_table .pricing' );
+				if ( checked ) {
+					target.show();
+				} else {
+					target.hide();
+				}
 			});
 		},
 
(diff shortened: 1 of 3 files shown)


includes/admin/discounts/add-discount.php
```diff
@@ -57,8 +57,8 @@
 				</th>
 				<td>
 					<input type="text" id="edd-amount" name="amount" value="" style="width: 40px;"/>
-					<p class="description edd-amount-description" style="display:none;"><?php printf( __( 'Enter the discount amount in %s', 'easy-digital-downloads' ), edd_get_currency() ); ?></p>
-					<p class="description edd-amount-description"><?php _e( 'Enter the discount percentage. 10 = 10%', 'easy-digital-downloads' ); ?></p>
+					<p class="description edd-amount-description flat" style="display:none;"><?php printf( __( 'Enter the discount amount in %s', 'easy-digital-downloads' ), edd_get_currency() ); ?></p>
```
chriscct7: Just thinking that flat might not be the best idea for a class name given some plugins might globally load css, and use flat for like flat design or something like that. Maybe `discount-type-flat` would be better?

assets/js/admin-scripts.js
```diff
@@ -773,8 +789,10 @@ jQuery(document).ready(function ($) {
 		type_select : function() {
 
 			$('#edd-edit-discount #edd-type, #edd-add-discount #edd-type').change(function() {
-
-				$('.edd-amount-description').toggle();
+				var val = $(this).val();
+				console.log(val);
```
chriscct7: Should this be removed?

includes/admin/discounts/add-discount.php
cklosowski: @chriscct7 I'll change `flat` to be a bit more defined.

assets/js/admin-scripts.js
cklosowski: Will remove this.

PR #5074, opened 2016-10-21
Update the current time to get GMT

SGr33n:
Hi,
If you live in a UTC+0 timezone this doesn't affect you, but if you live in a UTC+2 i.e. timezone, without getting GMT time, the first time the cron will start after two hours.

With the use of the true parameter the cron will start immediately.

### DIFF
--- a/includes/class-edd-cron.php
+++ b/includes/class-edd-cron.php
@@ -70,7 +70,7 @@ class EDD_Cron {
 	 */
 	private function weekly_events() {
 		if ( ! wp_next_scheduled( 'edd_weekly_scheduled_events' ) ) {
-			wp_schedule_event( current_time( 'timestamp' ), 'weekly', 'edd_weekly_scheduled_events' );
+			wp_schedule_event( current_time( 'timestamp', true ), 'weekly', 'edd_weekly_scheduled_events' );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


chriscct7:
I think this is exactly what is intended if I'm understanding your comment. The idea of the cron is that is will always run at 2am local time regardless of where you are in the world,.

SGr33n:
Sorry, I was not clear. wp_schedule_event first parameter is $timestamp, and this is the codex about it:

> (integer) (required) The time you want the event to occur. This must be in a UNIX timestamp format. **WP cron uses UTC/GMT time, not local time**. Use time(), which is always GMT in WordPress. (current_time( 'timestamp' ) is local time in WordPress.)
> 
>     Default: None 

The current version of EDD uses current_time( 'timestamp' ) to get the first argument, but current_time( 'timestamp' ) is the local time, so to get the UTC/GMT, as requested by the function wp_schedule_event you shuld use, instead, current_time( 'timestamp' , true ).

cklosowski:
I'm not sure this needs a 2.6.13. It could be fine in 2.7. Wether this runs at midnight or 2am isn't seemingly important, since all this is, is a scheduled task to run once a day. There is nothing that time sensitive about it running 2 hours late, IMHO. @pippinsplugins ?

pippinsplugins:
We might as well make it correct, so as long as this change is indeed right, let's do it.

michaelbeil:
Tracking issue https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5196.

sunnyratilal:
Milestone cleared and set on the tracking issue.

sunnyratilal:
Tested and confirmed it's working - thanks @SGr33n

Issue #5077, opened 2016-10-21
CSV import tool not importing Download Files


michaelbeil:
When mapping the Download Files to the Files CSV column, the Download Files (URLs) do not get imported to the download file meta box.

michaelbeil:
Just had another ticket come in regarding this.

pippinsplugins:
@michaelbeil could you link to the tickets here please?

michaelbeil:
1: https://secure.helpscout.net/conversation/266646277/45061/?folderId=725711
2: https://secure.helpscout.net/conversation/267628201/45192/?folderId=725711

pippinsplugins:
This was broken with https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4969

The logic that checks for a file condition is wrapped around the logic that sets up the file URL, so the array sent to `update_post_meta()` is empty.

pippinsplugins:
@michaelbeil ready for testing!

michaelbeil:
Tests pass here!

cklosowski:
Something isn't right here...it's again starting the price ID index at `0` on import instead of `1`, so it's broken the file import with conditions, going to fix it up quick.

cklosowski:
@michaelbeil is this still working for you with price conditions?

michaelbeil:
What did you do to test?

cklosowski:
My setup to export...then try and import making sure it still works:
Variations:
![screen shot 2016-10-27 at 11 44 56 am](https://cloud.githubusercontent.com/assets/1390124/19780660/d9621112-9c3a-11e6-92d4-9264229fb78d.png)

Files (focus on the conditions):
![screen shot 2016-10-27 at 11 44 50 am](https://cloud.githubusercontent.com/assets/1390124/19780674/e2fbb908-9c3a-11e6-914c-b93b108c3595.png)

michaelbeil:
Getting an issue where the starting price ID index is at 0 on import instead of 1. ![screen shot 2016-10-27 at 6 58 48 pm](https://cloud.githubusercontent.com/assets/4752949/19781792/5b7d7c22-9c8b-11e6-8878-11d60a1f6511.png)

cklosowski:
@michaelbeil verify your price IDs match too? Does the price IDs column of your export site and import site match?

michaelbeil:
[...]

PR #5083, opened 2016-10-25
Set file URL during import even if no condition value is available. #5077

pippinsplugins:
#5077


michaelbeil:
Restarted one test that failed. This should be good to merge once finished.

PR #5093, opened 2016-10-26
Fix #5092

### LINKED ISSUE #5092: Array to string conversion
## Steps to reproduce

- Install Fresh WP
- Enable `WP_DEBUG`
- Install EDD (stable or master trunk)
- Navigate to Downloads > Settings > Taxes
- Warnings under **Tax Rates** data table
screen shot 2016-10-26 at 20 54 14

## Tested with:
EDD: Stable (2.6.10) and latest master trunk
WP: 4.6.1

rahilwazir:
By default `selected` key in [`EDD_HTML_Elements::select`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L449) method
args is array which fails when there is no value provided on
call method.

Example: `includes/admin/settings/register-settings.php#1924`

Passing an empty value to `selected` key when no rates are defined.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -1926,6 +1926,7 @@ function edd_tax_rates_callback($args) {
 					echo EDD()->html->select( array(
 						'options'          => edd_get_country_list(),
 						'name'             => 'tax_rates[0][country]',
+						'selected'         => '',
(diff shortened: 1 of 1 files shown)


SDavisMedia:
#5092

PR #5095, opened 2016-10-26
Fix last text domain replacement

joseluis:
I was creating an online store and I was wondering why this string wasn't getting translated like the others. This one seems to have slipped through the cracks in the last text-domain change operation.


pippinsplugins:
Thanks!

Issue #5094, opened 2016-10-26
Block self-pings in tracking check ins


pippinsplugins:
We need to ensure that the tracking class cannot try and ping our own site. Fairly certain that has been the culprit in several server melt downs we've seen recently immediately following an EDD core update.

cklosowski:
We shouldn't be sending a checkin to ourselves at all: https://cl.ly/0H2r0G2o442p/Screen%20Shot%202016-10-26%20at%204.28.42%20PM.png

cklosowski:
Seems like with this disabled, we aren't pinging ourselves, even on upgrades...so that makes me think it's not the case.

pippinsplugins:
@cklosowski let's put the logic in any way to be safe.

cklosowski:
@pippinsplugins I added the logic to check if it's ourselves, but also added a filter so we could stop affwp, rcp, and pp from checking in.

pippinsplugins:
I want those two to track so all good :+1:

cklosowski:
Cool. Well we can leave it in there anyway, allows a quick way to disable it programattically if need be for any reason. Good to merge?

pippinsplugins:
Good to merge.

PR #5103, opened 2016-10-27
Don't call checkin on self, and a filter for our other sites #5094

cklosowski:
#5094

### DIFF
--- a/includes/admin/tracking.php
+++ b/includes/admin/tracking.php
@@ -114,6 +114,12 @@ class EDD_Tracking {
 	 */
 	public function send_checkin( $override = false, $ignore_last_checkin = false ) {
 
+		$home_url = trailingslashit( home_url() );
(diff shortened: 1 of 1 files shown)


includes/admin/tracking.php
```diff
@@ -114,6 +114,12 @@ private function setup_data() {
 	 */
 	public function send_checkin( $override = false, $ignore_last_checkin = false ) {
 
+		$home_url = trailingslashit( home_url() );
+		// Allows us to stop our own site from checking in, and a filter for our additional sites
+		if ( $home_url === 'https://easydigitaldownloads.com/' || apply_filters( 'edd_disable_tracking_checkin', false ) ) {
```
chriscct7: should this respect the override param?

pippinsplugins: No because we don't ever want to ping ourselves.

Issue #5096, opened 2016-10-26
Update the Insert Download media button icon


SDavisMedia:
<img width="500" alt="screenshot 2016-10-26 10 31 11" src="https://cloud.githubusercontent.com/assets/2359131/19737202/560c31e0-9b67-11e6-877e-4903320dc1bb.png">

Should probably be using the dashicon download icon. https://cl.ly/2B3e272o3Y23

michaelbeil:
Related: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/3052

michaelbeil:
Looks great here:
![screen shot 2016-10-27 at 2 35 16 pm](https://cloud.githubusercontent.com/assets/4752949/19767338/a3eb2eca-9c52-11e6-9fe3-cb25f2494b55.png)

sunnyratilal:
Unit tests need updating here because our CSS unit tests are failing in #5102 based on the changes introduced in df38b126a8b633c9b8ac61f46c3eecb9ffb18346

sunnyratilal:
Unit tests have been updated - waiting on Travis to pass.

sunnyratilal:
Travis tests have passed. Unit tests are hence fixed. PR is good to merge now.

PR #5106, opened 2016-10-27
Remove global variables from edd_admin_downloads_icon

chriscct7:
We don't use either of these in the actual function


chriscct7:
Chris K likes to try to beat me by 0.00002 seconds sometimes. This time he won. It's in #5096's pr

PR #5138, opened 2016-11-02
CSS Class Argument for Settings API Fields

d4mation:
Needed to add extra Class Names to Settings API Fields for CSS/JavaScript needs. This should allow that for every field, even the more specific ones like the callback for Tax Rates.

The `field_class` argument accepts both Strings and a basic Array.

Examples

`$args['field_class'] = super awesome class-name;`

```
$args['field_class'] = array(
    'super',
    'awesome',
    'class-name'
);
```

In cases where Chosen or something else would normally place a Class Name onto the Field this still works just fine. The additional Class Names just become appended to the String that gets created.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -198,6 +198,7 @@ function edd_register_settings() {
 				    'faux'          => false,
 				    'tooltip_title' => false,
 				    'tooltip_desc'  => false,
+				    'class'         => '',
 				) );
 
 				add_settings_field(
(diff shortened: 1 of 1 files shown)


cklosowski:
We can improve this a bit I think. We seem to be doing the same logic each time on all the types, by looking if the class is an array or string, and then imploding, etc.

We should move that into a private method of the HTML elements class, and within it, we need to be sure to use `sanitize_html_class` on each individual class.

Something like:
```
 		if ( is_string( $args['field_class'] ) ) {
 			$class = sanitize_html_class( $args['field_class'] );
 		} else if ( is_array( $args['field_class'] ) ) {
 			$classes = array_values( array_map( 'sanitize_html_class', $args['field_class'] );
 			$class = implode( ' ', array_unique( $classes ) );
 		}
         
```

Adding that to a method will make it so we don't have to repeat the logic as often and they'll always be sanitized.

Issue #5140, opened 2016-11-03
_edd_download_limit isn't checked for existence before accessing it, leading to fatal error


chriscct7:
In the save routine for the downloads metabox, _edd_download_limit isn't checked for existence before attempting to access it on line 119

chriscct7:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5141

pippinsplugins:
@chriscct7 can you give me steps to replicate the fatal?

chriscct7:
Removing the meta box from the page should do it

chriscct7:
(and then trying to save the download)

sunnyratilal:
Tested and working fine. Thanks @chriscct7

sunnyratilal:
Merged into `master`

PR #5141, opened 2016-11-03
Check _edd_download_limit exists before trying to access it in superglobal on metabox save routine

chriscct7:
Solves #5140 
In the save routine for the downloads metabox, _edd_download_limit isn't checked for existence before attempting to access it on line 119


cklosowski:
Removing milestone, since the attached issue has one.

PR #5161, opened 2016-11-05
Fix link to iOS addon

chriscct7:
The link to the iOS addon is incorrect. This simply fixes that typo


sunnyratilal:
Good catch, thanks @chriscct7

Issue #5148, opened 2016-11-04
[Hook Request] edd_profile_editor_before_name


rubengc:
Currently there is no form to add anything before name form in profile editor

I request this hook because I need to add an avatar field at top of this form

michaelbeil:
That seems like a definite possibility to add that hook. Would you care to send us a PR @rubengc?

SDavisMedia:
Agreed. Just _below_ this line, I'd say: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/templates/shortcode-profile-editor.php#L30

rubengc:
Added, on merge, I will close it

michaelbeil:
Added a PR: #5174

SDavisMedia:
@michaelbeil We already have a PR and ongoing discussion from the person contributing. https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5162

michaelbeil:
Oh, I did not see that. Closing in favor of #5162.

pippinsplugins:
PR on this looks fine. Moving to 2.6.12.

Issue #3904, opened 2015-10-12
Profile editor HTML structure should match the rest of EDD


spencerfinnell:
Super strange mix of markup in the profile editor form:

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/fbe20f79334ef6bd21f6a9810e8e5977bef116df/templates/shortcode-profile-editor.php#L28

Would be nice if each field was wrapped in a `p` individually like the login and register forms:

https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/fbe20f79334ef6bd21f6a9810e8e5977bef116df/templates/shortcode-register.php

pippinsplugins:
:+1:

pippinsplugins:
Before:
![screen shot 2016-08-12 at 11 32 35 am](https://cloud.githubusercontent.com/assets/1034109/17629539/7e803b8e-6080-11e6-9c71-6ed7f4952e79.png)

After:
![screen shot 2016-08-12 at 11 31 19 am](https://cloud.githubusercontent.com/assets/1034109/17629517/63488ef2-6080-11e6-9770-e880c855e84e.png)

Thoughts?

pippinsplugins:
See exact changes https://github.com/easydigitaldownloads/easy-digital-downloads/commit/80e633fe0bb112fd93abd7d264782e7a60f1b390

spencerfinnell:
Awesome, helps a lot. We'll remove our CSS shims in the next release of Marketify.

sumobi:
Great improvement 👍

SDavisMedia:
I'll have to update all of the themes but it's worth it. :+1:

pippinsplugins:
@spencerfinnell @sumobi @SDavisMedia happy with this PR? I'd like to merge it as soon as possible.

pippinsplugins:
Vendd before:
![screen shot 2016-10-14 at 11 25 13 am](https://cloud.githubusercontent.com/assets/1034109/19394964/e8447056-9200-11e6-9bd4-1f5089cd18a7.png)

Vendd after:
![screen shot 2016-10-14 at 11 25 45 am](https://cloud.githubusercontent.com/assets/1034109/19394986/fa0dfd02-9200-11e6-8d94-373cfa1f62cd.png)

Will definitely need some CSS updates.

SDavisMedia:
[...]

Issue #3568, opened 2015-07-05
<legend> output in forms


justintadlock:
It is my understanding that a `<fieldset>` should only have one `<legend>` and that should be the first child element.  Legends are currently wrapped in a `<span>`, so they're not the first child.

The fieldset and legend form a pair.  The legend is the "caption" for a particular fieldset.  You can have multiple `<fieldset>` elements, each with their `<legend>`, within a single `<form>` (e.g., would do this in the profile editor form).

I'm just bringing this up because how the `<legend>` element is done in the plugin messes up the design in every one of my themes.

I see two good options:

* Remove the `<span>` wrapped on `<legend>` elements and create extra `<fieldset>` elements in forms with multiple `<legend>` elements.
* Drop the `<legend>` element altogether and opt for an `<h2>` or something like that.

I tried to do a little research on this because I've never seen legends handled like this before. Here's some specs:

* http://www.w3.org/TR/html5/forms.html#the-legend-element
* http://www.w3.org/TR/html5/forms.html#the-fieldset-element
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

pippinsplugins:
Thanks for the report @justintadlock!

Tagged for 2.4.3.

pippinsplugins:
I vote we drop the span tags as that will have the least impact on existing styles.

pippinsplugins:
@SDavisMedia could you give this a quick test in your themes to make sure no styling is broken?

SDavisMedia:
It's definitely making a difference. Gonna look at some CSS now to see what it'll take to correct.

pippinsplugins:
I didn't see any differences in Twenty Fifteen. Also confirmed that EDD core doesn't style off of the `span` tags.

SDavisMedia:
I'm seeing Twenty Fifteen changes, the same changes I see in Vendd. I don't style off the `span` either. There's something going on here. What does the "Personal Info" legend text look like for you? Like this? http://glui.me/?i=85aki7ys6m7ktyl/2015-08-12_at_2.54_PM.png/

pippinsplugins:
Maybe it's coming from theme styles.

I'll check again.

pippinsplugins:
Punting to 2.5 since there are apparent style changes. Can't make style changes in a point release.

SDavisMedia:
[...]

PR #5162, opened 2016-11-05
Added edd_profile_editor_before_name action to profile editor

rubengc:
#5148

### DIFF
--- a/templates/shortcode-profile-editor.php
+++ b/templates/shortcode-profile-editor.php
@@ -28,8 +28,8 @@ if ( is_user_logged_in() ):
 
 	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post">
 		<fieldset>
-			<?php do_action( 'edd_profile_editor_before_name' ); ?>
 			<span id="edd_profile_name_label"><legend><?php _e( 'Change your Name', 'easy-digital-downloads' ); ?></legend></span>
+			<?php do_action( 'edd_profile_editor_before_name' ); ?>
 			<p id="edd_profile_name_wrap">
 				<label for="edd_first_name"><?php _e( 'First Name', 'easy-digital-downloads' ); ?></label>
 				<input name="edd_first_name" id="edd_first_name" class="text edd-input" type="text" value="<?php echo esc_attr( $first_name ); ?>" />
--- a/templates/shortcode-profile-editor.php
+++ b/templates/shortcode-profile-editor.php
@@ -24,9 +24,8 @@ if ( is_user_logged_in() ):
 
 	<?php edd_print_errors(); ?>
 
-	<?php do_action( 'edd_profile_editor_before' ); ?>
-
-	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post">
+	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post" <?php do_action( 'edd_profile_editor_form_tag' ); ?>>
+		<?php do_action( 'edd_profile_editor_before' ); ?>
 		<fieldset>
 			<span id="edd_profile_name_label"><legend><?php _e( 'Change your Name', 'easy-digital-downloads' ); ?></legend></span>
 			<?php do_action( 'edd_profile_editor_before_name' ); ?>
@@ -177,9 +176,8 @@ if ( is_user_logged_in() ):
 				<input name="edd_profile_editor_submit" id="edd_profile_editor_submit" type="submit" class="edd_submit" value="<?php _e( 'Save Changes', 'easy-digital-downloads' ); ?>"/>
 			</p>
 		</fieldset>
+		<?php do_action( 'edd_profile_editor_after' ); ?>
 	</form><!-- #edd_profile_editor_form -->
-
-	<?php do_action( 'edd_profile_editor_after' ); ?>
 
 	<?php
 else:
--- a/templates/shortcode-profile-editor.php
+++ b/templates/shortcode-profile-editor.php
@@ -28,6 +28,7 @@ if ( is_user_logged_in() ):
 
 	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post">
 		<fieldset>
+			<?php do_action( 'edd_profile_editor_before_name' ); ?>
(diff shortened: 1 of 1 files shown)


templates/shortcode-profile-editor.php
```diff
@@ -28,6 +28,7 @@
 
 	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post">
 		<fieldset>
+			<?php do_action( 'edd_profile_editor_before_name' ); ?>
```
pippinsplugins: Please move this below the leged

rubengc: Ok, but I want to add the section "Change your avatar" before "Change your name"

Would I add a new action for this? (Really I do not want to add an excesive number of actions)

and I just saw that the form does not have the attribute enctype, any idea on how to implement this only when necessary?

SDavisMedia:
@spencerfinnell @pippinsplugins Why does this need to be below the legend? Everywhere else on this form, you have the ability to add a completely new field/output unrelated to the existing fields (besides the naming scheme). If you place this hook below the legend, you tie it to the name fields and there's still no way to place something _unrelated_ above the name fields.

@rubengc's Request here is to to start the form with his own output. Placing a hook below the name legend is fine but that's a different goal and it would be incomplete as no other legend in this form are followed immediately by a hook.

spencerfinnell:
@SDavisMedia There is no need for it to be above. A legend will always appear at the top of the fieldset, and there should only be one legend per fieldset. 

A separate hook should be added to add the ability to create a separate fieldset/form if that is what is needed.

The only reason this works now is because of the `span` around the legend which should be removed as well. https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3568

http://codepen.io/anon/pen/ObVaYX

spencerfinnell:
Might make more sense to submit it against https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3904 as well.

SDavisMedia:
@spencerfinnell Good point.

@rubengc Is there a reason why you can't just use `edd_profile_editor_before`? https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/templates/shortcode-profile-editor.php#L27

rubengc:
I added a PR that could be a good point between I need and others could need

First I move edd_profile_editor_before and edd_profile_editor_after inside form
tag to allow add more fieldsets at the start and the end of the form

And I added an action inside form tag to allow add more attributes (like enctype, needle for file uploads)

With this changes we could remove edd_profile_editor_before_name because I think anyone will need more than first, last and display name here

What do you think?

SDavisMedia:
@rubengc That's not going to work. You can't change the location of hooks like that because the hooks are already in use. That change will affect sites in an undesirable way.

I see absolutely nothing wrong with the location (line number) you're asking for, but `edd_profile_editor_before` and `edd_profile_editor_after` need to stay exactly where they are.

Instead, create two new hooks located directly above the `<fieldset>` and directly after. Name them `edd_profile_editor_top` and  `edd_profile_editor_bottom`. That way there are hooks before and after the entire form as well as before and after the existing fields in the form.

rubengc:
Do you like last PR @SDavisMedia ?

SDavisMedia:
@rubengc That looks good to me. And it should behave just fine when #3904 is implemented.

@spencerfinnell See any issues with this approach?

rubengc:
@SDavisMedia I see [shortcode-register template](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/fbe20f79334ef6bd21f6a9810e8e5977bef116df/templates/shortcode-register.php)

Could be a better way use edd_profile_editor_fields_top and edd_profile_editor_fields_bottom to maintain same stucture as other shortcodes as points @spencerfinnell right?

SDavisMedia:
@rubengc That works. :+1:

spencerfinnell:
Seems like a good approach to me.

rubengc:
So now is ready to merge?

SDavisMedia:
@rubengc Yup, looking good here. Thanks! I'll let someone finalize it. Thumbs up from me. :+1:

templates/shortcode-profile-editor.php
```diff
@@ -26,7 +26,10 @@
 
 	<?php do_action( 'edd_profile_editor_before' ); ?>
 
-	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post">
+	<form id="edd_profile_editor_form" class="edd_form" action="<?php echo edd_get_current_page_url(); ?>" method="post" <?php do_action( 'edd_profile_editor_form_tag' ); ?>>
```
cklosowski: @SDavisMedia @spencerfinnell @rubengc I find it odd to have an action within an HTML tag. The only way this would be useful is to echo out information into the tag, but that's kind of not an ideal situation as there is no guarentee that the person using the hook is going to properly escape, or do proper sanitization of the form field tag.

I'm not sure I see the point of allowing someone to openly run an action on the attributes of a form field, unless I'm missing something.

rubengc:
I saw it before:
"And I added an action inside form tag to allow add more attributes (like enctype, needle for file uploads)"

I see this practice on other plugins (not from edd team)

There is another possibility to allow add custom attributes inside form tag? Or I could remove it and upload the image using ajax and adding here a hidden input with the url for the user meta

SDavisMedia:
@rubengc for something that custom that so few people would need, I'd rather see you just override the entire template in your theme: http://docs.easydigitaldownloads.com/article/1216-moving-edd-templates-to-your-theme

There's an argument that the other hooks needed to be added anyway but I don't think it's a good idea to put one directly in the form tag on any of our forms in core. Let's go ahead and remove that and we should be clear to merge.

rubengc:
I removed edd_profile_editor_form_tag action

Form enctype attribute is not for theme, It is for plugin support, but not problem, I will upload the image using ajax :)

SDavisMedia:
@rubengc I understand it's not for your theme. I'm saying if you wanted to add an attribute to the form tag, override the entire template file from your theme (see the documentation I linked) and add the attribute there. Then you'd still have to the option to either use ajax or enctype attribute. We just don't want to put it in EDD core.

I'll get this reviewed again and hopefully merged. Thanks again!

cklosowski:
@SDavisMedia looks good on my end, just let me know when yo'ure done with your testing and i'll merge it away.

SDavisMedia:
@cklosowski I'm already good to go. Merge away, sir!

rubengc:
👍

Issue #5160, opened 2016-11-05
Remove use of global $edd_options


chriscct7:
It's not uncommon for a developer to look at the source code of a parent project when trying to get best practice methods for writing an extension. 

In EDD a long time ago, we used to use global $edd_options to access settings. However, for years now, we've recommended people use the helper function as it's safer (don't have to exist check on the array when accessing it) and better (can easily set default).

Over time, many uses of  $edd_options have been removed from EDD core, but some have not. We should remove the last remaining ones.

pippinsplugins:
PR submitted that removes most of them.

I have not removed usage from tests.

chriscct7:
Still a couple uses in:
includes/admin/tools.php
includes/admin/tracking.php

sunnyratilal:
@chriscct7 I don't see these in the files you have mentioned - please can you add line numbers?

chriscct7:
In tracking file line 64, 77, 182, 202, etc it's all over that file

chriscct7:
In tools 381 and 409

sunnyratilal:
Are you looking at the correct branch? They have already been fixed. 

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/2.7/includes/admin/tracking.php#L76
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/2.7/includes/admin/tools.php#L407

chriscct7:
Oh, I thought these went out already. Theyre on the changelog for the .12
release so I was looking at master.

sonicviz:
I'm working on a project using EDD, FES and the Amazon S3 plugin and discovered the S3 plugin used edd_options in the init function

PR #5165, opened 2016-11-06
Remove more uses of $edd_options. #5160

pippinsplugins:
#5160

### DIFF
--- a/includes/admin/tracking.php
+++ b/includes/admin/tracking.php
@@ -193,15 +193,9 @@ class EDD_Tracking {
 	 * @return void
 	 */
 	public function check_for_optout( $data ) {
-
-		if( edd_get_option( 'allow_tracking', false ) ) {
-			edd_delete_option( 'allow_tracking' );
-		}
-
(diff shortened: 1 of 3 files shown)


includes/admin/tracking.php
```diff
@@ -199,10 +194,8 @@ public function check_for_optin( $data ) {
 	 */
 	public function check_for_optout( $data ) {
 
-		global $edd_options;
-		if( isset( $edd_options['allow_tracking'] ) ) {
-			unset( $edd_options['allow_tracking'] );
-			update_option( 'edd_settings', $edd_options );
+		if( edd_get_option( 'allow_tracking', false ) ) {
```
chriscct7: Couldn't this just be `edd_delete_option( 'allow_tracking' );`. We're opting them out regardless if they are already opted out or not. 

If currently opted in, saves 1 db call. If not opted in neutral.

pippinsplugins: Yes it could.

Issue #5169, opened 2016-11-07
Only require state if the Country has states


mintplugins:
Currently, when taxes are enabled and the customer is in a country without states, state is still required. This would require anyone from the UK to fill in a fake state in order to complete their purchase. 

We have a snippet in the library which disables the requirement for state but it would disable the requirement for countries which DO have states - which is not desirable for those countries.

I have found a work around for the snippet where the state is disabled ONLY if no states exist. 

```
function pw_edd_remove_state_from_required_fields( $fields ) {

	$states = edd_get_shop_states( $_POST['billing_country'] );

	if( array_key_exists( 'card_state', $fields ) && empty( $states ) ) {
		unset( $fields['card_state'] );
	}
	return $fields;
}
add_filter( 'edd_purchase_form_required_fields', 'pw_edd_remove_state_from_required_fields' );
```

I feel this should be default functionality for EDD as no one would (by default) want to require customers to fill in a state when no states actually exist for them. 

Proposal:
Only require state if the country has states.

pippinsplugins:
I agree.

mintplugins:
PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5170

mintplugins:
To test:

On Master Branch:
1. Enable Taxes.
2. Add product to cart.
3. Switch Country to USA.
4. Leave State Blank.
5. Click Purchase. 
6. Confirm the state error shows up. 
7. Switch to UK
8. Leave State Blank.
9. Complete Purchase. 
10. Confirm state requirement still shows up. 

On Branch 5169:
1. Enable Taxes.
2. Add product to cart.
3. Switch Country to USA.
4. Leave State Blank.
5. Click Purchase. 
6. Confirm the state error shows up. 
7. Switch to UK
8. Leave State Blank.
9. Complete Purchase. 
10. Confirm purchase goes through successfully. 

Additional test to make sure nothing broke:
2. Add product to cart.
3. Switch Country to USA.
4. Choose a state.
5.  Confirm purchase goes through successfully.

pippinsplugins:
I had a customer test this and they confirmed it worked properly.

PR looks good to me too.

michaelbeil:
All works well here. Ready for merge.

PR #5170, opened 2016-11-07
Check if the Customer's Country has no states.

mintplugins:
Check if the Customer's Country has been passed in and if it has no states. If the country has no states, remove the requirement for state. Otherwise leave it in.

Addresses: #5169

### DIFF
--- a/includes/process-purchase.php
+++ b/includes/process-purchase.php
@@ -390,6 +390,17 @@ function edd_purchase_form_required_fields() {
 			'error_id' => 'invalid_state',
 			'error_message' => __( 'Please enter billing state / province', 'easy-digital-downloads' )
 		);
+
+		// Check if the Customer's Country has been passed in and if it has no states.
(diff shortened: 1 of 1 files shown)


includes/process-purchase.php
```diff
@@ -390,6 +390,17 @@ function edd_purchase_form_required_fields() {
...
 		);
+
+		// Check if the Customer's Country has been passed in and if it has no states.
+		if ( isset( $_POST['billing_country'] ) ){
+			$customer_billing_country = sanitize_text_field( $_POST['billing_country'] );
+			$states = edd_get_shop_states( $customer_billing_country );
+
+			// If this country has no states, remove the requirement of a card_state.
+			if ( empty( $states ) ){
+				unset( $required_fields['card_state'] );
```
pippinsplugins: This should have a check to ensure `card_state` is set in `$required_fields`

Issue #5166, opened 2016-11-07
Icons in the frontend and performance


samikeijonen:
There are couple of icons in the frontend which are used as custom font icon or dashicons. We could probably move away from icon fonts and use SVGs only.

- First example is from checkout page. There we load all dashicons if SSL is around just for `padlock` icon. We could use SVG icon instead.
- Second example is the Ok icon before the Added to chart text. We could use SVG for that.
- Third example is the spinner. We could probably do that with CSS only.

I'm not sure are the icons `edd-icon-remove`, `edd-icon-remove-circle`, and `edd-icon-ok-circle` used anywhere.

I can create PR for this. I do understand that this might have backwards compatibility problems with themes. Another solution is to add some hooks where we can insert icons without JS.

pippinsplugins:
I like this idea.

I doubt there'd be much BC concern here.

samikeijonen:
Ok. I create PR for this for testing in couple of days.

pippinsplugins:
Since this does have potential backwards compatibility ramifications, I'm going to milestone it for 2.7.

samikeijonen:
Anybody had time to test this?

pippinsplugins:
I have not yet but will soon.

SDavisMedia:
I was able to test this a bit. The spinner is super slick and the Added to Cart check is working just fine for me. Couple points:

1. I'm not set up to check the SSL lock so someone else needs to grab that please.

2. How about the loading.gif usage for the gateway picker switching and discount form submission loading (and probably other places)? I know that's not an icon font. But are we looking to adjust those too or were they intentionally left out?

samikeijonen:
2. I just probably didn't notice them. I'll check if we can tackle them
also.

samikeijonen:
@SDavisMedia: In commit 9e23214f36a0c5d7784a7117f585719790e74b39 there is first pass of replacing `loading.gif` with the same spinner.

I'm not perfectly happy with small up/down movement in discount code but run out of time for today.

pippinsplugins:
This is looking great!

The padlock on the CC form is definitely not working though.

Master:
![screen shot 2016-12-07 at 2 15 43 pm](https://cloud.githubusercontent.com/assets/1034109/20984876/b85bdef4-bc87-11e6-92c3-ba81d363a148.png)

Issue branch:
[...]

PR #5177, opened 2016-11-08
Issue/5166 - Icons in the frontend and performance

samikeijonen:
Needs lot's of testing and polishing but here is the first pass for issue #5166. I could not test lock icon in checkout page at all because I don't have SSL in my localhost.


sunnyratilal:
Reverted until unit tests are fixed

Issue #5187, opened 2016-11-10
Settings API Checkbox Field doesn't allow a default 'checked' state


d4mation:
The Checkbox Settings API Field only applied `checked` based on the saved value rather than also allowing a default state.

PR Incoming.

SDavisMedia:
Working well for me. :+1:

cklosowski:
I have a bit of a concern with this actually. It's coded just fine, but the idea is actually something a little deeper, and I'd love to get your opinion @pippinsplugins.

EDD's Settings in core and extensions always default the standard behavior of a setting controlled by a checkbox to the `unchecked` state. That is, any setting that's a checkbox is an `opt-in` setting. There is a little bit of a drift then in the DB and what your settings would be at.

Let's say for instance that you register a checkbox setting with a standard value of checked. If a user goes to the settings, sees that checked, they will assume that the value is on. In reality the setting isn't in the Database at all, and it's up to a developer to be sure that they always supply `true` as the default value passed into `edd_get_option`.

That seeing an option checked isn't actually checked until I hit save. That's my biggest concern with it. Thoughts?

d4mation:
Ah, that would be an issue. I made this PR for a use case where basically EDD Settings API Fields are being used within a Repeater where a New Row would have the Checkbox default to `checked`. In that case the fact that it doesn't exist in the database wouldn't be a problem as it shouldn't until that Row's data got saved anyway. I can definitely see how this is would be a problem in every other instance though.

Outside of providing a default for `edd_get_option()` like you said I'm not sure if there's any other way to handle it. Maybe it should just be added into the documentation/doc block as a note if `std` is set that using the default in `edd_get_option()` would be necessary? It may just have to be up to the Dev within their Extension or within Core to ensure an opt-out feature is handled correctly.

pippinsplugins:
Point brought up by @cklosowski is definitely valid.

@d4mation with a repeater field is there any reason not to just set the boxes to checked via JS when inserting new rows?

d4mation:
@pippinsplugins That is definitely an option. I figured it would be easier overall if Checkboxes had the ability to set a default state, but I completely understand that with database assumptions that it could potentially be confusing.

pippinsplugins:
I like the idea of supporting a default state but I think we need to think it through further and ensure that `edd_get_option()` reacts accordingly.

d4mation:
[...]

PR #5188, opened 2016-11-10
Allow default for Settings API Checkbox

d4mation:
The Checkbox Settings API Field only applied `checked` based on the saved value rather than also allowing a default state. Referencing #5187


SDavisMedia:
@d4mation while we haven't been great about getting EDD fully up to speed, let's be sure to use the WP coding standard for brace style when adding new code: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style

d4mation:
@SDavisMedia Just fixed the coding styling for the braces.

PR #5190, opened 2016-11-10
Check file size against memory limit prior to attempting a forced download #3310

cklosowski:
#3310

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -148,8 +148,8 @@ function edd_process_download() {
 
 		// If the file isn't locally hosted, process the redirect
 		if ( filter_var( $requested_file, FILTER_VALIDATE_URL ) && ! edd_is_local_file( $requested_file ) ) {
-			edd_deliver_download( $requested_file, true );
-			exit;
+			//edd_deliver_download( $requested_file, true );
+			//exit;
 		}
 
 		if( 'x_sendfile' == $method && ( ! function_exists( 'apache_get_modules' ) || ! in_array( 'mod_xsendfile', apache_get_modules() ) ) ) {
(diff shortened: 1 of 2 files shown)


includes/process-download.php
```diff
@@ -233,13 +233,25 @@ function edd_process_download() {
 
 				}
 
+				// Determine if the file we're attempting to use direct on is too large for the memory limit to use direct
+				$memory_limit = @ini_get( 'memory_limit' );
```
pippinsplugins: This should probably only happen if `$direct` is `true`

includes/process-download.php
```diff
@@ -233,13 +233,25 @@ function edd_process_download() {
...
+				if ( $memory_limit ) {
+					$memory_limit_bytes = edd_determine_size_in_bytes( $memory_limit );
+					$file_size          = edd_determine_size_in_bytes( @filesize( ABSPATH . $file_path ) );
```
pippinsplugins: Above we do `@filesize( $file_path )` so `ABSPATH` shouldn't be used here I don't think.

includes/process-download.php
```diff
@@ -838,3 +850,23 @@ function edd_set_requested_file_scheme( $requested_file, $download_files, $file_
 
 }
 add_filter( 'edd_requested_file', 'edd_set_requested_file_scheme', 10, 3 );
+
+/**
+ * Given a string for a file size, dtermeine the size in bytes
```
michaelbeil: s/dtermeine/determine

sunnyratilal: Fixed

includes/process-download.php
sunnyratilal: Yep `ABSPATH` isn't required here

sunnyratilal: Fixed

includes/process-download.php
sunnyratilal: Yep, fixed it up.

Issue #5163, opened 2016-11-05
edd_get_users_purchases docs needs to be updated


chriscct7:
Cross referencing from https://github.com/easydigitaldownloads/edd-recurring/issues/562

The docs of the `$status` variable need to be updated to say it supports as an array. Maybe even the same casting logic from https://github.com/easydigitaldownloads/edd-recurring/pull/563 could be used, and make the default an array. Also, unit tests should be added that use that function as an array.

sunnyratilal:
PR ready for testing

sunnyratilal:
Merged into `master`

PR #5195, opened 2016-11-11
Update doc for edd_get_users_purchases() and allow array to be passed as $status

sunnyratilal:
PR for #5163

### DIFF
--- a/includes/user-functions.php
+++ b/includes/user-functions.php
@@ -24,7 +24,7 @@ if ( ! defined( 'ABSPATH' ) ) exit;
  * @param int $user User ID or email address
  * @param int $number Number of purchases to retrieve
  * @param bool pagination
- * @param string|array $status Either an array of statuses or a single status as a string literal
+ * @param string|array $status Either an array of statuses, a single status as a string literal or a comma separated list of statues
  *
  * @return bool|object List of all user purchases
  */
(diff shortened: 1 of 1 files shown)


includes/user-functions.php
```diff
@@ -38,7 +37,14 @@ function edd_get_users_purchases( $user = 0, $number = 20, $pagination = false,
 		return false;
 	}
 
-	$status = $status === 'complete' ? 'publish' : $status;
+	if ( ! is_array( $status ) && is_string( $status ) ) {
```
chriscct7: Couldn't you just do is_string here?

cklosowski: Yeah the `! is_array` isn't needed, should we support a comma separate list of statuses here too? So something like:
```
if ( is_string( $status ) ) {
    if ( strpos( ',', $status ) ) {
        $status = explode( ',', $status );
    } else {
        $status = $status === 'complete' ? 'publish' : $status;
        $status = array( $status );
    }
}

cklosowski: Only reason I say this is that many functions in WP Core allow an array, or comma separated list of statuses. Just makes it consistant and doesn't take much on our end.

cklosowski:
@chriscct7 Look better? @sunnyratilal if you are good with this merge when you review.

sunnyratilal:
All good here

PR #5197, opened 2016-11-11
Issue/4936

sunnyratilal:
PR for #4936

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -65,16 +65,42 @@ function edd_get_discounts( $args = array() ) {
 function edd_has_active_discounts() {
 	$has_active = false;
 
-	$discounts  = edd_get_discounts();
+	$page = 1;
+	do {
+		/*
+		 * Get active discounts in bulk of 100
+		 * we can do such a large number because we are only fetching ids
+		 *
+		 * We need to fetch them all to do a manual check for expiration.
+		 */
+		$discounts = edd_get_discounts(
+			array(
+				'post_status'    => 'active',
+				'posts_per_page' => 100,
+				'paged'          => $page,
+				'fields'         => 'ids'
+			)
+		);
 
-	if ( $discounts) {
+		// When there are no discounts found anymore there are no active ones.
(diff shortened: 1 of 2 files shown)


includes/discount-functions.php
```diff
@@ -383,15 +408,16 @@ function edd_discount_exists( $code_id ) {
  *
  * @since 1.0
```
cklosowski: Can we add a 2nd `@since 2.6.11` tag here, and specify that it's when the `update` attribute was added @sunnyratilal

includes/discount-functions.php
```diff
@@ -579,11 +605,12 @@ function edd_is_discount_not_global( $code_id = 0 ) {
  * Checks whether a discount code is expired.
  *
```
cklosowski: Same as above, can we add another since here?

tests/tests-discounts.php
```diff
@@ -84,6 +84,11 @@ public function test_discounts_exists() {
 		$this->assertTrue( edd_has_active_discounts() );
 	}
 
+	public function test_is_discount_active() {
+		$this->assertTrue( edd_is_discount_active( $this->_post_id, true ) );
+		$this->assertTrue( edd_is_discount_active( $this->_post_id, false ) );
```
cklosowski: This test needs to also verify that the status wasn't changed when we send in a `false`. So build one out with an expired discount, check that `is_active` with a `false` still returns it as `active` in the DB?

includes/discount-functions.php
sunnyratilal: Added in 25ee7424e5c199db155e5406847148424ee4db04

includes/discount-functions.php
sunnyratilal: Added in 25ee7424e5c199db155e5406847148424ee4db04

tests/tests-discounts.php
sunnyratilal: Added in5a0edd47777235b03dcd6015986feb9c98f53d3a

Issue #5142, opened 2016-11-03
Many {non-string compatible} to String conversion Warnings in EDD HTML Select function


chriscct7:
When an array of options are sent to the EDD HTML select function in the $args param key `options` during the `foreach ( $args['options'] as $key => $option ) {` [1] if either:
- $args key for `multiple` has either been passed in as `false` or the default value (which is `false`) 
- $args key for `multiple` has been passed in as true and the `selected` $args value has either been passed in as a non-array such as string or an object, the function runs `$selected = selected( $args['selected'], $key, false );` [2] which will result in the `Array to string conversion` warning.

The easiest way to replicate this is to open up Software Licensing, make a new download with 1 variable option, save the download and then look at the upgrade path metabox (note, PHP Error need to be set to show strict mode errors)

[1 ]https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L527
[2] https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L532

A combination of similar PHP warnings and fatal errors can be trivially generated by getting a non-array value in any of these:
- https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L520
- https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L509

or by getting a non-string value for `selected` here:
- https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L522

Any time `selected` it needs to be checked for type in this function.

sunnyratilal:
@chriscct7 Even with strict mode on, I'm unable to replicate this error and it the `License Upgrade Paths` metabox works as expected too.

sunnyratilal:
@chriscct7 Replicated.

sunnyratilal:
@chriscct7 PR added at #5199. Please could you review when you get a chance.

chriscct7:
Same issue in both places but if multiple was false but selected is an array then you'll get a strict error https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5199/files#diff-d0193f2d42f7e8a5e59caca5e3dcd583R521 & https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5199/files#diff-d0193f2d42f7e8a5e59caca5e3dcd583R532

[...]

PR #5199, opened 2016-11-12
Issue/5142

sunnyratilal:
PR for #5142


pippinsplugins:
@cklosowski which bug is that?

PR #5207, opened 2016-11-14
Fix "Notice: Undefined offset".

remcotolsma:
Notice: Undefined offset: 7 in
wp-content/plugins/easy-digital-downloads/includes/admin/reporting/graph
ing.php on line 75.


cklosowski:
@sunnyratilal mind looking this over while you're doing reporting changes?

sunnyratilal:
Thanks @remcotolsma  but this has already been fixed in b38435823798642adc8cfc3767b4953aab83cd94

Issue #5154, opened 2016-11-05
When single day ranges are used on Payments list table, order status counts not calculated


chriscct7:
If you do a single date range, ex Jan 1 - Jan 1 of the same year on the payments list table, all payment status counts show 0. If range is Jan 1 - Jan 2 however, they work

PR #5212, opened 2016-11-14
Issue/5154

sunnyratilal:
PR for #5154

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -518,13 +518,9 @@ function edd_count_payments( $args = array() ) {
 
 		$is_date    = checkdate( $month, $day, $year );
 		if ( false !== $is_date ) {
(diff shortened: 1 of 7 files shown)


includes/payments/functions.php
```diff
@@ -520,6 +520,11 @@ function edd_count_payments( $args = array() ) {
 		if ( false !== $is_date ) {
 
 			$date   = new DateTime( $args['end-date'] );
+
+			if ( $args['end-date'] == $args['start-date'] ) {
+				$date->add( new DateInterval( 'P1D' ) );
```
cklosowski: `DateInterval` is a `PHP 5.3+` only class. `DateTime` (used above) is `PHP 5.2+` We need a solution that works on HP 5.2+

Issue #5206, opened 2016-11-14
PHP Warning:  Illegal string offset 'email'


pippinsplugins:
This is showing up in error logs:

PHP Warning:  Illegal string offset 'email' in /dom8333/wp-content/plugins/easy-digital-downloads/includes/payments/class-edd-payment.php on line 1609

cklosowski:
It seems odd that this is failing:
`$current_meta['user_info']['email']  = $meta_value;`

Need to see if I can reproduce this in some way.

pippinsplugins:
I saw this on the live EDD site, if that helps.

sunnyratilal:
Looks like for some reason a `string` is being returned from the `get_meta()` call for it to report an `Illegal string offset`. Best way around it would just be to wrap [this](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/payments/class-edd-payment.php#L1608) in an `is_array()` check.

sunnyratilal:
Any further thoughts @easydigitaldownloads/core-devs?

pippinsplugins:
I think adding type checks will be fine.

PR #5213, opened 2016-11-15
Type check to avoid string offset error #5206

sunnyratilal:
PR for #5206

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -1605,10 +1605,8 @@ class EDD_Payment {
 			$meta_value = apply_filters( 'edd_edd_update_payment_meta_' . $meta_key, $meta_value, $this->ID );
 			update_post_meta( $this->ID, '_edd_payment_user_email', $meta_value );
 
-			$current_meta = $this->get_meta();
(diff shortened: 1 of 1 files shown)


includes/payments/class-edd-payment.php
```diff
@@ -1606,7 +1606,9 @@ public function update_meta( $meta_key = '', $meta_value = '', $prev_value = ''
 			update_post_meta( $this->ID, '_edd_payment_user_email', $meta_value );
 
 			$current_meta = $this->get_meta();
-			$current_meta['user_info']['email']  = $meta_value;
+			if ( is_array( $current_meta ) ) {
+				$current_meta['user_info']['email']  = $meta_value;
```
pippinsplugins: We need a fallback here I think in the case $current_meta is not an array. Should fall back to the email meta field and then the customer email if the one on payment is empty.

sunnyratilal:
Type checking isn't actually necessary.  We can just force an array to be returned from the `get_meta()` function by setting `$single` to `false`.  This means an array will *always* be returned.

cklosowski:
PHPUnit is passing, I think i'm happy with this and merging.

Issue #5247, opened 2016-11-23
Add support for user-defined locales


ramiabraham:
WordPress 4.7 introduces `get_user_locale()` ([changeset](https://core.trac.wordpress.org/changeset/38705)), which falls back to the standard `get_locale()`.

Areas affected:

- The EDD `load_textdomain()` loader method ([line link](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/easy-digital-downloads.php#L384)) to support user locales.

- `EDD_Tracking::setup_data()`, specifically `$data['locale'] `.

ramiabraham:
Added a PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5248

sunnyratilal:
Merged into `master`

PR #5248, opened 2016-11-23
Add support for user admin locales

ramiabraham:
Adds support for user locales:
- The primary EDD `load_textdomain()` loader
- EDD_Tracking::setup_data()
- Related inline hook doc

#5247

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -381,7 +381,21 @@ final class Easy_Digital_Downloads {
 		$edd_lang_dir  = apply_filters( 'edd_languages_directory', $edd_lang_dir );
 
 		// Traditional WordPress plugin locale filter.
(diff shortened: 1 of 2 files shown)


easy-digital-downloads.php
```diff
@@ -381,7 +381,21 @@ public function load_textdomain() {
 		$edd_lang_dir  = apply_filters( 'edd_languages_directory', $edd_lang_dir );
 
 		// Traditional WordPress plugin locale filter.
-		$locale        = apply_filters( 'plugin_locale',  get_locale(), 'easy-digital-downloads' );
+
+		$get_locale = get_locale();
+
+		if ( $wp_version >= 4.7 ) {
```
pippinsplugins: `$wp_version` global has to be defined.

Issue #5082, opened 2016-10-24
Show gateway selector upon error


michaelbeil:
We should show the gateway selector upon error, whilst still keeping the currently selected payment gateway in the query string. 

To replicate:
1. Enable both Stripe and PayPal
2. Checkout with Stripe
3. Add a wrong zip code
4. Page refreshes, an error shows, with /checkout/?payment-mode=stripe in the URL
5. You cannot switch gateways

Ref: https://secure.helpscout.net/conversation/267010262/45104/?folderId=180504

Perhaps related to https://github.com/easydigitaldownloads/easy-digital-downloads/issues/2222.

SDavisMedia:
We definitely need this, especially for 2Checkout + Recurring Payments users: https://github.com/easydigitaldownloads/edd-recurring/issues/531#issuecomment-255210144

pippinsplugins:
I agree, let's keep the gateway picker even after error.

pippinsplugins:
PR submitted.

This does a few things:

1. Keeps gateway selector anytime there is more than one gateway
2. Resolves and issue with the gateway fields loading the fields for a gateway other than the one specified with `?payment-mode={gateway}`
3. Resolves an issue that permitted a non-enabled gateway's fields to be loaded
4. Resolves and issue with the proper radio field in the gateway select from being checked on page load when `?payment-mode={gateway}` is present

SDavisMedia:
I just tested this in relation to the following ticket: https://secure.helpscout.net/conversation/290741250/49034/?folderId=180499

I am getting the payment gateway selector after reload and the proper gateway is selected. One thing I noticed is that while testing with Stripe, the credit card info was lost after refresh. Intentional? It makes sense, just want to make sure.

pippinsplugins:
@SDavisMedia card data isn't saved yet and will not be until #4220

SDavisMedia:
Got it. Thanks.

pippinsplugins:
This customer had an interesting issue: https://secure.helpscout.net/conversation/296058902/49682/?folderId=180505

When using Conditional Gateways, Recurring Payments, Stripe, and PayPal, it was not possible to have a subscription product that was limited to just PayPal purchases. Anytime someone tried to make a purchase, it would show an error about billing address being required.

Our changes here resolve that customer's issue as well as the original bug here due to the changes in `edd_get_chosen_gateway()` that make it work more reliably. 💯

cklosowski:
[...]

PR #5253, opened 2016-11-24
Keep gateway selector on checkout error. - issue/5082

pippinsplugins:
Also resolves issue with edd_get_chosen_gateway() not working reliably. Also makes checkout load the proper gateway based on the ?payment-mode parameter. #5082


cklosowski:
No sure what happened with this PR, going to close it and open a new one using the cherry pick method of the 2 commits that @pippinsplugins made, so we can isolate the changes.

PR #5257, opened 2016-11-25
Fixes issue #5256 with http_build_query with bad arg_seperator.output setting

### LINKED ISSUE #5256: Force http_build_query to use `&` as the arg_seperator
HostEurope and a couple other bad-actor hosts change the `arg_seperator.output` PHP ini setting from the default `&` to `&amp;`. This setting is used in PHP core functions like `http_build_query` and a couple other functions and the change breaks the function from creating a redirectable url. 

When using `http_build_query` and related functions, the third param (optional) is the arg seperator, and only if left unused it will default to the PHP ini setting. When using it, you should always force the third param to be & for good measure.

2 of 3 EDD core uses of it do not force it:
- https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b37acf76690486409e50f5542ae430e6b9c6f405/includes/gateways/paypal-standard.php#L287
- https://github.com/easydigitaldownloads/easy-digital-downloads/blob/ad537475d57fffb880a4619501c1f6dd59d6ddb1/includes/api/class-edd-api.php#L1786

Importantly the first one contains the following line:
```
$paypal_redirect = str_replace( '&amp;', '&', $paypal_redirect );
```
This sorta fixes the issue, but we should fix it properly and then remove this line (as it would no longer be needed for the intended purpose and keeping it would breaks products or more likely fees containing an escaped ampersand for display purposes).

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b37acf76690486409e50f5542ae430e6b9c6f405/includes/gateways/paypal-standard.php#L290

chriscct7:
Fixes #5256 

Fixes issue #5256 with http_build_query with bad arg_seperator.output

### DIFF
--- a/includes/api/class-edd-api.php
+++ b/includes/api/class-edd-api.php
@@ -1780,10 +1780,11 @@ class EDD_API {
 			'purchasekey' => isset( $wp_query->query_vars['purchasekey'] ) ? $wp_query->query_vars['purchasekey'] : null,
 			'email'       => isset( $wp_query->query_vars['email'] )       ? $wp_query->query_vars['email']       : null,
 		);
-
+		
+		$arg_separator    = edd_get_php_arg_separator_output();
 		$log_data = array(
 			'log_type'     => 'api_request',
-			'post_excerpt' => http_build_query( $query ),
+			'post_excerpt' => http_build_query( $query, '', $arg_separator ),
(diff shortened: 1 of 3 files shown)


chriscct7:
We're going to be updating the pr later tonight as discussed in slack

sunnyratilal:
Closing PR as the tracking issue is closed.

Issue #3498, opened 2015-06-11
Right-click "Save Link As" in Firefox on WPEngine results in files named "index.php"


pippinsplugins:
There's a strange bug that results in files getting named "index.php".

It appears to only happen in Firefox when customers right-click and choose "Save Link As". It also appears to be limited to WPEngine, or perhaps NGinx.

Ticket that reported the issue: https://secure.helpscout.net/conversation/96429143/10523/

ankitrox:
I am having same issue, but my site is hosted on AWS EC2 instance. I tried to rename that index.php file to index.zip and surprisingly zip file was ok.

so only issue is extension of the file is getting changed.

SDavisMedia:
Think this is related? https://secure.helpscout.net/conversation/106380533/13238/

pippinsplugins:
@SDavisMedia Very possibly.

mindctrl:
Another one in Safari 9. https://secure.helpscout.net/conversation/136337769/20553/

mindctrl:
Some additional info from customer:

> "I have discovered that if you click "Download Linked File" (and not "Download Linked File As..." this can be seen in one of my screenshots) the file downloads with the correct filename."

akyusa01:
Here's another ticket: https://secure.helpscout.net/conversation/179645160/29659/

cklosowski:
There isn't really a way around this. The browser simply looks at the path in the URL (of which ours is `index.php?...` so it's naming it what the file technically is. The only way to do this would be to name the endpoint the `.zip` file which isn't possible as it isn't a real file (to secure it).

sunnyratilal:
I agree with @cklosowski - this is more of a browser/server issue than an issue on our end that we can fix.

Issue #5259, opened 2016-11-27
Add download attribute to downloads


schlessera:
Given the name of the plugin, it should assume that the links it generates within receipt emails and shortcodes should be identified as downloads through the HTML5 `download` attribute on that `<a>` tag.

This tag does two things:

1. It allows to specify a clean filename that is used for saving the file on the customer's computer, instead of the random string that gets generated for security purposes.
2. It forces the browser to consider this link to be a download, instead of a file that should be played inline if possible. This avoids such scenarios where clicking on a purchased media file just plays that media file in the browser instead of downloading it.

Usage example:
```HTML
<a href="https://example.domain/index.php?eddfile=830%3A360%3A1%3A0&ttl=1480163556&file=1&token=33b98290c9ada42f94ef10ac92ae9441" download="myfile.mp3">
```

This will let the browser immediately download the file instead of playing it in the browser window, and it will save the file as `myfile.mp3` instead of whatever the file happened to be called on the server side.

I was hoping that I could just use a filter to add this attribute, but the link generation seems to be hard-coded without obvious extension points.

chriscct7:
Duplicates of #2794 and #2798

pippinsplugins:
I really wish we could add this but it breaks download link error messages.

PR #5260, opened 2016-11-27
Add download attribute to file download links. See #5259

schlessera:
As EDD already seems to already default to a clean file name, I opted to only use the `download` attribute without the optional filename override.

See: http://www.w3schools.com/TAgs/att_a_download.asp

#5259

### DIFF
--- a/includes/emails/class-edd-email-tags.php
+++ b/includes/emails/class-edd-email-tags.php
@@ -427,7 +427,7 @@ function edd_email_tag_download_list( $payment_id ) {
 					if ( $show_links ) {
 						$download_list .= '<div>';
 							$file_url = edd_get_download_file_url( $payment_data['key'], $email, $filekey, $item['id'], $price_id );
-							$download_list .= '<a href="' . esc_url( $file_url ) . '">' . edd_get_file_name( $file ) . '</a>';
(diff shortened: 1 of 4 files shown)


cklosowski:
Renaming the PR since `purchase links` is the term used for the purchase buttons, not the links to download files.

@pippinsplugins I thought we tried this once already but ran into issues. Maybe @chriscct7 remembers? Maybe it could fix #3498?

chriscct7:
@cklosowski see https://github.com/easydigitaldownloads/easy-digital-downloads/pull/3681#issuecomment-137552441

pippinsplugins:
#https://github.com/easydigitaldownloads/easy-digital-downloads/pull/3681#issuecomment-137552441 explains it.

We unfortunately cannot do this, though I really wish we could.

schlessera:
@pippinsplugins Ok, I understand why you cannot fix this. But currently, there's no hooks in there to do such a change on a site-by-site base either.
It'll have to wait until after WCUS, but would you be interested in a pull request that adds a filter around the link itself so that developers have control over the markup of the link?

pippinsplugins:
@schlessera I'd definitely be interested in getting the necessary hooks and filters in place to support this. If you have specific hooks / filters you'd like, we'd be more than happy to consider them.

Issue #5268, opened 2016-11-30
If download file doesn't exist on server, an unusable file, with correct name is downloaded


Shelob9:
Steps to reproduce:
1) Migrate your site, but don't copy uploads/edd (BTW I think this is an issue with Migrate DB Pro's media files add-on)
2) Go to purchase history and click a download link
3) In my case I get a zip file with the right name, but its empty and therefore GNOME Archive Manger (and probably other zip apps) reports an error

pippinsplugins:
@Shelob9 Are you able to provide a dump of post meta from the product for us?

Shelob9:
@pippinsplugins  Here is output of `wp post meta list` for an 1 download that was screwed up by my mistake: https://pastebin.com/xi9LMa4R

To be clear, the issue was that when I migrated the server, the EDD files did not come with it. So there is no way EDD should have been able to produce a proper download file.

I think the bug is bad error handling. It shouldn't list the download file in purchase details, it should show something like `Error with download file, email Josh and tell him to fix it.`

pippinsplugins:
Specifically, the problem is this: `"attachment_id":"33038"`. I think any way.

We should add better checks to determine if the attachment exists.

pippinsplugins:
@Shelob9 did that attachment ID get migrated but without the file or was the attachment ID invalid as well?

pippinsplugins:
PR attached that checks if the file exists. It does not trigger an error message but does then default back to trying to use the file URL (which could still exist) inside of post meta.

We could potentially go a step further by doing a wp_remote_head() request to the file URL to confirm the file does not return a 404 error.

Shelob9:
@pippinsplugins Correct, the attachment got migrated, but the actual file did not.

pippinsplugins:
@Shelob9 the site URL was unchanged, right?

Shelob9:
The site URL was changed. WP Migrate DB Pro did update the meta properly to the new URL, it just didn't pull the download files properly, which I don't really consider a bug in WP Migrate DB Pro, since I had set nginx to block access to EDD dir.

pippinsplugins:
Thoughts on approach in #5275?

It introduces a new action hook that provides an easy point at which we can inject remote head request to determine if the URL is accessible.

pippinsplugins:
Quick way to test:

1. Add a file to a product that is valid
2. Add a file URL to a product that is invalid (any fake or protected URL will work)
[...]

PR #5275, opened 2016-12-01
Confirm file exists before setting it to $requested_file - #5268

pippinsplugins:
#5268

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -90,6 +90,11 @@ function edd_process_download() {
 
 				$attached_file = get_attached_file( $attachment_id, false );
 
+				// Confirm the file exists
+				if( ! file_exists( $attached_file ) ) {
+					$attached_file = false;
+				}
+
 			}
 
 			if( $attached_file ) {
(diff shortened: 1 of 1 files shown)


includes/process-download.php
```diff
@@ -838,3 +850,30 @@ function edd_set_requested_file_scheme( $requested_file, $download_files, $file_
...
+ * @return void
+ */
+function edd_check_file_url_head( $requested_file, $args, $method ) {
+
+	// If this is a file URL (not a path), perform a head request to determine if it's valid
+	if( filter_var( $requested_file, FILTER_VALIDATE_URL ) ) {
+
+		$request = wp_remote_head( $requested_file );
+
+		if( is_wp_error( $request) ) {
```
cklosowski: I don't believe that a `404` returns `WP_Error` but requires checking the `wp_remote_retrieve_response_code`.

Should we do something like this instead?
`( is_wp_error( $request ) || (int) wp_remote_retrieve_response_code( $request ) === 404 )`

cklosowski:
Doh, looks like we need to update something in our unit tests for `latest`. I'll rerun the `latest` tests for WP Core and see if that fixes it, and we were just in a weird state for the release cycle.

includes/process-download.php
pippinsplugins: I was having a hard time triggering a 404 error code. It was strange, even files that were clearly 404'ing were returning a 200 code with wp_remote_head(). No idea why.

Checking for a 404 code does seem like a good idea.

pippinsplugins: @cklosowski Added 404 check.

Issue #4824, opened 2016-07-25
Product Import allows assigning non-existent author


cklosowski:
When you import downloads and use the Author field, it's possible to have the downloads assigned to an author ID that's non-existant. We should verify that a user with the ID exists before using it in the Post Data, and if it is not, maybe resort to the admin Id.

cklosowski:
I think best solution here is that if we don't identify an author with that ID, we just assign the post to the user importing it?

pippinsplugins:
>>>I think best solution here is that if we don't identify an author with that ID, we just assign the post to the user importing it?

I agree.

PR #5285, opened 2016-12-08
Detect missing users while importing #4824

cklosowski:
#4824

### DIFF
--- a/includes/admin/import/class-batch-import-downloads.php
+++ b/includes/admin/import/class-batch-import-downloads.php
@@ -106,6 +106,17 @@ class EDD_Batch_Downloads_Import extends EDD_Batch_Import {
 					}
 				}
 
+				if ( empty( $args['post_author'] ) ) {
+					$user = wp_get_current_user();
+					$args['post_author'] = $user->user_login;
+				} else {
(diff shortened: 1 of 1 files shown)


includes/admin/import/class-batch-import-downloads.php
```diff
@@ -106,6 +106,17 @@ public function process_step() {
 					}
 				}
 
+				if ( empty( $args['post_author'] ) ) {
+					$user = wp_get_current_user();
+					$args['post_author'] = $user->user_login;
+				} else {
+					$user = get_user_by( 'user_login', $args['post_author'] );
```
pippinsplugins: This should detect if `post_author` has been specified with an ID or email.

cklosowski: @pippinsplugins yeah that's an improvement. Originally our code simply worked with `user_login`, not any other fields, I was going for minimal change set.

pippinsplugins: :+1:

cklosowski:
Have a look at that @pippinsplugins.

Issue #5292, opened 2016-12-13
Download links in emails over escaped


Shelob9:
My email receipts are over-escaped and therefore don't work, which is bad.

Here is an example, the & is being encoded.

https://new-cf.dev/wp/index.php?eddfile=34476%3A565%3A1%3A1&#038;ttl=1481992962&#038;file=1&#038;token=84c31976f264e5b067f8ad02426ac356

My solution (PR incoming) is to use `esc_url_raw` instead of `esc_url` in `edd_email_tag_download_list` ~L430

Shelob9:
System info: 
production -https://pastebin.com/rREvJb47
development -  https://pastebin.com/8cLszVHy

This wasn't happening on old site, which was PHP5. Hard to say if PHP7 is to blame, as many things changed.

Shelob9:
BTW, my quick fix until this is fixed in main plugin (easier than switching my composer.json to use my fork) https://pastebin.com/3wjYhW1u

pippinsplugins:
@Shelob9 We'll get this included with 2.6.14, thanks!

PR #5293, opened 2016-12-13
Use esc_url_raw to prevent over escaping of links in emails

Shelob9:
#5292


Shelob9:
BTW Here is my system info:
Development: https://pastebin.com/8cLszVHy
Production: https://pastebin.com/rREvJb47

Issue was happening on both. Both are PHP7 (old site was 5.5 and this didn't happen, but tons of things have changed as well so not sure that has anything to do with this.

Issue #5294, opened 2016-12-13
No-transient issue caused by W3TC causes update checks on every admin page


webzunft:
A lot of EDD users, especially those with the Software Licensing enabled use your class from https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/EDD_SL_Plugin_Updater.php.

This class is great (especially after a recent fix for duplicate requests), but there is one issue I see a lot with my own implementation and I bet it is happening here as well.

A lot of users requests updates on every admin page impression because transients are returning null due to an usage of the W3 Total Cache plugin.

It basically happens when "Object Cache" in W3TC is enabled and the problem is described in many threads, e.g. https://wordpress.org/support/topic/w3-total-cache-transient-admin-issue/.

A solution I believe might work would be to not use the Transient API. Maybe set an option like `edd_last_update_check` with the timestamp of the last check in it and then check and update it based on the expiry time. It can also be just an exception when Object Cache is enabled in W3TC, but I searched for months after the source why some users have a very slow backend after enabling our plugins and also generate plenty of calls to our store, while others don’t.

Since you actually use this class as an integration example for customers, I wanted to get your feedback before implementing my own hack to solve it in time.

shazahm1:
I was experiencing the same issue with my commercial addon for my plugin. My site was be bombarded with update checks with sites running W3TC.

I ended writing a replacement upgrade check class which interfaced with WP REST API endpoints I wrote.

My replacement class, you can found here (which is a drop in replacement for the EDD-SL Plugin Updater):

https://github.com/Connections-Business-Directory/Connections/blob/develop/includes/admin/class.plugin-updater.php

Here's the link to my WP REST API endpoints required for my plugin updater class:

https://github.com/shazahm1/EDD-SL-Plugin-Update-REST-API

Instead of using transients, I use the option API, the same way WP core does to save its plugin updates meta using the same timeouts for the option, set using a time stamp (more or less). This prevents the double hit from the filter being run twice and limits update checks to only occur when needed.

I did a heck of a lot of testing before deployment and it has had a net positive effect. Unfortunately this does not and can not fix the users who have not updated so I still get a bunch of unnecessary updates checks but they are much more manageable now.

Oh, there is one issue... EDD-SL does have some logic to deal with licensing for WP multisite which I have not yet implemented in my plugin updater.

Not saying you should use my code as a solution, but maybe there's bit and pieces you'll find useful / or not.

Hope this helps!

[...]

PR #5302, opened 2016-12-16
Replace transients with options as caching mechanism in EDD_SL_Plugin_Updater - issue/5294

pippinsplugins:
#5294

### DIFF
--- a/includes/EDD_SL_Plugin_Updater.php
+++ b/includes/EDD_SL_Plugin_Updater.php
@@ -98,12 +98,13 @@ class EDD_SL_Plugin_Updater {
 			return $_transient_data;
 		}
 
-		$version_info = get_transient( $this->cache_key );
+		$version_info = $this->get_cached_version_info();
 
 		if ( false === $version_info ) {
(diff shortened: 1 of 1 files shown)


includes/EDD_SL_Plugin_Updater.php
```diff
@@ -260,7 +260,7 @@ public function plugins_api_filter( $_data, $_action = '', $_args = null ) {
 
 		$cache_key = 'edd_api_request_' . md5( serialize( $this->slug . $this->api_data['license'] ) );
 
-		//Get the transient where we store the api request for this plugin for 24 hours
+		// Get the transient where we store the api request for this plugin for 24 hours
 		$edd_api_request_transient = get_site_transient( $cache_key );
```
ghost1227: @pippinsplugins Shouldn't this have been replaced as well? We're not setting this anymore, so I'm guessing that'll always return nothing.

pippinsplugins: Good catch, thanks. Updated.

Issue #5003, opened 2016-10-04
When adding fee to cart, precision in decimal spot can be lost


chriscct7:
Inserting a fee with the amount of 50.00 will result in the fee getting stored as 50.0 due to PHP rounding it. We should ensure when decimal place 1 is set, that decimal place 2 is a zero if not set so people can do basic precision reliant comparisons.

sunnyratilal:
PR opened at #5315

pippinsplugins:
@sunnyratilal we'll need unit tests added for this before we can merge it.

PR #5315, opened 2016-12-25
Force fee amount to have proper number of decimal places #5003

sunnyratilal:
PR for #5003


includes/class-edd-fees.php
```diff
@@ -107,6 +107,9 @@ public function add_fee( $args = array() ) {
 		// Sanitize the amount
 		$args['amount'] = edd_sanitize_amount( $args['amount'] );
 
+		// Force the amount to have 2 decimal places
+		$args['amount'] = number_format( (float) $args['amount'], 2, '.', '' );
```
cklosowski: we should use the `edd_currency_decimal_filter()` method here. It'll make sure that sites using accuracy different than `2` will work.

cklosowski: @sunnyratilal By using that then we can also write some unit tests around other decimal precisions as well.

Issue #5217, opened 2016-11-15
Introduce EDD_Discount


cklosowski:
In the spirit of abstracting all the object data we can. We need an `EDD_Discount` object that allows us to work through a custom table.

sunnyratilal:
Initial work for this has begun.  An `EDD_Discount` class has been created and methods have been written for all the discount meta.

pippinsplugins:
WIP PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5318

pippinsplugins:
@sunnyratilal let's make each of the class properties `protected` to enforce the usage of the `__get()` magic method.

pippinsplugins:
That will make it more consistent with `EDD_Payment` as well.

sunnyratilal:
All the discount functions have been ported over to make use of `EDD_Discount`. I am now going to begin initial testing of all the functions.

Once I've done the base testing, I'll mark this issue as `Needs Testing`.

sunnyratilal:
@easydigitaldownloads/core-devs Ready for general testing.

sunnyratilal:
PR at #5318

pippinsplugins:
This is looking really good. Left a small change request on the PR.

Now, a question that could result in sizable change.

In `EDD_Payment` we support setting individual properties and then calling `save()`, like this:

```php
$payment = new EDD_Payment( 34 );
$payment->total = 35;
$payment->save();
```

It shouldn't take a ton of work to make that work with `EDD_Discount` and it would make the behavior consistent. So question for @sunnyratilal @cklosowski is "should we?"

sunnyratilal:
Definitely agree that it the behaviour should be similar to that of `EDD_Payment`.

It shouldn't be too big of a change - it's effectively just creating another `add()` method which makes use of object variables in lieu of a passed array of arguments.

pippinsplugins:
Let's keep `add()` and `update()` as is so they can also be used, unless either of you disagree?

sunnyratilal:
Agreed - we should keep them as is for backwards compatibility reasons.

`edd_store_discount()` currently calls the `add()` method which requires an array to be passed. We should only alter the `add()` method when we move to custom tables.

sunnyratilal:
`save()` method has been created and `EDD_Discount` now has similar behaviour to that of `EDD_Payment`

pippinsplugins:
[...]

PR #5318, opened 2016-12-26
EDD_Discount - Issue/5217

pippinsplugins:
#5217


includes/class-edd-discount.php
```diff
@@ -0,0 +1,583 @@
...
+	/**
+	 * Retrieve the name of the discount.
+	 *
+	 * @since 2.7
+	 * @access public
+	 *
+	 * @return string Name of the download.
+	 */
+	public function get_name() {
+		$this->name = get_the_title( $this->ID );
```
pippinsplugins: This should check if the class property is null and only retrieve the title if it is null. That will ensure calling `$discount->get_name()` multiple times only does a single query.

Applies here and throughout the class.

includes/class-edd-discount.php
```diff
@@ -0,0 +1,583 @@
...
+
+		if ( ! is_string( $field ) ) {
+			return false;
+		}
+
+		switch ( strtolower( $field ) ) {
+			case 'code':
+				break;
+
+			case 'id':
```
chriscct7: We should be really careful about this. Do we really need to allow post ID querying? If we're moving to a custom table this could get really confusing if its just called "id". Imagine a year from now if these are stored in custom tables. Is this a row ID in the db table or the post ID, despite the fact we won't be storing them in posts anymore.

sunnyratilal: This is something I had a few concerns about too - currently, only part of the `edd_get_discount_by()` function has been ported over.  I would remove the `id` from the `switch` and only keep it `edd_get_discount_by()`.

So if a discount is requested by it's `ID` (via `edd_get_discount_by()`), we only create a new instance of `EDD_Discount` and return that.

includes/class-edd-discount.php
```diff
@@ -0,0 +1,1743 @@
...
+			 *     @type array  $excluded_products Product IDs excluded from this discount code.
+			 *     @type bool   $is_not_global     If the discount code is not globally applied to all products. Defaults to false.
+			 *     @type bool   $is_single_use     If the code cannot be used more than once per customer. Defaults to false.
+			 * }
+			 * @param int $ID The ID of the discount that was inserted.
+			 */
+			do_action( 'edd_post_insert_discount', $meta, $this->ID );
+
+			// Discount code created
+			return $this->ID;
```
pippinsplugins: We need to re-run `setup_discount()` before returning (here and with update()) to ensure the properties are updated with the proper values.

sunnyratilal: Addressed in c5100e0d332b2c3ac354849638ab99b4087d5864

cklosowski:
Merged in `release/2.7` so that unit tests are updated to work with WP Core 4.7

PR #5340, opened 2017-01-04
Fixing unit tests for useage of _delete_all_data

cklosowski:
@sunnyratilal @pippinsplugins would love you two to look this over. It fixes the Unit tests for WordPress 4.7. They introduced a `_delete_all_data` which runs after each class is completed, so naturally, we have to run `edd_install()` before each class starts it's run.

While doing this, I found a unique case where the installer wouldn't install a missing page, since all we did was check if the `purchase_page` was set, so I did a little work and made `edd_install()` check each of our pages individually and crates them if any number of them are missing in the options, it also makes sure that the ID set for the page, exists. If not it recreates a page

### DIFF
--- a/includes/install.php
+++ b/includes/install.php
@@ -84,7 +84,7 @@ function edd_run_install() {
 	$current_options = get_option( 'edd_settings', array() );
 
 	// Checks if the purchase page option exists
-	if ( ! array_key_exists( 'purchase_page', $current_options ) ) {
+	if ( ! array_key_exists( 'purchase_page', $current_options ) || empty( get_post( $current_options['purchase_page'] ) ) ) {
 		// Checkout Page
 		$checkout = wp_insert_post(
 			array(
@@ -100,7 +100,9 @@ function edd_run_install() {
 		$options['purchase_page'] = $checkout;
 	}
 
-	if ( ! array_key_exists( 'success_page', $current_options ) ) {
+	$checkout = isset( $checkout ) ? $checkout : $current_options['purchase_page'];
+
+	if ( ! array_key_exists( 'success_page', $current_options ) || empty( get_post( $current_options['success_page'] ) ) ) {
 		// Purchase Confirmation (Success) Page
 		$success = wp_insert_post(
 			array(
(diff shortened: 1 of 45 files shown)


sunnyratilal:
All looks good to me! Great work @cklosowski!

I'd say let's get this merged ASAP and fast-forward our issue branches against release/2.7 to ensure unit tests work for existing and new PRs

pippinsplugins:
@cklosowski could we please reduce some of the code duplication by adding our own class that extends `WP_UnitTestCase` and have it take care of the `edd_install()`? Each of our test classes can then extend the new `EDD_UnitTestCase`.

cklosowski:
Asked drew about that and it's possible that the instantiation does not
work that way based off the base class. I'll try, but they way they do it,
it only nests one deep I believe.

On Jan 4, 2017 8:13 AM, "Pippin Williamson" <notifications@github.com>
wrote:

@cklosowski <https://github.com/cklosowski> could we please reduce some of
the code duplication by adding our own class that extends WP_UnitTestCase
and have it take care of the edd_install()? Each of our test classes can
then extend the new EDD_UnitTestCase.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5340#issuecomment-270393985>,
or mute the thread

cklosowski:
@pippinsplugins the exact quote was:
`Depends on how you’re extending the core unit test case class. That pair of static methods only works one extension deep iirc.`

So I'll have to verify that. Also, I'm still working through if we actually NEED to run install on each of our unit tests class. If we don't, then we should limit it to only classes that need it, and therefore, not build a base test class.

cklosowski:
Woot! It DOES work actually. Code duplication reduced.

cklosowski:
ok @pippinsplugins ready when you are! once merged I'm going to go into all open PRs for 2.7 and update the branches to have the unit tests so we can have confidence in their passing.

pippinsplugins:
Excellent!

Issue #3516, opened 2015-06-17
Disable buy now buttons if multiple payment gateways are enabled


pippinsplugins:
It doesn't make sense to support buy now buttons if multiple payment gateways are activated.

chriscct7:
We have 3 options:
- Disable buy now
- Let user pick which gateway to use for buy now (which might not be a bad idea) given we're going to have 2 core gateways with buy now support
- Show multiple buy now buttons (like when you go to a site and they have a buy now with paypal, buy now with amazon and add to cart button.

pippinsplugins:
Punting to 2.6. This isn't a tiny feature and we haven't had any recent tickets from this.

mintplugins:
Customer Ticker regarding this here:
https://secure.helpscout.net/conversation/213683592/36044/?folderId=180499

sunnyratilal:
@easydigitaldownloads/core-devs Could we vote on the best way forward with this please based on the three options @chriscct7 has provided.

My personal preference is option 1.

pippinsplugins:
I vote option 1 at this time.

SDavisMedia:
I'm with option 1 for now. Option 2 with a super sweet modal might be nice in the future.

pippinsplugins:
>>>Will we be incorporating any note/message to admins to communicate that buy now buttons are disabled when they turn on that second gateway?

Instead of just hiding the settings, I think we should gray them out and provide a tooltip with an explanation of why they are disabled.

PR #5341, opened 2017-01-04
Disable buy now text and checkboxes when multiple gateways enabled #3516

cklosowski:
#3516

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -898,6 +898,12 @@ function edd_get_registered_settings() {
 		)
 	);
 
+	if ( ! edd_shop_supports_buy_now() ) {
+		$edd_settings['misc']['button_text']['buy_now_text']['disabled']      = true;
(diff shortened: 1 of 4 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -898,6 +898,12 @@ function edd_get_registered_settings() {
 		)
 	);
 
+	if ( ! edd_shop_supports_buy_now() ) {
+		$edd_settings['misc']['button_text']['buy_now_text']['disabled']      = true;
```
pippinsplugins: Any reason you're adding these disabled input fields instead of graying out the `Button Behavior` option on the Download metabox?

cklosowski:
We are doing both.

includes/admin/settings/register-settings.php
cklosowski: @pippinsplugins we're doing both. This way, if they look at the item itself they see it grayed out (with updated tooltip), and if they go to the button text b/c they think the text isn't set correctly, they see the reason why it's not changing.

pippinsplugins: Ah! I see what's happening now. I like it.

Issue #2714, opened 2014-10-13
Add option to complete abandoned/pending purchases


ghost1227:
Users should be able to complete abandoned/pending purchases on subsequent visits. In order to facilitate this, a status of 'processing' should be added which will be used by eChecks instead of the 'pending' status used now. This will allow pending purchases to be completed without risking affecting eCheck purchases which are pending processing.

@pippinsplugins has indicated that WooCommerce does this well and we should take inspiration from them.

pippinsplugins:
Let's hold off on this until https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/2667 is completed so that we are not refactoring payment creation / updates twice.

danieliser:
Isn't this partially already covered by several extensions aimed at recovering them? Maybe would be well suited as a feature for those plugins instead. Like the idea to clean up the pending transfers though.

Also consider starting down the path with this will likely eventually result in those extensions being merged into core as first comes showing people those transactions next comes reminding them by email since that would be a simple step with this in place as well.

pippinsplugins:
No, not really. The current extensions generate new purchase records. We want to complete the existing pending record.

danieliser:
Oh sorry, I was under the assumption from their descriptions that they took abandoned and pending carts and notified users to come back to finish them.

pippinsplugins:
They do notify them but they just recreate the purchase.

danieliser:
Ok then ignore my previous statements.

KingYes:
@pippinsplugins: For now, can I continue pending purchase manually?

pippinsplugins:
@KingYes

KingYes:
How?

pippinsplugins:
Oh, sorry, I just reread what you wrote.

No, that's not supported yet.

KingYes:
Okay.
Do you know  when it's will be supported?

pippinsplugins:
Not yet.

pippinsplugins:
This is too large of a feature to finish up for 2.4 since we're just a few days away from beta. Punting to 2.5.

cklosowski:
Blocked by #2667

cklosowski:
As much as we want this in 2.5, we're just running to close to the `release/2.5` string freeze for translation so I'm going to punt this to 2.6.

arielk:
I'm waiting for this feature for too long :(

pippinsplugins:
[...]

PR #5351, opened 2017-01-06
Complete abandoned and pending payments #2714

cklosowski:
#2714

### DIFF
--- a/includes/admin/payments/view-order-details.php
+++ b/includes/admin/payments/view-order-details.php
@@ -90,6 +90,16 @@ $customer       = new EDD_Customer( $payment->customer_id );
 												?>
 												<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php echo $status_help; ?>"></span>
 											</p>
+
+											<?php if ( $payment->is_recoverable() ) : ?>
+											<p>
+												<span class="label"><?php _e( 'Recovery URL', 'easy-digital-downloads' ); ?>:</span>
+												<?php $recover_help = __( 'Pending and abandoned payments can be resumed by the customer, using this custom URL. Payments can be resumed only when they do not have a transaction ID from the gateway.', 'easy-digital-downloads' ); ?>
+												<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php echo $recover_help; ?>"></span>
+
+												<input type="text" class="large-text" readonly="readonly" disabled="disabled" value="<?php echo $payment->get_recovery_url(); ?>" />
+											</p>
+											<?php endif; ?>
 										</div>
 
 										<div class="edd-admin-box-inside">
--- a/includes/admin/payments/view-order-details.php
+++ b/includes/admin/payments/view-order-details.php
@@ -97,7 +97,7 @@ $customer       = new EDD_Customer( $payment->customer_id );
 												<?php $recover_help = __( 'Pending and abandoned payments can be resumed by the customer, using this custom URL. Payments can be resumed only when they do not have a transaction ID from the gateway.', 'easy-digital-downloads' ); ?>
 												<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php echo $recover_help; ?>"></span>
 
-												<input type="text" class="large-text" readonly="readonly" disabled="disabled" value="<?php echo $payment->get_recovery_url(); ?>" />
+												<input type="text" class="large-text" readonly="readonly" value="<?php echo $payment->get_recovery_url(); ?>" />
 											</p>
 											<?php endif; ?>
 										</div>
--- a/includes/payments/actions.php
+++ b/includes/payments/actions.php
@@ -357,6 +357,10 @@ add_action( 'edd_daily_scheduled_events', 'edd_cleanup_stats_transients' );
  */
 function edd_recover_payment() {
 
+	if ( empty( $_GET['payment_id'] ) ) {
+		return;
+	}
+
 	$payment = new EDD_Payment( $_GET['payment_id'] );
 	if ( $payment->ID !== (int) $_GET['payment_id'] ) {
 		return;
@@ -364,6 +368,12 @@ function edd_recover_payment() {
 
 	if ( ! $payment->is_recoverable() ) {
 		return;
+	}
+
+	if ( is_user_logged_in() && $payment->user_id != get_current_user_id() ) {
+		$redirect = get_permalink( edd_get_option( 'purchase_history_page' ) );
+		edd_set_error( 'edd-payment-recovery-user-mismatch', __( 'Error resuming payment.', 'easy-digital-downloads' ) );
+		wp_redirect( $redirect );
 	}
 
 	// Empty out the cart.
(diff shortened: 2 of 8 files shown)


templates/history-purchases.php
```diff
@@ -8,38 +8,46 @@
...
+	if ( $payments ) :
 		do_action( 'edd_before_purchase_history' ); ?>
 		<table id="edd_user_history" class="edd-table">
 			<thead>
 				<tr class="edd_purchase_row">
 					<?php do_action('edd_purchase_history_header_before'); ?>
 					<th class="edd_purchase_id"><?php _e('ID','easy-digital-downloads' ); ?></th>
 					<th class="edd_purchase_date"><?php _e('Date','easy-digital-downloads' ); ?></th>
-					<th class="edd_purchase_amount"><?php _e('Amount','easy-digital-downloads' ); ?></th>
+					<th class="edd_purchase_amount"><?php _e('Amount','easy-digital-dxownloads' ); ?></th>
```
pippinsplugins: Incorrect text domain here.

templates/history-purchases.php
```diff
@@ -8,38 +8,46 @@
...
 					</td>
 					<td class="edd_purchase_details">
-						<?php if( $post->post_status != 'publish' ) : ?>
-						<span class="edd_purchase_status <?php echo $post->post_status; ?>"><?php echo edd_get_payment_status( $post, true ); ?></span>
-						<a href="<?php echo esc_url( add_query_arg( 'payment_key', edd_get_payment_key( $post->ID ), edd_get_success_page_uri() ) ); ?>">&raquo;</a>
+						<?php if( $payment->status != 'publish' ) : ?>
+							<span class="edd_purchase_status <?php echo $payment->status; ?>"><?php echo $payment->status_nicename; ?></span>
+							<?php if ( $payment->is_recoverable() ) : ?>
+								<?php
+								$query_args = array(
```
pippinsplugins: Let's turn this into a function `edd_get_payment_recovery_url()` (or a method of EDD_Payment). 

That will make it much easier for devs to extend and for us to reuse it (such as on View Order Details).

cklosowski: See my reply about why they are nonce'd

includes/admin/payments/view-order-details.php
```diff
@@ -90,6 +90,16 @@
...
 												<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php echo $status_help; ?>"></span>
 											</p>
+
+											<?php if ( $payment->is_recoverable() ) : ?>
+											<p>
+												<span class="label"><?php _e( 'Recovery URL', 'easy-digital-downloads' ); ?>:</span>
+												<?php $recover_help = __( 'Pending and abandoned payments can be resumed by the customer, using this custom URL. Payments can be resumed only when they do not have a transaction ID from the gateway.', 'easy-digital-downloads' ); ?>
+												<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php echo $recover_help; ?>"></span>
+
+												<input type="text" class="large-text" readonly="readonly" disabled="disabled" value="<?php echo $payment->get_recovery_url(); ?>" />
```
pippinsplugins: Disabled means we can't click into it to copy the URL. Should just be read only.

includes/payments/actions.php
```diff
@@ -348,3 +348,108 @@ function edd_cleanup_stats_transients() {
...
+
+	$redirect_args = array( 'payment-mode' => $payment->gateway );
+	$redirect      = add_query_arg( $redirect_args, edd_get_checkout_uri() );
+	wp_redirect( $redirect );
+	exit;
+}
+add_action( 'edd_recover_payment', 'edd_recover_payment' );
+
+/**
+ * If the payment trying to be recovered has a User ID associated with it, we need them to login.
```
pippinsplugins: `login` should be `log in`

includes/payments/actions.php
```diff
@@ -348,3 +348,108 @@ function edd_cleanup_stats_transients() {
...
+ *
+ * @since  2.7
+ * @return void
+ */
+function edd_recovery_force_login_fields() {
+	$resuming_payment = EDD()->session->get( 'edd_resume_payment' );
+	if ( $resuming_payment ) {
+		$payment = new EDD_Payment( $resuming_payment );
+		if ( ! empty( $payment->user_id ) && ( ! is_user_logged_in() || $payment->user_id != get_current_user_id() ) ) {
```
pippinsplugins: If you are logged in and try and resume a payment that is connected to a different user, you will see a log in form.

That's weird.

I don't think we should allow someone to resume a payment that's not their's, but I also don't think showing the log in form to a logged in user is right.

includes/payments/class-edd-payment.php
```diff
@@ -1657,6 +1678,27 @@ public function update_meta( $meta_key = '', $meta_value = '', $prev_value = ''
 		return update_post_meta( $this->ID, $meta_key, $meta_value, $prev_value );
 	}
 
+	public function is_recoverable() {
```
pippinsplugins: Needs PHPdoc

includes/payments/class-edd-payment.php
```diff
@@ -1657,6 +1678,27 @@ public function update_meta( $meta_key = '', $meta_value = '', $prev_value = ''
...
+
+		$recoverable_statuses = apply_filters( 'edd_recoverable_payment_statuses', array( 'pending', 'abandoned' ) );
+		if ( in_array( $this->status, $recoverable_statuses ) && empty( $this->transaction_id ) ) {
+			$recoverable = true;
+		}
+
+		return $recoverable;
+	}
+
+	public function get_recovery_url() {
```
pippinsplugins: Needs PHPdoc

includes/payments/functions.php
```diff
@@ -110,7 +110,30 @@ function edd_insert_payment( $payment_data = array() ) {
 		return false;
 	}
 
-	$payment = new EDD_Payment();
+	$existing_payment = EDD()->session->get( 'edd_resume_payment' );
+	if ( ! empty( $existing_payment ) ) {
+
+		$payment = new EDD_Payment( $existing_payment );
+
```
pippinsplugins: We need to check that the payment is recoverable before continuing here.

includes/admin/payments/view-order-details.php
cklosowski: Disabled works fine in Chrome. I was using it to test. I'll remove it so other browsers don't fail.

includes/payments/actions.php
cklosowski: I was following what we did for EDD Recurring. When a log in is required for those users, we display the login, I see no reason we shouldn't use the same behavior.

cklosowski: I think from a user perspective I'd be more frustrated if I went all the way through the checkout, got an error (like the one we show if they are not logged in to the user that the payment is for), and all my checkout details are gone.

This notice and form allows us to let them login, without going through the process of typing in all their information, and just end right back up here when done.

includes/payments/actions.php
```diff
@@ -348,3 +348,108 @@ function edd_cleanup_stats_transients() {
...
 add_action( 'edd_daily_scheduled_events', 'edd_cleanup_stats_transients' );
+
+/**
+ * Process an attempt to complete a recoverable payment.
+ *
+ * @since  2.7
+ * @return void
+ */
+function edd_recover_payment() {
+
```
pippinsplugins: Let's go ahead and bail early if `$_GET['payment_id']` is empty.

includes/payments/actions.php
pippinsplugins: I think maybe you misunderstood my comment.

With the current iteration it is possible to see a log in form while _already_ logged in. That should not happen.

cklosowski: Ah I did misunderstand, sorry about that.

includes/payments/actions.php
```diff
@@ -366,6 +370,12 @@ function edd_recover_payment() {
 		return;
 	}
 
+	if ( is_user_logged_in() && $payment->user_id != get_current_user_id() ) {
+		$redirect = get_permalink( edd_get_option( 'purchase_history_page' ) );
+		edd_set_error( 'edd-payment-recovery-user-mismatch', __( 'Error resuming payment.', 'easy-digital-downloads' ) );
```
pippinsplugins: Let's add more explanation in the error, otherwise it will be had for store admins to assist customers and for them to understand why the error was shown.

includes/payments/actions.php
```diff
@@ -366,6 +370,12 @@ function edd_recover_payment() {
...
+		wp_redirect( $redirect );
```
pippinsplugins: This needs an `exit`

Issue #5338, opened 2017-01-03
Undefined index notice from dashboard widget notice


Shelob9:
Stack trace:

```
[03-Jan-2017 18:00:38 UTC] PHP Notice:  Undefined index: id in /srv/www/new-cf/public/content/plugins/easy-digital-downloads/includes/admin/dashboard-widgets.php on line 178
[03-Jan-2017 18:00:38 UTC] PHP Stack trace:
[03-Jan-2017 18:00:38 UTC] PHP   1. {main}() /srv/www/new-cf/public/wp/wp-admin/admin-ajax.php:0
[03-Jan-2017 18:00:38 UTC] PHP   2. do_action($tag = 'wp_ajax_edd_load_dashboard_widget', $arg = *uninitialized*) /srv/www/new-cf/public/wp/wp-admin/admin-ajax.php:91
[03-Jan-2017 18:00:38 UTC] PHP   3. WP_Hook->do_action($args = array (0 => '')) /srv/www/new-cf/public/wp/wp-includes/plugin.php:453
[03-Jan-2017 18:00:38 UTC] PHP   4. WP_Hook->apply_filters($value = '', $args = array (0 => '')) /srv/www/new-cf/public/wp/wp-includes/class-wp-hook.php:323
[03-Jan-2017 18:00:38 UTC] PHP   5. edd_load_dashboard_sales_widget('') /srv/www/new-cf/public/wp/wp-includes/class-wp-hook.php:298
[03-Jan-2017 18:00:38 UTC] PHP Notice:  Undefined index: id in /srv/www/new-cf/public/content/plugins/easy-digital-downloads/includes/admin/dashboard-widgets.php on line 178
```

pippinsplugins:
I've seen this a few times too.

pippinsplugins:
PR https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5352

PR #5352, opened 2017-01-07
Issue #5338

sanketio:
Fix - Undefined index `id` for dashboard widget #5338


includes/admin/dashboard-widgets.php
```diff
@@ -175,7 +175,7 @@ function edd_load_dashboard_sales_widget( ) {
 									<?php echo get_the_title( $payment->ID ) ?>
 									&mdash; <?php echo $payment->user_info['email'] ?>
 								</a>
-								<?php if ( $payment->user_info['id'] > 0 ) {
+								<?php if ( ! empty( $payment->user_info['id'] ) && ( $payment->user_info['id'] > 0 ) ) {
```
pippinsplugins: This fix works but let's go a step further and use `$payment->user_id` instead. That helps move us away from the `user_info` array and it's value will always be set.

sanketio: @pippinsplugins 
Should I replace for every instance in that portion or just the condition?

pippinsplugins: Just this one.

includes/admin/dashboard-widgets.php
```diff
@@ -175,7 +175,7 @@ function edd_load_dashboard_sales_widget( ) {
 									<?php echo get_the_title( $payment->ID ) ?>
 									&mdash; <?php echo $payment->user_info['email'] ?>
 								</a>
-								<?php if ( ! empty( $payment->user_info['id'] ) && ( $payment->user_info['id'] > 0 ) ) {
+								<?php if ( ! empty( $payment->user_id ) && ( $payment->user_info['id'] > 0 ) ) {
```
pippinsplugins: `&& ( $payment->user_info['id'] > 0 )` needs to be adjusted too.

cklosowski:
This PR has a dirty commit history. I'm going to make a new one, preserving the commit history of @sanketio so it's a cleaner merge changeset

PR #5365, opened 2017-01-13
Issue/5166

sunnyratilal:
Updated PR for #5166 with fixed unit tests


cklosowski:
Thanks for fixing up the Unit tests.

Issue #4284, opened 2016-02-10
Session cookies are created on every page load


mindctrl:
We've had two reports recently about the fact that a session cookie gets created on every page load, making it hard to cache. One came through a Pagely customer and another was from Pressjitsu.

One report:

> The use of sessions in EDD prevents any page from being cached because it's issuing a unique cookie to every visitor. It would be best if EDD initialized this unique session ONLY when the user has taken any sort of action, like add something to the cart. That way, the majority of the users can enjoy fast cached responses throughout the whole website, while those with intent to buy something will get their unique cookies.

Other report:

> The site is setting edd_wp_session cookies for every visitor, which makes thing a lot more painful, as nothing could be cached at the Nginx level & every request is dynamic.

If you use `EDD_USE_PHP_SESSIONS` set to false and fall back to `WP_Session`, you always get a cookie because it creates one in `__construct()`. [Link](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/2.5.6/includes/libraries/class-wp-session.php#L105).

cklosowski:
We've got to make sure if we make any changes to this, that it doesn't break extensions that require the sessions, things like User History.

pippinsplugins:
Let's try and resolve this in a point release that has almost nothing else in it. For now, milestoning for 2.5.9

cklosowski:
looking into this I can almost get there. I can get sessions/cookies to only be created when adding to the cart, however for some reason the cookies don't carry over to page refreshes. All AJAX requests know about the current cart contents, but the front of site has no idea about it, which is strange.

chriscct7:
This is going to break any extensions that rely on EDD sessions from the sound of it outside of the normal checkout flow

pippinsplugins:
@chriscct7 Yes it will and we'll have to deal with that. Unfortunately, breaking caching on _every_ single page of _every single_ EDD site is far worse.

cklosowski:
Every time I try and get this done, something major breaks in the cart system. I've taken a few stabs at it between getting other issues resolved. I'm going to punt (for now) and we can look to get this in a future release.

It's not that it's not important, it's just important to get right.

bmoredrew:
Thanks for the effort everyone. Is there something we can do on an individual site level? My client has a pretty large site in terms of traffic that just falls over on Pagely when we try to implement EDD, so all the pieces are in place just kind of sitting there for a one day hopeful fix/use.

[...]

PR #5381, opened 2017-01-15
Issue/4284 - Rules for sessions being created

sunnyratilal:
#4284

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -332,14 +332,13 @@ class EDD_Session {
 		 *  2. An EDD AJAX request is currently running
 		 *  3. If we are on the checkout page, or any child pages
 		 */
-		if ( isset( $_COOKIE['edd_items_in_cart'] ) || ( defined( 'DOING_AJAX' ) && DOING_AJAX && false !== strpos( $_REQUEST['action'], 'edd_' ) ) || ( false !== strpos( $_SERVER['REQUEST_URI'], '/checkout' ) ) {
+		if ( isset( $_COOKIE['edd_items_in_cart'] ) || ( defined( 'DOING_AJAX' ) && DOING_AJAX && false !== strpos( $_REQUEST['action'], 'edd_' ) ) || ( is_page( edd_get_option( 'purchase_page' ) ) ) ) {
 			$start_session = true;
 		} else {
 			$start_session = false;
 		}
 
 		return apply_filters( 'edd_start_session', $start_session );
-
 	}
 
 	/**
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -99,11 +99,7 @@ class EDD_Session {
 
 		}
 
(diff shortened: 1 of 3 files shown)


includes/class-edd-session.php
```diff
@@ -310,23 +310,32 @@ public function use_cart_cookie() {
...
 			}
+		}
 
+		/**
+		 * Sessions should only be allowed if:
+		 * 	1. A user has an item in their cart
+		 *  2. An EDD AJAX request is currently running
+		 *  3. If we are on the checkout page, or any child pages
+		 */
+		if ( isset( $_COOKIE['edd_items_in_cart'] ) || ( defined( 'DOING_AJAX' ) && DOING_AJAX && false !== strpos( $_REQUEST['action'], 'edd_' ) ) || ( false !== strpos( $_SERVER['REQUEST_URI'], '/checkout' ) ) {
```
pippinsplugins: The checkout screen isn't always `/checkout`. We'll need to detect when it's been changed to something else.

chriscct7: couldn't you just use edd_is_checkout()?

pippinsplugins: @chriscct7 no because it's too early in the load process

chriscct7: Somehow that function is going to need to be used. It's the only reliable way to determine the checkout page in EDD because you can filter it at the bottom of the function.

pippinsplugins: We could pull the page ID and then grab the slug from that. Or perhaps we should have a dedicated option that contains the checkout slug.

sunnyratilal: I'll get the PR updated

includes/class-edd-session.php
```diff
@@ -332,14 +332,13 @@ public function should_start_session() {
 		 *  2. An EDD AJAX request is currently running
 		 *  3. If we are on the checkout page, or any child pages
 		 */
-		if ( isset( $_COOKIE['edd_items_in_cart'] ) || ( defined( 'DOING_AJAX' ) && DOING_AJAX && false !== strpos( $_REQUEST['action'], 'edd_' ) ) || ( false !== strpos( $_SERVER['REQUEST_URI'], '/checkout' ) ) {
+		if ( isset( $_COOKIE['edd_items_in_cart'] ) || ( defined( 'DOING_AJAX' ) && DOING_AJAX && false !== strpos( $_REQUEST['action'], 'edd_' ) ) || ( is_page( edd_get_option( 'purchase_page' ) ) ) ) {
```
pippinsplugins: `is_page()` cannot be used this early.

includes/class-edd-session.php
chriscct7: We should at minimum run it over the filter in edd_is_checkout somehow, or at least we should publish a dev note on this as there exists a filter in EDD to define the checkout page via a filter, so its likely people are using it

Issue #5325, opened 2016-12-29
Link to Discount edit screen from payment record


SDavisMedia:
It would be great if the discount code(s) on the payment record were linked to their edit screens, as the record is a hub for everything else related to the payment.

<img width="307" alt="screenshot 2016-12-29 10 47 43" src="https://cloud.githubusercontent.com/assets/2359131/21552128/6203a422-cdb4-11e6-9387-a2a57672f92f.png">

If everyone agrees, this would be a great addition once `EDD_Discount` is ready. https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5217

sunnyratilal:
Yep, it's something small but a nice addition at the same time. We'll write the code up once #5217 is merged.

sunnyratilal:
In #5217, I've added an `edit_url()` function to make this even easier.

sunnyratilal:
Commit ref: 30c42bc7e817e7b0cfd80b3ad7fc9b78ffc0a1c8

cklosowski:
Reopening for some visual updates.

sunnyratilal:
PR tested and I'm happy with it 👍

SDavisMedia:
Beautiful. Thumbs up from me as well. :+1:

sunnyratilal:
Merged into `release/2.7`

pippinsplugins:
Very nice!

ghost1227:
@SDavisMedia Improved discount list ftw!

<img width="288" alt="screen shot 2017-01-17 at 10 24 35 am" src="https://cloud.githubusercontent.com/assets/36433/22029142/398d81b2-dc9f-11e6-8d3a-96f4b26bc5c0.png">

PR #5395, opened 2017-01-17
Rework discount list

evertiro:
RE: #5325

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -136,6 +136,23 @@ a.edd-delete:hover {
 	width: 15px;
 }
 
+.edd-order-discount-item pre {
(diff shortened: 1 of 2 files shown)


SDavisMedia:
@evertiro I do like this but I think it could end up causing unnecessary trouble. For example, the more complex display won't handle long discount codes very well.

screenshot 2018-02-14 10 06 36

Having floated elements involved definitely complicates things in a tight space. An obvious solution is to not float the link icons and let them stay inline. But then at that point you end up with a somewhat random looking icon display _and_ pretty much the same thing as the enhancement we already made.

I think it's best to stick with inline plain text links.

screenshot 2018-02-14 10 09 04

PR #5428, opened 2017-01-31
Make file protected instead of private

### LINKED ISSUE #5427: Member has private access
RE: https://github.com/easydigitaldownloads/EDD-Commissions/issues/159

In the EDD_Batch_Export class we are defining `$file` as private, but attempting to access it in extensions. This should be made protected.

ghost1227:
Fixes #5427


cklosowski:
@ghost1227 I assume you're wanting to set this to protected b/c your `extend` this class? is there a need to extend in your case?

cklosowski:
Seems to work fine with the other export methods.

Issue #5465, opened 2017-02-14
edd_register_batch_exporters may fire before includes in extensions


mintplugins:
The EDD hook which sets up batch exporters fires upon plugins_loaded. However, if an extension also includes extension-integration scripts upon plugins_loaded, if the integration has a batch exporter, it won't be included before edd_register_batch_exporters is fired. Thus it won't get registered and causes a tough-to-track 500 error in the ajax call while doing a batch export. 

To fix this, I propose setting the priority of this hook to 99:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/admin/reporting/export/export-actions.php#L80

pippinsplugins:
We could lower the priority of `edd_register_batch_exporters` so it fires earlier on `plugins_loaded`.

mintplugins:
That's an even better idea 👍

mintplugins:
Actually on second though, I don't think that would work. 

We need edd_register_batch_exporters to fire later as opposed to earlier.

ghost1227:
Derp. Commented on the PR instead of the issue. @pippinsplugins His method works well in my tests.

pippinsplugins:
This breaks the `EDD_Batch_Sales_Export` export option/

pippinsplugins:
Nevermind, it was already broken on `master`.

ghost1227:
99 little bugs in the code, 99 little bugs...

PR #5473, opened 2017-02-15
Set priority of edd_register_batch_exporters to 99

mintplugins:
https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5465


ghost1227:
I've been working off this since @mintplugins figured it out yesterday. Works great in all the tests I've thrown at it! @pippinsplugins is there any reason we _shouldn't_ do it this way?

PR #5476, opened 2017-02-15
Register the sales export batch process. Fixes #5475

### LINKED ISSUE #5475: EDD_Batch_Sales_Export broken. Class not found error.
When attempting to export sale records with the `EDD_Batch_Sales_Export` class, a `class not found` fatal error is returned.

pippinsplugins:
#5475

### DIFF
--- a/includes/admin/reporting/export/export-actions.php
+++ b/includes/admin/reporting/export/export-actions.php
@@ -177,6 +177,29 @@ function edd_include_file_downloads_batch_processer( $class ) {
 }
 
 /**
+ * Register the sales batch exporter
+ *
+ * @since  2.7
+ */
+function edd_register_sales_export_batch_export() {
+	add_action( 'edd_batch_export_class_include', 'edd_include_sales_export_batch_processer', 10, 1 );
+}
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/export/export-actions.php
```diff
@@ -177,6 +177,29 @@ function edd_include_file_downloads_batch_processer( $class ) {
...
 
 /**
+ * Register the sales batch exporter
+ *
+ * @since  2.7
+ */
+function edd_register_sales_export_batch_export() {
+	add_action( 'edd_batch_export_class_include', 'edd_include_sales_export_batch_processer', 10, 1 );
+}
+add_action( 'edd_register_batch_exporter', 'edd_register_sales_export_batch_export', 10 );
```
cklosowski: Should we drop this to a lower priority like `9` or something? That way no matter what the load order is of the plugins on the site, we're running before default priority?

pippinsplugins: Does it matter? Is there a case you've thought of where the load order for export-specific classes would fail?

cklosowski: Not off hand, just trying to be preventitive.

PR #5487, opened 2017-02-16
Ensure edd_update_payment_status uses EDD_Payment for hooks

JeroenSormani:
Heya!

I noticed in regards of one of my plugins that not always the payment status update hooks were triggered that are available in the `EDD_Payment()->update_status()` method. 

This ensures it goes through the right route so it triggers those hooks

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -268,10 +268,7 @@ function edd_update_payment_status( $payment_id = 0, $new_status = 'publish' ) {
 	$payment = new EDD_Payment( $payment_id );
 
 	if( $payment && $payment->ID > 0 ) {
-
-		$payment->status = $new_status;
-		$updated = $payment->save();
-
+		$updated = $payment->update_status( $new_status );
 	}
 
 	return $updated;


JeroenSormani:
@pippinsplugins @cklosowski any update on this?
I've just had another user report an issue with me that I tracked back to the `edd_update_payment_status()` func not triggering hooks. 

It also appears that making this change will add additional benefits such as trigger the re-count of sales/stats when doing a refund through PayPal directly.

pippinsplugins:
I'm fine with this change, though I am concerned about the underlying reason this is necessary. It implies that logic is incorrect in our save() method that detects when the status has changed.

@cklosowski how about we merge as is and then investigate the underlying reason further later on?

JeroenSormani:
Might've been wrong too, I always forget about the magic methods (and why I dislike them) and how it saves as it seems like properties are assigned directly vs. going through the method.
 
Looking at the code it does appear to go through the `EDD_Payment::update_status()` method properly when performing the `save()`.

I'll close for now and re-investigate my issue.

Issue #5448, opened 2017-02-08
Introduce class `EDD_Customers_Query`


felixarntz:
Queries against the customers table are currently run in singular methods of the `EDD_DB_Customers` class. A dedicated query class like WordPress core's ones (for things like posts, comments,...) could improve the performance, extendability and possibilities to query without bloating the `EDD_DB_Customers` class itself. The `get_customers()` method and several other methods of the class could then become wrappers for making a query through an instance of `EDD_Customers_Query`. The new class would also bring enhanced caching for all queries (I was just wondering why https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-db-customers.php#L510 is in there --> were there issues with caching?).

I would suggest building the class after one of the more recent core classes, like `WP_Comment_Query` or `WP_Site_Query`. I'd be happy to work on it in a PR if this idea gets positive feedback.

pippinsplugins:
I _love_ this idea.

felixarntz:
Great! I'll start working on it this weekend or early next week.

pippinsplugins:
@felixarntz I've milestoned this for 2.8. We'll do some testing and see about getting it in. Thanks!

SDavisMedia:
Everything I'm checking seems to be good on `issue./5448`. Can't find any breaking output or functionality. :+1:

sunnyratilal:
Working well and as expected 👍

sunnyratilal:
Merged into `release/2.8`

PR #5514, opened 2017-02-25
Introduce class `EDD_Customer_Query`

felixarntz:
This PR introduces the new `EDD_Customer_Query` class (as described in #5448 before) and makes use of it in the `EDD_DB_Customers` methods `get_customer_by()`, `get_customers()` and `count()`.

The class uses a simple caching layer for its queries to boost performance, therefore a bit of other related functionality has been introduced as well (in particular the methods `set_last_changed()` and `get_last_changed()` in `EDD_DB_Customers`. A few new properties on the `EDD_DB_Customers` class have been introduced as well.

I tried to make the query class more flexible than the current query method, thus handled a few query variables differently. However, when creating a query instance through one of the methods that already existed before, all the arguments are automatically "translated" to the `EDD_Customer_Query` ones so that things are backward-compatible.

I did not do extensive testing yet, only basic tests in EDD, so that still needs to be done. So there still needs to be some work done for this PR, but I already opened it for first review/feedback.

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -259,6 +259,7 @@ final class Easy_Digital_Downloads {
 		require_once EDD_PLUGIN_DIR . 'includes/class-edd-db.php';
 		require_once EDD_PLUGIN_DIR . 'includes/class-edd-db-customers.php';
 		require_once EDD_PLUGIN_DIR . 'includes/class-edd-db-customer-meta.php';
+		require_once EDD_PLUGIN_DIR . 'includes/class-edd-customer-query.php';
 		require_once EDD_PLUGIN_DIR . 'includes/class-edd-customer.php';
 		require_once EDD_PLUGIN_DIR . 'includes/class-edd-discount.php';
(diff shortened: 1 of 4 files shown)


felixarntz:
I noticed several unit tests started failing after this. I fixed a few issues I encountered and looked into what might cause these issues, but couldn't figure it out yet. When using the class and its surrounding functions in the admin, things worked fine.

I'm not familiar with how the EDD test suite is set up, some help would be much appreciated. :)

sunnyratilal:
Looks like a corrupt `MySQL` table

cklosowski:
Changed the base to `release/2.8`

cklosowski:
Going to start looking into this PR, might be necessary to close it and pull it into an issue branch within the organization, but rest assured @felixarntz your commit history will stay in tact if I do that. I appreciate the work you've done here. 👍 🔥

felixarntz:
@cklosowski Thanks! There was one issue I couldn't resolve when I wrote the code - let me know if you have questions about anything.

cklosowski:
Opening up as an issue branch to fix a few things.

Issue #5307, opened 2016-12-20
Proper validation for zip/postal code when billing country is Ireland.


pderksen:
Not sure if I'm 100% correct on this, but I believe we shouldn't require a zip code for Irish customers. Or possibly the validation should require but is not doing so correctly.

I've had several (potential) customers send this complaint to me the last few months.

One customer:

> I live in Ireland and here we do not have 5 digits zip/postal code.
> Tried with 5 zeros but kept failing zip input validation.

Another:

> Irish post codes are like the following: T12 Y275 - which is the one for my own card.
> When I enter this, I get: Error: The zip code you supplied failed validation

From my own research, it looks like there's a mix of both no postal codes and a new "Eircode" system introduced in 2014 spanning the country.

WooCommerce issue discussion & commit:

https://github.com/woocommerce/woocommerce/issues/10663
https://github.com/woocommerce/woocommerce/commit/3112c769395763617725add855e1a5857c2ac007

I'm trying to first nail down what it _should_ validate if Ireland is the selected country.

It doesn't look like there's any special case logic for Ireland in EDD core so far (country `IE`).

I'm also using EDD-Stripe and not requiring a full billing address, just country and zip. Not sure if this is related.

Thanks!

pippinsplugins:
Thanks for the report @pderksen, we'll get this addressed shortly after 2.7.

ghost1227:
@pippinsplugins OK, I've been trying to figure this out, but I'm at a loss on two fronts. First, if there's a chance of no zip code, we can't REQUIRE the zip field, but I don't see an existing way to disable that requirement. We have an existing way of removing state from the require list, but I don't see a way for zip... though I can't believe there isn't one. Second, I've determined I still suck at regex. Mind giving me some guidance on how to do the validation for that string?

pippinsplugins:
@ghost1227 we can make the field not required here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/process-purchase.php#L393

Perhaps we should add a new `edd_maybe_require_zip()` function that then programmatically determines if the zip code should be required based on the country that is submitted.

ghost1227:
@pippinsplugins I THINK this does what you want in terms of helper function flexibility. Right now, only excludes IE, but can certainly add others

pippinsplugins:
@ghost1227 see PR review for a couple of minor changes./

ghost1227:
@pippinsplugins Changes made!

[...]

PR #5541, opened 2017-03-06
Dont require zip code from IE, now with helper function and filter 

ghost1227:
First steps towards #5307

### DIFF
--- a/includes/checkout/functions.php
+++ b/includes/checkout/functions.php
@@ -583,3 +583,21 @@ function edd_purchase_form_validate_cc_exp_date( $exp_month, $exp_year ) {
 	return $expiration >= time();
 
 }
+
+/**
+ * Determine if zip code should be required by country
+ *
+ * @since 2.7.5
+ * @param string $country_code The country code to check
+ * @return bool true if required, false otherwise
+ */
+function edd_maybe_require_zip( $country ) {
+	$required = true;
+
+	switch( sanitize_text_field( $country ) ) {
+		case 'IE':
+			$required = false;
+	}
+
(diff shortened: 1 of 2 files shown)


includes/checkout/functions.php
```diff
@@ -583,3 +583,21 @@ function edd_purchase_form_validate_cc_exp_date( $exp_month, $exp_year ) {
...
+ * @since 2.7.5
+ * @param string $country_code The country code to check
+ * @return bool true if required, false otherwise
+ */
+function edd_maybe_require_zip( $country ) {
+	$required = true;
+
+	switch( sanitize_text_field( $country ) ) {
+		case 'IE':
+			$required = false;
```
pippinsplugins: This needs a break statement please.

pippinsplugins: Let's also set it to `false` if `$country` is empty.

includes/checkout/functions.php
```diff
@@ -583,3 +583,21 @@ function edd_purchase_form_validate_cc_exp_date( $exp_month, $exp_year ) {
...
 }
+
+/**
+ * Determine if zip code should be required by country
+ *
+ * @since 2.7.5
+ * @param string $country_code The country code to check
+ * @return bool true if required, false otherwise
+ */
+function edd_maybe_require_zip( $country ) {
```
pippinsplugins: Please default `$country` to `$country = ''`

includes/checkout/functions.php
ghost1227: Done

includes/checkout/functions.php
ghost1227: Makes sense!

Issue #5547, opened 2017-03-07
EDD Payment Query does not have adequate handling of a payment lookup failure


chriscct7:
Copying over relevant section from https://github.com/easydigitaldownloads/edd-fes/issues/1174

It doesn't handle the rare edge case that edd_get_payment as documented, can return false (or the rarer chance that the filter returns a non-object to use). This leads to a payment query where you have say 10 post objects and sandwhiched in the middle one false. If $payment is not an object, it should not be added to the $this->payments array.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb765e03ca311742d8be78b810fb78bd58f65a4c/includes/payments/class-payments-query.php#L171

It doesn't handle the rare edge case that edd_get_payment as documented, can return false (or the rarer chance that the filter returns a non-object to use). This leads to a payment query where you have say 10 post objects and sandwhiched in the middle one false. If $payment is not an object, it should not be added to the $this->payments array.

chriscct7:
PR https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5548

pippinsplugins:
Good catch!

sunnyratilal:
Closing this out as it is no longer valid as I've refactored `EDD_Payments_Query` in #4576.

PR #5548, opened 2017-03-07
Only add payment to payments query results if it is an object

chriscct7:
See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5547


chriscct7:
This might not be a bad thing to seperately add a unit test for

sunnyratilal:
Closing as this no longer valid.

PR #5569, opened 2017-03-13
Add template tag support to email header

### LINKED ISSUE #4886: Allow email template tags to be added to the Purchase Email Heading
The {payment_id} template tag has never been usable in the purchase email heading. Is this something we want to add support for? Perhaps this should be expanded to all email template tags?

ghost1227:
Fixes #4886


includes/emails/functions.php
```diff
@@ -43,6 +43,7 @@ function edd_email_purchase_receipt( $payment_id, $admin_notice = true, $to_emai
 
 	$heading      = edd_get_option( 'purchase_heading', __( 'Purchase Receipt', 'easy-digital-downloads' ) );
 	$heading      = apply_filters( 'edd_purchase_heading', $heading, $payment_id, $payment_data );
+	$heading      = edd_do_email_tags( $heading, $payment_id );
```
pippinsplugins: This needs to be done on _all_ emails, not just receipts, so it should be done via the EDD_Emails class.

Issue #5346, opened 2017-01-05
Multiple EDD_Payments_Query's affect each other. 


mintplugins:
If you instantiate 2 objects of EDD_Payments_Query, because the values are hooked to edd_pre_get_payments, any custom values you set up for the first object continue to be hooked for any subsequent objects. 

**For example:** 
The following code snippet will cause the payment history page in the WordPress dashboard to show no results (notice how we don't return any values or do anything with our returned data here - in theory, because of the word "get", this shouldn't be affecting anything):
```
function mess_with_edd_get_payments() {
	$args = array(
		'end_date' => time()
	);

	$payments = edd_get_payments( $args );
}
add_action( 'admin_init', 'mess_with_edd_get_payments' );
```

The problem is that in the EDD_Payments_Query class, the values are hooked to edd_pre_get_payments. So when the payment history page runs this in a separately instatiated object, the hook still runs which sets the "end_date" - even though we had no intention of setting the end_date for the Payment History page. 

**Proposed Solution:**
Instead of hooking the custom values to edd_pre_get_payments, we should simply call each method from within the get_payments method in EDD_Payments_Query.

pippinsplugins:
Let's look at this soon after 2.7.

mintplugins:
One other approach might be to remove the edd_pre_get_payments hooks created by the class by hooking to edd_post_get_payments and unhooking the calls there.

MythThrazz:
Hi Pippin,
Could you please give it a slightly higher priority? We have a ready plugin which relies heavily on the fix for this issue.

pippinsplugins:
I've bumped it to 2.7.3.

ghost1227:
@pippinsplugins Do we have a preferred method for fixing this? The method proposed by @mintplugins in the initial issue seems cleanest to me.

pippinsplugins:
@ghost1227 go with @mintplugins's proposal.

mintplugins:
With guidance from @ghost1227 I'm going to take a shot at this

ghost1227:
@mintplugins If you have any questions/need help feel free to ping me!

mintplugins:
This is working in my tests but could definitely use more tests. Try anything/everything that would use a payment query.

sunnyratilal:
I've run multiple tests and all good here but would still like prefer final review from @pippinsplugins or @cklosowski

ghost1227:
Another 👍 from me!

[...]

PR #5578, opened 2017-03-14
Call query modifiers directly in EDD_Get_Payments

mintplugins:
Instead of hooking to edd_pre_get_payments, call the query modifiers directly.


mintplugins:
https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5346

PR #5590, opened 2017-03-21
Ensure EDD_Payments_Query does not skip renewal payments #5377

### LINKED ISSUE #5377: EDD_Payments_Query "downloads" filter skips renewal payments.
If you want to query payments based on the download(s) purchased, you would do something like this

```
$args = array(
	'download' => array( DOWNLOAD_ID_HERE, ANOTHER_DOWNLOAD_ID_HERE ),
	'number'   => -1,
);
$query    = new EDD_Payments_Query( $args );
$payments = $query->get_payments();
```

However, because the query uses "[post_parent](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/payments/class-payments-query.php#L521)" and  "[post_parent__in](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/payments/class-payments-query.php#L534)", payments that share the same post parent are not returned (For example, renewal payments from EDD Recurring).

In order to make those renewal payments get returned as well, we would have to use "post__in" instead of "post_parent" and "post_parent__in" (see above links).

This section could then be removed as well:
```
if ( is_array( $this->args['download'] ) ) {
	unset( $args['post_parent'] );
	$args['post__in'] = $this->args['download'];
}
```

**Question:**
Is the current way expected? That is, is it expected that EDD_Payments_Query::downloads only gets the subscription's original payment? Personally I would expect all matching payments to be returned whether they are renewal payments or original payments.

sunnyratilal:
Fixes #5377

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -528,7 +528,7 @@ class EDD_Payments_Query extends EDD_Stats {
 		global $edd_logs;
 
 		$args = array(
-			'post_parent'            => $this->args['download'],
+			'post__in'               => is_array( $this->args['download'] ) ? $this->args['download'] : array( $this->args['download'] ) ,
 			'log_type'               => 'sale',
 			'post_status'            => array( 'publish' ),
 			'nopaging'               => true,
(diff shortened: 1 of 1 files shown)


pippinsplugins:
PR is invalid. See #5377

sunnyratilal:
Closing PR because this PR needs a refresh if the issue still persists.

PR #5602, opened 2017-03-26
Empty check breaks file assignments for variation price index 0 on old downloads

chriscct7:
See https://github.com/easydigitaldownloads/edd-fes/issues/1152

#5610


chriscct7:
As a sidenote, this section is unused: 
image

Issue #5635, opened 2017-04-04
Tax rate of 1 or less amounts to incorrect total


mintplugins:
If you enter a tax rate (country or fallback) of 1 or less, the calculation is incorrect. Confirmed by @SDavisMedia

mintplugins:
The tax calculation is working for me now. 

However, I'm not sure if this is related but if I enter a tax rate for a country/state, it is falling back to the fallback rate even when it shouldn't.

cklosowski:
@mintplugins can you replicate that on master?

mintplugins:
Hmm - it's not happening for me on master or `issue/5635` for me now. Maybe it was a weird cached situation or something. Disregard - this looks fine to me now 👍

SDavisMedia:
This is working as it should for me for both tax rate entries. :+1:

PR #5641, opened 2017-04-05
Allow tax rates lower than 1% #5635 issue/5635

cklosowski:
#5635

### DIFF
--- a/includes/tax-functions.php
+++ b/includes/tax-functions.php
@@ -89,8 +89,9 @@ function edd_get_tax_rate( $country = false, $state = false ) {
 					}
 				} else {
 
-					if( empty( $tax_rate['state'] ) || strtolower( $state ) != strtolower( $tax_rate['state'] ) )
+					if( empty( $tax_rate['state'] ) || strtolower( $state ) != strtolower( $tax_rate['state'] ) ) {
 						continue;
+					}
 
 					$state_rate = $tax_rate['rate'];
 					if( 0 !== $state_rate || ! empty( $state_rate ) ) {
@@ -101,10 +102,9 @@ function edd_get_tax_rate( $country = false, $state = false ) {
 		}
 	}
 
-	if( $rate > 1 ) {
-		// Convert to a number we can use
-		$rate = $rate / 100;
-	}
+	// Convert to a number we can use
+	$rate = $rate / 100;
+
 	return apply_filters( 'edd_tax_rate', $rate, $country, $state );
 }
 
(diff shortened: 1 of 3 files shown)


tests/tests-tax.php
```diff
@@ -124,6 +125,10 @@ public function test_get_tax_rate() {
 		$this->assertEquals( '0.036', edd_get_tax_rate( 'HK' ) );
 	}
 
+	public function test_get_tax_rate_less_than_one() {
```
pippinsplugins: To be extra thorough, let's get at least 5 different tests added for 5 different rates.

pippinsplugins: Let's also add tests that test the tax amount calculation.

cklosowski:
@pippinsplugins updated the tests. look good?

includes/tax-functions.php
```diff
@@ -101,10 +102,9 @@ function edd_get_tax_rate( $country = false, $state = false ) {
 		}
 	}
 
-	if( $rate > 1 ) {
-		// Convert to a number we can use
-		$rate = $rate / 100;
```
pippinsplugins: This was originally done to enforce rates to be entered as a whole percentage and not `0.05` (to represent 5%).

If allowing `0.05`, we need to add a notice that shows a message to site owners when a rate is detected lower than that after installing this update so they can verify it's entered correctly, otherwise we could end up with a bunch of sites that are suddenly calculating very wrong tax amounts.

sunnyratilal:
@pippinsplugins After deliberation with @SDavisMedia, he suggests that we shouldn't have a notice and we should make use of tooltips and labels due to this being quite a big edge case.

Proposal:
1. Have a tooltip on the `Rate` column which specifies the rate entered is treated as a percentage.
2. Have a `placeholder` or another tooltip on the actual rate input saying `6 = 6%`.

Is this ok with you?

SDavisMedia:
@sunnyratilal I actually don't agree with what you suggest on point 2. My suggestions are:

1. Add a tooltip to the **Tax Rates -> Rate** column which specifies the rate entered is treated as a percentage.
2. Take the same new tool tip text and _add_ it to the **Fallback Tax Rate** setting tooltip. Also note that we kind of already say this on that label itself.
3. Adjust the **Tax Rates** label to give an example like we always do. For example, enhance that label to be `Enter tax rates for specific regions. Enter a percentage, such as 6.5.`

I _do not_ like the idea of putting the example in a placeholder. 

I don't think an entire notice is necessary. If people were putting in things like `0.05` up to this point, then we _already_ have a bunch of sites that are calculating wrong tax amounts.

sunnyratilal:
Thanks for the clarification @SDavisMedia - I misunderstood from what we discussed in Slack. :)

pippinsplugins:
I agree with @SDavisMedia

SDavisMedia:
@sunnyratilal @pippinsplugins just pushed up the adjustments I suggested.

PR #5649, opened 2017-04-09
issue/5637 - Fix payment history date filter not working as expected

### LINKED ISSUE #5637: Payment History date filter not working
Reported by two separate customers today so far and confirmed on my local dev site and a live site. If you select a start and end date when viewing payment history, the results will get filtered but they will be wrong. Still trying to figure out a pattern. So far the actual results which get returned are confusing. But so far, very consistently wrong as in not actually from the specified date range.

sunnyratilal:
Fixes #5637


sunnyratilal:
Unit tests added

PR #5663, opened 2017-04-14
issue/5620 - Prepend site URL if using a relative path

### LINKED ISSUE #5620: Forward Slashes for Download Files create a new Product
A customer is experiencing an error with CSV product imports. The Download Files field is supposed to allow for forward slashes for the location of the file. 

The documentation states that 
```
This field can accept a valid web address of a file, an absolute path 
on a filesystem, or the bucket/filename.ext on an Amazon S3 instance. 
Muitlple files are separated by vertical bar like |. 
Note that file names can not be entered separately but will be 
automatically generated using the file's name.
```

However, when using forward slashes, they are acting the way `|` is supposed to act and creating new files for each

For example, this in the CSV file:
screen shot 2017-03-30 at 2 12 05 pm

Results in this in EDD after import:
screen shot 2017-03-30 at 2 12 48 pm

sunnyratilal:
Fixes #5620

### DIFF
--- a/includes/admin/import/class-batch-import-downloads.php
+++ b/includes/admin/import/class-batch-import-downloads.php
@@ -192,6 +192,10 @@ class EDD_Batch_Downloads_Import extends EDD_Batch_Import {
 				// setup files
 				if( ! empty( $this->field_mapping['files'] ) && ! empty( $row[ $this->field_mapping['files'] ] ) ) {
 
+					if ( ! preg_match('#^http[s]{0,1}://#', $row[ $this->field_mapping['files'] ] ) ) {
+						$row[ $this->field_mapping['files'] ] = get_site_url( null, $row[ $this->field_mapping['files'] ] );
+					}
+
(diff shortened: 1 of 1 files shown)


sunnyratilal:
@pippinsplugins @cklosowski Not the solution I feel most _comfortable_ with - would either of you have anything "better" in mind?

cklosowski:
@sunnyratilal My only concern here, is that we also allow direct file path, which looks the same as a relative URL, so we'd have no way to determine if this is supposed to be a relative URL or an absolute path.

sunnyratilal:
@cklosowski Is there any harm in converting the absolute path to a relative URL? The absolute path would be _surely_ be prepended with a  `/` ?

chriscct7:
Except on Windows where that's not necessarily true

pippinsplugins:
I've submitted an alternate version that is safer. See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5620#issuecomment-298645750

PR #5664, opened 2017-04-14
issue/5646 - Update customer data based on info added at checkout 

### LINKED ISSUE #5646: If Customer Name/Info blank upon new purchase, use data from checkout.
Currently, if a customer has no name entered (Unnamed Customer), and then they buy a new product and fill out their name on checkout, that information does not get added to the customer record. 

I propose that if a Customer has blank data and completes a payment, that newly entered info be added to the customer.

sunnyratilal:
Fixes #5646


includes/process-purchase.php
```diff
@@ -74,6 +74,31 @@ function edd_process_purchase_form() {
 		'address'    => $user['address']
 	);
 
+	// Update a customer record if they have added/updated information
+	if ( is_user_logged_in() ) {
```
pippinsplugins: This shouldn't require the customer be logged in. As soon as we have a customer record identified, we can update it, logged in or logged out.

sunnyratilal: Change implemented in 2a96efd2125267dc900b64ef4298b813a1750759

includes/process-purchase.php
```diff
@@ -74,6 +74,30 @@ function edd_process_purchase_form() {
 		'address'    => $user['address']
 	);
 
+	// Update a customer record if they have added/updated information
+	$customer = new EDD_Customer( $user_info['user_id'], true );
```
pippinsplugins: This should be instantiated by email, not user ID, so that it works for guest purchases.

PR #5668, opened 2017-04-16
Typo: "succesfully" instead of "successfully"

boluda:
Just a tiny misspelling. Three lines with "succesfully" instead of "successfully".


sunnyratilal:
Good catch - thanks!

PR #5697, opened 2017-05-01
Allow requiring payment info on 100% discounted carts

### LINKED ISSUE #5686: Allow requiring credit cards, even with 100% discount applied
_I'm not certain if this should go here or in the `edd-recurring` repo; happy to move it._

I'm working with a client who'd like to be able to require credit card information, even if a 100% discount is applied. This is because with a recurring payment, they may provide a 100% off discount code for the first payment, but want to be able to charge the buyer's card at the next subscription billing period.

It looks like this is mostly controlled in `edd-checkout-global.js` [here](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/0787a913a6c33843eb8cdde92548b11cc00e94a4/assets/js/edd-checkout-global.js#L149-L162).

I'm thinking a reasonable approach will be to add a global EDD option toggling whether to require CC information regardless of cart value (for now; we may want to try per-product settings later on), and if that option is set, then have `discount_response` include that flag, and do e.g.

```js
if( true == discount_response.require_cc ) {

  if (!inputs.is('.card-address-2')) {
    inputs.attr('required','required');
  }
  $('#edd_cc_fields,#edd_cc_address').slideDown();
  
}

else if( '0.00' == discount_response.total_plain ) {

	$('#edd_cc_fields,#edd_cc_address,#edd_payment_mode_select').slideUp();
	inputs.removeAttr('required');
	$('input[name="edd-gateway"]').val( 'manual' );

} else {

	if (!inputs.is('.card-address-2')) {
		inputs.attr('required','required');
	}
	$('#edd_cc_fields,#edd_cc_address').slideDown();

}
```

That feels a bit hacky/non-DRY, but hopefully conveys the intent.

If this is kosher, and the desired approach, I'll move forward and then open a PR for this feature.

Thanks!

tnorthcutt:
Closes #5686

Adds a setting which, when true, will result in payment information
being required on checkout regardless of the value of a cart. This is
specifically to allow requiring payment info for a recurring subscription
when a coupon for 100% off the first billing period has been applied. Before
this patch, that situation would result in the buyer having to revisit
the site later on to provide payment information for their subscription to be
able to continue after the first billing period.

### DIFF
--- a/assets/js/edd-checkout-global.js
+++ b/assets/js/edd-checkout-global.js
@@ -146,7 +146,7 @@ window.EDD_Checkout = (function($) {
 
 						var inputs = $('#edd_cc_fields .edd-input, #edd_cc_fields .edd-select,#edd_cc_address .edd-input, #edd_cc_address .edd-select,#edd_payment_mode_select .edd-input, #edd_payment_mode_select .edd-select');
 
-						if( true == discount_response.require_cc ) {
+						if( true == discount_response.require_payment_info ) {
 
 							if (!inputs.is('.card-address-2')) {
 								inputs.attr('required','required');
--- a/assets/js/edd-checkout-global.js
+++ b/assets/js/edd-checkout-global.js
@@ -146,7 +146,16 @@ window.EDD_Checkout = (function($) {
 
 						var inputs = $('#edd_cc_fields .edd-input, #edd_cc_fields .edd-select,#edd_cc_address .edd-input, #edd_cc_address .edd-select,#edd_payment_mode_select .edd-input, #edd_payment_mode_select .edd-select');
 
-						if( '0.00' == discount_response.total_plain ) {
+						if( true == discount_response.require_cc ) {
+
+							if (!inputs.is('.card-address-2')) {
+								inputs.attr('required','required');
+							}
+							$('#edd_cc_fields,#edd_cc_address').slideDown();
+
+						}
+
+						else if( '0.00' == discount_response.total_plain ) {
 
 							$('#edd_cc_fields,#edd_cc_address,#edd_payment_mode_select').slideUp();
 							inputs.removeAttr('required');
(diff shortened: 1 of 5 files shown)


tnorthcutt:
N.B. I ran the grunt makepot task, since this introduced new strings. However, it appears as though that hadn't been run for some amount of time, as it resulted in several unrelated updates to the `.pot` file. I left those as-is; I wasn't sure what the proper protocol is for committing those unrelated changes.

pippinsplugins:
@tnorthcutt can you remove the pot file changes? We don't ever update pot files until just before release.

tnorthcutt:
@pippinsplugins sorry I never got back to you on this! This is on hold on my end, pending client go-ahead 😄

tnorthcutt:
@pippinsplugins I believe this is ready to go now; there are some failing checks, but they're confined to the HHVM builds. Is that a blocker?

tnorthcutt:
Oh, now I see the error message for those builds:

`HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with 'dist: trusty'.`

So, that explains that 😁

cklosowski:
@tnorthcutt Yeah, HHVM is out, we've removed it from our 2.8 branch, which should be coming up here any day so it won't be an issue.

tnorthcutt:
@cklosowski Awesome! So, does that mean this is a candidate to go ahead and get pulled in soon?

tnorthcutt:
@cklosowski Anything I need to do on this to get it merged in?

assets/js/edd-checkout-global.js
```diff
@@ -146,7 +146,16 @@ window.EDD_Checkout = (function($) {
 
 						var inputs = $('#edd_cc_fields .edd-input, #edd_cc_fields .edd-select,#edd_cc_address .edd-input, #edd_cc_address .edd-select,#edd_payment_mode_select .edd-input, #edd_payment_mode_select .edd-select');
 
-						if( '0.00' == discount_response.total_plain ) {
+						if( true == discount_response.require_payment_info ) {
+
+							if (!inputs.is('.card-address-2')) {
```
pippinsplugins: @tnorthcutt can you explain why this is necessary?

tnorthcutt: Sure! That's what keeps the credit card form from being hidden (and not required), IF a filter has been added to require payment info even on 100% discounted carts. That filter can be added like so:

```
add_filter( 'edd_require_payment_info', '__return_true' );
```

Without adding that filter, behavior is completely "normal".

tnorthcutt:
closing in favor of https://github.com/easydigitaldownloads/edd-stripe/issues/295

Issue #5291, opened 2016-12-12
Replace Paypal "Page Style" with "Image URL"


mintplugins:
According to my personal tests and a report from a customer, Paypal Standard is removing support for custom checkout Page Styles. 

Our customer received a response from Paypal tech support stating it was no longer compatible with the platform. While Paypal's documentation still contains a reference to the page_style attribute, it does not appear to actually work and, again, Paypal's tech support straight-up says it doesn't work anymore:
https://secure.helpscout.net/conversation/290837461/49045/?folderId=749231

Their advice is to use the image_url paramater so you can show your own logo instead. 

Proposal:
Remove support for Paypal Page Styles and replace with an option to upload a 150 by 150 px image which would be passed to the image_url paramater in Paypal's API:

https://developer.paypal.com/webapps/developer/docs/classic/paypal-payments-standard/integration-guide/Appx_websitestandard_htmlvariables/#paypal-checkout-page-variables

mintplugins:
https://secure.helpscout.net/conversation/290837461/49045/?folderId=180504

pippinsplugins:
Customer report of this: https://secure.helpscout.net/conversation/348838465/57297/?folderId=180499

cklosowski:
This is ready for some testing @mintplugins Keep in mind that if you are using a local dev environment it doesn't seem to load it (since it's a 404 to PayPal's servers). But if you view the source of the checkout page but you can see the URL in the `meta` tag.

pippinsplugins:
@cklosowski have you been able to get the image to show up? It won't show up for me on PayPal.

cklosowski:
@pippinsplugins were you testing on a live site? If it's a local site, it may not show up because they do a `404` check on it before using it and since it's not publicly accessible, it won't display.

cklosowski:
Not a perfect example....but it shows it displaying.

![screen shot 2017-05-12 at 3 09 00 pm](https://cloud.githubusercontent.com/assets/1390124/26018644/f9335a84-3724-11e7-84ed-e01460086e54.png)

pippinsplugins:
I use a live URL.

I'm happy with it if you're showing it though :+1:

cklosowski:
Merging but leaving open so we can update our docs.

PR #5723, opened 2017-05-09
Change to using image_url instead of page_style #5291

cklosowski:
#5291


includes/gateways/paypal-standard.php
```diff
@@ -729,14 +729,14 @@ function edd_get_paypal_redirect( $ssl_check = false, $ipn = false ) {
 }
 
 /**
- * Set the Page Style for PayPal Purchase page
+ * Get the image for the PayPal purcahse page.
```
pippinsplugins: `purcahse` should be `purchase`

Issue #5239, opened 2016-11-22
Don't get notes twice in new EDD_Customer()


stephywells:
The database queries when a customer is fetched look like this:
SELECT * FROM wp_edd_customers WHERE user_id = '6' LIMIT 1
SELECT notes FROM wp_edd_customers WHERE id = '2' LIMIT 1;

Would it cause any problems if the get_raw_notes function looks like this to prevent the extra query?

	private function get_raw_notes() {

		if ( isset( $this->notes ) ) {
			$all_notes = $this->notes;
		} else {
			$all_notes = $this->db->get_column( 'notes', $this->id );
		}

		return (string) $all_notes;

	}

pippinsplugins:
No issues at all that I see!

stephywells:
I've got a PR ready for this.

pippinsplugins:
@stephywells After seeing the unit tests fail due to the changes, I think we should alter this slightly. See my comment on the PR. Thanks!

SDavisMedia:
Ping.

Removing `Needs Testing` until the PR is updated.

PR #5758, opened 2017-05-22
Add a raw_notes internal property so we can only get it once and update it when needed #5239

cklosowski:
#5239


pippinsplugins:
If tests pass, looks good to me!

Issue #5774, opened 2017-05-30
WP Cli Sales Command


bryancisler:
I was looking through the EDD WP CLI commands this weekend and was wondering what your thoughts were about adding an enhancement to the wp edd sales command? I have a PR coming that would rework the functionality so someone could filter sales by email. Something like wp edd sales –email=bryan.cisler@strategynewmedia.com

pippinsplugins:
@bryancisler that would be a great improverment!

sunnyratilal:
PR checked and it's good to be merged

PR #5775, opened 2017-05-30
Add email parameter to wp edd sales command.

bryancisler:
Issue: #5774

### DIFF
--- a/includes/class-edd-cli.php
+++ b/includes/class-edd-cli.php
@@ -372,7 +372,11 @@ class EDD_CLI extends WP_CLI_Command {
 	 */
 	public function sales( $args, $assoc_args ) {
 		
-		$email = isset( $assoc_args ) && array_key_exists( 'email', $assoc_args ) ? $assoc_args['email'] : false;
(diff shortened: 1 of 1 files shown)


includes/class-edd-cli.php
```diff
@@ -363,19 +363,39 @@ public function customers( $args, $assoc_args ) {
...
+		$email = isset( $assoc_args ) && array_key_exists( 'email', $assoc_args ) ? $assoc_args['email'] : false;
 
 		$sales = $this->api->get_recent_sales();
 
 		if( empty( $sales ) ) {
 			WP_CLI::error( __( 'No sales found', 'easy-digital-downloads' ) );
 			return;
 		}
+		
+		if( !empty($email)) {
```
cklosowski: By doing it after the fact, it's possible that you might not get the correct number of items back since you are filtering after the query has been run. The `get_recent_sales()` method in the API Looks for an email address in the `$wp_query->query_vars['email']` values, so it'd be better to simply do something like this before the line that calls `get_recent_sales()`

```
global $wp_query;
$wp_query->query_vars['email'] = $email;
```
Can you give that a shot instead of filtering after?

cklosowski:
Closing as it's been updated to #5780.

Issue #5404, opened 2017-01-19
Remove "Settings" Suffix in Settings Subsection Titles


spencerfinnell:
https://cloudup.com/cwqU_UZPSgC

We already know we are in the Settings section. And it's inconsistent across tabs. I think "Misc" should be "Misc." with a period as well.

spencerfinnell:
This PR removes extraneous "Settings" suffixes on subpanel labels, and removes section titles on pages that only have one section to avoid duplicate (and usually inconsistent) titles.

sunnyratilal:
Merged

PR #5776, opened 2017-05-31
Add consistency to settings labels 

spencerfinnell:
and remove section labels on tabs that only have one section. See #5404

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -1229,7 +1229,7 @@ function edd_get_registered_settings_sections() {
 
 	$sections = array(
 		'general'    => apply_filters( 'edd_settings_sections_general', array(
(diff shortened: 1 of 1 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -1283,35 +1216,35 @@ function edd_get_registered_settings_sections() {
 
 	$sections = array(
 		'general'    => apply_filters( 'edd_settings_sections_general', array(
-			'main'               => __( 'General Settings', 'easy-digital-downloads' ),
-			'currency'           => __( 'Currency Settings', 'easy-digital-downloads' ),
-			'api'                => __( 'API Settings', 'easy-digital-downloads' ),
+			'main'               => _x( 'General', 'easy-digital-downloads' ),
```
pippinsplugins: This should be `__()` or needs to have the context parameter added.

Issue #5041, opened 2016-10-13
Introduce shortcode-download.php Template File


spencerfinnell:
I've been harping about this forever now but don't think there is an open issue for tracking. 

As EDD makes a push for more themes (Themedd!) and continued development of many extensions that output downloads for various reasons I think this becomes even more important. There are a few issues with the current way the `[downloads]` shortcode loads multiple template files.
1. Any theme or plugin that wants to output downloads is forced to use the `[downloads]` shortcode (not necessarily a bad thing) or call the `edd_downloads_query()` function directly to achieve the same output as their store page.
   
   In my mind this is the preferred approach that themes and plugins should be taking. 
2. Any theme or plugin that doesn't use the `[downloads]` shortcode has to either duplicate the code inside of `edd_downloads_query()` and then selectively include the separate template files. Often times this leaves out many of the filters that are in the shortcode callback, and if they do exist it's different so extra code still needs to be added.
   
   This has caused issues with current extensions that output lists of downloads completely differently than what a theme has styled for the shortcode.
   
   e.g: https://github.com/easydigitaldownloads/edd-recommended-products/issues/15
3. `edd_downloads_query()` is **massive** and is certainly not just querying downloads.

There are many pros to calling these multiple template files from a singular file:
1. A theme can override one file to remove calls to the other pieces without having to filter the shortcode attributes, plugin filters, etc.
2. It can be done with complete backwards compatibility since all existing template files remain the exact same. 

The bad things:
1. Existing themes/plugins that do not call the shortcode directly will not be affected until they update their code (this is also really a good thing).
2. `edd_get_template_part()` can't pass variables so the existing `$atts` array would need some finagling. 

---

My proposal would be to replace https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/shortcodes.php#L564-L587 with an action that calls a template file. This will allow the `$atts` variable to be dealt with.

```
function edd_download_item( $atts ) {
    global $atts;

    edd_get_template_part( 'shortcode', 'download' );
}
add_action( 'edd_download_item', 'edd_download_item' );
```

```
<?php
/**
 * A download for the [downloads] grid.
 */

global $atts;

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/shortcodes.php#L564-L587
```

---

So there are technically no immediate benefits for existing themes/plugins but it will certainly help things a lot moving forward. 

Thoughts?

[...]

PR #5778, opened 2017-05-31
Introduce a single template file for use inside of the [downloads] shortcode

spencerfinnell:
See #5041

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -1111,13 +1111,13 @@ function edd_get_bundle_item_price_id( $bundle_item ) {
  * @param int   $i The current item count.
  */
 function edd_download_shortcode_item( $atts, $i ) {
-	global $edd_download_shortcode_item_atts, $edd_download_shortcode_item_$i;
+	global $edd_download_shortcode_item_atts, $edd_download_shortcode_item_i;
 
 	/**
 	 * The variables are registered as part of the global scope so the template can access them.
 	 */
 	$edd_download_shortcode_item_atts = $atts;
-	$edd_download_shortcode_item_$i = $i;
+	$edd_download_shortcode_item_i = $i;
 
 	edd_get_template_part( 'shortcode', 'download' );
 }
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -1111,13 +1111,13 @@ function edd_get_bundle_item_price_id( $bundle_item ) {
  * @param int   $i The current item count.
  */
 function edd_download_shortcode_item( $atts, $i ) {
-	global $edd_download_shortcode_item_atts, $edd_download_shortcode_item_i;
+	global $edd_download_shortcode_item_atts, $edd_download_shortcode_item_$i;
 
 	/**
 	 * The variables are registered as part of the global scope so the template can access them.
 	 */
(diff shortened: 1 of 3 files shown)


rubengc:
I love this one

But instead of atts parameter and their string expected values, will be awesome make It in a php standard way

Current way to say you want output thumbnail and excerpt:
```php
edd_download_shortcode_item( array(
    'thumbnails' => 'true',
    'excerpt' => 'yes'
) );
```


Proposal:
```php
edd_download_shortcode_item( array(
    'thumbnail' => true,
    'excerpt' => true
) );
```

Or included turn It into WordPress common fields (post_excerpt, post_thumbnail)

This could be turned in the standard way to output a download(s), so, also, it opens the possibility to add a method like 'edd_render_download( $args, $context )' to allow pass as context the first parameter that will be passed to 'edd_get_template_part' (shortcode, widget, etc)

includes/template-functions.php
```diff
@@ -1097,3 +1097,28 @@ function edd_get_bundle_item_price_id( $bundle_item ) {
...
+ * shortcode's attributes can be passed to the template file via
+ * a global variable.
+ *
+ * @since 2.9.0
+ *
+ * @param array $atts The [downloads] shortcode attributes.
+ * @param int   $i The current item count.
+ */
+function edd_download_shortcode_item( $atts, $i ) {
+	global $edd_download_shortcode_item_atts, $edd_download_shortcode_item_i;
```
cklosowski: This last global should be `$edd_download_shortcode_item_$i`.

As it stands all items will be stored in the same global b/c it's using `i` as a string.

includes/template-functions.php
```diff
@@ -1097,3 +1097,28 @@ function edd_get_bundle_item_price_id( $bundle_item ) {
...
+
+	/**
+	 * The variables are registered as part of the global scope so the template can access them.
+	 */
+	$edd_download_shortcode_item_atts = $atts;
+	$edd_download_shortcode_item_i = $i;
```
cklosowski: This needs to be `$edd_download_shortcode_item_$i`, as above.

templates/shortcode-download.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @since 2.9.0
+ *
+ * @package EDD
+ * @category Template
+ * @author Easy Digital Downloads
+ * @version 1.0.0
+ */
+
+global $edd_download_shortcode_item_atts, $edd_download_shortcode_item_i;
```
cklosowski: Save as above for `$edd_download_shortcode_item_$i`

cklosowski:
@sunnyratilal @spencerfinnell I left a note on the PR here...am I correct that these globals should be unique to each part of the iteration, or are they ok to be overwritten at each step in the process? Let me know, if I'm wrong (which is possible :) ) i'll approve as this looks solid.

sunnyratilal:
@cklosowski Updated the variable name and the globals are overwritten in each iteration as per how @spencerfinnell envisaged this

cklosowski:
I talked to @spencerfinnell and the way we had it before the most recent change was intentional. i'll revert that and we should be good.

PR #5781, opened 2017-06-02
Prevent extreme duplicate SQL queries

JeroenSormani:
When a discount is applied to the cart it will call the `edd_get_discounts` a lot of times. This function calls the WP_Query which causes duplicate queries. When there's only one product in the cart its 'only' 82 queries instead of ~32. When there's more it will multiply quickly and with a test of 9 products in the cart it will send 500+ duplicate queries to the DB.

https://cl.ly/262J0O1i0X2V

--------

I've also investigated the roots of it being called that many times; this is something else that I'd like to discuss sometime soon. The results of some quick tests seem very promising for improvements in the checkout page speed. 

Before going from 2.2~2.3 seconds (with pretty regular peeks of ~5+ seconds); going to 1.4-1.6 seconds without those spikes.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -35,7 +35,7 @@ function edd_get_discounts( $args = array() ) {
 	$discounts_hash = md5( json_encode( $args ) );
 	if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
 		$discounts = get_posts( $args );
-		wp_cache_set( $discounts_hash, $discounts, 'edd-discounts' );
+		wp_cache_set( $discounts_hash, $discounts, 'edd-discounts', HOUR_IN_SECONDS );
 	}
 
 
@@ -53,7 +53,7 @@ function edd_get_discounts( $args = array() ) {
 		$discounts_hash = md5( json_encode( $args ) );
 		if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
 			$discounts = get_posts( $args );
-			wp_cache_set( $discounts_hash, $discounts, 'edd-discounts' );
+			wp_cache_set( $discounts_hash, $discounts, 'edd-discounts', HOUR_IN_SECONDS );
 		}
 	}
 
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -33,7 +33,7 @@ function edd_get_discounts( $args = array() ) {
 	$args = wp_parse_args( $args, $defaults );
 
 	$discounts_hash = md5( json_encode( $args ) );
-	$discounts = $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' );
+	$discounts = wp_cache_get( $discounts_hash, 'edd-discounts' );
 	if ( false === $discounts ) {
 		$discounts = get_posts( $args );
 		wp_cache_set( $discounts_hash, $discounts, 'edd-discounts', HOUR_IN_SECONDS );
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -33,7 +33,8 @@ function edd_get_discounts( $args = array() ) {
 	$args = wp_parse_args( $args, $defaults );
 
 	$discounts_hash = md5( json_encode( $args ) );
-	if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
+	$discounts = $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' );
+	if ( false === $discounts ) {
 		$discounts = get_posts( $args );
 		wp_cache_set( $discounts_hash, $discounts, 'edd-discounts', HOUR_IN_SECONDS );
 	}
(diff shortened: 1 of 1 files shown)


includes/discount-functions.php
```diff
@@ -32,19 +32,29 @@ function edd_get_discounts( $args = array() ) {
 
 	$args = wp_parse_args( $args, $defaults );
 
-	$discounts = get_posts( $args );
+	$discounts_hash = md5( json_encode( $args ) );
+	if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
+		$discounts = get_posts( $args );
+		wp_cache_set( $discounts_hash, $discounts, 'edd-discounts' );
```
sunnyratilal: An expiry needs to be set to `3600` otherwise it default to `0` which means no expiration.

includes/discount-functions.php
```diff
@@ -32,19 +32,29 @@ function edd_get_discounts( $args = array() ) {
...
 		$args['meta_key']     = '_edd_discount_code';
 		$args['meta_value']   = $args['s'];
 		$args['meta_compare'] = 'LIKE';
 		unset( $args['s'] );
-		$discounts = get_posts( $args );
+
+		$discounts_hash = md5( json_encode( $args ) );
+		if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
+			$discounts = get_posts( $args );
+			wp_cache_set( $discounts_hash, $discounts, 'edd-discounts' );
```
sunnyratilal: Same here - an expiry needs to be set to `3600` otherwise it default to `0` which means no expiration.

includes/discount-functions.php
```diff
@@ -32,19 +32,29 @@ function edd_get_discounts( $args = array() ) {
...
 	if( ! $discounts && ! empty( $args['s'] ) ) {
-		// If no discounts are found and we are searching, re-query with a meta key to find discounts by code
 		$args['meta_key']     = '_edd_discount_code';
 		$args['meta_value']   = $args['s'];
 		$args['meta_compare'] = 'LIKE';
 		unset( $args['s'] );
-		$discounts = get_posts( $args );
+
+		$discounts_hash = md5( json_encode( $args ) );
+		if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
```
sunnyratilal: In order to maintain compatibility with PHP 5.3, please declare and call the function outside the `if` statement and then reference the variable.

i.e.
```php
$discounts = wp_cache_get( $discounts_hash, 'edd-discounts' )
if ( false === $discounts ) {
```

includes/discount-functions.php
```diff
@@ -32,19 +32,29 @@ function edd_get_discounts( $args = array() ) {
 
 	$args = wp_parse_args( $args, $defaults );
 
-	$discounts = get_posts( $args );
+	$discounts_hash = md5( json_encode( $args ) );
+	if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
```
sunnyratilal: In order to maintain compatibility with PHP 5.3, please declare and call the function outside the `if` statement and then reference the variable.

i.e.
```php
$discounts = wp_cache_get( $discounts_hash, 'edd-discounts' )
if ( false === $discounts ) {
```

JeroenSormani: `if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {` 
is valid for php 5.3 (verified), but I can understand if this is not a preferred method. Can you confirm if this needs changing either way?

includes/discount-functions.php
JeroenSormani: Thanks for spotting that, fixed it!

includes/discount-functions.php
sunnyratilal: Yes please - it does need changing

JeroenSormani: Just pushed the change :-)

includes/discount-functions.php
```diff
@@ -33,7 +33,8 @@ function edd_get_discounts( $args = array() ) {
 	$args = wp_parse_args( $args, $defaults );
 
 	$discounts_hash = md5( json_encode( $args ) );
-	if ( ! $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' ) ) {
+	$discounts = $discounts = wp_cache_get( $discounts_hash, 'edd-discounts' );
```
sunnyratilal: Needs changing to `$discounts = wp_cache_get( $discounts_hash, 'edd-discounts' );`

JeroenSormani: Oops! Fixed that.

pippinsplugins:
I'd like to see a few unit added for `edd_get_discounts()` before this merged in.

cklosowski:
If we're doing this, we need cache busting on adding/updating a discount.

cklosowski:
Actually i'm going to pull this into an issue branch and work up some unit tests on it as well.

cklosowski:
Issue #6039 created, closing this. your commit history will remain in tact @JeroenSormani

PR #5782, opened 2017-06-02
Use document.body where possible for performance

JeroenSormani:
Using the more efficient `$( document.body )` vs `$( 'body' )` for example.

As a extend to the previous issue; https://github.com/easydigitaldownloads/easy-digital-downloads/pull/3707

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -1411,7 +1411,7 @@ jQuery(document).ready(function ($) {
 			} );
 		},
 		recount_stats : function() {
-			$( 'body').on( 'change', '#recount-stats-type', function() {
+			$( document.body).on( 'change', '#recount-stats-type', function() {
 
 				var export_form   = $('#edd-tools-recount-form');
 				var selected_type = $('option:selected', this).data('type');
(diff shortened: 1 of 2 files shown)


spencerfinnell:
Will changing the selector on the `.trigger()` calls affect extensions that are binding to `'body'`?

JeroenSormani:
They aren't affected by this change.

Here's the validation of it;
https://cl.ly/2c3T3U1X212h

cklosowski:
This seems to be fine, the `trigger` references still seem to fire. I'm going to merge it into `release/2.8`.

I'm hesitant to do this in a minor point release. A larger release is better suited for something of this magnitude so we can get some people beta testing it in many different scenarios.

cklosowski:
I've had to revert this in `release/2.8` as it's caused some problems with our other gateways that hook into the JS, like Stripe.

cklosowski:
I've had to revert the changes here:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5782/files#diff-28089abe109462083feccaa41af37d26R307

and here:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5782/files#diff-28089abe109462083feccaa41af37d26R364

As the change caused some bad effects with gateways.

PR #5783, opened 2017-06-02
Remove HHVM from unit tests tests

JeroenSormani:
I'd suggest to follow WordPress' lead here and remove HHVM from the unit tests. 
So far I've only seen them give issues and make the tests fail. Would be neat to actually see accurate Travis results in the PR overview.

https://make.wordpress.org/core/2017/05/25/hhvm-no-longer-part-of-wordpress-cores-testing-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ matrix:
   include:
     - php: 5.3
       env: WP_VERSION=latest WP_MULTISITE=1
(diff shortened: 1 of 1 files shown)


cklosowski:
@JeroenSormani We've already done this for the upcoming 2.8 release in #5771

PR #5784, opened 2017-06-02
Fixed incomplete and failing Tests_API unit tests

DevinWalker:
## PR Overview

1. Fixed `test_get_queried_version()`
2. Fixed `test_missing_auth()  `
3. Fixed `test_invalid_key() `
4. Fixed `test_invalid_auth()`
5. Fixed `test_info()`
6. Fixed `test_process_query()`


cklosowski:
@DevinWalker we had a bunch of failed tests thanks to the HHVM stuff being removed. We've got that slated for `release/2.8` which is coming up soon. I've closed this PR and moved it to #5786, which is based from the release branch we're working up. Since the API class is changing some returns (instead of not returning at all) I think it should go in a major version. I cannot think of anything this should break, but just in case for some odd reason someone is looking for a return where there didn't used to be.

This should allow the unit tests to pass for the branch and if all is well, I'll merge for `release/2.8`. Superb man, I really appreciate this.

DevinWalker:
Great thanks @cklosowski

Issue #5802, opened 2017-06-12
Really Simple SSL: Download links do not work.


cklosowski:
We've had a few reports of this, and we've narrowed it down but it took a release for RSSSL to make it something we could address. Now that [the bug has been fixed](https://github.com/rlankhorst/really-simple-ssl/issues/30), we can address it within our code to address it.

PR #5803, opened 2017-06-12
Avoid Really Simple SSL mixed content buffer on file delivery #5802

cklosowski:
#5802


cklosowski:
Once CI passes I'll merge and get 2.7.10 out.

Issue #5809, opened 2017-06-14
Shortcode [downloads] doesn't support slugs that begin with numbers


mintplugins:
If you have a category or tag that starts with a number (like `7thgrade`), the downloads shortcode will think you passed an ID of `7`, rather than treating it like a slug - resulting in incorrect results. 

This could be fixed by using `is_numeric` as the check instead of `(int)`.

mintplugins:
To test, make a category and a tag which both start with numbers (like 7th grade). Then use `category`, `exclude_category`, `tag`, and `exclude_tag` using both the term id and the slug. Make sure results are as expected. 

For example:
```
[downloads category="1math" exclude_tags="7thgrade" orderby="title" order="ASC"]
```

spencerfinnell:
Related https://github.com/easydigitaldownloads/easy-digital-downloads/issues/1922

pippinsplugins:
Also related https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3638, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/2970, and https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3001.

mintplugins:
Closing this now because of https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3638. Our official position is now that you MUST use the ID of the category or tag and the [doc](http://docs.easydigitaldownloads.com/article/224-downloads) has been updated to remove any reference to the use of slugs.

PR #5810, opened 2017-06-14
Use is_numeric instead of (int) for slug checking

mintplugins:
#5809


mintplugins:
See #5809 for why this was closed.

Issue #5716, opened 2017-05-08
Amazon Payments Stops ALL payments (all gateways) and even shows error on Profile page


jcollake:
AWS is throwing a 'misconfigured' error, despite being configured properly (or it was before the last update weeks/months ago, afaik). It was popularly used. Then an update came and broke the site and it has never been resolved. I've got little to not clues, because I've found none.

pippinsplugins:
Hi there!

Please go ahead and open a support ticket through our support portal: https://easydigitaldownloads.com/support

we use GitHub for development tracking only.

cklosowski:
@pippinsplugins I've been working with him on this, I cannot find anything that causes it.

The error message received is:
`ERROR: here is an error in your Amazon Payments configuration`

jcollake:
Any old data that might cause it to think it is misconfigured? Aside from
that, when the error occurs, as you can surely make happen, it blocks ALL
gateways. That is the more serious issue really if the config is isolated
to me.

#mobile

amdrew:
Similarly, I'm seeing this message in many places that you shouldn't, except I don't have it configured at all. I purposely enabled `Amazon` from Downloads -> Settings -> Payment Gateways but haven't entered anything in the Amazon Payments section.

It affects quite a few areas:

___

It kills FES's `Vendor Details` page:

<img width="1639" alt="screenshot 2017-06-18 22 48 33" src="https://user-images.githubusercontent.com/52581/27260136-46319858-5478-11e7-8212-3a6faa392302.png">

___

The Download grid. Though the amazon notice will only show when no checkout page is configured in EDD.

<img width="889" alt="screenshot 2017-06-18 22 50 21" src="https://user-images.githubusercontent.com/52581/27260149-9aaa4812-5478-11e7-998b-f09b85fe9e8f.png">

___

Profile editor shortcode
<img width="884" alt="screenshot 2017-06-18 22 53 21" src="https://user-images.githubusercontent.com/52581/27260156-f9f8a318-5478-11e7-9026-5b95ac9bbaf1.png">

___

Register shortcode

<img width="862" alt="screenshot 2017-06-18 22 52 50" src="https://user-images.githubusercontent.com/52581/27260157-04423eb0-5479-11e7-8a65-2b866d3b8f96.png">

___

Login shortcode

[...]

PR #5816, opened 2017-06-18
Check for Amazon config during purchase processing instead of on init - issue/5716

pippinsplugins:
#5716


cklosowski:
Once CI passes I'm good with this.

Issue #5791, opened 2017-06-06
Add "Enable Product Reviews" link to product edit screen


pippinsplugins:
To help encourage people to add the Product Reviews plugin, let's add a `Enable Product reviews` metabox / link to the product edit screen when the EDD Reviews plugin is not activated.

danieliser:
I will say that these in dash contextual upsell links generate a substantial portion of Popup Makers revenue. Much more than say the Extensions/Addons page in dash.

cklosowski:
Ok updated this to a metabox. It has 2 statuses:
1) When `EDD-Reviews/edd-reviews.php` or `edd-reviews/edd-reviews.php` is a valid plugin installed (but not activated), it shows a button to `Activate Reviews`
![screen shot 2017-06-26 at 12 13 06 pm](https://user-images.githubusercontent.com/1390124/27558709-f7201784-5a72-11e7-92f1-6cb697ab35ee.png)

2) If the plugin is not found, it gives this message to send them to our sales store with UTM tags in a new tab:
![screen shot 2017-06-26 at 1 24 43 pm](https://user-images.githubusercontent.com/1390124/27558730-113bcfd2-5a73-11e7-912c-3a64a3a02393.png)

cklosowski:
Updated to be `Product Reviews`

cklosowski:
Once CI Passes, I'll merge this.

PR #5823, opened 2017-06-22
issue/5791 - Add `Enable Product Reviews` button

sunnyratilal:
#5791


cklosowski:
If we're adding a button here when Reviews does not exist, we should also add it when it does exist, but instead of linking, it should actually enable reviews. I'd be pretty confused if I didn't have Reviews installed, went and purchased, installed and activated, and now that button was missing.

So I think that Reviews needs to have an update to make a media button as well, but when activated it allows you to one click enable them.

pippinsplugins:
I don't think I agree with putting this in the media buttons. Instead I think it should be a metabox like the standard reviews box (enable / close) is when Reviews is active.

cklosowski:
@pippinsplugins that's what @sunnyratilal and I agreed to within Slack the other day too. It simply just makes the same metabox that would exist if Reviews is active, and once Reviews is active, EDD Core does not register it.

pippinsplugins:
:+1:

Issue #5636, opened 2017-04-04
`EDD_Payments_Query::get_payments()` may mess up global `$post`


felixarntz:
`EDD_Payments_Query::get_payments()` uses `wp_reset_postdata()` to reset the original post (as well as other globals), which generally works fine as long there _is_ a global `$wp_query` that has been run.

This is not always the case though. When editing a post in the admin, the `$post` global as well as other related globals are set manually. Since `wp_reset_postdata()` only looks at `$wp_query`, it doesn't work in this case, resulting in the last payment result of the query suddenly being the `$post`.

I had some custom logic in the admin editing screen for my downloads, and was querying payments for that download. This caused most functionality to break, since the `$post` was suddenly set to the payment, not the download anymore.

The following is an alternate version, that fixes that bug. Not optimal, but that's what I used for a quick fix (sorry, I altered the code directly, since it was critical :) ); we can probably improve that a bit more.

        public function get_payments() {

		do_action( 'edd_pre_get_payments', $this );

		$query = new WP_Query( $this->args );

		$custom_output = array(
			'payments',
			'edd_payments',
		);

		if ( ! in_array( $this->args['output'], $custom_output ) ) {
			return $query->posts;
		}

		if ( $query->have_posts() ) {
			$has_post = isset( $GLOBALS['wp_query'] ) && isset( $GLOBALS['wp_query']->post );
			$original_post = isset( $GLOBALS['post'] ) ? $GLOBALS['post'] : null;

			while ( $query->have_posts() ) {
				$query->the_post();

				$payment_id = get_post()->ID;
				$payment    = edd_get_payment( $payment_id );

				if ( edd_get_option( 'enable_sequential' ) ) {
					// Backwards Compatibility, needs to set `payment_number` attribute
					$payment->payment_number = $payment->number;
				}

				$this->payments[] = apply_filters( 'edd_payment', $payment, $payment_id, $this );
			}

			if ( $has_post ) {
				wp_reset_postdata();
			} elseif ( $original_post ) {
				$GLOBALS['post'] = $original_post;
			} else {
				unset( $GLOBALS['post'] );
			}
		}

		do_action( 'edd_post_get_payments', $this );

		return $this->payments;
	}

cklosowski:
I've replicated the problem with the EDD Dev Tools plugin (which now contains the last 5 payments of a download) when editing the download.

And after the table is done and the class is called `edd_payment` is the value for `$post->post_type`, not `download`.

[...]

PR #5829, opened 2017-06-23
Store and restore any existing globals for post and wp_query #5636

cklosowski:
#5636

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -50,6 +50,24 @@ class EDD_Payments_Query extends EDD_Stats {
 	public $payments = array();
 
 	/**
+	 * Holds a boolean to determine if there is an existing $wp_query global.
+	 *
+	 * @var bool
+	 * @access private
+	 * @since 2.8
+	 */
+	private $existing_query;
+
+	/**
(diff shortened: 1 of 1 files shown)


includes/payments/class-payments-query.php
```diff
@@ -180,7 +203,8 @@ public function get_payments() {
 				$this->payments[] = apply_filters( 'edd_payment', $payment, $payment_id, $this );
 			}
 
-			wp_reset_postdata();
+			$this->maybe_reset_globals();
```
pippinsplugins: Doesn't this need to be after line 210?

cklosowski: Probably a safe bet yeah. I'll move that down.

includes/payments/class-payments-query.php
```diff
@@ -180,11 +203,12 @@ public function get_payments() {
...
 			}
 
-			wp_reset_postdata();
 		}
 
 		add_action( 'edd_post_get_payments', array( $this, 'date_filter_post' ) );
 
+		$this->maybe_reset_globals();
+
 		do_action( 'edd_post_get_payments', $this );
```
pippinsplugins: I thin this do_action should fire before the rest.

Issue #5690, opened 2017-04-29
Add EDD_Customer object to complete purchase hooks


pippinsplugins:
Actions like `edd_complete_purchase` currently do not provide any access to the `EDD_Customer` object. The payment ID is available but accessing the customer requires setting up a new instance of `EDD_Customer`.

We should add the customer object as a parameter to the `edd_complete_purchase` hook and possibly others.

cklosowski:
If we do this we'll need to be sure we add it also inside #5453

sunnyratilal:
I've added the objects to the hook - we could refactor `edd_email_purchase_receipt()` to accept an `EDD_Payment` object as a parameter which would reduce the calls to `edd_get_payment_` methods from within the method.

cklosowski:
@sunnyratilal I think we still need to support the ID for backwards compatibility, but I see no harm in avoiding all the calls to `edd_get_payment_` instead of just instantiating a single `EDD_Payment` object once we're within the function.

sunnyratilal:
The `$payment_id` is still passed as the first parameter.

cklosowski:
@sunnyratilal correct, but what i'm saying is all the calls to `get_` could be avoided if we simply did ```
$payment = edd_get_paymetn( $payment_id );
$payment->email
```

The only difference is that we need to maintain the main `meta` content for all the hooks.

sunnyratilal:
Ah, I misunderstood :)

PR #5835, opened 2017-06-27
Issue/5690

cklosowski:
#5690

### DIFF
--- a/includes/emails/actions.php
+++ b/includes/emails/actions.php
@@ -16,19 +16,23 @@ if ( ! defined( 'ABSPATH' ) ) exit;
  * Triggers Purchase Receipt to be sent after the payment status is updated
  *
  * @since 1.0.8.4
- * @param int $payment_id Payment ID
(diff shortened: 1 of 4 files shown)


includes/emails/actions.php
```diff
@@ -16,19 +16,23 @@
...
- * @param int $payment_id Payment ID
+ * @since 2.8 - Add parameters for EDD_Payment and EDD_Customer object.
+ *
+ * @param int          $payment_id Payment ID.
+ * @param EDD_Payment  $payment    Payment object for payment ID.
+ * @param EDD_Customer $customer   Customer object for associated payment.
  * @return void
  */
-function edd_trigger_purchase_receipt( $payment_id ) {
+function edd_trigger_purchase_receipt( $payment_id, $payment, $customer ) {
```
cklosowski: We need to set some defaults here, since we're adding parameters that haven't existed before.

Just in case someone is calling this directly, even if we just set them to `null` so they trigger the `gets` in the end result of the action.

Issue #4736, opened 2016-06-29
Add support for "multiple" to chosen drop down in settings API


d4mation:
Currently we're able to make Select fields into Chosen fields via `'chosen' => true`, but if we want to allow multiple items to be selected via Chosen's native functionality (Or the less intuitive, basic HTML interface) that isn't currently possible.

I did some hacky regex in the meantime to enable it within my extension, but the data does not save:

```
add_filter( 'edd_after_setting_output, function( $html, $args ) {

    if ( $args['chosen'] === true ) {

        if ( $args['section'] == 'my-settings-section' ) {

            $html = preg_replace( '/(<select.*?[^>])*/i', '$1multiple="true" ', $html, 1 );

        }

    }

    return $html;

}, 10, 2 );
```

Screenshot (Before Saving):
![image](https://cloud.githubusercontent.com/assets/7770631/16461428/7eb4c99c-3dfa-11e6-9019-8a6044577911.png)

I'm now realizing that I could have just as easily made a custom Callback function rather than writing that regex, but the saving issue would have still persisted. `edd_get_option( 'my-field-id' )` always returns 1 value in this case.

I've been poking around and I believe it is the sanitization filters that are to blame. However, I can't seem to modify the data properly. Even throwing a `die()` in there doesn't seem to stop execution:

```
add_filter( 'edd_settings_sanitize_select', function( $value, $key ) {

    die( $value );

}, 10, 2 );
```

Ideally, having another option for Select fields to set `'multiple' => true` to enable this functionality would be best. I'm thinking this could be a pretty big addition for a lot of Extensions.

pippinsplugins:
@d4mation We actually do already support allowing the `multiple` option.

Example:

```php
echo EDD()->html->select( array( 'chosen' => true, 'multiple' => true ) );
```

pippinsplugins:
Oh wait, you mean in the `select` option callback in the settings API, yes?

d4mation:
Yep! I am talking about the Settings API. Looking around there isn't any support for it there, and when I hacked it in there it wouldn't save the data properly. I figure this has to do with `edd_settings_sanitize()` and/or its filters, but I can't seem to pin it down.

pippinsplugins:
Thanks! We can get support added for that.

d4mation:
I figured this out for a project and made it a Custom Field Type.

Went ahead and added that functionality to the Select Field Type in PR #5683.

[...]

PR #5837, opened 2017-06-27
Support 'multiple' in select settings API #4736

cklosowski:
#4736

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -979,10 +979,6 @@ function edd_settings_sanitize( $input = array() ) {
 						unset( $output[ $key ] );
 					}
 					break;
-				case 'select':
-					// Ensures that Multiple Selects that have been cleared out save as Empty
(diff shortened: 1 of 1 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -979,10 +979,6 @@ function edd_settings_sanitize( $input = array() ) {
 						unset( $output[ $key ] );
 					}
 					break;
-				case 'select':
-					// Ensures that Multiple Selects that have been cleared out save as Empty
-					if ( empty( $_POST['edd_settings'][ $key ] ) ) $output[ $key ] = '';
-					break;
```
d4mation: @cklosowski With this bit of code removed, if a value is saved into a Multiple Select and then the User attempts to clear out all of its values and Save those changes it will not work.

I see the problem that you mentioned in #4736 since this _is_ a fairly heavy-handed approach and apparently effects Select Fields not on the same Page. However, the inability to clear out the value isn't very good :/

We may need to find another way.

Issue #5841, opened 2017-06-28
Support Variable Price IDs in the payment import/export


cklosowski:
As it stands you can only import the item amount and tax. The tax is denoted inside `{}` and is always included. We could just append another item that gives the price ID...this allows an import and export of payments with prices, tax and price IDs for variable pricing.

cklosowski:
For testing, the new Raw Products format would be:
`Download Name|Price{tax}{price_id}` or for an actual item:
`Download Name|199.00{19.90}{1}` (for Download Name price ID 1)

pippinsplugins:
Looks good, merged.

PR #5842, opened 2017-06-28
Allow variable price IDs in import/export of payments #5841

cklosowski:
#5841


includes/admin/import/class-batch-import-payments.php
```diff
@@ -314,11 +315,15 @@ public function create_payment( $row = array() ) {
 
 					$item_price = ! isset( $price ) ? edd_get_download_price( $download_id ) : $price;
 					$item_tax   = ! isset( $tax ) ? ( $download_count > 1 ? 0.00 : $payment->tax ) : $tax;
+					$price_id   = ! isseT( $price_id ) ? false : $price_id;
```
pippinsplugins: `isseT` should be `isset`

Issue #5851, opened 2017-07-03
Fatal error: Uncaught Error: Call to undefined function edd_extension_has_beta_support()


pippinsplugins:
We've seen a few reports of Fatal error: Uncaught Error: Call to undefined function edd_extension_has_beta_support() happening when the updater class is loaded on the frontend.

The edd_extension_has_beta_support() is only registered in /wp-admin so it fatals if called on the frontend.

While I think it's odd to call the updater on the frontend, let's get a function exists check added to ensure it doesn't fatal.

PR #5852, opened 2017-07-03
Add function_exists() check to auto_updater() call for edd_extension_has_beta_support()

pippinsplugins:
#58512


SDavisMedia:
#5851

PR #5885, opened 2017-07-21
Small fees cleanup to make the code more readable

JeroenSormani:
Should make the code more readable;
The diff looks a bit wonky for this PR, but its only bunch of duplicate `! empty( $fees ) &&` checks

### DIFF
--- a/includes/class-edd-fees.php
+++ b/includes/class-edd-fees.php
@@ -189,12 +189,16 @@ class EDD_Fees {
 	public function get_fees( $type = 'fee', $download_id = 0, $price_id = NULL ) {
 		$fees = EDD()->session->get( 'edd_cart_fees' );
 
+		if ( ! is_array( $fees ) ) {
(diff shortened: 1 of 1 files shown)


JeroenSormani:
@pippinsplugins Since we force it to be an array here; https://github.com/easydigitaldownloads/easy-digital-downloads/pull/5885/files#diff-d77cb480639c965722daaf17f3545991R192
and unsetting elements from an array, even when emptying the `$fees` var out, it will remain an array that is valid to be used in foreach (won't do anything though). 

Double checked on 5.3 and 7.0 with the following code
```
$fees = array(
	array( 'amount' => 1 ),
	array( 'amount' => 2 ),
);

foreach ( $fees as $k => $fee ) {
	unset( $fees[$k]);
}

var_dump( $fees );
foreach ( $fees as $k => $fee ) {
	// Should give no error
}
```

PR #5886, opened 2017-07-21
Ensure EDD_Cart::get_discounts() return type is consistent

JeroenSormani:
Ensure the return type of the `EDD_Cart::get_discounts()` method is always an `array` as specified in the docblock. 
Which is also consistent with the older `edd_get_cart_discounts()` function.

Also two other minor changes in the EDD_Cart class
- the `quantity` property is already set the `get_quantity` method
- make the `get_contents()` return type accurate

### DIFF
--- a/includes/cart/class-edd-cart.php
+++ b/includes/cart/class-edd-cart.php
@@ -322,7 +322,8 @@ class EDD_Cart {
 
 		$has_discounts = false;
 
-		if ( ! empty( $this->get_discounts() ) ) {
+		$discounts = $this->get_discounts();
+		if ( ! empty( $discounts ) ) {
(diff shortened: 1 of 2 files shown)


includes/cart/class-edd-cart.php
```diff
@@ -322,7 +322,7 @@ public function has_discounts() {
 
 		$has_discounts = false;
 
-		if ( $this->get_discounts() ) {
+		if ( ! empty( $this->get_discounts() ) ) {
```
pippinsplugins: This fatal errors on PHP 5.3 and earlier.

JeroenSormani: Fixed that.

PR #5888, opened 2017-07-21
Refactor EDD_Stats::convert_date() method

JeroenSormani:
This is kinda like a proposal PR. Its probably best to implement some tests before actually merging, but before I write those I want to be sure you're on track with changing it to this as I've already spend a good amount of time to do this.

This refactor should make it easier to read / maintain the conversion of the dates. 

The PR is a result of a earlier bug (already resolved AFAIk) from some months ago where the dashboard was showing incorrect stats. 

Could be made even simpler by removing the `date( 'U', ` part as `strtotime` returns a UNIX timestamp, but I thought it could be a good addition for possibly the future where also a `$format` param for example may be returned. 

Let me know your thoughts!

### DIFF
--- a/includes/class-edd-stats.php
+++ b/includes/class-edd-stats.php
@@ -102,7 +102,9 @@ class EDD_Stats {
 	 *
 	 * @access public
 	 * @since 1.8
-	 * @return void
+	 *
+	 * @param string $_start_date Range it should get the date for.
+	 * @param bool $_end_date True when it should return the end date from the period.
 	 */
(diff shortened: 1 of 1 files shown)


JeroenSormani:
These conversions are used on the dash widget (not reporting). I used this script to override the `dashboard-widgets.php` to add all ranges to the dash + dates to verify.
https://gist.github.com/anonymous/ae0c15c16a0a6168bbce8737165f9802

Issue #5896, opened 2017-07-27
edd_receipt_shortcode calls edd_get_purcahse_id_by_key twice


cklosowski:
There is a duplicate query that get's run when you view a purchase receipt.

The specific set of code that causes it is:
```
$payment_id    = edd_get_purchase_id_by_key( $payment_key );
$user_can_view = edd_can_view_receipt( $payment_key );
```

The problem is that we get the `$payment_id` via the `edd_get_purchase_id_by_key` function, but then call `edd_can_view_receipt`, which calls the same function `edd_getpurchase_id_by_key`, resulting in two direct calls to the database:
`$purchase = $wpdb->get_var( $wpdb->prepare( "SELECT post_id FROM $wpdb->postmeta WHERE meta_key = '_edd_payment_purchase_key' AND meta_value = %s LIMIT 1", $key ) );`

PR #5904, opened 2017-07-29
use a global to prevent getting the same purchase ID in the same thread #5896

cklosowski:
#5896

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -1430,11 +1430,19 @@ function edd_set_payment_transaction_id( $payment_id = 0, $transaction_id = '' )
  */
 function edd_get_purchase_id_by_key( $key ) {
 	global $wpdb;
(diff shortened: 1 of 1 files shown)


includes/payments/functions.php
```diff
@@ -1430,11 +1430,19 @@ function edd_set_payment_transaction_id( $payment_id = 0, $transaction_id = '' )
  */
 function edd_get_purchase_id_by_key( $key ) {
 	global $wpdb;
+	$global_key_string = 'edd_purchase_id_by_key' . $key;
+	global $$global_key_string;
```
pippinsplugins: This is fugly but I kind of like it :D

cklosowski: @pippinsplugins yeah, I'm not a huge fan of the variable variables...but I think it's just the simplest method here.

PR #5909, opened 2017-07-30
issue/5846 - Unneeded validation of billing fields when cart total is $0

### LINKED ISSUE #5846: Unneeded validation of billing fields when cart total is $0
Initially reported here: https://github.com/easydigitaldownloads/edd-stripe/issues/234

The function `edd_purchase_form_validate_cc` runs Zip Code validation against a country, even when the cart is $0. This needs to be removed. There is no reason to run this validation unless those fields are present on the screen.

To replicate do the following ( I used stripe ):
1) Add product to the cart
2) Use Stripe
3) Set a postal code/zip
4) Remove country and state (set to blank)
5) Apply a 100% discount
6) Purchase.

What's happening is that in some cases, someone might change the fields that are required for checkout. If they initially used no fields (like stripe allows), when the stripe card/customer is created it may not have those fields, however, if you then change the validation to require country/zip or full address, the card data may not exist for it, and since the field forms are empty, it won't pass CC validation, even though it's free.

sunnyratilal:
Fixes #5846

### DIFF
--- a/includes/process-purchase.php
+++ b/includes/process-purchase.php
@@ -877,7 +877,7 @@ function edd_purchase_form_validate_cc() {
 	$card_data = edd_get_purchase_cc_info();
 
 	// Validate the card zip
-	if ( ! empty( $card_data['card_zip'] ) && ! edd_get_cart_total() > 0.00 ) {
+	if ( ! empty( $card_data['card_zip'] ) && edd_get_cart_total() > 0.00 ) {
 		if ( ! edd_purchase_form_validate_cc_zip( $card_data['card_zip'], $card_data['card_country'] ) ) {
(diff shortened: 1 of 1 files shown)


includes/process-purchase.php
```diff
@@ -877,7 +877,7 @@ function edd_purchase_form_validate_cc() {
 	$card_data = edd_get_purchase_cc_info();
 
 	// Validate the card zip
-	if ( ! empty( $card_data['card_zip'] ) ) {
+	if ( ! empty( $card_data['card_zip'] ) && 0.00 !== edd_get_cart_total() ) {
```
pippinsplugins: I'd like to see this changed to `&& ! edd_get_cart_total() > 0` to ensure extensions that filter the cart amount cannot cause this to not work due to the strict type check.

PR #5910, opened 2017-07-30
issue/5684 - Discounts with multiple exclusions can be applied to excluded products

### LINKED ISSUE #5684: Discounts with multiple exclusions can be applied to excluded products
When a discount code has multiple downloads excluded, the code can be applied on the checkout when an excluded item is in the cart. The total amount isn't reduced, but the discount code is attached to the payment and counted as being used.

Steps to reproduce:
- Create new discount code, add multiple items to the Exclude Downloads list.
- Add one of the excluded items to the cart.
- Apply the discount code. Note that the code is applied.
- Complete purchase.

When viewing the payment details, you can see the code attached to the payment. The discount code's uses count is increased as well.

When a discount code has only one exclusion, it is properly blocked from being applied.

sunnyratilal:
Fixes #5684

### DIFF
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -1651,18 +1651,11 @@ class EDD_Discount {
 		}
 
 		if ( ! empty( $excluded_ps ) ) {
-			foreach ( $excluded_ps as $download_id ) {
-				if ( empty( $download_id ) ) {
-					continue;
-				}
+			if ( count( array_intersect( $cart_ids, $excluded_ps ) ) == count( $cart_ids ) ) {
+				$return = false;
 
-				if ( edd_item_in_cart( $download_id ) ) {
-					if ( $set_error ) {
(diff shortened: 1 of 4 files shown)


includes/class-edd-discount.php
```diff
@@ -1606,72 +1606,57 @@ public function is_product_requirements_met( $set_error = true ) {
...
-			switch( $this->product_condition ) {
-
+			switch ( $this->product_condition ) {
 				case 'all' :
-
 					// Default back to true
 					$return = true;
 
 					foreach ( $product_reqs as $download_id ) {
-
```
pippinsplugins: Please don't remove this kind of whitespace. Let the code breath.

sunnyratilal:
@pippinsplugins Whitespace re-added.

pippinsplugins:
Thanks.

Let's hold off on merging this until tests are fixed. CC @cklosowski

Issue #5891, opened 2017-07-24
Allow multiple checkout pages


tnorthcutt:
This is a for client project that @tubiz and I are going to be working on, and will be PRing into EDD core.

The title says it all: right now, EDD only supports one checkout page; we'll be adding support for multiple checkout pages.

SDavisMedia:
First of all, 😍 ... and also, thanks so much for the contribution! I'm going to post remarks below as if we wrote the code. I'll let the lead devs decide how to proceed. I'm just here to test. 😄 

Okay, I'm getting a JS error on the Downloads -> Settings -> General page the moment I click into the Chosen field:
```
Uncaught TypeError: Cannot read property 'length' of undefined
http://localhost/edd/wp-content/plugins/easy-digital-downloads/assets/js/admin-scripts.js?ver=2.8.1 line 1289
```

The pages save just fine and each of them seems to be able to process a payment. Getting to "a" checkout page might be a problem, though. Not sure how we plan to make `edd_get_checkout_uri()` backwards compatible? It's being used in a number of different ways, including the Add to Cart button. It just returns `false` now.

I really can't think of anything else to test. The behavior should be the same across all checkout pages. So it's just a matter of when to use what checkout page we'll have to figure out. 🤔

tubiz:
Thanks for the comment Sean. I will look into the JS error on the console I was able to replicate it on my end too. 

I only checked `edd_is_checkout()` to make sure it still work with the multiple checkout pages. I totally forget about `edd_get_checkout_uri()`. 

I am working on making `edd_get_checkout_uri()` backwards compatible as  well. A single uri will have to be returned from the multiple pages set.

SDavisMedia:
Very much appreciated, @tubiz!

tnorthcutt:
@SDavisMedia regarding `edd_get_checkout_uri()`, do you have a preference on how returning a single uri is handled? For instance, should there be some UI to indicate a "primary" checkout page?

SDavisMedia:
@tnorthcutt the thought of intentionally setting a primary checkout page did cross my mind. I don't think it's safe to automatically assume one for the user. He or she should make that decision, in my opinion.

Would love to hear from @pippinsplugins and @cklosowski here as well since this could potentially complicate a very basic UI.

cklosowski:
I have a suggestion that I think might work but let me get non-mobile so
that I can get code snippets in easily.

[...]

PR #5914, opened 2017-08-01
Allow multiple checkout pages #5891

tubiz:
This allow for multiple checkout pages to be set in EDD. Presently the `edd_global_vars` JavaScript variable is only added to the checkout page that is set on the settings page and this doesn't allow multiple pages to be used. 
So if `[download_checkout]` shortcode is added to a different page the customer won't be able to checkout. 

#5891


cklosowski:
Closing in favor of #6083.

Issue #5918, opened 2017-08-02
Strict in_array() comparison for Settings API Select Multiple


d4mation:
Unless you do a strict comparison for `in_array()` it can cause unexpected `<option>`s to be selected once the callback function completes.

For example, if a Setting has the following saved:

```
$value = array(
    '15-1',
);
```

And the following is available as `$args['options']`:

```
$args['options'] = array(
    '15' => 'Should NOT be selected',
    '15-1' => 'Should be selected',
    '15-2' => 'Does not get selected',
)
```

The`<option>`s with values `15` AND `15-1` get selected. I suppose this is due to `15` not containing any _different_ characters than `15-1` while `15-2` does contain different characters. According to the [PHP Docs](http://php.net/manual/en/function.in-array.php), setting it to a Strict Comparison is meant to check against the Type of the Needle but it also forces it to match the entire String it would seem.

PR Incoming~

d4mation:
Figured I'd add screenshots of my example:

In this case, only `15-1` should be selected. However, without adding the Strict Comparison Flag, it ends up selecting `15` as well.

![image](https://user-images.githubusercontent.com/7770631/28889614-8c699b30-7792-11e7-95c8-feb71580e1c7.png)

![image](https://user-images.githubusercontent.com/7770631/28889633-9b306f90-7792-11e7-98f8-833e505bc9e3.png)

PR #5919, opened 2017-08-02
issue/5918 - Strict in_array() comparison for Settings API Select Multiple

d4mation:
Unless you do a strict comparison for `in_array()` it can cause unexpected `<option>`s to be selected once the callback function completes.

For example, if a Setting has the following saved:

```
$value = array(
    '15-1',
);
```

And the following is available as `$args['options']`:

```
$args['options'] = array(
    '15' => 'Should NOT be selected',
    '15-1' => 'Should be selected',
    '15-2' => 'Does not get selected',
)
```

The`<option>`s with values `15` AND `15-1` get selected. I suppose this is due to `15` not containing any _different_ characters than `15-1` while `15-2` does contain different characters. According to the [PHP Docs](http://php.net/manual/en/function.in-array.php), setting it to a Strict Comparison is meant to check against the Type of the Needle but it also forces it to match the entire String it would seem.

#5918

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -1796,7 +1796,7 @@ function edd_select_callback($args) {
 			$html .= '<option value="' . esc_attr( $option ) . '" ' . $selected . '>' . esc_html( $name ) . '</option>';
 		} else {
 			// Do an in_array() check to output selected attribute for Multiple
-			$html .= '<option value="' . esc_attr( $option ) . '" ' . ( ( in_array( $option, $value ) ) ? 'selected="true"' : '' ) . '>' . esc_html( $name ) . '</option>';
+			$html .= '<option value="' . esc_attr( $option ) . '" ' . ( ( in_array( $option, $value, true ) ) ? 'selected="true"' : '' ) . '>' . esc_html( $name ) . '</option>';
 		}
 		
 	}


d4mation:
I see. I saw you mentioned casting the Needles to Strings for comparison purposes but still saving/utilizing as Mixed Types for every other purpose. Is there a particular problem with doing that? Is it simply a performance concern?

cklosowski:
@d4mation I think the casting change for *comparison only* _might_ get us past the hurdles there. I think we were already past the point of a major overhaul on the old data. The key is testing with more than just EDD Core. Saving options with another data set that might store the data as an int or a string.

d4mation:
I looked into casting all the Keys for the stored value to Strings (Which I figured would allow better comparison against the Needle) and it turns out that [it isn't possible because PHP is PHP](https://stackoverflow.com/a/23478614). Tried it myself to confirm and PHP auto-detects that something _could_ be an integer and immediately converts it without asking.

_However_, casting just the Needle to a String works without needing to use a Strict Comparison for `in_array()`. This may solve the problem as a String that is equivalent to an Int should still return `true`, but an _actual_ String like `15-1` should return `false` for an Int like `15`.

SDavisMedia:
Where does this stand at the moment?

PR #5951, opened 2017-08-14
Flex recovery rules when guest checkout is enabled #5926

### LINKED ISSUE #5926: Cart recovery requires login when it could be smoother
Currently, if you are recovering a cart and are logged out, you are required to login - even if guest mode is enabled. 

@pippinsplugins suggested that we only require login if the email entered on checkout is different than the one attached to the cart recovery payment record. 

The arose out of [this ticket](https://secure.helpscout.net/conversation/407870291/64228/?folderId=749231), where the customer noticed that they have to login (guest mode enabled), if they cancel while at paypal and then immediately try and checkout again.

cklosowski:
resolves #5926

### DIFF
--- a/includes/error-tracking.php
+++ b/includes/error-tracking.php
@@ -27,16 +27,25 @@ if ( ! defined( 'ABSPATH' ) ) exit;
 function edd_print_errors() {
 	$errors = edd_get_errors();
 	if ( $errors ) {
+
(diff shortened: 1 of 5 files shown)


cklosowski:
I'm merging this, the broken build is from #5953, which I'll fix in master directly.

PR #6015, opened 2017-09-22
Add country name to the payments export, with helper function and tests #5629

### LINKED ISSUE #5629: Payment Export: Country Name - Report export improvement
Having the ISO country code is technically more accurate than the country name, but for ease of reading (and filing taxes) it's quicker to have the actual country name. Could an additional column be added called "Country Name" to the report export?

cklosowski:
resolves #5629

### DIFF
--- a/includes/country-functions.php
+++ b/includes/country-functions.php
@@ -1670,6 +1670,6 @@ function edd_get_country_name( $country_code = '' ) {
 	$country_list = edd_get_country_list();
 	$country_name = isset( $country_list[ $country_code ] ) ? $country_list[ $country_code ] : $country_code;
 
-	return apply_filters( 'edd_country_label', $country_name, $country_code );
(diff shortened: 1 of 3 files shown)


includes/country-functions.php
```diff
@@ -1658,3 +1658,18 @@ function edd_get_spain_states_list() {
...
+ * @since 2.8.7
+ * @param string $country_code The ISO Code for the country
+ *
+ * @return string
+ */
+function edd_get_country_name( $country_code = '' ) {
+	$country_list = edd_get_country_list();
+	$country_name = isset( $country_list[ $country_code ] ) ? $country_list[ $country_code ] : $country_code;
+
+	return apply_filters( 'edd_country_label', $country_name, $country_code );
```
pippinsplugins: Can we please make the filter name match the function name?

cklosowski: Gah sorry It did at one point, just decided to change the function name :)

Issue #4864, opened 2016-08-11
Add a debugging class


cklosowski:
The EDD error handling needs to be improved and following the footsteps of AffiliateWP we can introduce a logging class (https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/class-logging.php) that will allow us to easily log errors and exceptions. This will be great for core and extensions as it gives us a place to store those exceptions and off cases so we can better offer support later.

Off the top I think it should have a things (up for discussion):
- Allow a 'context' reference (Core, Software Licensing, Mailchimp), which would ake it easier to see where the log originated from
- Auto timestamp the errors coming in
- Add a method for popping the last log off the top
- Allow some sort and count parameters when getting logs

michaelbeil:
Do you think we should add a debug assistant tool a la AffiliateWP as well? 
![screen shot 2016-08-11 at 12 15 54 pm](https://cloud.githubusercontent.com/assets/4752949/17597861/61cb0d34-5fbd-11e6-85b8-2d0ba4927df6.png)

cklosowski:
The goal is to do all of that yes.

Chris Klosowski
cklosowski@gmail.com
@cklosowski

michaelbeil:
Right on.

cklosowski:
Some more thoughts on this that we had in a discussion on Slack with @chriscct7 

- Still use the wp_log type, which allows easy use of the wp list tables and searching and bulk actions.
- {expand/delete buttons} |Importance Level | Name of Issuer | Action when issuing | Title (character limited to post_title in the unminimized view) |  Date/Time of Event
- Possibly we could do like i’m doing with PPP as well, using the post_content as json https://thomasgriffin.io/wordpress-performance-outside-box-handling-retrieving-data/
- Use the post_author to report the person logged in (or customer) triggering the action
- Also instead of a importance level column, we'd do 1 for importance and 1 for type so you can distinguish an important notice from an urgent error

Chris Klosowski [4:00 PM]  
this all seems like a great idea :slightly_smiling_face: I think we should do it
- we should also let the issuer provide an expiry, default false, so by default a logged event would be kept forever, but extensions could make notices that are cleared from the post table. That way if someone wanted tehy could build a stream like system for logging edd stuff. After expiry, a daily cron would remove the expired log items, so that we could keep some logs forever, but let informational ones expire

[4:04]  
[...]

PR #6020, opened 2017-09-25
Debug mode - Issue/4864

pippinsplugins:
#4864

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -671,7 +671,7 @@ function edd_get_registered_settings() {
 					'debug_mode' => array(
 						'id'   => 'debug_mode',
 						'name' => __( 'Debug Mode', 'easy-digital-downloads' ),
-						'desc' => __( 'Check this box to enable debug mode. When enabled, debug messages will be logged and shown in Downloads &rarr; Tools &rarr; Debug Mode.', 'easy-digital-downloads' ),
+						'desc' => __( 'Check this box to enable debug mode. When enabled, debug messages will be logged and shown in Downloads &rarr; Tools &rarr; Debug Log.', 'easy-digital-downloads' ),
 						'type' => 'checkbox',
 					),
 					'uninstall_on_delete' => array(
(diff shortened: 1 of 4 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -668,6 +668,12 @@ function edd_get_registered_settings() {
 						'desc' => sprintf(__('Allow quantities to be adjusted when adding %s to the cart, and while viewing the checkout cart.','easy-digital-downloads' ), edd_get_label_plural( true ) ),
 						'type' => 'checkbox',
 					),
+					'debug_mode' => array(
+						'id'   => 'debug_mode',
+						'name' => __( 'Debug Mode', 'easy-digital-downloads' ),
+						'desc' => __( 'Check this box to enable debug mode. When enabled, debug messages will be logged and shown in Downloads &rarr; Tools &rarr; Debug Mode.', 'easy-digital-downloads' ),
```
sunnyratilal: Needs changing to `Downloads &rarr; Tools &rarr; Debug Log`

sunnyratilal:
@pippinsplugins Is it necessary to have [this](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/88d6dfcbb1799f267778b2086e45d122af53aa5e/includes/class-edd-logging.php#L483). Instead, could it be added as an object to the `Easy_Digital_Downloads` class like we do with our other classes?  Just so it keeps everything consolidated?

sunnyratilal:
Besides my above comments, I've tested the functionality and it seems to be working well 👍

pippinsplugins:
@sunnyratilal that global has been there since _way_ back and is unrelated to this change so I'd rather we not change it.

chriscct7:
It'd be nice if edd_is_debug_mode could be additionally turned on via a constant

```
$ret = edd_get_option( 'debug_mode', false );
$ret = $ret || ( is_defined( 'EDD_DEBUG_MODE' ) && EDD_DEBUG_MODE );
```

chriscct7:
Personally, re:global, I'd remove it like sunny suggested (self::$instance->logs = new EDD_Logging(); ) and then in the bottom of EDD main file do:

$GLOBALS['edd_logs'] = EDD()->logs;

That way you can use the logs off the global object (expected) and if you have a person editing the log object it will also affect the global in EDD so that other plugins don't have unexpected results.

That's fully backwards compatible, and is a lot more expected

PR #6021, opened 2017-09-25
Strict type casting for stats updates.

### LINKED ISSUE #5623: Lots of 'A non-numeric value encountered' errors with php 7.1
As Travis shows here. https://travis-ci.org/easydigitaldownloads/easy-digital-downloads/jobs/217205739

cklosowski:
Resolves #5623

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ php:
 - 5.5
 - 5.6
 - 7.0
+- 7.1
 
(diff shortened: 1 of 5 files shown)


cklosowski:
There is apparently an issue with WP Core on PHP 7.1 on versions lower than `WP 4.7` so Removing them from the test matrix.

PR #6040, opened 2017-09-28
Performance improvements and unit tests for edd_get_discounts

### LINKED ISSUE #6039: Performance improvements to edd_get_discounts
as reported by @JeroenSormani in #5781.

Opening up as an issue so we can make a new branch and finish it up.

cklosowski:
Resolves #6039 

Props @JeroenSormani


includes/admin/discounts/discount-actions.php
```diff
@@ -76,6 +76,7 @@ function edd_add_discount( $data ) {
 
 		if ( edd_store_discount( $posted ) ) {
 
+			wp_cache_flush();
```
pippinsplugins: This flushes the entire cache, not just the one for discounts. Could we avoid that?

sunnyratilal:
Yeah, this should be using `wp_cache_delete` in lieu of `wp_cache_flush`.

But we do run into problem due to the cache key. It's hashed using `md5( json_encode( $args ) )`.€##€

includes/admin/discounts/discount-actions.php
cklosowski: Nope the only cache deletion available requires a hash and since the hashes are based on args the only way to guarantee it's  a cleared is to kill it all  There is not a way to delete an entire group. We _could_ look at supporting that non-persistent flag, I linked you to yesterday.

pippinsplugins: Then  how about replacing the hash with a predictable cache key? Deleting _all_ caches is just being a bad citizen.

JeroenSormani:
It is possible in A way to remove/clear a cache group, there's just not a WP function available for it. 
Doing the following should clear out the group:
```
global $wp_object_cache;

unset( $wp_object_cache->cache['edd-discounts'] );
```

Haven't tested it yet, but if it doesn't work (not entirely use due to magic `__unset()` method being used), there are also other ways to clear it.

cklosowski:
That doesn't take care of persistent caches though in my experience.

cklosowski:
A predictable hash doesn't work either because any number of hashes could exist due to it being derived from args

cklosowski:
I think our best bet is to just make it a cache used non persistently.

cklosowski:
wp_cache_add_non_persistent_groups

https://codex.wordpress.org/Class_Reference/WP_Object_Cache

JeroenSormani:
Not sure what you're thinking about, but another alternative would be to not add it to a group, but create our own array of discounts in a single `edd-discounts` as the `key` in `wp_cache_set( $key, $val )`

JeroenSormani:
Re non_persistent_groups, would that work for the std wp cache given this note on the docs:
> NB: only some caching plugins implement this function!

cklosowski:
I'm not 100% positive on the size of the object cache, especially since it's possible for caching plugins and infrastructures to be offloading it to other platforms like redis or memcache, so I'm thinking we just use a global, which will resolve it for page loads, but not between diffrent sessions, plus let's us avoid messing with WP Cache.

sunnyratilal:
This looks good @cklosowski. Just tested and working as expected.

Issue #6049, opened 2017-09-30
Fix a typo in the docs for the payments create CLI command


DrewAPicture:
Discovered a typo today in the command examples for `edd payments create`. "completed" isn't a valid status ;)

DrewAPicture:
PR: #6050

sunnyratilal:
Merged, thank you @DrewAPicture

PR #6050, opened 2017-09-30
issue/6049 – Use a valid status in the payments create command docs.

DrewAPicture:
Issue: #6049


sunnyratilal:
Thanks @DrewAPicture

PR #6053, opened 2017-10-02
issue/6031 - Newly register user's ID is not set on pre-existing payments

### LINKED ISSUE #6031: Newly register user's ID is not set on pre-existing payments
`edd_add_past_purchases_to_new_user()` doesn't do what it's supposed to; any past guest purchases are left with their user ID set to 0.

PR forthcoming.

sunnyratilal:
Resolves #6031


sunnyratilal:
@cklosowski unit tests have been added

PR #6057, opened 2017-10-03
Add link to product to product name

X-Raym:
Add link to product to product name, using permalink

Very useful to have this backlink, I think it should be in the default template.

### DIFF
--- a/templates/shortcode-receipt.php
+++ b/templates/shortcode-receipt.php
@@ -147,7 +147,7 @@ $status    = edd_get_payment_status( $payment, true );
 						?>
 
 						<div class="edd_purchase_receipt_product_name">
-							<?php echo esc_html( $item['name'] ); ?>
+							<a href="<?php echo( get_post_permalink( $item['id'] ) ); ?>"><?php echo esc_html( $item['name'] ); ?></a>
 							<?php if ( edd_has_variable_prices( $item['id'] ) && ! is_null( $price_id ) ) : ?>
 							<span class="edd_purchase_receipt_price_name">&nbsp;&ndash;&nbsp;<?php echo esc_html( edd_get_price_option_name( $item['id'], $price_id, $payment->ID ) ); ?></span>
 							<?php endif; ?>


cklosowski:
@X-Raym is there a specific use case where you need to send the person back to the sales copy after purchase?

by default the permalink generated would be to the place where it was purchased from. In cases like using the Content Restriction plugin that would be useful, but there are many cases where it isn't desired. Think on a single product site like Affiliate WP or Restrict Content pro (just as examples) where the product landing page is actually even hidden using the `EDD Hide Download` add-on, since it's a single product site and the sales copy is a custom page.

I think in 'some' cases it makes sense to link this, I'm just curious what your specific case is.

X-Raym:
@cklosowski Hi ! Apart some WordPress plugins and service, I sell scripts for an audio software. Scripts are hosted on a private GitLab repo. To download the scripts, the users only have to copy paste a link in the software (then the software download the scripts). 
Untill today, I had a note on the Purchase Confirmation page to tell the client to go back on the product page to see instructions. On the product page, instructions were displayed for user with a current valid license (thanks to custom shortcode): a unique special url is given (the url contains the user license key).

I added this installation instructions on the product page cause it was simpler to get the product ID and set the integration than on the product confirmation page (which implies using actions to add data, modifying the templates, return the original product page etc...).

In my case, adding a backlin right on the purchase confirmation page makes sense.

Users were a bit confuses by this system though, and I get emails about that. So I decided to make it easier for them and added a backlink right on the Product Confirmation page.

I finally even get further and add installation instructions with the custom url+key right on the confirmation page too.

I though that the backlink could be useful so that the user can go back on the product page right on the product confirmation page, and as post link is generated by WordPress (not recorded in database), I don't think it can break things. Note that I also use Hide Download and Affiliate WP.

But you are right, I didn't think about the case where the download is embedded as shortcode in another post. In this case, then my pull request may not fit.

It was just a proposition, as you already have a template  system which allows users to modify the display based on their own needs, it is ok if you don't accept this pull request. I think in most case (the download purchase is made right on the download post) it could be useful, but it may not fit all the case indeed.

cklosowski:
Thanks for the detailed explanation. I really appreciate you taking the time to do that. I think there is a valid use case for it, however I don't think it should be the default, but if we added a filter to this, it would allow extensions to modify it programatically.

You think you could update the PR to use a filter? I think I'd like to see it do the following:
1) Filter of `edd_purchase_receipt_product_title_url`, with `$item` being passed into the filter.
2) By default the returned value from the filter is `''` or empty string.
3) Wrap the title in the anchor tags _only_ if the `$url` isn't empty.

This would allow you to programmatically add the link without modifying the EDD template (and missing any updates we do to it, without manually applying the ), and it would also allow other extensions to do work here in cases where it makes sense.

What are your thoughts there?

X-Raym:
@cklosowski Understood!
Indeed, filters are always appreciated 👍 

Rather than having a `$url` variable and having to make a comparison and add `<a>` tags etc..., why not simply just put the filter right on `esc_html( $item['name'] )`, with `$item` as extra parameters ? It will then simply return the string if there no filter. Would be one line long fix only 🍰

PR #6071, opened 2017-10-10
Expose entire customer address to tax rate filter

petemolinero:
Exposing the entire customer address to the tax rate filter makes proper sales tax calculation possible for U.S. destination-based tax states.

In the US, there are 34 states that used destination-based tax rates. For those states, the tax region is usually determined by county and city jurisdiction. Unfortunately, there's no way to identify a customer's proper tax rate without looking at the customer's entire address, since some zip codes and cities span multiple tax regions.

Currently, there's no way in EDD to properly figure sales tax for U.S. destination-based states since the filter only exposed state and country. By exposing the entire customer address to the tax rate filter, as I've done in this pull request, it allows people like myself to write custom filters that can figure out the proper tax rate for a customer based on their entire address (using a service like [TaxCloud](https://taxcloud.net/)).

Note: I wasn't sure what build process you typically use for minifying `edd-checkout-global.js`, so I simply minified it and pasted it in. I'm happy to correct this if there's another minification process you prefer.


cklosowski:
We minify prior to release. If you could revert any commits to the minified
file that would be ideal and reduce conflicts.

petemolinero:
Sounds good! I've reverted all changes to the minified file.

includes/tax-functions.php
```diff
@@ -52,6 +52,11 @@ function edd_get_tax_rate( $country = false, $state = false ) {
 
 	$user_address = edd_get_customer_address();
 
+	$address_line_1 = !empty( $_POST['card_address'] ) ? $_POST['card_address'] : '';
+	$address_line_2 = !empty( $_POST['card_address_2'] ) ? $_POST['card_address_2'] : '';
+	$city = !empty( $_POST['card_city'] ) ? $_POST['card_city'] : '';
+	$zip = !empty( $_POST['card_zip'] ) ? $_POST['card_zip'] : '';
```
cklosowski: Let's get these `=` signs lined up to match our preferred whitespace alignment.

assets/js/edd-checkout-global.js
```diff
@@ -344,15 +344,25 @@ function recalculate_taxes(state) {
...
 	var postData = {
 		action: 'edd_recalculate_taxes',
-		billing_country: $edd_cc_address.find('#billing_country').val(),
+		card_address: card_address,
+		card_address_2: card_address_2,
+		card_city: card_city,
+		card_zip: card_zip,
 		state: state,
-		card_zip: $edd_cc_address.find('input[name=card_zip]').val()
+		billing_country: billing_country
```
cklosowski: We need to leave `card_zip` in this array of things being posted so that anyone triggering on this JS can still use it, if they are expecting it.

petemolinero: Hi @cklosowski , `card_zip` is still in there, just above the `state: state`, item. It's just less noticeable because I moved `card_zip = $edd_cc_address.find('#card_zip').val();` up to the top for consistency's sake (I did the saem for `state = $edd_cc_address.find('#card_state').val();`).

cklosowski: @petemolinero Ah, sorry I totally missed that. Thanks! This is fine then.

For future reference, moving things for 'consistency' is better left unless the whole section is refactored as now when trying to find when things were add/removed requires jumping a few different 'blame' commits back. It's not a technical issue, just makes managing change history harder when debugging.

petemolinero: @cklosowski Got it! Will remember that in the future!

petemolinero:
@cklosowski Equals sign alignment updated. This should be good to go.

includes/tax-functions.php
```diff
@@ -52,6 +52,11 @@ function edd_get_tax_rate( $country = false, $state = false ) {
 
 	$user_address = edd_get_customer_address();
 
+    $address_line_1 = !empty( $_POST['card_address'] ) ? $_POST['card_address'] : '';
+    $address_line_2 = !empty( $_POST['card_address_2'] ) ? $_POST['card_address_2'] : '';
+    $city           = !empty( $_POST['card_city'] ) ? $_POST['card_city'] : '';
+    $zip            = !empty( $_POST['card_zip'] ) ? $_POST['card_zip'] : '';
```
cklosowski: Sorry one more change here @petemolinero totally missed it last time, can we drop a `sanitize_text_field()` around these assignments? Just to be sure we're sanitizing the `$_POST` data before assigning to the variables `$address_line_1, $address_line_2, $city, $zip`?

petemolinero: @cklosowski Done!

PR #6083, opened 2017-10-16
Add support for multiple checkouts #5891

### LINKED ISSUE #5891: Allow multiple checkout pages

cklosowski:
resolves #5891


tnorthcutt:
@cklosowski anything I can do from here to help make sure this is ready for merge?

cklosowski:
@tnorthcutt if you want to test it out, that'd be awesome. I think I worked through some changes with @pippinsplugins into this solution, so if he's good with it as well.

includes/checkout/functions.php
```diff
@@ -27,15 +27,18 @@ function edd_is_checkout() {
 	$is_checkout      = is_page( edd_get_option( 'purchase_page' ) );
 
 	if( ! $is_object_set ) {
-
 		unset( $wp_query->queried_object );
-
+	} else if ( is_singular( apply_filters( 'edd_page_metabox_post_types' , array( 'page' ) ) ) ) {
```
pippinsplugins: This filter name doesn't make sense to me. Is this meant to indicate this is a post type that is supported as an EDD checkout screen?

cklosowski: Appears that my previous solution has a left over filter. The goal was to only allow specific types of custom posts to trigger the condition.

Yes. The goal is to make it so that a specific set of post types could trigger this condition.

That filter was left over from having the checkbox in a custom meta box to define something as a checkout page. But since we swayed away from that...thoughts on a better filter name?

tnorthcutt: My initial reaction is that it seems fine to say that a checkout "page" can only be a WP Page.

However, if you want to use a filter to allow other post types to function as a checkout "page", perhaps `edd_checkout_post_types` would be a good filter name.

pippinsplugins: `edd_checkout_post_types` is definitely better.

mindctrl: Does it matter what the post type is? Why not just support anything people want to add the shortcode to? The function already has a filter anyway.

cklosowski: Yeah I like that.

tnorthcutt:
shia-clapping

PR #6116, opened 2017-10-27
Honeypot validation during registration. Fixes #6115

### LINKED ISSUE #6115: Registration Form Honeypot Check
Hi,

The registration form (_[edd_register]_) includes a [Honeypot field](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb881ef270d43d815754690270074940e9158faa/templates/shortcode-register.php#L43) however I see no validation of this field during the registration process.

I'm being hit with a large number of registrations despite having processes in place to protect against BOT registrations on the main WP registration forms. I guess therefore that these registrations are coming from the EDD form.

PR submitted adding a new `edd_do_honeypot_check()` function and being referenced during registration process

Thanks
Mike

mikeyhoward1977:
The honeypot field generated via `[edd_register]` is never validated.

Added `edd_do_honeypot_check()` function to check for value within this field.

Added call to `edd_do_honeypot_check()` during registration process


cklosowski:
I'm going to merge this into an issue branch and rework it a little bit. By doing this you'll still retain the original props @mikeyhoward1977

PR #6118, opened 2017-10-28
readmefinal.txt

Sulekhiya:
just removed spaces


sunnyratilal:
The space are there to keep the document organised and also allow for easier reading hence I'm closing this PR.

The savings that would occur for the size of the file are negligible.

PR #6120, opened 2017-10-28
smallest code possible

Sulekhiya:


SDavisMedia:
Thanks @Sulekhiya but I'm going to close this PR. There's an unnecessary tab on the line, which is not a big deal, but we don't mind the empty line. If we modify that file in the future, the tab will be removed.

PR #6130, opened 2017-11-01
Added United Kingdom States list

davidsherlock:
"States" in the context of the United Kingdom are technically 'counties". 

I have added all the "counties" including ISO codes.


cklosowski:
This is a blocker for #6133

pippinsplugins:
Merging into issue branch then updating since tags. Otherwise PR looks good here.

PR #6136, opened 2017-11-02
Make Download Log IP addresses clickable to ARIN website

davidsherlock:
As per the title, make the file download log IP addresses clickable through to the ARIN website.


includes/admin/reporting/class-file-downloads-logs-list-table.php
```diff
@@ -127,6 +127,8 @@ public function column_default( $item, $column_name ) {
 				return '<a href="' . add_query_arg( 'user', $item[ 'customer' ]->email ) . '">' . $item['customer']->name . '</a>';
 			case 'payment_id' :
 				return $item['payment_id'] !== false ? '<a href="' . admin_url( 'edit.php?post_type=download&page=edd-payment-history&view=view-order-details&id=' . $item['payment_id'] ) . '">' . edd_get_payment_number( $item['payment_id'] ) . '</a>' : '';
+			case 'ip' :
+				return '<a href="http://whois.arin.net/rest/ip/' . $item['ip']  . '" target="_blank">' . $item['ip']  . '</a>';
```
sunnyratilal: Not too keen on have `target="_blank"` as using it is a known security issue.

PR #6143, opened 2017-11-05
Simplify edd_get_price_name

chriscct7:
Made the function more safe by actually making sure $options has a price_id, not just an array of other details. Also removed the unneeded $name variable, and collapsed the logic of the if statement to a single line.

### DIFF
--- a/includes/cart/functions.php
+++ b/includes/cart/functions.php
@@ -205,14 +205,11 @@ function edd_get_cart_item_tax( $download_id = 0, $options = array(), $subtotal 
  */
 function edd_get_price_name( $download_id = 0, $options = array() ) {
 	$return = false;
-	if( edd_has_variable_prices( $download_id ) && ! empty( $options ) ) {
+	if ( edd_has_variable_prices( $download_id ) && ! empty( $options ) && ! empty( $options['price_id'] ) ) {
 		$prices = edd_get_variable_prices( $download_id );
-		$name   = false;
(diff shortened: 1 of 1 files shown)


cklosowski:
Been thinking about this lately. Should this be a method of EDD_Download? This helper can call the class, but this seems like something that belongs in the download object.

chriscct7:
No I don't think so, because a download object is an object of a complete download, not of a specific variation. What this function does doesn't make sense in the context of the download object. I think a helper function could be added called get_variation_name which takes in an index to return the name of.

However, loading an EDD Download object adds a ton of often unneeded DB calls. Every time we make some of these simpler functions do new EDD_Download and wrap them into that, we're causing functions to become a lot less performant than before, and particularly for simple, basic functions that will be used often in templating, that's not a really good practice.

includes/cart/functions.php
```diff
@@ -205,14 +205,11 @@ function edd_get_cart_item_tax( $download_id = 0, $options = array(), $subtotal
  */
 function edd_get_price_name( $download_id = 0, $options = array() ) {
 	$return = false;
-	if( edd_has_variable_prices( $download_id ) && ! empty( $options ) ) {
+	if ( edd_has_variable_prices( $download_id ) && ! empty( $options ) && ! empty( $options['price_id'] ) ) {
```
cklosowski: The price ID could be `0` here for old stores.

Issue #6124, opened 2017-10-31
Performance issues/duplicate queries


MythThrazz:
Hello,

We've been investigating the performance of our page lately because we've started to have slowdowns, and while checking the backend we've found out that the Software Licensing backend page is doing a lot of duplicate queries (literally thousands).

Before the changes the Licenses page was loaded in nearly 7s and required over 3,6k queries. After few small tweaks we managed to get down to around 2s and around 100queries. Mainly by removing unnecessary updates and selects, which were repeated hundreds of times.

The changes are really small in most cases it's just saving the values statically within the function/method instead of calling the database multiple times.  

**easy-digital-downloads\includes\payments\class-edd-payment.php**

```
	private function setup_user_id() {
		$user_id  = $this->get_meta( '_edd_payment_user_id', true );
        /*
         * Cminds change
         */
        $old_user_id = $user_id;
		$customer = new EDD_Customer( $this->customer_id );

		// Make sure it exists, and that it matches that of the associted customer record
		if( empty( $user_id ) || ( ! empty( $customer->user_id ) && (int) $user_id !== (int) $customer->user_id ) ) {

			$user_id = $customer->user_id;

            /*
            * Cminds change
            */
            if($old_user_id != $user_id){
			// Backfill the user ID, or reset it to be correct in the event of data corruption
                $this->update_meta( '_edd_payment_user_id', $user_id );
            }

		}

		return $user_id;
	}
```

**easy-digital-downloads\includes\class-edd-customer.php**

         /**
	 * Given the customer data, let's set the variables
	 *
	 * @since  2.3
	 * @param  object $customer The Customer Object
	 * @return bool             If the setup was successful or not
	 */
	private function setup_customer( $customer ) {

		if ( ! is_object( $customer ) ) {
			return false;
		}

		foreach ( $customer as $key => $value ) {

            /*
            * Cminds change - removed the unnecessary notes loading
            */
            $this->$key = $value;

		}

		$this->emails   = (array) $this->get_meta( 'additional_email', false );
		$this->emails[] = $this->email;

		// Customer ID and email are the only things that are necessary, make sure they exist
		if ( ! empty( $this->id ) && ! empty( $this->email ) ) {
			return true;
		}

		return false;

	}

	/**
	 * Get things going
	 *
	 * @since 2.3
	 */
	public function __construct( $_id_or_email = false, $by_user_id = false ) {
        /*
         * Cminds change - added the static saving of the instances
         */
[...]

PR #6147, opened 2017-11-05
Only load additional emails when requested

chriscct7:
Part of #6124

### DIFF
--- a/includes/class-edd-customer.php
+++ b/includes/class-edd-customer.php
@@ -714,7 +714,7 @@ class EDD_Customer {
             		return $this->data['emails'];
         	} else {
 			$this->data['emails']   = (array) $this->get_meta( 'additional_email', false );
-			$this->data['emails'][] = $this->email;
+			array_unshift( $this->data['emails'], $this->email );
 			return $this->data['emails'];
 		}
 	}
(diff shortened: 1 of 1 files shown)


includes/class-edd-customer.php
```diff
@@ -699,6 +703,22 @@ public function decrease_value( $value = 0.00 ) {
...
+	 *
+	 * @since  x.y.z
+	 * @return array           The emails of the customer.
+	 */
+	public function get_emails() {
+		if ( array_key_exists( 'emails', $this->data  ) ) {
+            		return $this->data['emails'];
+        	} else {
+			$this->data['emails']   = (array) $this->get_meta( 'additional_email', false );
+			$this->data['emails'][] = $this->email;
```
cklosowski: While we're in here, one thing that's always annoyed me is that the primary email is the `last` email in the array...can we make it so `$this->email` is the first value in the array of emails returned? It makes more sense for that to be first than last.

chriscct7: @cklosowski It's been updated to support this

cklosowski: 👍

chriscct7:
@cklosowski just an fyi the since docbloc wasn't updated on this one for the new function

PR #6148, opened 2017-11-05
Only backfill _edd_payment_user_id if its not the same as the one alr…

chriscct7:
…eady there

 Backfill _edd_payment_user_id if its not the same as the one already there and also corrects typo.

Part of #6124


cklosowski:
This is causing some issues with fatal errors, going to revert to test out.

PR #6173, opened 2017-11-12
issue/5087 - Add support to resume downloads of large files

### LINKED ISSUE #5087: Improve edd_readfile_chunked function to support resume
Suggested: https://secure.helpscout.net/conversation/268267024/0/

We should see if we can support resuming with the chunked readfile. Will need to verify this works with SL

sunnyratilal:
Resolves #5087

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -733,11 +733,18 @@ function edd_get_file_ctype( $extension ) {
  * See http://codeigniter.com/wiki/Download_helper_for_large_files/
  *
  * @access   public
- * @param    string  $file      The file
- * @param    boolean $retbytes  Return the bytes of file
+ *
+ * @param    string  $file     The file
+ * @param    boolean $retbytes Return the bytes of file
+ *
  * @return   bool|string        If string, $status || $cnt
  */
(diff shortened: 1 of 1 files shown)


cklosowski:
This all seems to work surprisingly well. To test it out I used `curl` with things like the following:

Getting the first 2KB
`curl --header "Range: bytes=0-2048" http://edd.local/index.php\?eddfile\=4808%3A4673%3A1%3A1\&ttl\=5120864898\&file\=1\&token\=2505a34acf3277f11de987fb6eee3800 -o part1`

Getting the second 2KB
`curl --header "Range: bytes=2049-4096" http://edd.local/index.php\?eddfile\=4808%3A4673%3A1%3A1\&ttl\=5120864898\&file\=1\&token\=2505a34acf3277f11de987fb6eee3800 -o part2`

This produced two files, both not equal in contents, but the expected sizes. I also tested downloading without a header and received the full file with no issues.
`curl http://edd.local/index.php\?eddfile\=4808%3A4673%3A1%3A1\&ttl\=5120864898\&file\=1\&token\=2505a34acf3277f11de987fb6eee3800 -o file.pdf`

PR #6186, opened 2017-11-16
Issue/6185 - Defensive coding for edd_add_past_purchases_to_new_user 

### LINKED ISSUE #6185: Defensive coding for edd_add_past_purchases_to_new_user
We've had an [issue reported by a customer](edd_add_past_purchases_to_new_user) which is causing errors when EDD is attaching past purchases to users.

I have spent a fair amount of time trying to replicate but am unable to do so.

Each time I've tried, EDD has successfully linked past purchases to the new user.

It's actually doing it in `edd_get_payments()` which is a bit odd but nevertheless, in terms of the scope of this issue, we should wrap the code to check the payment objects are indeed instances of `EDD_Payment`.

sunnyratilal:
Resolves #6185

### DIFF
--- a/includes/user-functions.php
+++ b/includes/user-functions.php
@@ -483,7 +483,7 @@ function edd_add_past_purchases_to_new_user( $user_id ) {
 		edd_send_user_verification_email( $user_id );
 
 		foreach( $payments as $payment ) {
-			if ( $payment instanceof EDD_Payment ) {
+			if ( is_object( $payment ) && $payment instanceof EDD_Payment ) {
 				if ( intval( edd_get_payment_user_id( $payment->ID ) ) > 0 ) {
 					continue; // This payment already associated with an account
 				}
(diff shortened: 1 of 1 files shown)


includes/user-functions.php
```diff
@@ -483,12 +483,14 @@ function edd_add_past_purchases_to_new_user( $user_id ) {
 		edd_send_user_verification_email( $user_id );
 
 		foreach( $payments as $payment ) {
-			if( intval( edd_get_payment_user_id( $payment->ID ) ) > 0 ) {
-				continue; // This payment already associated with an account
-			}
+			if ( is_object( $payment ) && $payment instanceof EDD_Payment ) {
+				if ( intval( edd_get_payment_user_id( $payment->ID ) ) > 0 ) {
```
cklosowski: Since we have a known `EDD_Payment` object, can we just use `$payment->user_id`? `edd_get_payment_user_id()` does this, which means we're kicking up two instances of the same payment.

pippinsplugins: Yeah we should use `$payment->user_id`.

cklosowski:
Good now @pippinsplugins ?

PR #6188, opened 2017-11-16
Add a filter for the web accept variables #6184

### LINKED ISSUE #6184: Allow filtering of the PayPal IPN data
A point was raised in the case that we should allow filtering the data sent in from the PayPal Standard web-accept IPN call.

Location:
/includes/gateways/paypal-standard.php:460

Prior to the first `if` statement add:
`$data = apply_filters( 'edd_paypal_web_accept_and_cart_data', $data ); `

cklosowski:
resolves #6184

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -457,6 +457,19 @@ add_action( 'edd_verify_paypal_ipn', 'edd_process_paypal_ipn' );
  * @return void
  */
 function edd_process_paypal_web_accept_and_cart( $data, $payment_id ) {
+
+	/**
+	 * PayPal Web Accept Data
+	 *
+	 * Allows filtering the Web Accept data that PayPal passes back in via IPN with PayPal Standard
+	 *
+	 * @since 3.8.13
+	 *
+	 * @param array $data      The PayPal Web Accept Data
+	 * @param int  $payment_id The Payment ID associated with this IPN request
+	 */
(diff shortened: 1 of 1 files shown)


includes/gateways/paypal-standard.php
```diff
@@ -457,6 +457,19 @@ function edd_process_paypal_ipn() {
  * @return void
  */
 function edd_process_paypal_web_accept_and_cart( $data, $payment_id ) {
+
+	/**
+	 * PayPal Web Accept Data
+	 *
+	 * Allows filtering the Web Accept data that PayPal passes back in via IPN with PayPal Standard
+	 *
+	 * @since 3.8.13
```
pippinsplugins: This should be `2.8.13`.

includes/gateways/paypal-standard.php
```diff
@@ -457,6 +457,19 @@ function edd_process_paypal_ipn() {
...
+	 *
+	 * @param array $data      The PayPal Web Accept Data
+	 * @param int  $payment_id The Payment ID associated with this IPN request
+	 */
+	$data = apply_filters( 'edd_paypal_web_accept_and_cart_data', $data, $payment_id );
```
pippinsplugins: Could we change this so that it's a filter available to all IPN types instead of just web_accept and cart_data?

cklosowski: @pippinsplugins yeah we can add more than this one filter. I'll add a few more for the paypal standard IPN responses.

Issue #6197, opened 2017-11-19
PayPal Standard: add support for Payment Data Transport (PDT)


pippinsplugins:
Payment Data Transport is an option PayPal supports in Standard that sends data about the purchase back to the website via the return URL. The data can then be used to look up the payment via the API to confirm it was completed successfully.

This provides an alternative to the PayPal IPN and, for accounts that have it enabled, should make the completion process more reliable.

pippinsplugins:
See https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/paymentdatatransfer/

pippinsplugins:
@easydigitaldownloads/core-team this is ready for testing!

1. Please read https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/paymentdatatransfer/ for an overview of what PDT is and how to enable it in your account.
2. Enter your identity token in Downloads > Settings > Payment Gateways > PayPal Standard
3. Complete a purchase using PayPal Standard
4. Immediately upon return from PayPal, the payment record should be marked as completed and the transaction ID stored. No IPN processing required, so this works on localhosts too.

We'll obviously update our documentation to explain how to enable PDT for customers before this goes live.

pippinsplugins:
I've sent this to two customers that are experiencing problems with the PayPal IPN to confirm it works for them.

pippinsplugins:
One customer has confirmed this works for them. It was a customer that couldn't get IPN to work.

PR #6198, opened 2017-11-19
Add support for PayPal PDT. #6197

pippinsplugins:
#6197

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -847,7 +847,10 @@ function edd_paypal_process_pdt_on_return() {
 			)
 		);
 
-		edd_debug_log( 'Attempting to verify PayPal payment with PDT. Args: ' . print_r( $remote_post_vars, true ) );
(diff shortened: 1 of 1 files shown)


includes/gateways/paypal-standard.php
```diff
@@ -780,6 +793,87 @@ function edd_paypal_success_page_content( $content ) {
...
+			),
+			'sslverify'   => false,
+			'body'        => array(
+				'tx'  => sanitize_text_field( $_GET['tx'] ),
+				'at'  => $token,
+				'cmd' => '_notify-synch',
+			)
+		);
+
+		edd_debug_log( 'Attempting to verify PayPal payment with PDT. Args: ' . print_r( $remote_post_vars, true ) );
```
mindctrl: I don't think you'll want to log the token in an accessible log file.

cklosowski:
@mindctrl fixed it so the token get's sanitized to only show the last 6 characters, to help with debugging.

PR #6201, opened 2017-11-22
Add text to clarify when discount codes expire

tnorthcutt:
Without this text, the time a discount code expires is rather ambiguous.
If you choose a date with the datepicker, you don't see until coming
back to edit the discount code that in fact a date and time (the last second of the
day) have been saved. This new text explicitly states that the discount
code will expire at the end of the day chosen.

### DIFF
--- a/includes/admin/discounts/add-discount.php
+++ b/includes/admin/discounts/add-discount.php
@@ -131,7 +131,7 @@ if ( ! defined( 'ABSPATH' ) ) exit;
 				</th>
 				<td>
 					<input name="expiration" id="edd-expiration" type="text" class="edd_datepicker"/>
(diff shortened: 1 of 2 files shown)


mindctrl:
What happens if someone manually supplies a different end time? It is used or does it get overwritten to be end of day?

tnorthcutt:
Sounds like it’s possible to set a different time, but it’s finicky:

a4097255-6903-449c-92cd-61c00cda6232

JJJ:
Supporting times here is not difficult, if we want to.

I'm not opposed to adding more descriptive text here, otherwise.

pippinsplugins:
Let's add support for time.

sunnyratilal:
Closing as EDD 3.0 allows for start and expiration times to be set for discount codes.

Issue #6203, opened 2017-11-24
Add a helper to retrieve and filter discount status labels


DrewAPicture:
In working to implement discounts in SellBird, we discovered that the status labels for discounts are essentially hard-coded into the list table class. I hesitate to _also_ hard-code derivation of the status label as it could easily fall out of date with EDD core should other statuses come in later, or some add-on needs to change something in the future.

The attached PR introduces `edd_get_discount_status_label()` which wraps `EDD_Discount::get_status_label()`. The `get_status_label()` method introduces a `edd_get_discount_status_label` filter.

Also includes four tests for `edd_get_discount_status_label()`.

DrewAPicture:
PR: #6204

pippinsplugins:
PR looks good to me. @sunnyratilal @cklosowski let's get the since tags updated and merged in!

sunnyratilal:
Merged

PR #6204, opened 2017-11-24
Issue/6203 – Discount status label helper

DrewAPicture:
Issue: #6203


DrewAPicture:
@sunnyratilal Note that I left all the `@since` tags as version x.x because I didn't know when it might be merged, so you'll want to update those to whatever release you're targeting after merge (or I can do it) ;)

PR #6221, opened 2017-12-01
Adding payments to before/after purchase history action

brianhogg:
Avoids a second query when using the action, where access to the payments would normally be needed


pippinsplugins:
Looks good, thanks @brianhogg!

Issue #6182, opened 2017-11-13
Email taken error when adding to customer includes emails for users who are not customers


brashrebel:
When editing a customer and adding more email addresses to their profile, if you attempt to add one which is taken by a _user_ who is not a customer you will still get the message:

> Email address is already associated with another customer.

This is confusing because if you then go looking for this "customer" you won't find them since they are not a customer.

The solution could be as simple as changing the text to read "user" instead but that probably won't do as it leaves the user with no suggested action. The message could be modified to suggest editing the customer and attaching them to the user if applicable.

It would be nice to make this message more helpful all around. We're saying that the email is taken, why don't we provide a link directly to the account it is taken by? If the email is taken by a customer, link to that customer profile. If it is taken by a user who is not a customer, link to the user and suggest linking to the customer.

pippinsplugins:
I like the idea of making the message more descriptive and also linking to the found user. Perhaps we could even show a message to "link anyway?" that allows an admin to force the connection to be made.

SimonChaw:
Hi there, I took a crack at this.

So for this example we have the following user:
<img width="874" alt="users" src="https://user-images.githubusercontent.com/7470359/33520089-7bd569cc-d78a-11e7-9ed1-b28f69f78f68.png">

And the two following customers:
<img width="897" alt="customers" src="https://user-images.githubusercontent.com/7470359/33520090-86956cea-d78a-11e7-918e-0c3f6f6eadd9.png">

In this example John Doe is a user and a customer and the admin wants to add his user email to his customer object.
<img width="1078" alt="add duplicate user email pre" src="https://user-images.githubusercontent.com/7470359/33520092-8e569076-d78a-11e7-8c2f-c806f4de5e87.png">

The admin is informed that the email already belongs to a user, and is given the option to force adding the email:
<img width="891" alt="add duplicate user email post" src="https://user-images.githubusercontent.com/7470359/33520094-9ac3475a-d78a-11e7-81de-bc32073921f9.png">

This will work just fine, and so will adding an existing customer email:
<img width="857" alt="add duplicate customer email pre" src="https://user-images.githubusercontent.com/7470359/33520097-ab389e5a-d78a-11e7-96cc-d26780460ce7.png">

But you will not be able to set that duplicate customer email as the primary email because of the  UNIQUE email constraint on the wp_edd_customers tab
[...]

PR #6222, opened 2017-12-03
Issue/6182

SimonChaw:
#6182 

Admins now have the option to view the customer/user that are tied to the email they are trying to add to a customer. If the email belongs to a user they can force association, but not if it belongs to another customer.


includes/scripts.php
```diff
@@ -171,7 +171,7 @@ function edd_load_admin_scripts( $hook ) {
 	$css_dir = EDD_PLUGIN_URL . 'assets/css/';
 
 	// Use minified libraries if SCRIPT_DEBUG is turned off
-	$suffix  = ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ? '' : '.min';
+	$suffix  = ''; //( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ? '' : '.min';
```
cklosowski: This needs to be reverted. It will currently make it so no minified scripts load.

assets/js/admin-scripts.js
```diff
@@ -1853,19 +1853,33 @@ jQuery(document).ready(function ($) {
...
+				if ( e.target.id === 'add-customer-email-force' ){
+					forceLink = true;
+				}
 
 				var postData = {
 					edd_action:  'customer-add-email',
 					customer_id: customer_id,
 					email:       email,
 					primary:     primary,
+					forceLink:   forceLink,
```
cklosowski: Our preferred variable format is snake case, not camel case. Could we change that to `force_link`, so that we don't have to guess what version we are using in cases like this where we now have 2 different cases of how `forceLink and force_link` mean the same thing.
`$force_link  = 'true' === $args['forceLink'] ? true : false;`

includes/admin/customers/customer-actions.php
```diff
@@ -231,10 +232,25 @@ function edd_add_customer_email( $args ) {
...
+				$user_type = 'customer';
+				$user = new EDD_Customer( $email );
+
+				if( empty($user->email) ){
+
+					$user = get_user_by( 'email', $email );
+					$user_type = 'user';
+				}
+
+				$link = (($user_type === 'customer') ?
```
cklosowski: The formatting on this seems off... `$link = ( (condition) ? case 1 : case 2 )`

This should really just be `$link = condition ? case 1 : case 2`

includes/admin/customers/customer-actions.php
```diff
@@ -231,10 +232,25 @@ function edd_add_customer_email( $args ) {
...
+				admin_url( 'edit.php?post_type=download&page=edd-customers&view=overview&id=' . $user->id ) :
+				admin_url( 'user-edit.php?user_id=' . $user->id ));
 
 				$output = array(
 					'success' => false,
-					'message' => __( 'Email address is already associated with another customer.', 'easy-digital-downloads' ),
+					'message' => __( 'Email address is already associated with another', 'easy-digital-downloads' ),
```
cklosowski: `message` here has been modified but is now incomplete:
`Email address is already associated with another` is not a complete sentence.

includes/class-edd-customer.php
```diff
@@ -303,9 +303,10 @@ public function update( $data = array() ) {
 	 * @since  2.6
 	 * @param  string $email The email address to remove from the customer
 	 * @param  bool   $primary Allows setting the email added as the primary
+	 * @param  bool		$force_link Allows the admin to force an email to be linked to a customer even if it is attached to another user or customer
```
cklosowski: Alignment issue needs to be corrected.

includes/class-edd-customer.php
```diff
@@ -396,8 +397,8 @@ public function set_primary_email( $new_primary_email = '' ) {
 		// Remove new primary from list of additional emails
 		$remove = $this->remove_email( $new_primary_email );
 
-		// Add old email to additional emails list
-		$add = $this->add_email( $old_email );
+		// Add old email to additional emails list and if the email was forcibly put there, make sure it is re-added
+		$add = $this->add_email( $old_email, false, ( email_exists( $old_email ) ) );
```
cklosowski: The `email_exists( $old_email )` check needs to be pulled out of the `add_email` method in order to main PHP 5.2 compatibility. Otherwise we'll get the fatal error of using a function in a write context.

includes/scripts.php
SimonChaw: I could have sworn I changed this back, sorry.

SimonChaw:
Hi @cklosowski! I've made the requested changes. Thank you for your feedback, here's what I came up with for your UI request:
screen shot 2017-12-03 at 5 28 23 pm

The user had two clear options now, if they hit cancel it will remove the notice and clear the email text box. Also I figure since it's not _really_ an error why not use the warning notice?

And this is what it will look like when you run into an existing customer email:
screen shot 2017-12-03 at 5 28 39 pm

cklosowski:
@SimonChaw oh yeah I like that look much better, nice work. Did something else change? It looks like the full error message is now a link? I believe we should still have a link to 'View user/customer' when appropriate, instead of making the entire message a link.

SimonChaw:
Well I saw that it seems like the best practice was to not include HTML when preparing to respond to an AJAX call:
`$output = array(
					'success' => false,
					'message' => sprintf( __( 'Email address is already associated with another %s', 'easy-digital-downloads' ), $user_type ),
					'link'    => $link,
					'user_type'=> $user_type,
				);`

So before this I was passing back an incomplete message and then from the client side adding the user_type to the message and wrapping it in an <a> tag. That way only the user type was clickable.

SimonChaw:
@cklosowski So I can either wrap the user type in an `<a>` tag server side or do it client side. Which would be preferred?

cklosowski:
@SimonChaw Adding marking in minor context like links is allowed. There is a further problem though in that you are now not allowing `user` and `customer` to be translated. Since those strings are put in 'raw' without any translation string, it's possible for translators to translate the entire string except for the `$user_type`, since that isn't wrapped in context. You could put a `_x` around it to give them context, but then you run into a problem with the line that determines the link to provide, since if `customer or user` is translated, the `$link = 'customer' === $user_type` will fail since `customer` could be translated.

I think something like the following would be more beneficial to the translators
```
// Find out if this email is taken by a user or a customer and then include a link to the correct admin view.
$customer    = new EDD_Customer( $email );
$is_customer = true;

if( empty( $customer->email ) ){
	$user        = get_user_by( 'email', $email );
	$is_customer = false;
}

if ( $is_customer ) {
	$link          = admin_url( 'edit.php?post_type=download&page=edd-customers&view=overview&id=' . $customer->id );
	$message = sprintf( __( 'Email address is already associated with another <a href="%s">customer</a>', 'easy-digital-downloads' ), $link );
} else {
	$link          = admin_url( 'user-edit.php?user_id=' . $user->id );
	$message = sprintf( __( 'Email address is already associated with another <a href="%s">user</a>', 'easy-digital-downloads' ), $link );
}

$output = array(
	'success'  => false,
	'message' => $message,
);


```

SimonChaw:
@cklosowski Oh okay, I didn't even factor in translation. It's a foreign concept to me. 

No worries on the delay! I'll make these changes.

cklosowski:
Looks better @SimonChaw I'll test this out tomorrow.

One thing I noticed is that you've got changes to both the includes/scripts.php and the minified JS file. These appear to be file permission changes, and no actual changes. If you could un-stage those changes that'd be preferred.

In the future when you are working on a project doing the following command when in the root of a git project can help prevent those file permissions from getting committed:
`> git config core.fileMode false`

SimonChaw:
@cklosowski Sorry about that. Also I had to commit twice, git wouldn't pick up the file permission change I made to includes/scripts.php.

I will keep that command in mind, thank you!

cklosowski:
Perfect @SimonChaw, sorry for the late change, do you think it's possible for you to update your branch with the latest that is in our `release/2.9` branch?

The issue this is associated with is set for our 2.9 release, not the next major, so I had to update your PR to base against `release/2.9`.

If you aren't sure how to do that let me know and we'll find a way to walk/talk through it if need be.

SimonChaw:
@cklosowski If you could run me through it that would be great 😄

cklosowski:
@SimonChaw No problemo!

First, if you haven't setup the main EDD repository as an `upstream` remote, follow this guide:
https://help.github.com/articles/configuring-a-remote-for-a-fork/

Once you do you should be able to do the following:

Checkout your issue branch from your fork
`> git checkout issue/6182'

Then `fetch` the branches in `upstream` (or whatever you name the primary EDD repo from the steps in the link above:
`> git fetch upstream`

A bunch of stuff will probably shoot out, but in the end all you care about is getting the `release/2.9` branch merged:
`> git pull upstream release/2.9`

Then once the merge is happy, you can simply do a `> git push origin issue/6182` and it'll push it to this PR.

If we're lucky, there will be no conflicts. Give that a try, but if you have problems or it just isn't working right, what I can do is actually `fetch` your pull request and then make any necessary changes, but by doing so, it will retain your commit history so you keep the credit for the work you did ;).

Also, Thanks for your first PR contribution!
tada

cklosowski:
As a side note, by setting up the `upstream` you can then keep your local development version of EDD updated easier by just doing this:
```
> git checkout master (which will be your fork's master branch)
> git pull upstream master
> git push origin master
```

This will basically bring your fork's `master` branch up to speed with EDD's `master` branch. Then before you make new issue branches, you can just make sure you do that same step above. I'm going to work on getting our [CONTRIBUTING.md](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/CONTRIBUTING.md) updated with some of this into to make it easier in the future.

SimonChaw:
@cklosowski Thank you for the help! I'll do that now.

Just curious, shouldn't this be `master` to `release/2.9`?
screen shot 2017-12-13 at 3 46 01 pm

cklosowski:
Yes, sorry...i failed that one by mistyping the search for the re-base. Looks like you don't need anything else here. My apologies :)

Will test this out and if all is well, merge in for 2.9

SimonChaw:
Awesome! @cklosowski

SimonChaw:
@cklosowski Could you shoot me an e-mail when you get a chance? I have a couple questions. 

chawblah@gmail.com

cklosowski:
We're super close here with the work @SimonChaw has done. i'm going to close this PR and make an issue branch from it which will keep your commit history in tact @SimonChaw.

From there I'll work out the last few things and get a new PR opened.

PR #6231, opened 2017-12-13
Change home_url() to site_url() for WPML compatibility

dgwatkins:
The function edd_is_local_file() uses home_url() to check if a file is local.

Most of the times home_url() and site_url() are the same, but when they are different the correct function is site_url(). site_url() returns the URL for WordPress core files (like for example the uploads folder):
https://codex.wordpress.org/Function_Reference/site_url

This little difference makes WPML fail this check because home_url() has the language code attached while site_url() doesn't have it.

The problem was originally reported here:
https://wpml.org/forums/topic/edd-frontend-submission-form-translation-wrong-behaviour/

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -366,10 +366,10 @@ function edd_deliver_download( $file = '', $redirect = false ) {
  * @return bool                   If the file is hosted locally or not
  */
 function edd_is_local_file( $requested_file ) {
-	$home_url       = preg_replace('#^https?://#', '', home_url() );
+	$site_url       = preg_replace('#^https?://#', '', site_url() );
 	$requested_file = preg_replace('#^(https?|file)://#', '', $requested_file );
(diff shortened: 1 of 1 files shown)


JJJ:
Confirmed, and the PR looks good to me. Thanks @dgwatkins! This is safe to merge IMO.

This is a frequent and old enough problem that it's likely to keep coming back even after we fix this here, but do we know if there are any other places in EDD that could use a similar switch?

dgwatkins:
Not that I know.

But I will raise here any other issues that our customers report.

JJJ:
> But I will raise here any other issues that our customers report.

Thanks! We appreciate it greatly.

I'll make sure this gets merged in.

JJJ:
Updated the destination branch to `release/3.0` and will review & merge momentarily.

Issue #6253, opened 2017-12-26
Introduce a base Registry class


DrewAPicture:
Tracking issue: #6236

In order to facilitate registering reports tabs and their tiles, we'll need a base Registry class to register against in memory.

DrewAPicture:
PR: #6256 

Just need tests.

DrewAPicture:
Added 24 tests (full coverage) for `EDD_Registry`. @cklosowski I'm probably just gonna do `EDD_` prefixing on class names for now until you decide on what (if any) kind of namespacing scheme you'd like to use.

DrewAPicture:
Merged #6256.

PR #6256, opened 2017-12-27
issue/6253 – Registry class.

DrewAPicture:
Issue: #6253

### DIFF
--- a/includes/class-edd-exception.php
+++ b/includes/class-edd-exception.php
@@ -29,33 +29,26 @@ class EDD_Exception extends \Exception {
 	 *                      alongside the exception message.
 	 */
 	public function log( $extra = '' ) {
-		/**
-		 * Filters whether to skip logging EDD exceptions to the debug log.
-		 *
-		 * @since 3.0
-		 *
-		 * @param bool $skip Whether to skip logging exceptions. Default false (proceed).
-		 */
-		if ( false === apply_filters( 'edd_skip_logging_exceptions', false ) ) {
+		if ( $this->getCode() ) {
 
-			if ( $this->getCode() ) {
+			$message = sprintf( 'Exception Code: %1$s – Message: %2$s',
+				$this->getCode(),
+				$this->getMessage()
+			);
 
-				edd_debug_log( sprintf( 'Exception Code: %1$s – Message: %2$s %3$s',
-					$this->getCode(),
-					$this->getMessage(),
-					$extra
-				) );
+		} else {
 
-			} else {
-
-				edd_debug_log( sprintf( 'Exception Message: %1$s %2$s',
-					$this->getMessage(),
-					$extra
-				) );
-
-			}
+			$message = sprintf( 'Exception Message: %1$s',
+				$this->getMessage()
+			);
 
 		}
+
+		if ( ! empty( $extra ) ) {
+			$message = sprintf( "{$message} %s", $extra );
+		}
+
+		edd_debug_log( $message );
 	}
 
 	/**
--- a/includes/class-edd-exception.php
+++ b/includes/class-edd-exception.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * Exception object for EDD
+ *
+ * This class handles connecting exception handling to the debug log.
+ *
+ * @package     EDD
+ * @subpackage  Classes/Utilities
+ * @copyright   Copyright (c) 2018, Pippin Williamson
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       3.0
+ */
+
+/**
+ * Implements some EDD-specific logic for exception handling.
+ *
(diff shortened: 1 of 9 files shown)


DrewAPicture:
@mindctrl Some notes:

* An earlier version of this class is already in use in AffiliateWP core. [Source](https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/abstracts/class-affwp-registry.php).
* The main difference between this one and the AffiliateWP version is that some returns have been adjusted to leverage `WP_Error` and this one also extends `ArrayObject`.
* If you want to confirm the tests pass, you'll need to run them yourself as this branch isn't included in the Travis config

DrewAPicture:
OK @mindctrl, I've pushed up the changes we discussed yesterday.

mindctrl:
@DrewAPicture changes look good. Will give this a proper look and test tomorrow.

tests/tests-registry.php
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	/**
+	 * @covers \EDD_Registry::add_item()
+	 * @expectedException EDD_Exception
+	 */
+	public function test_add_item_with_empty_attributes_should_return_false() {
+		$this->assertFalse( $this->mockRegistry->add_item( 'foo', array() ) );
+	}
+
```
mindctrl: @DrewAPicture can you explain this test? I'm curious why we test a `false` return here? `add_item()` returns an exception if passed empty attributes, and this test passes due to the `@expectedException` tag, which is tested in `test_add_item_with_empty_attributes_should_throw_exception()` below.

includes/class-edd-exception.php
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	public function log( $extra = '' ) {
+		/**
+		 * Filters whether to skip logging EDD exceptions to the debug log.
+		 *
+		 * @since 3.0
+		 *
+		 * @param bool $skip Whether to skip logging exceptions. Default false (proceed).
+		 */
+		if ( false === apply_filters( 'edd_skip_logging_exceptions', false ) ) {
```
mindctrl: @DrewAPicture I'm wondering if this filter could cause confusion or problems. Example: say I'm using/extending this class and I always want to log exceptions. Other code could break my exception logging, and anyone else's exception logging. Errors could possibly disappear into a black hole. Is there a solid use case for disabling exception logging when all other debug logging is enabled?

tests/tests-registry.php
DrewAPicture: The exception doesn't get _returned_ it gets thrown. One of the primary benefits of throwing exceptions vs returning `WP_Error` objects is that it allows you to return consistent types while still flagging errors. In this case, `add_item()` consistently returns `true` or `false`.

includes/class-edd-exception.php
DrewAPicture: By allowing exceptions to be logged to the debug log, there's potential to _quickly_ fill up logs with said exceptions if there's a problem. I think in any situation where many pieces of code can have their hands in the same pot there are going to be cases where one preempts the other and that's a reality we have to be OK with, especially in something as yuge as EDD.

tests/tests-registry.php
mindctrl: I understand `return` vs `throw`, but what I'm not getting is why we test for a `false` return when passing empty attributes? An exception is thrown when attributes is empty, it doesn't return false. Without the `@expectedException ` tag, the assertion fails. But if we're testing for an exception by using the phpdoc tag, how is it different than the test below that tests the same exception with `$this->setExpectedException()`?

I guess I'm asking, what's the real difference between these two tests?

includes/class-edd-exception.php
mindctrl: Large log files could definitely happen, but EDD debug mode is meant for debugging, not full-time production use. If support asks customers to enable debug logging and exceptions aren't logged, we're missing some important info that quickly solve problems.

tests/tests-registry.php
DrewAPicture: I understand your confusion.

It's about _literally_ testing the expected _default_ in lieu of any other known or unknown factors. In the case of `add_item()` the default is `false` and we're testing that the default is false. Code would never be literally checking for a `false` value against `add_item()` because of the need for a try/catch, it's merely testing that the code works as written in the spirit of completeness. We don't always test code 100 percent for real world use, some of it is just to test that the code is doing what it's supposed to be doing.

includes/class-edd-exception.php
DrewAPicture: The filter exists only for flexibility. which EDD _inherently_ needs in order to exist in the ecosystem it does. I'd be fine with removing it.

tests/tests-registry.php
mindctrl: Ahh, I see. Thanks for explaining!

Issue #6236, opened 2017-12-18
Build out API for creating report sections, tiles, and trend graphs


pippinsplugins:
As part of the 3.0 development push, we're going to create a new API that allows core and extensions to easily create new report options.

Tracking PR | Project |
---------------- | -----------
| ~#6273~ | [3DD™: Reporting API Reboot](https://github.com/easydigitaldownloads/easy-digital-downloads/projects/2)

This API will comprise three primary pieces:

**Infrastructure**
* [x] #6253 – API to register (any) stuff
* [x] #6253 – API to handle exceptions
* [x] #6260 – Concrete class to initialize the Reports API

**Report (tabs)**
* [x] #6254 – API to register report tabs (PR: #6263)
* [x] #6284 – `Report` object (PR: #6289)
* [x] #6284  – API to build and render reports (PRs: #6289, #6329)
* [x] #6334 – Integrate the Carbon library to simplify date filtering (PR: #6353)
* [x] #6257 – API to register filters against report tabs (PR: #6361)
* [x] #6360 – Deprecate old report view/tab display actions using a callback (PR: #6371)
* [x] #6212 – Audit test coverage for Reports API Reboot pre-merge (PR: #6520)

**Endpoints**
* [x] #6280 – API for Endpoints (PR: #6283)
* [x] #6285 – Add tile support to the Endpoints API (PR: #6313)
* [x] #6287 – Add chart support to the Endpoints API
* [x] #6286 – Add table support to the Endpoints API (PR: #6339)
* [x] #6511 – Add pie chart support to the chart endpoint (PR: #6515)

**Nice to Have**
* [ ] #6281 – Expose endpoints via REST

The reports screen from AffiliateWP represents a fairly good view of what this API will ultimately provide:

![AffiliateWP Reports](https://user-images.githubusercontent.com/1034109/34129740-d4ea758e-e40a-11e7-8982-b3fc4855b764.png)

DrewAPicture:
PR submitted to @mindctrl for review on #6253.

DrewAPicture:
The empty `EDD_Reports` class has been merged to `issue/6236` via #6260. And so it goes.

DrewAPicture:
It's worth noting that while we're in the process of moving the Logs tab from Reports to Tools, this tab largely comprises graphs that will (likely) be directly integrated into new Reports views, so may actually be able to just completely deprecate the Logs tab altogether at the end of this thing.

[...]

PR #6273, opened 2018-01-01
Issue/6236 – Reports API Reboot

DrewAPicture:
#6236 is the main tracking issue for this aspect of 3DD.

Creating this PR so that we can get a 10,000 ft view of the overall progress the Reports API reboot.


tests/reports/data/tests-endpoint-registry.php
```diff
@@ -0,0 +1,433 @@
...
+		$this->registry->exchangeArray( array() );
+
+		parent::tearDown();
+	}
+
+	/**
+	 * @covers \EDD\Reports\Data\Endpoint_Registry::$item_error_label
+	 */
+	public function test_item_error_label_should_be_reports_endpoint() {
+		$this->assertSame( 'reports endpoint', $this->registry::$item_error_label );
```
mindctrl: `$this->registry::$item_error_label` breaks the unit tests, as it's only available on PHP 7+.

includes/reports/reports-functions.php
```diff
@@ -0,0 +1,453 @@
...
+ * @param \EDD\Reports\Data\Report $report Report object.
+ */
+function default_display_tables_group( $report ) {
+	if ( $report->has_endpoints( 'tables' ) ) :
+		$tables = $report->get_endpoints( 'tables' );
+		?>
+		<div id="edd-reports-tables-wrap">
+
+			<?php foreach ( $tables as $endpoint_id => $table ) : ?>
+				<h3><?php echo esc_html( $table->get_label() ); ?></h3>
```
mindctrl: You probably know this already, but `$table` is an array here so this fatals.

DrewAPicture: `get_endpoints()` returns an array of keyed objects. If `$table` is an array here, then something else is wrong. I think I may have addressed the problem in 8f19622607156b2b83e5a67d3931585856418942.

tests/reports/data/tests-endpoint-registry.php
DrewAPicture: Out of curiosity, what's the error message? I don't see anything about this that would be PHP 7 specific.

includes/reports/reports-functions.php
mindctrl: Looks like that commit fixed it.

tests/reports/data/tests-endpoint-registry.php
mindctrl:
```
Parse error: syntax error, unexpected '::' (T_PAAMAYIM_NEKUDOTAYIM) in /home/travis/build/easydigitaldownloads/easy-digital-downloads/tests/reports/data/tests-endpoint-registry.php on line 66
```

includes/reports/data/class-endpoint-registry.php
```diff
@@ -0,0 +1,334 @@
...
+		);
+
+		$attributes = array_merge( $defaults, $attributes );
+
+		$attributes['id']    = $endpoint_id;
+		$attributes['views'] = Reports\parse_endpoint_views( $attributes['views'] );
+
+		// Bail if this endpoint ID is already registered.
+		if ( $this->offsetExists( $endpoint_id ) ) {
+			$message = sprintf( 'The \'%1$s\' endpoint already exists and cannont be registered.', $endpoint_id );
```
mindctrl: Misspelled `cannot` here.

includes/reports/class-registry.php
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * Retrieves all registered items with a given sorting scheme.
+	 *
+	 * @since 3.0
+	 *
+	 * @param string $sort Optional. How to sort the list of registered items before retrieval.
+	 *                     Accepts 'priority' or 'ID' (alphabetized by item ID), or empty (none).
+	 *                     Default empty.
+	 */
```
sunnyratilal: Missing `@return` tag in PHPDoc.

includes/reports/data/charts/v2/class-dataset.php
```diff
@@ -0,0 +1,318 @@
+<?php
+/**
+ * Reports API - Dataset class
+ *
+ * @package     EDD
+ * @subpackage  Reports\Data\Charts
+ * @copyright   Copyright (c) 2018, Pippin Williamson
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       3.0
+ */namespace EDD\Reports\Data\Charts\v2;
```
sunnyratilal: Could this `namespace` be moved on to a separate line please?

sunnyratilal:
@DrewAPicture I've looked over the codebase extensively.  I can't see anything that would constitute as _incorrect_ with any of the methods, or classes.  The unit tests definitely vouch for that.

I've left some class formatting comments but otherwise, this is good to merge. 👍 

Again, truly amazing work done here. Kudos. 💯

Issue #6278, opened 2018-01-04
Introduce global notes table


sunnyratilal:
Introduce a global notes table which will handle payment and discount notes.

Two new tables will be introduced: `wp_edd_notes` and `wp_edd_notemeta`.

pippinsplugins:
What's the use case for note meta?

pippinsplugins:
The notes table should also be used for customer and subscription notes, though those can be migrated later on.

sunnyratilal:
@JJJ said that we should have note meta for future third party opportunities.

pippinsplugins:
I can think of a few potential use cases so works for me!

pippinsplugins:
@sunnyratilal let's get payment notes and discount notes connected to this. Once that's done the upgrade routine can be worked on.

We should also then consider whether it's worthwhile putting customer notes in during 3.0. Unsure yet.

sunnyratilal:
@pippinsplugins Not quite sure I follow when you say "connected".  It would be better to link up discounts in #5794 once #5277 and this is merged to allow us to work with the updated version of `EDD_Discount`.

I can go ahead and rewrite the implementation for `edd_*_payment_note` to fetch from the custom table and set up the BC layers if that's what you mean?

sunnyratilal:
And also add helper methods to get/insert/delete discount notes

pippinsplugins:
Ah yes, doing it in the other issues is better, sorry :)

Let's do payments once that migration is done too.

sunnyratilal:
Once #6277 is merged, we need to update the `EDD_Note` object as the column has been renamed from `note` to `content`, which makes more sense as a column name.

JJJ:
Notes tables are in, and Customers/Discounts/Payments are setup to use them now.

Let's close this issue, and open new ones for any regressions.

If something is majorly fubar, pleas reopen and blame @sunnyratilal ❤️

PR #6279, opened 2018-01-04
issue/6278 - Introduce global notes table

sunnyratilal:
#6278


includes/payments/functions.php
```diff
@@ -1511,17 +1510,24 @@ function edd_get_payment_notes( $payment_id = 0, $search = '' ) {
...
+	}
 
 	do_action( 'edd_pre_insert_payment_note', $payment_id, $note );
 
-	$note_id = wp_insert_comment( wp_filter_comment( array(
+	/**
+	 * For backwards compatibility purposes, we need to pass the data to wp_filter_comment in the event that the note
+	 * data is filtered using the WordPress Core filters prior to be inserted into the database.
+	 */
+	$filtered_data = wp_filter_comment( array(
```
pippinsplugins: Do we have a realistic case where this is actually needed?

includes/class-edd-note.php
```diff
@@ -0,0 +1,490 @@
...
+			foreach ( get_object_vars( $note ) as $key => $value ) {
+				$this->$key = $value;
+			}
+		}
+
+		/**
+		 * We fill object vars which have the same name as the object vars in WP_Comment for backwards compatibility
+		 * purposes.
+		 */
+		$this->comment_ID       = absint( $this->id );
```
pippinsplugins: I think we can get away with not maintaining these properties.

includes/payments/functions.php
sunnyratilal: It's here purely for backwards compatibility on the off chance that it gets filtered.  Highly unlikely that it will ever be  needed but doesn't hurt to keep it there?

sunnyratilal:
@pippinsplugins I've reverted 2652d054b20f61be602f385c5b442264729edccf because it breaks here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/gateways/paypal-standard.php#L958 which subsequently causes [this test](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/tests/tests-payments.php#L239) to fail.

Issue #6277, opened 2018-01-04
Base class for all custom tables


JJJ:
See: #6224, #5277, #4576, #3841, #6278 

With the number of custom database tables that EDD 3.0 will introduce, the codebase will benefit from a simple base class to manage the database tables themselves.

Not to be confused with an interface for CRUD actions, this base class is literally only for automating the scope and maintenance of custom database tables themselves.

My plan is as follows:
* Introduce `/includes/database/` as a new directory for the base class to start in
* That directory will eventually contain 1 class per additional database table (including meta tables)
* Each object table should have a complimentary `meta` table for plugin support, following the standard WordPress meta table schema. If that schema ever changes, we would want ours to follow suit.

Table logic:
* Each table will be versioned in `wp_options`, and schema changes will trigger automatic upgrades by comparing versions and calling appropriate methods
* Each table can be independently upgraded as the needs of the application change
* Each table is automatically added to the `$wpdb` interface, so querying them is magically supported
* Each table could be marked as global for explicit multisite support. By default, tables are not global, meaning in a multisite installation EDD is per-site. We could choose to have a canonical EDD site, or a different schema for global tables that includes a `site_id` column. We won't right away, but who knows.

History:
* This same approach is already pushed to Sugar Calendar and a number of my personal plugins, and works great
* I've opened a core ticket to consider moving all core tables to a similar approach. See: https://core.trac.wordpress.org/ticket/37692
* BuddyPress has not moved to this yet, but it probably should 💯 

----

Work on this issue will begin immediately in an issue branch, and I'll defer final merge back to someone else after a thorough review. 👍

sunnyratilal:
Closing this out as we have classes for all the custom tables and they are all working sufficiently.

PR #6308, opened 2018-01-14
issue/6277 - Base class for all custom tables

sunnyratilal:
**WIP - Do not merge.**

#6277


includes/database/class-edd-db-table-logmeta.php
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$max_index_length = 191;
+		$this->schema     = "meta_id bigint(20) unsigned NOT NULL auto_increment,
+			log_id bigint(20) unsigned NOT NULL default 0,
```
sunnyratilal: This needs to be `edd_log_id`.

includes/database/class-edd-db-table-logmeta.php
```diff
@@ -0,0 +1,59 @@
...
+			meta_key varchar(255) DEFAULT NULL,
+			meta_value longtext DEFAULT NULL,
+			PRIMARY KEY log_id (log_id),
```
sunnyratilal: This needs to be `edd_log_id`.

includes/database/class-edd-db-table-discounts.php
```diff
@@ -0,0 +1,71 @@
...
+			status varchar(20) NOT NULL default '',
+			type varchar(20) NOT NULL default '',
+			scope varchar(20) NOT NULL default 'all',
+			amount mediumtext NOT NULL default '',
+			description longtext NOT NULL default '',
+			max_uses bigint(20) unsigned NOT NULL default 0,
+			use_count bigint(20) unsigned NOT NULL default 0,
+			once_per_customer int(1) NOT NULL default 0,
+			min_cart_price mediumtext NOT NULL default '',
+			date_created datetime NOT NULL default '0000-00-00 00:00:00',
```
sunnyratilal: We're missing the `product_condition` column here.

includes/database/class-edd-db-table-discountmeta.php
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$max_index_length = 191;
+		$this->schema     = "meta_id bigint(20) unsigned NOT NULL auto_increment,
+			discount_id bigint(20) unsigned NOT NULL default 0,
```
sunnyratilal: This needs to be `edd_discount_id`.

includes/database/class-edd-db-table-discountmeta.php
```diff
@@ -0,0 +1,59 @@
...
+			meta_key varchar(255) DEFAULT NULL,
+			meta_value longtext DEFAULT NULL,
+			PRIMARY KEY discount_id (discount_id),
```
sunnyratilal: This needs to be `edd_discount_id`.

includes/database/class-edd-db-table-notemeta.php
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$max_index_length = 191;
+		$this->schema     = "meta_id bigint(20) unsigned NOT NULL auto_increment,
+			note_id bigint(20) unsigned NOT NULL default 0,
```
sunnyratilal: This needs to be `edd_note_id`.

includes/database/class-edd-db-table-notemeta.php
```diff
@@ -0,0 +1,59 @@
...
+			meta_key varchar(255) DEFAULT NULL,
+			meta_value longtext DEFAULT NULL,
+			PRIMARY KEY note_id (note_id),
```
sunnyratilal: This needs to be `edd_note_id`.

includes/database/class-edd-db-table-notes.php
```diff
@@ -0,0 +1,63 @@
...
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$this->schema = "id bigint(20) unsigned NOT NULL auto_increment,
+			object_id bigint(20) unsigned NOT NULL default '0',
+			object_type varchar(20) NOT NULL default '',
+			author_id bigint(20) unsigned NOT NULL default '0',
+			author_ip varchar(60) unsigned NOT NULL default '',
```
sunnyratilal: Mentioned previously in the commit but not sure `author_ip` is necessary for a note as it has not been used previously and I can't seem to think why it would be necessary considering only users with certain roles will be able to make notes on certain objects (i.e. payments, discounts).

includes/database/class-edd-db-table-payment-itemmeta.php
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$max_index_length = 191;
+		$this->schema     = "meta_id bigint(20) unsigned NOT NULL auto_increment,
+			payment_item_id bigint(20) unsigned NOT NULL default 0,
```
sunnyratilal: Needs to be `edd_payment_item_id`.

includes/database/class-edd-db-table-payment-itemmeta.php
```diff
@@ -0,0 +1,59 @@
...
+			meta_key varchar(255) DEFAULT NULL,
+			meta_value longtext DEFAULT NULL,
+			PRIMARY KEY payment_item_id (payment_item_id),
```
sunnyratilal: Needs to be `edd_payment_item_id`.

includes/database/class-edd-db-table-paymentmeta.php
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$max_index_length = 191;
+		$this->schema     = "meta_id bigint(20) unsigned NOT NULL auto_increment,
+			payment_id bigint(20) unsigned NOT NULL default 0,
```
sunnyratilal: Needs to be `edd_payment_id`.

includes/database/class-edd-db-table-paymentmeta.php
```diff
@@ -0,0 +1,59 @@
...
+			meta_key varchar(255) DEFAULT NULL,
+			meta_value longtext DEFAULT NULL,
+			PRIMARY KEY payment_id (payment_id),
```
sunnyratilal: Needs to be `edd_payment_id`.

sunnyratilal:
Left some comments @JJJ.

includes/database/class-edd-db-table-discountmeta.php
```diff
@@ -0,0 +1,59 @@
...
+// Exit if accessed directly
+defined( 'ABSPATH' ) || exit;
+
+if ( class_exists( 'WP_DB_Table' ) ) :
+/**
+ * Setup the global "edd_discountmeta" database table
+ *
+ * @since 3.0.0
+ */
+final class EDD_DB_Table_Discountmeta extends WP_DB_Table {
```
mindctrl: Aren't we namespacing all this new code?

includes/database/class-edd-db-table-discounts.php
```diff
@@ -0,0 +1,71 @@
...
+		$this->schema = "id bigint(20) unsigned NOT NULL auto_increment,
+			name varchar(200) NOT NULL default '',
+			code varchar(50) NOT NULL default '',
+			status varchar(20) NOT NULL default '',
+			type varchar(20) NOT NULL default '',
+			scope varchar(20) NOT NULL default 'all',
+			amount mediumtext NOT NULL default '',
+			description longtext NOT NULL default '',
+			max_uses bigint(20) unsigned NOT NULL default 0,
+			use_count bigint(20) unsigned NOT NULL default 0,
```
mindctrl: Do we need `bigint` for these use counters? That's a huge number and if I understand correctly requires more storage space.

includes/database/class-edd-db-table-payment-items.php
```diff
@@ -0,0 +1,68 @@
...
+	/**
+	 * Setup the database schema
+	 *
+	 * @since 3.0.0
+	 */
+	protected function set_schema() {
+		$this->schema = "id bigint(20) unsigned NOT NULL auto_increment,
+			payment_id bigint(20) unsigned NOT NULL default '0',
+			product_id bigint(20) unsigned NOT NULL default '0',
+			price_id bigint(20) unsigned NOT NULL default '0',
```
mindctrl: `bigint` seems like overkill for price option ids.

includes/database/class-edd-db-table-payments.php
```diff
@@ -0,0 +1,76 @@
...
+			tax double NOT NULL default '0',
+			discounts double NOT NULL default '0',
+			total double NOT NULL default '0',
+			PRIMARY KEY (id),
+			KEY number (number),
+			KEY status (status(20)),
+			KEY user_id (user_id),
+			KEY customer_id (customer_id),
+			KEY email (email(100)),
+			KEY ip (ip(60)),
```
mindctrl: Asked this on a commit, but do we need to index ip?

includes/database/class-wp-db-table.php
```diff
@@ -0,0 +1,340 @@
...
+
+	/**
+	 * Add class hooks to WordPress actions
+	 *
+	 * @since 1.1.0
+	 */
+	private function add_hooks() {
+
+		// Activation hook
+		register_activation_hook( __FILE__, array( $this, 'maybe_upgrade' ) );
```
mindctrl: I don't think this will work since this file doesn't get activated.

includes/database/class-edd-db-table-payment-itemmeta.php
```diff
@@ -0,0 +1,59 @@
...
+// Exit if accessed directly
+defined( 'ABSPATH' ) || exit;
+
+if ( class_exists( 'WP_DB_Table' ) ) :
+/**
+ * Setup the global "edd_payment_itemmeta" database table
+ *
+ * @since 3.0.0
+ */
+final class EDD_DB_Table_Payment_Itemeta extends WP_DB_Table {
```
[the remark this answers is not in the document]
mindctrl: On that note, I think we should split the words with an underscore, like we're doing in `EDD_DB_Discount_Meta`.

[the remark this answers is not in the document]
JJJ: Doh. Yes. Good catch.

includes/database/class-edd-db-table-payments.php
JJJ: Consensus in Slack seems like no.

I added it thinking it might be nice to search by or query by IP. Would help for preventing duplicate transactions, detecting spam/abuse, etc... Maybe that's already more easily handled elsewhere or not important for now.

includes/database/class-edd-db-table-discounts.php
JJJ: Probably not, but the compelling reason to stick with `bigint(20)` everywhere is because it's what WordPress uses almost everywhere else for numeric values.

WordPress also uses `int(11)` in some places that are `rating` or `karma` related, but they aren't actually used in core anymore. `user_status` still uses this, incorrectly (should be a string key with an API like posts.)

See this post for more on storage: https://stackoverflow.com/questions/4769416/mysql-bigint-vs-int

> The difference is purely in the maximum value which can be stored

4-byte vs. 8-byte per character. Since everything is UTF8 going forward, it's probably best to stay with `bigint` but the `(20)` for `ZEROFILL` is a stylistic decision that I don't feel strongly about.

includes/database/class-edd-db-table-discountmeta.php
JJJ: `WP_DB_Table` is a base class that won't need namespacing. We could if we really want to be overly cautious, but changing or aliasing it later is a non-breaking change also. 50/50.

includes/database/class-edd-db-table-payments.php
JJJ: I removed this index.

includes/database/class-wp-db-table.php
JJJ: Fixed in c7b25eeafb48ad0123af6546759398aa3d0cb7f9.

JJJ:
I _think_ this is ready to merge, but would like eyes on it before hitting go.

A few things:

* These files need to be grafted in. They aren't included, so they don't actually execute yet.
* Once they're included, the load order will need to be adjusted to be as early as possible
* The new `$files` class variables will need to be added to point to the file we use as the activation hook for creating these tables. This might need tweaking/removing depending on how third-party plugin support looks.
* If we want to tweak the new database table schemas after merge but before 3.0 ships, we'll either need to `DROP` them ourselves or create class methods with the necessary `ALTER`s. Either way is fine with me, and that's why these classes exist.

sunnyratilal:
@JJJ This is looking great. I've pulled down and checked through everything and it all looks good. We should definitely add in the function that loads these new classes and creates the tables before we merge in so we can get each of the other migrations back to a testable, and usable state.

One last thing - are we namespacing these new classes? Perhaps `EDD\Database`?

JJJ:
Sounds good. I’ll write that function tonight and add it here.

Re: namespacing - What do you think about separating database tables from database queries? I did this for Sugar Calendar, but I think it actually makes more sense in EDD where there are so many of both. Maybe `EDD\Database\Tables` and `Queries`?

Issue #6290, opened 2018-01-09
Bump the PHP minimum to 5.3


DrewAPicture:
I think it's safe to assume that to truly bump the minimum to PHP 5.3 in 3.0, we'll likely need to handle it similarly to how we've done in AffiliateWP: keeping the main plugin file 5.2 compatible, throwing up an admin notice, and killing further execution, then do any 5.3+ stuff in other files.

afragen:
Clearly this provides the opportunity to create an autoloader and namespacing. I’d love to help with a PR but I have a couple of questions.

* What do you see as the primary namespace? `EDD` or `Easy_Digital_Downloads`
* Should file restructuring and renaming be considered to more easily conform to PSR-4? This could mean renaming files from `includes/class-edd-customer-query.php` containing `class EDD_Customer_Query` to `src/EDD/Customer_Query.php` containing `namespace EDD` and `class Customer_Query`.

Really I’m looking for some guidelines for helping with an autoloader and file/directory restructuring to help, but these are decisions that @pippinsplugins et al would need to make.

DrewAPicture:
We're talking about selectively autoloading some things, but not the whole codebase. The top-level namespace I've been using so far in the reports API reboot for the 3.0 sprint is EDD

sunnyratilal:
Just to add to that: we've also decided against renaming/restructuring due to backwards compatibility purposes.

pippinsplugins:
Renaming / re-structuring will have to happen slowly so as to prevent problems for existing integrations. We can, however, begin to introduce namespaces for all new code.

I would like the primary namespace to be `EDD`.

afragen:
Well I was looking for direction and got it in spades. Thanks everyone.

cklosowski:
@JeroenSormani At this time our active user base reporting PHP versions is too high for 5.3 for us to cut support for it. We closely monitor this number and when we hit a number we're comfortable removing support for, we can start that process.

JeroenSormani:
Are those numbers about PHP versions for up-to-date installs? Active installs often are also outdated, e.g. someone running WP 3.7 and EDD 2.1 with PHP 5.3 shouldn't be really taken into consideration, IMO 😇

[...]

PR #6311, opened 2018-01-16
issue/6290 - (WIP) Initial pass at autoloader

afragen:
Still very rough.

#6290

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -31,432 +31,60 @@
 // Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) exit;
 
-if ( ! class_exists( 'Easy_Digital_Downloads' ) ) :
+if ( version_compare( '5.3.0', PHP_VERSION, '>=' ) ) {
+	?>
+	<div class="error notice is-dismissible">
+		<p>
+			<?php printf( esc_html__( 'Easy Digital Downloads cannot run on PHP versions older than %s. Please contact your hosting provider to update your site.', 'easy-digital-downloads' ), '5.3.0' ); ?>
+		</p>
+	</div>
+	<?php
 
-/**
- * Main Easy_Digital_Downloads Class.
- *
- * @since 1.4
- */
-final class Easy_Digital_Downloads {
-	/** Singleton *************************************************************/
-
-	/**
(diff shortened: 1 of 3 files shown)


afragen:
@cklosowski @DrewAPicture wanna take a look at this. Working fine on my end. 

I/We may need to make a few adjustments to use same autoloader for EDD_SL but maybe just include a slightly different version. I haven't worked on that part yet.

DrewAPicture:
@cklosowski @JJJ can correct me if I'm wrong, but I was under the impression initial plan was introduce the autoloader much lower in the stack to autoload _specific_ files or directories only and no rework the entire base class. This feels like a pretty heavy approach.

afragen:
Sorry if I over extended. This autoloader doesn't autoload all files, only classes.

afragen:
I guess I mean allow for PSR-4 loading if that sort of file renaming occurs.

afragen:
I think I've actually gotten it to work with EDD and EDD_SL.

JJJ:
Hey all,

I've reviewed this PR, and while I agree with @DrewAPicture that it's a lot to take in, the work done here is both impressive and comprehensive. It would be a shame for it to go unnoticed, so I want to thank @afragen for taking the time to work this all out.

The only reason that I do not want to merge this as is, is because the long term goal is to normalize _all_ of the file names and locations into an easier to understand and work-with pattern, and that will offset the need for a lot of the complexity that was necessary here.

----

There are still a number of add-ons that choose to include EDD files directly, and while I personally do not want to endorse that type of coding for very much longer, we will need time to communicate to those plugin authors to avoid breaking sites for users.

As we continue to work more on EDD 3.0, this PR will become increasingly more stale, and merging is going to become more and more difficult, and realistically, unlikely.

I don't want to abandon Andy's efforts, but I also don't want to keep expectations lingering around that this will be the approach that we end up taking when it seems like it may not be.

I'd like to recommend that we close this pull request without merging, and revisit the approach used here for EDD 3.1 or after. Pinging @pippinsplugins @cklosowski for a consult.

pippinsplugins:
@JJJ I can agree on all your sentiments above. Thanks @afragen!

afragen:
@pippinsplugins @jjj your are both/all quite welcome. It was a fun exercise and I learned a bunch in doing it.

Issue #6306, opened 2018-01-13
Edit download lifetime checkbox does not save first time around


scottrpaterson:
Editing a download with variable pricing enabled and the software licensing add-on and you click "Add New Price".

In this newly created price option box, if you check the "lifetime" checkbox and then save the page, it does not save that checkbox value. You have to go back and click it again, then save a second time.

![untitled](https://user-images.githubusercontent.com/11578353/34900123-9cee1aaa-f7ba-11e7-9c27-4fe1cffd9953.png)

cklosowski:
Just making a note, that while this references Software Licensing, it is possible that this is due to something in EDD Core not fixing the `index` of the input when the clone happens on the row. So before we close and move to SL, we should just verify if this belongs in Core or SL.

SDavisMedia:
I think this is core. It happens with the All Access "Exclude from All Access?" setting as well. I can only replicate the behavior with checkbox settings. Everything else seems fine.

To be clear, we want all fields in their default state for a newly created row, correct? @cklosowski

cklosowski:
@SDavisMedia correct. they should be in their unmodified state, and their `index` updated for the new row ID.

SDavisMedia:
Just taking a stab at starting this. Not sure if there's a better way, but the above PR will uncheck any checkboxes when the repeatable row is cloned.

**The problem with the checkbox not saving on the first save is still present, though.**

cklosowski:
i feel like this line is ithe one that's failing @SDavisMedia https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6315/files#diff-4e07975920906985a5fbbe3afd720629R56

Will do some testing today to see what I can figure out.

cklosowski:
Ignore my previous statement, the problem is here:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6315/files#diff-4e07975920906985a5fbbe3afd720629R50

We are setting the `value` attribute of all `inputs` to `''` or empty string, meaning that the checkbox is getting an empty value, instead of keeping it at `1`, so saving doesn't work. We need a way to break that out so that we don't reset the `value` on checkboxes, but still do all the key updates.

SDavisMedia:
@cklosowski I think you are absolutely right. I should be able to sort that out then. I have two approaches in mind already. I'll play with them and push something up for feedback later on.

Thanks!

SDavisMedia:
[...]

PR #6315, opened 2018-01-17
issue/6306 - manually uncheck checkboxes when cloning repeatable row #6306

SDavisMedia:
#6306

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -36,6 +36,14 @@ jQuery(document).ready(function ($) {
 				$( this ).val( row.find( 'select[name="' + $( this ).attr( 'name' ) + '"]' ).val() );
 			});
 
+			/** manually uncheck any checkboxes */
+			clone.find( 'input[type="checkbox"]' ).each(function() {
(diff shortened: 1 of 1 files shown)


assets/js/admin-scripts.js
```diff
@@ -61,6 +56,20 @@ jQuery(document).ready(function ($) {
...
+				$( this ).val( row.find( 'select[name="' + $( this ).attr( 'name' ) + '"]' ).val() );
+			});
+
+			/** manually uncheck any checkboxes */
+			clone.find( 'input[type="checkbox"]' ).each(function() {
+				var checked = $(this).is(':checked');
+				if ( checked ) {
+					$(this).prop('checked', false);
+				}
+				$(this).val(1);
```
cklosowski: @SDavisMedia Let's drop a comment before this line detailing why we are setting the value (due to the foreach above).

Should just make less confusion if someone looks and sees the unset above, and then us setting it here. Other than that, looks/works great! Nice job.

SDavisMedia:
@cklosowski how's that?

cklosowski:
Working perfectly. Merging.

Issue #6317, opened 2018-01-18
PHP 7.2 fatal error


afragen:
There's a PHP fatal error in PHP 7.2 when `count()` is not a countable object or array, like when it's null.

This issue can exist for https://github.com/easydigitaldownloads/easy-digital-downloads/blob/c91cfd9cd984c856b345f88fa9fc0b00887a737c/includes/emails/class-edd-email-tags.php#L211

JJJ:
I just ran into this today, while testing on 7.2.

pippinsplugins:
Put into the 3.0 milestone for now. If we encounter this more, we can move to a point release if necessary.

sunnyratilal:
All unit tests that use `count()` need to be updated as well - PHP 7.2 changed the way the `count()` method is used: https://wiki.php.net/rfc/counting_non_countables.

sunnyratilal:
Currently we have 61 unit tests failing on PHP 7.2 with the same error as above - I'll go ahead and open a PR with so that our test matrix can also test against PHP 7.2 as it was released as of 30 November 2017.

sunnyratilal:
Core build now passes with PHP 7.2

sunnyratilal:
Closing this out as the build currently passes with no issues when tested against PHP 7.2.

PR #6318, opened 2018-01-18
fix for PHP 7.2 fatal error if return value is null

afragen:
Not sure which branch to make this against. #6317


JJJ:
Fixed in `release/3.0` by casting upstream a bit farther than @afragen did in the original PR.

Issue #6331, opened 2018-01-26
3.0 Queries: Base query class (meta, dates, caching)


JJJ:
Tracking issue: #4576

Pull request: #6333 

# TL;DR

Transition Payments/Discounts/Logs/Notes/Orders/Order-Items post-types to custom component database tables, most paired with meta. Architect a traditional WordPress query interface that's simple and DRY for all of them to maintain developer happiness.

# Summary

One technical challenge for EDD3 is the introduction of approximately 12 new database tables (for a total of 14.) The reason for this move is to ultimately provide a more scalable application without a _requirement_ for third-party horizontal scaling techniques (multiple servers; WordPress drop-in plugins; cache/db/lb/web/etc...)

By scaling the databases horizontally by default, EDD can better control the flow of data internally without compromising performance. Put another way, by not storing everything as a custom post type & post-meta, EDD can avoid hurting the performance of other post-type related data, while also improving its own efficiency.

It's a win-win for everyone, but with a hidden caveat...

We learned first-hand (via BuddyPress and Multisite) that the above approach works well-enough at providing a stable foundation for scaling a complex WordPress application, but we also learned having more than a half-dozen different interfaces for interacting with new & unique components in WordPress is a hugely negative experience for traditional WordPress developers.

Each new component/meta relationship is a new hurdle to jump, and too many hurdles leaves developers exhausted and unhappy.

I'm proposing (and working on) a mostly-magic base query-class for all custom database tables to extend, that will automatically make every new table queryable using the same basic parameters, provide object caching for those individual items, without compromising either application performance or developer joy.

Kinda like an [ORM](https://en.wikipedia.org/wiki/Object-relational_mapping), but without the need to relearn the WordPress way of querying for things.

## How does it work?

1. Component classes provide an array of column definitions and other component attributes
1. Column definitions help determine the shape of the component item
1. That shape helps define the queryable columns
1. Queries are cached by an md5 hash of parameters used
1. Items are cached by their ID

# Components

Here is the master list of components, alphabetically. Each will get checked when complete:

- [x] Customers
- [x] Discounts
- [x] Logs (General)
- [x] Logs (API Requests, No Meta)
- [x] Logs (Downloads, No Meta)
- [x] Notes
- [x] Orders
- [x] Order Items (Linked to Orders)

# Orders (Previously Payments)

Table/schema completed (see #6277)

### Query vars
- [x] `id`
- [x] `id__in`
- [x] `id__not_on`
[...]

PR #6333, opened 2018-01-27
issue/6331 - Query class, Meta query, Date queries, Caching

sunnyratilal:
**WIP - Do not merge**

Issue: #6331


JJJ:
Updated title to reflect changes to #6331.

includes/database/queries/class-edd-db-column.php
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	 * Return array of allowed column types.
+	 *
+	 * @since 3.0.0
+	 * @access private
+	 * @return array
+	 */
+	private function get_allowed_types() {
+		return (array) apply_filters( 'get_allowed_types', array(
```
mindctrl: If we keep this filter and the one below, we need to prefix them with `edd_<something>`.

includes/database/queries/class-edd-db-column.php
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	 * Type of database column
+	 *
+	 * @since 3.0.0
+	 * @access public
+	 * @var string
+	 */
+	public $type = '';
+
```
mindctrl: What's the reason for these properties being `public`? Can we make them `protected` or `private` and use a method to retrieve them instead?

includes/database/queries/class-edd-db-order-query.php
```diff
@@ -0,0 +1,272 @@
...
+		),
+
+		// total
+		array(
+			'name'       => 'total',
+			'type'       => 'double',
+			'default'    => '0'
+		)
+	);
+
```
mindctrl: Same question as above regarding `public`.

includes/database/queries/class-edd-db-column.php
JJJ: I think I'll probably remove them for EDD3, and work on them later if there's a business need. We can just use the appropriate MySQL column types until then, and worry about this bit of sanity checking once we identify if this is even a problem.

includes/database/queries/class-edd-db-column.php
JJJ: The reason is it's easier for me to architect these bits fully public, and then protect or privatize them later once it's all working.

Protected is probably best IMO. I think we can add methods to get private vars once we start enforcing the shape/sanitization/validation of them.

JJJ: Just tested setting these as `protected` and remembered that even with magic getters, WordPress's list utility functions and methods use `array_key_exists()` internally, so these will need to either:

* Remain explicitly public
* Not be defined at all, which will default them to public when they are set

I think being explicit here is better, even if it's exposing these vars a bit.

includes/database/queries/class-edd-db-column.php
JJJ: Opted to remove this filter entirely.

JJJ:
Deleted the `issue/6331` branch.

PR #6362, opened 2018-02-07
Add hooks before and after downloads shortcode item

kiarashi:
If it's not too much trouble, I added two new hooks;
`edd_download_shortcode_item_before` & `edd_download_shortcode_item_before`
inside the `edd_downloads_query` function. With these two new hooks I (and others) can add extra code before and after the loop.

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -579,11 +579,16 @@ function edd_downloads_query( $atts, $content = null ) {
 		$wrapper_class = 'edd_download_columns_' . $atts['columns'];
 		ob_start(); ?>
 		<div class="edd_downloads_list <?php echo apply_filters( 'edd_downloads_list_wrapper_class', $wrapper_class, $atts ); ?>">
+			
+			<?php do_action( 'edd_download_shortcode_item_before', $atts, $i ); ?>
+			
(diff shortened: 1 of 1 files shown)


cklosowski:
@kiarashi In order to be properly named, these would need to be within the `while` loop. The naming suggest you want one to fire before and after each item is displayed? If so, they are in the wrong place currently. This would always just fire before the loop and after the loop, rending sending in the `$i` unnecessary.

I would suggest that these exist before and after the line that reads:
`<?php do_action( 'edd_download_shortcode_item', $atts, $i ); ?>`

and within the `while` loop.

kiarashi:
It's actually the opposite; I need these hooks outside the `while` loop.  By having these extra hooks outside, I can add extra code to place the edd downloads inside a custom grid or table list. 
If I wish to add any code inside the loop, I can use the shortcode-xxx.php templates for that. :)

cklosowski:
@kiarashi ok, then the naming needs to change.

`edd_download_shortcode_item_before` as a name makes it appropriate for a single item, where the action actually runs outside of the item scope.

kiarashi:
Feel free to name the hooks as you see fit. :)

SDavisMedia:
New PR with @kiarashi commit carried over: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6364

PR #6366, opened 2018-02-08
Save page permalinks for performance #6369

JeroenSormani:
Hi Guys!

In the opportunity to reduce DB queries I've found that a `get_post` call was done every pageload if a cart widget is shown somewhere to get the permalink of the checkout page. I've worked to make a workaround for that by saving the permalinks in the settings instead and use that when possible.

Should be fully BC and nicely save the settings during the upgrade process.

PS. the other settings/pages didn't make a call by default, but for consistency I did also save those permalinks

Addresses #6369

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -2357,3 +2357,35 @@ function edd_add_setting_tooltip( $html, $args ) {
 	return $html;
 }
 add_filter( 'edd_after_setting_output', 'edd_add_setting_tooltip', 10, 2 );
+
+/**
+ * Save permalinks to save permalink only gueries in the future.
+ *
+ * @since 2.9.0
+ *
+ * @param mixed $value Current option value.
+ * @param mixed $old_value Old option value.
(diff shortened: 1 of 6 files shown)


cklosowski:
This seems to have broken the unit tests.

It's likely due to the tests being isolated, so the settings have to be remade every time. I'll move this over to an issue branch and see if we can work from there. @JeroenSormani I'm going to close this PR once I get the issue branch made and keep your commit history intact.

JeroenSormani:
Should also add code to generate the options at installation

cklosowski:
@JeroenSormani that's probably where the tests are failing. I'll fix that up. thanks!

JeroenSormani:
yeah figured that out too just now, I've already written the code, but having issues running the unit tests locally atm

cklosowski:
no worries @JeroenSormani. I have them running locally here, and can take care of it if you want.

JeroenSormani:
@cklosowski just pushed the changes, I'm also on Slack / pinged you there.

PR #6379, opened 2018-02-13
Make sure there are tags, before we try and count them #6372

### LINKED ISSUE #6372: Warning: count(): Parameter must be an array or an object that implements Countable in widgets.php on line 311
Did a fresh install of WordPress, EDD with Themedd locally and setup the default widgets. On download pages, this error is displaying:

`Warning: count(): Parameter must be an array or an object that implements Countable in /app/public/wp-content/plugins/easy-digital-downloads/includes/widgets.php on line 311`

cklosowski:
Resolves #6372

### DIFF
--- a/includes/widgets.php
+++ b/includes/widgets.php
@@ -302,15 +302,36 @@ class EDD_Product_Details_Widget extends WP_Widget {
 		echo $purchase_button;
 
 		// categories and tags.
-		$category_list     = $categories ? get_the_term_list( $download_id, 'download_category', '', ', ' ) : '';
-		$category_count    = count( get_the_terms( $download_id, 'download_category' ) );
-		$category_labels   = edd_get_taxonomy_labels( 'download_category' );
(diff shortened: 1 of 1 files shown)


spencerfinnell:
@cklosowski This won't fully solve #6372 I don't think. `get_the_terms()` can still return `false` or `WP_Error`. You're just checking if the widget option is set before counting, not actual returned tags.

Should be something like

```php
$tags          = $instance['tags'] ? $instance['tags'] : '';
$tag_terms = get_the_terms( $download_id, 'download_tag' );

if ( $tags && $tag_terms && ! is_wp_error( $tag_terms ) ) {
// make list
// count
// etc
}
```

cklosowski:
Ah, yes @spencerfinnell I had written all that prior to doing this, and forgot what `$tags` was associated with, I'll get it cleaned up.

cklosowski:
@spencerfinnell this looking better for you?

pippinsplugins:
Looks good to me, merging.

PR #6380, opened 2018-02-13
Ignore the reqeust for checkin, and just move on #6367

### LINKED ISSUE #6367: Tracking timeout and blocking
We don't really care about the server result of the optin tracking nor does it need to be blocking.

We could/should change the wp_remote_post to blocking->false and timeout to either 3 or 5 seconds

cklosowski:
Resolves #6367

### DIFF
--- a/includes/admin/tracking.php
+++ b/includes/admin/tracking.php
@@ -138,7 +138,7 @@ class EDD_Tracking {
 
 		wp_remote_post( 'https://easydigitaldownloads.com/?edd_action=checkin', array(
 			'method'      => 'POST',
-			'timeout'     => 20,
+			'timeout'     => 8,
 			'redirection' => 5,
 			'httpversion' => '1.1',
 			'blocking'    => false,
--- a/includes/admin/tracking.php
+++ b/includes/admin/tracking.php
@@ -110,7 +110,11 @@ class EDD_Tracking {
 	 * Send the data to the EDD server
 	 *
 	 * @access private
-	 * @return void
+	 *
+	 * @param  bool $override If we should override the tracking setting.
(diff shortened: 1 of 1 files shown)


includes/admin/tracking.php
```diff
@@ -132,20 +136,16 @@ public function send_checkin( $override = false, $ignore_last_checkin = false )
 
 		$this->setup_data();
 
-		$request = wp_remote_post( 'https://easydigitaldownloads.com/?edd_action=checkin', array(
+		wp_remote_post( 'https://easydigitaldownloads.com/?edd_action=checkin', array(
 			'method'      => 'POST',
 			'timeout'     => 20,
```
chriscct7: is there a reason the timeout needs to be so high on this? Since not all transport methods support non-blocking requests, we should really lower this to 3 or 5 seconds

cklosowski: I think we could, my concern would be that it aborts on the user end if we make it too short, but then it also by default would kill the PHP thread on our end as well.

I'll set it to 8 seconds, which is far less than 20, but in the future we'll get the following into the server end of this telemetry data:
`ignore_user_abort(true);`

With that, at least if the client end aborts, we'll still try and process.

cklosowski: Worst case scenario, we miss some telemetry data here and there until then.

chriscct7: Right, I think as a policy we should default to not getting data for a week, vs possibly affecting people's sites. I also think eventually EDD will rebuild the server side in non-WordPress and then you don't need to have such a large processing time since you can just get the data and store it incredibly fast.

PR #6384, opened 2018-02-14
fix image src when plugins are not inside wp-content, fixes #6044

### LINKED ISSUE #6044: Image paths not resolved in non standard setup
In my setup, the `WP_PLUGIN_DIR` is not inside my `WP_CONTENT_DIR` (but on the same level). I do not have any images/icons, e.g., payment icons in my theme dir.

In https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/checkout/template.php#L694, for example, you're trying to replace the plugins path with the `content_url()`. But since the plugins dir does not have to be inside the `WP_CONTENT_DIR` this will not work in all non-default setups.

alpipego:
Instead of replacing the content dir with the content url, this could be done on the plugins path/url respectively.

Addresses #6044

### DIFF
--- a/includes/checkout/template.php
+++ b/includes/checkout/template.php
@@ -681,7 +681,7 @@ function edd_show_payment_icons() {
 			} else {
 
 				$image       = edd_locate_template( 'images' . DIRECTORY_SEPARATOR . 'icons' . DIRECTORY_SEPARATOR . $card . '.png', false );
-				$content_dir = WP_CONTENT_DIR;
+				$content_dir = WP_PLUGIN_DIR;
 
 				if( function_exists( 'wp_normalize_path' ) ) {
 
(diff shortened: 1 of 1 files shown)


cklosowski:
@alpipego Is this the PR associated with issue #6044?

cklosowski:
I've tested this out, and cannot find any issues that it is causing for the cards. I even included 2Checkout in my testing since it uses a custom image, and it still continues to work.

alpipego:
@cklosowski Yes this is a fix for #6044 

It's not about custom images but custom `WP_CONTENT_DIR` and `WP_PLUIGN_DIR` constants. 

In [`includes/checkout/templates.php`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/checkout/template.php#L694) you're assuming that the plugin is always inside of the `WP_CONTENT_DIR`.

cklosowski:
@alpipego Right i was just testing with something that i know uses that line of code, and the custom payment icons are affected by it :+1:

PR #6394, opened 2018-02-19
Return expected array when cart contents are empty, not false

JeroenSormani:
Caused an error for me when I doing `reset( $content )` when the cart is empty because it returned `false`. 

Checking the EDD core code it only does `empty( $contents )` and ` if ( $contents )` type of checks, which should all react the same for returning `array()` as returning `false`

### DIFF
--- a/includes/cart/class-edd-cart.php
+++ b/includes/cart/class-edd-cart.php
@@ -216,7 +216,7 @@ class EDD_Cart {
 		global $edd_is_last_cart_item, $edd_flat_discount_total;
 
 		if ( empty( $this->contents ) ) {
-			return false;
+			return array();
 		}
 
 		$details = array();


cklosowski:
While semantically yes, returning an `array()` here seems logical, I do have some major concerns on it being a breaking change with extensions, etc.

@sunnyratilal you mind taking a look when you get a few minutes as well?

sunnyratilal:
@cklosowski I've taken a look and we should be fine with backwards compatibility here.

cklosowski:
Looked through and the only things I see are checking if the item returns, and if it's not `false` it does a `foreach`. I searched all of my installed plugins which is quite a few. My only concern is 3rd party ones.

Since this is a potentially 'breaking' change 2.9 is a valid place to put it, we'll just have to voice the change in the release.

PR #6403, opened 2018-02-27
fixes 0 length downloads for larger files

Ratcreamsoup:
We used to have problems with files of more than 50MB in size; for direct download method, users would receive a file with 0 bytes. The patch adresses this issue by making shure that the buffer is being flushed to the client during chunked downloads.


pippinsplugins:
@Ratcreamsoup this has already been fixed for the upcoming release. See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6387

PR #6411, opened 2018-03-05
fix incorrect link paths #6410

### LINKED ISSUE #6410: Admin Extensions page uses old /extensions URL
Mainly the 'Browse All Extensions' buttons leading back to the main site.

SDavisMedia:
Resolves #6410


SDavisMedia:
Didn't expect that! Good catch. :+1:

PR #6427, opened 2018-03-12
Provide better error messages when attempting to add email addresses to customers. #6182

cklosowski:
Addresses #6182


cklosowski:
Closing and deleting this branch. The attached Issue is slated for 3.0, and all this work is from the 2.9.x branching, which handles customer email addresses in a different way. WE should come back to this when 3.1 starts developmetn and look over tehse cahnges.

Issue #6442, opened 2018-03-19
Unnecessary update of _edd_payment_user_id when constructing EDD_Payment


jonmarkhall:
In EDD_Payment::setup_user_id(), the payment's _edd_payment_user_id field is updated to match the user ID of the customer associated with the payment if that payment meta field is currently empty. However, if the customer has no associated user ID, that payment meta field _should_ be empty, and updating it is unnecessary and could negatively affect performance, particularly if EDD_Payment objects are constructed multiple times for such guest payments in a single page load.
PR #6444

PR #6444, opened 2018-03-19
Fix unnecessary payment meta update

jonmarkhall:
Don't update the payment user ID to match the customer user ID if the customer user ID is empty
#6442


cklosowski:
Thanks @

Issue #6439, opened 2018-03-19
Update Grunt configuration


SDavisMedia:
I just need a placeholder issue.

SDavisMedia:
This is becoming difficult to manage because testing the config is causing changes (that's a good thing) I don't want to be part of this issue. 😄 So pushing up and opening a PR now to try to keep this stuff isolated.

This covers just the basics:

* Minify CSS files (individually) in the `assets/css` folder, excluding `.min.css` files and files with `jquery` in the name.
* Minify CSS files (individually) in the `templates` folder.
* Uglify JS files (individually) in the `assets/js` folder, excluding `.min.js` files and files with `jquery` in the name.
* Check all plugin translation functions for missing or incorrect text domains. Since there are a few that are missing on purpose, the command will continue to run, but the "problems" will display in command line just so you know where they are.
* Update the .pot file
* Build a `build/` directory in the root of the plugin that includes a clean copy of the version of EDD you're releasing as well as a .zip copy. This directory will not include any dev/tool related files and it will be ignored by git. So it's basically just for you locally (can easily be used to drop over an SVN checkout to update the files... until we have a better solution for that process).

-----

With NPM and Grunt set up for the project and everything updated based on the new config, the following command should handle all of the above:

`grunt build`

SDavisMedia:
I've actually decided to approach minifying the CSS files from the opposite direction.

In both the `assets/css` folder and the `templates` folder, there's actually just one file in each that we edit. The rest are libraries. I left the file specifications in long form (doesn't hurt anything) because we could end up adding more files. Plus, I'm curious if we usually minify those library files ourselves.

`chosen.css`, for example... do we usually download both versions of the file or do we get the regular file and then minify it "manually"? If the latter, I can adjust this to minify them with the grunt script. I just wanted to be sure first. Same is true of the other files that are libraries (both CSS and JS).

SDavisMedia:
Been trying to find a good tool for building a changelog rough draft based on Issue titles under a specific milestone. Not having much luck.

Feel free to drop ideas, preferences, and requirements... anything to make the release process easier to handle.

cklosowski:
@SDavisMedia this _might_ work, but i've not tested. https://github.com/conventional-changelog/grunt-conventional-github-releaser

spencerfinnell:
Linking to the comparison view in Github also provides helpful information on point releases. 

[...]

PR #6445, opened 2018-03-19
Issue/6439 (WIP) - update Grunt configuration

SDavisMedia:
See #6439

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,36 +1,36 @@
 module.exports = function(grunt) {
 
-// Load multiple grunt tasks using globbing patterns
-require('load-grunt-tasks')(grunt);
+	// Load multiple grunt tasks using globbing patterns
+	require('load-grunt-tasks')(grunt);
 
-// Project configuration.
-grunt.initConfig({
-	pkg: grunt.file.readJSON('package.json'),
+	// Project configuration.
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -81,103 +121,60 @@ grunt.initConfig({
...
-				cmd: 'tx push -s'
-			},
+		// Clean up build directory
+		clean: {
+			main: ['build/<%= pkg.name %>']
 		},
 
-		dirs: {
-			lang: 'languages',
+		// Copy the theme into the build directory
```
pippinsplugins: `theme` should be `plugin`

SDavisMedia: oooops... fixed

spencerfinnell:
The `copy` task might be simpler if you used a whitelist instead of a blacklist. Seems like theres just a few top level files and then a few directories vs excluding all the dotfiles.

https://plugins.trac.wordpress.org/browser/easy-digital-downloads/trunk

8 includes vs many excludes.

SDavisMedia:
@spencerfinnell Good call. I agree. I ultimately decided to make the same type of change (conceptually) for the `cssmin` and `uglify` tasks. I'll revisit `copy` later on tonight.

Issue #6447, opened 2018-03-21
Pippin's Disappearing Customers (VARCHAR issue with customers table email column)


chriscct7:
We've had an issue in EDD for a while where customers would disappear in EDD and could not be created. 

Today I finally figured it out. EDD sets the email address column for the customers column to a *really* short 50 characters. However, emails can be longer than this, in fact, WordPress allows for a VARCHAR(100). 

What happens if a user enters an email address > VARCHAR 50 atm is the order goes through, but the customer isn't created and then thus not associated with the order due to a silently handled failed DB insert (COLUMN VALUE EXCEEDS SIZE RESTRICTIONS) into the edd_customers table.

Changing this column to a VARCHAR 100 will solve this problem for the future.

We however will also need:
- ALTER table for existing EDD sites
- A process that finds orders where the EDD email address meta key value is > 50 characters, and for each of those orders, if an attached customer is not there, then create & associate the order.

Ideally this should be done before EDD 3.0 to allow for data normalization.

chriscct7:
This where the current VARCHAR is set (and where it needs to be updated): https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-db-customers.php#L589

cklosowski:
@JJJ Ping for EDD 3.0 please? Make sure we make this part of the DB class. I'm going to get something together for the current releases.

cklosowski:
@chriscct7 ah yes. good call. I'll work up an Upgrade routine that only shows up if there are payments with an email address over 50 chars.

cklosowski:
Actually, I think we should hold off here. I know it's a bug...but with 3.0 we'll be doing a TON of db upgrades, and new table formatting classes. Since we'll have to itterate over everything to migrate the payments, we should just take care of the customer creation during that. Instead of pushing a fast fix, i'd rather go for the solid fix where we're already going through each paymetn with a fine tooth comb.

PR #6449, opened 2018-03-21
Change email address column to varchar(100) #6447

cklosowski:
Addresses #6447


cklosowski:
I'm going to close this PR. We should do this alongside 3.0 as it's going to take analyzing all the data for the payments to know of any customers we should create in the process.

PR #6450, opened 2018-03-21
Always treat sections data as an array, instead of mixed #6434

### LINKED ISSUE #6434: Warning: count(): Parameter must be an array or an object that implements Countable in display-settings.php
As per the title, when you visit the "Licenses" tab this PHP error is triggered. Across the board, PHP 7.2 compatibility need's some work in EDD in relation to the usage of `count()`. Will keep an eye out for anywhere else I see it flagging up.

cklosowski:
resolves #6434


SDavisMedia:
@cklosowski you already know I accidentally approved this PR earlier. My environment is working now, though, and I was able to replicate and test. So, officially all good here. :+1:

Issue #5797, opened 2017-06-09
Auto-map fields during CSV import


davidsherlock:
I had to move products from one EDD install to another, and while the export/import functions work well, I feel mapping shouldn't be required (or at least pre populated) if the structure (headings) are using the EDD export schema.

The fields should be automapped (and selected) to save time and potential errors while importing to speed the process up.

pippinsplugins:
This is definitely something we wanted to make happen when first building the import tool but it didn't happen.

👍

pippinsplugins:
This is _lovely_!

PR #6453, opened 2018-03-26
Try and pre-select column when importing downloads/payments

JeroenSormani:
This will try to select the column when it has the right header (which is used when exporting payments/downloads from EDD). 

Selecting the columns once isn't that bad, but if you have to run a import like me more then once due to testing / developing this will be a hellish task to select each column every single import (test).

Before:
screen recording 2018-03-26 at 07 08 pm

After:
screen recording 2018-03-26 at 07 05 pm

If the field doesn't exist it won't break or anything, just won't pre-select it.


SDavisMedia:
#5797

PR #6456, opened 2018-03-26
Docblock improvements - proper @return types + remove @access public

JeroenSormani:
Well this PR escalated quickly.. 😅

Started with some `@return` type changes but ended up removing all `@access public` tags as these are considered implicit. 

- https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.access.pkg.html
- https://core.trac.wordpress.org/ticket/41452 (not exactly the one I was looking for - there's also a ticket somewhere for `access` specifically I believe


SDavisMedia:
All of this looks good to me. Thanks @JeroenSormani. 👌

PR #6457, opened 2018-03-27
Make string RTL-compatible

aristath:
Makes the `Customer since {$date}` string is RTL-compatible (in some languages the date should precede the "customer since" part of the string).
Also makes sure translations are properly escaped.


cklosowski:
Thanks @aristath!

PR #6466, opened 2018-03-30
issue/6436 - add support for custom classes on downloads shortcode

### LINKED ISSUE #6436: Add class parameter support to [downloads] short-code arguments
Having the ability to append an additional (_optional_) class onto the `[downloads]` wrapper when outputting the  short-codes/grids would be extremely useful when you want to apply different styling to different page elements.

SDavisMedia:
Fixes #6436

Proposed Changes:
1. Add support for `class` parameter in `[downloads]` shortcode.

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -590,8 +590,9 @@ function edd_downloads_query( $atts, $content = null ) {
 	if ( $downloads->have_posts() ) :
 		$i = 1;
 		$columns_class   = array( 'edd_download_columns_' . $atts['columns'] );
-		$wrapper_classes = array_unique( array_merge( $columns_class, explode( ',', $atts['class'] ) ) );
-		$wrapper_classes = rtrim( implode( ' ', $wrapper_classes ) );
+		$custom_classes  = array_filter( explode( ',', $atts['class'] ) );
+		$wrapper_classes = array_unique( array_merge( $columns_class, $custom_classes ) );
+		$wrapper_classes = implode( ' ', $wrapper_classes );
 		ob_start(); ?>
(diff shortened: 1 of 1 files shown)


includes/shortcodes.php
```diff
@@ -588,9 +589,11 @@ function edd_downloads_query( $atts, $content = null ) {
 	$downloads = new WP_Query( $query );
 	if ( $downloads->have_posts() ) :
 		$i = 1;
-		$wrapper_class = 'edd_download_columns_' . $atts['columns'];
+		$columns_class   = array( 'edd_download_columns_' . $atts['columns'] );
+		$wrapper_classes = array_unique( array_merge( $columns_class, explode( ',', $atts['class'] ) ) );
```
cklosowski: I would suggest we do something like:
`$args_classes = array_filter( explode( ',', $atts['class'] ) );`

And then:
`$wrapper_classes = array_unique( array_merge( $columns_class, $args_classes ) );`

Reason being that if you have no classes defined in `$args` and run an `explode` on it, you end up with an array with a single entry of an empty value, which means there _is_ something in the array, instead of being an `empty` array.

By doing the `array_filter` after the `explode` but before the `array_merge` we'd be merging an actual empty array, instead of an array with a single empty value:

```
array (size=1)
  0 => string '' (length=0)
```

vs.
```
array (size=0)
  empty
```

SDavisMedia: SMART. I think that actually solves the problem I attempted to solve with `rtrim()`, since having no classes on the shortcode led to an extra space. Will update!

SDavisMedia:
Okay check that out @cklosowski.

PR #6467, opened 2018-03-30
issue/6400 - correctly count cart columns based on hook usage

### LINKED ISSUE #6400: edd_checkout_cart_columns *always* returns 5
Unless `do_action( 'edd_checkout_table_header_first' )` or `do_action( 'edd_checkout_table_header_last' )` get called a couple of times, the following will always return 5:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/cff792c5d7887195b07af4d814b94953a2609a16/includes/cart/template.php#L140-L146

In fact you're adding `colspan="5"` on every cart, although the table is only three columns wide.

SDavisMedia:
Fixes #6400

Proposed Changes:
1. Instead of using `did_action()` to see if hooks are being used, use `$wp_filter` to retrieve the callbacks of the hooks. _If_ callbacks exist, count those to determine the number of columns to add to the default column count (3).

### DIFF
--- a/includes/cart/template.php
+++ b/includes/cart/template.php
@@ -138,13 +138,15 @@ add_action( 'edd_cart_empty', 'edd_empty_checkout_cart' );
  * @return int The number of columns
  */
 function edd_checkout_cart_columns() {
-	global $wp_filter;
+	global $wp_filter, $wp_version;
 
 	$columns_count = 3;
 
 	if ( ! empty( $wp_filter['edd_checkout_table_header_first'] ) ) {
 		$header_first_count = 0;
-		foreach ( $wp_filter['edd_checkout_table_header_first']->callbacks as $callback ) {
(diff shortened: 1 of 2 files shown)


cklosowski:
Looks like this isn't working for PHP 4.4, 4.5 or 4.6. Investigating and fixing what's necessary.

cklosowski:
the issue is due to a change in WP 4.7:
https://make.wordpress.org/core/2016/09/08/wp_hook-next-generation-actions-and-filters/

Will fix it up

cklosowski:
Details of the reason it broke Pre-4.7 versions is that the `$wp_filter` global changed in 4.7. Pre-4.7 it was a simple array of hooks and priorities. In 4.7 it was changed to an object with all the priorities existing in a `callbacks` property.

We're just doing a simple version check of WP Core now to know what format that data is in.

SDavisMedia:
@cklosowski makes perfect sense to me! :+1:

Issue #6471, opened 2018-04-02
Search not fully working


JeroenSormani:
It stopped working >= 2.8.13 due to a change made for this issue: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6178

If sequential order numbers are enabled nothing below that `if` is executed, causing the search functionality in the payments (and licenses among others probably) to not fully work as expected.

Searching for a any generic `s` (non-user/email), `discount:` or  `#` download doesn't work.

cklosowski:
I'd say let's see if we can test this once 3.0 is at a more stable point. With the new tables we're building out our own DB patterns that should, _in theory_, give us far more flexibility in the search patterns.

@JJJ @sunnyratilal heads up, let's see if we can support this in the native query classes we're building for payments.

JJJ:
Acknowledged. We can trick this out very easily now, so I’ll think on it and own this for 3.0.

JJJ:
Here's where we are right now:

* For 3.0 specifically, `get_search_sql()` looks like:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/queries/class-base.php#L1028

* It's looped through in our `parse_where()` method:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/queries/class-base.php#L1090

* Searchable columns are filterable:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/queries/class-base.php#L1243

* Specifically, `wp_edd_orders:order_number` is marked as searchable 👍

* Meta queries use `WP_Meta_Query` so `LIKE` comparisons are possible.

* We also allow for custom `groupby` clauses, which we use internally inside of wp-admin for grouping discounts and customers currently. Support for multiple `groupby`'s is coming eventually.

If I understand the original problem correctly, the next steps are making sure that our admin area screens and forms are properly passing their `s` values through.

We _could_ go a step farther and add some _very basic_ of parameterization to our search boxes. Say, searching for: `order_number=2,status=pending`. If we think we want that in a future version other than 3.0, I can get `DB\Query\Base` ready now so we don't need any breaking changes later.

All ears!

Issue #6178, opened 2017-11-13
Searching by order number only works if a postfix or prefix is included


pippinsplugins:
When searching payment history, the search will return no results if a sequential order number is entered and the store does not use prefixes nor postfixes for order numbers.

It fails because of our logic in the payments query class:

```php
} elseif (
	edd_get_option( 'enable_sequential' ) &&
	(
		false !== strpos( $search, edd_get_option( 'sequential_prefix' ) ) ||
		false !== strpos( $search, edd_get_option( 'sequential_postfix' ) )
	)
) {

	$search_meta = array(
		'key'     => '_edd_payment_number',
		'value'   => $search,
		'compare' => 'LIKE'
	);

	$this->__set( 'meta_query', $search_meta );
	$this->__unset( 's' );

}
```

If sequential order numbers are enabled, I think we should just assume a numerical number is an order number and always search in `_edd_payment_number`.

PR #6472, opened 2018-04-02
Fix search not fully working

JeroenSormani:
Fixes https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6471

Reverts the change made in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6178 and adds additional code to search within sequential payment numbers when entering just the numeric ID.


JeroenSormani:
@cklosowski I don't believe that is relevant for this code. I've just done a test to verify on 5.2 and it doesn't give me a notice.
The relevant part is only restored from a older version of EDD, so that should be all tested and working already ;-)

Issue #6324, opened 2018-01-23
Group backwards compatibility code together


sunnyratilal:
As part of the migrations, we are going to have a lot of code that ensures backwards compatibility.  A lot of this will be repeated and be placed across multiple files.

We should introduce a new file that has all of the code for backwards compatibility and group together where possible.

E.g. https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/5277/includes/discount-functions.php#L1173 can be changed to work with `edd_discount` and `edd_payment` without requiring two separate methods.

sunnyratilal:
Closing all of the backwards compatibility code has been grouped together.

PR #6475, opened 2018-04-03
issue/6324 - Phase 1 of Backwards Compatibility

sunnyratilal:
Phase 1 of back compat for 3DD. #6324


includes/compat/class-base.php
```diff
@@ -0,0 +1,53 @@
...
+
+// Exit if accessed directly
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Base class.
+ *
+ * @since 3.0.0
+ */
+abstract class Base {
```
DrewAPicture: Can we add an abstract `hooks()` method to `Base` to enforce the contract for subclasses?

includes/compat/class-customer.php
```diff
@@ -0,0 +1,170 @@
...
+ */
+class Customer extends Base {
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.0.0
+	 */
+	public function __construct() {
+		$this->set_component( 'customer' );
```
DrewAPicture: Why not just adjust `Base` to inject a required `$component` via a constructor then set `$component there directly? Making subclasses set things that _every_ subclass needs to set just encourages unnecessary duplication, imo.

includes/compat/class-base.php
```diff
@@ -0,0 +1,53 @@
...
+	 *
+	 * @since 3.0.0
+	 * @access private
+	 * @var string
+	 */
+	private $component;
+
+	/**
+	 * Setter for component.
+	 *
```
DrewAPicture: Missing `@since` version

includes/compat/class-customer.php
```diff
@@ -0,0 +1,170 @@
...
+	 * @return mixed Dependent on the method being dispatched to.
+	 */
+	public function __call( $name, $arguments ) {
+		switch ( $name ) {
+			case 'add':
+			case 'insert':
+				return edd_add_customer( $arguments[0] );
+				break;
+			case 'update':
+				return edd_update_customer( $arguments[0], $arguments[1] );
```
DrewAPicture: I think we should be careful about making assumptions that arguments will be defined here and some of the other cases. Both arguments for `edd_update_customer()` are technically optional, for instance.

tests/tests-customers-db.php
```diff
@@ -141,8 +151,7 @@ public function test_installed() {
 	public function test_get_by() {
 		$customer = edd_get_customer_by( 'email', 'testadmin@domain.com' );
 
-		$this->assertInternalType( 'object', $customer );
-		$this->assertObjectHasAttribute( 'email', $customer );
+		$this->assertGreaterThan( 0, $customer->id );
```
DrewAPicture: Shouldn't we be testing that a specific customer was returned since we're passing a specific email? ;-)

PR #6482, opened 2018-04-06
issue/6477 - build payment icon URL regardless of file location

### LINKED ISSUE #6477: Payment Method Icons Using Filesystem Path Instead of URL
### Expected behavior
Payment Icons to actually be displayed on checkout page.

### Actual behavior
404 error instead of icons

1) Go to the checkout page

PHP Version: not sure

EDD Version (or branch): 2.9

WordPress Version: 4.9.4

The icons used to display perfectly until upgrading to a recent version of EDD.  The code is setting the images src attribute to the absolute path on the server's filesystem.

icons-missing
src-full-path

SDavisMedia:
Fixes #6477 

Proposed Changes:
1. Remove dependency on content directory (the old way) or the plugin directory (added here #6384) when building payment icon absolute paths. 
2. Once path is known and normalized, convert absolute paths to URLs at the `site_url()` level.
3. Add `alt` attribute to image tag to improve handling of broken payment icon paths.

### DIFF
--- a/includes/checkout/template.php
+++ b/includes/checkout/template.php
@@ -680,28 +680,25 @@ function edd_show_payment_icons() {
 
 			} else {
 
-				$image       = edd_locate_template( 'images' . DIRECTORY_SEPARATOR . 'icons' . DIRECTORY_SEPARATOR . $card . '.png', false );
-				$content_dir = WP_PLUGIN_DIR;
+				// Get the absolute path to the payment icon
+				$image = edd_locate_template( 'images' . DIRECTORY_SEPARATOR . 'icons' . DIRECTORY_SEPARATOR . $card . '.png', false );
 
 				if( function_exists( 'wp_normalize_path' ) ) {
 
-					// Replaces backslashes with forward slashes for Windows systems
-					$image = wp_normalize_path( $image );
-					$content_dir = wp_normalize_path( $content_dir );
-
+					// Convert absolute path to URL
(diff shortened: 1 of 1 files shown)


SDavisMedia:
@alpipego Unfortunately, #6384 broke the system that allows users to place their own payment icons in their themes [like so](https://docs.easydigitaldownloads.com/article/1216-moving-edd-templates-to-your-theme). Using the plugin path/URL breaks the string replace that attempts to convert absolute paths to URLs because the string being searched for will now always include the plugins directory. This only happens when you've moved the icons to your own theme, which is why it wasn't noticed before.

In this PR, I've attempted to remove the dependency on content/plugin directory. The full change is on branch `issue/6477`. Would you mind testing to see if you run into any issues? I think I have it all covered but let's be sure.

Thanks!

mfisher31:
@SDavisMedia - I can confirm this is working.  I tested in my local dev environment.  I'll report back if something goes wrong after deploying to the live site.  Thanks again.

alpipego:
@SDavisMedia thanks for pointing out, that I missed that.

Unfortunately, your fix does not work in a non-default setup. By using `ABSPATH`, you're assuming that the plugins folder is beneath the `ABSPATH`. This is not the case in:

+-- web
      +-- wordpress (`ABSPATH`)
      +-- plugins (`WP_PLUGIN_DIR`)
      +-- content (`WP_CONTENT_DIR`)
      |      +-- themes 

      etc.

I'll update my solution, considering the template overrides.

PR #6485, opened 2018-04-06
Support a separate privacy policy on checkout.

### LINKED ISSUE #6484: Add support for a Privacy Policy checkbox
We should add support for a separate Privacy Policy checkbox. This checkbox should be allowed to be independent of the Terms of Agreement checkbox.

It should also allow the support of a single checkbox "Terms of Agreement" but also display the content or a Privacy Policy on checkout.

It should support pulling the content of any created Privacy Policy page, which is being added to WordPress core, so that a store owner does not have to keep to copies active. It can simply pull the `page_for_privacy_policy` option, and use `get_post_field( 'post_content'...` to get the information for it.

### Justification or use case
To allow compliance for GDPR

cklosowski:
Fixes #6484 

Proposed Changes:
1. Add settings for privacy policy agreement, privacy policy agreement text, privacy policy page, and show privacy policy on checkout
2. Includes logic to handle showing 1 checkbox but two agreements/policies, or 2 individually.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -873,6 +873,30 @@ function edd_get_registered_settings() {
 						'type' => 'rich_editor',
 					),
 				),
+				'privacy_policy'     => array(
+					'show_agree_to_privacy_policy' => array(
+						'id'   => 'show_agree_to_privacy_policy',
+						'name' => __( 'Agree to Privacy Policy', 'easy-digital-downloads' ),
+						'desc' => __( 'Check this to show an agree to privacy policy on the checkout that users must agree to before purchasing.', 'easy-digital-downloads' ),
+						'type' => 'checkbox',
+					),
+					'agree_privacy_label' => array(
+						'id'   => 'privacy_agree_label',
+						'name' => __( 'Agree to Privacy Policy Label', 'easy-digital-downloads' ),
+						'desc' => __( 'Label shown next to the agree to privacy policy check box.', 'easy-digital-downloads' ),
+						'type' => 'text',
+						'size' => 'regular',
+					),
+					'agree_privacy_page' => array(
+						'id'   => 'privacy_agree_page',
+						'name' => __( 'Privacy Agreement Page', 'easy-digital-downloads' ),
+						'desc' => __( 'If Agree to Privacy Policy is checked, enter page for the Privacy Agreement Here.', 'easy-digital-downloads' ),
+						'type'        => 'select',
+						'options'     => edd_get_pages(),
+						'chosen'      => true,
+						'placeholder' => __( 'Select a page', 'easy-digital-downloads' ),
+					),
+				),
 			)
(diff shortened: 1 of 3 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -872,6 +880,34 @@ function edd_get_registered_settings() {
...
+					'show_privacy_policy_on_checkout' => array(
+						'id'   => 'show_to_privacy_policy_on_checkout',
+						'name' => __( 'Show the privacy policy on checkout', 'easy-digital-downloads' ),
+						'desc' => __( 'Display your privacy policy on checkout.', 'easy-digital-downloads' ),
+						'type' => 'checkbox',
+					),
+					'agree_privacy_page' => array(
+						'id'   => 'privacy_agree_page',
+						'name' => __( 'Privacy Agreement Page', 'easy-digital-downloads' ),
+						'desc' => __( 'If Agree to Privacy Policy is checked, enter page for the Privacy Agreement Here.', 'easy-digital-downloads' ),
```
pippinsplugins: `enter` should be `select` here.

includes/checkout/template.php
```diff
@@ -770,21 +770,45 @@ function edd_terms_agreement() {
...
+				$agree_page      = edd_get_option( 'privacy_agree_page', get_option( 'page_for_privacy_policy' ) );
+				$agree_label     = edd_get_option( 'privacy_agree_label', __( 'Agree to Terms?', 'easy-digital-downloads' ) );
+				$agreement_text  = get_post_field( 'post_content', $agree_page );
+				?>
+
+				<?php if ( ! empty( $agreement_text ) ) : ?>
+					<div id="edd-privacy-policy" class="edd-terms" style="display:none;">
+						<?php
+						do_action( 'edd_before_privacy_policy' );
+						echo wpautop( stripslashes( $agreement_text ) );
```
pippinsplugins: This should either call `do_shortcode` or should run the `the_content` filter to ensure shortcodes on the page get parsed.

cklosowski: :+1:

includes/admin/settings/register-settings.php
cklosowski: :+1:

includes/checkout/template.php
SDavisMedia: @cklosowski I'm not seeing shortcodes working here. Am I understanding correctly that they are supposed to now?

cklosowski:
@SDavisMedia Yeah sorry I assumed it would just work. Silly me. the problem is that this is populated via AJAX, and AJAX requests have no access to the Shortcodes API. I'll come up with a work around.

cklosowski:
@SDavisMedia all set now. Sorry I had only added `do_shortcode()` to one of the sections, not both.

cklosowski:
@pippinsplugins we good here now?

PR #6488, opened 2018-04-07
Payment icon template overrides

### LINKED ISSUE #6477: Payment Method Icons Using Filesystem Path Instead of URL

alpipego:
Fixes #6477
Supersedes and closes #6482

Proposed Changes:
* first attempts to replace `WP_PLUGIN_DIR` with `WP_PLUGIN_URL`, to get the default payment icons
* then attempts to replace `WP_CONTENT_DIR` with `WP_CONTENT_URL` to account for template overrides
* removes `function_exists` check for `wp_normalize_path` (has been added to core in 3.9)


SDavisMedia:
@alpipego I like it! I actually didn't realize I needed to consider paths completely outside of WordPress. Rookie mistake. Good call on removing `wp_normalize_path()`. I wanted to ditch it but when I saw it was still in the Codex example, I left it.

I've tested all typical scenarios. If it's working with your configuration as well, I'm calling this good to go.

Your turn, @cklosowski!

PR #6517, opened 2018-04-20
Issue/6376 -  add hooks before and after EDD download list output

### LINKED ISSUE #6376: Add action hook above short code download list
Issue #6363 currently add's action hooks before and after the loop, but these hooks appear _within_ the product grid, rather than out with causing problems if you want to add a search box or similar above the download list.

I propose we add a new action hook alongside these before the download list to allow easy third-party integration. This is useful for instances such as my own where I am building AJAX filters and search functionality for EDD that I want to appear above the grid.

I propose we use the following hook name:

```
<?php do_action( 'edd_downloads_list_before', $atts ); ?>
```

I will submit PR for this.

SDavisMedia:
Fixes #6376 

Proposed Changes:
1. Adds a hook pair before and after the output of the `[downloads]` shortcode and `No {plural_label} found` output.

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -600,7 +600,7 @@ function edd_downloads_query( $atts, $content = null ) {
 
 		<div class="edd_downloads_list <?php echo apply_filters( 'edd_downloads_list_wrapper_class', $wrapper_classes, $atts ); ?>">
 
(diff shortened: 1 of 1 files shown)


SDavisMedia:
@cklosowski ^^^

PR #6526, opened 2018-04-26
Remove qualifying selector on purchase form submit button.

### LINKED ISSUE #6525: Remove Qualifying Element in Checkout Javascript
### Expected behavior

When integrating a custom checkout I would like to be able to use a `button` to submit the checkout form.

### Actual behavior

I am forced to use an `input`.

Remove the `input` qualifying element in the Javascript selector.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/edd-ajax.js#L384

Another option would be to hook in to the form's `submit` event rather than watching the button specifically. Though that would be a much larger change with potential side effects.

### Justification or use case

The JS selector for submitting the form is already very specific (with two IDs) and adding the `input` forces that element to be used -- when a `button` is often much more flexible.

spencerfinnell:
Fixes #6525

Proposed Changes:
1. Remove qualifying selector in Javascript and CSS for the checkout form.

### DIFF
--- a/assets/js/edd-ajax.js
+++ b/assets/js/edd-ajax.js
@@ -381,7 +381,7 @@ jQuery(document).ready(function ($) {
 	}
 
 	// Process checkout
-	$(document).on('click', '#edd_purchase_form #edd_purchase_submit input[type=submit]', function(e) {
+	$(document).on('click', '#edd_purchase_form #edd_purchase_submit [type=submit]', function(e) {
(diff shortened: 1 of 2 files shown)


JJJ:
Seems OK to me also.

JJJ:
Merged to `master` and also upstream to `release/3.0`.

PR #6532, opened 2018-04-27
Add String, domain, and email masking functions

### LINKED ISSUE #6502: Add email psudo-masking helper function
## Enhancement Request
For some upcoming changes across EDD and it's add-ons, we need a build in way to anonymize an email address, and we can use psudo-masking to make this possible. This masking should take an email like:
`customer@example.org`

and convert it to:
`c******r@e*****e.org`

### Justification or use case
This is to facilitate some changes upcoming in EDD for [GDPR](#6216 )

cklosowski:
Fixes #6502 

Proposed Changes:
1. Adds 3 new functions `edd_mask_string`, `edd_mask_domain`, and `edd_mask_email`.
2. The mask email function uses the mask domain, which uses the mask string. DRY style.
3. Adds in unit tests for cases I could think of

Note:
The domain parsing supports domains like `example.org`, `example.org.uk`, `test.example.com` and `test.example.org.uk`.

### DIFF
--- a/includes/privacy-functions.php
+++ b/includes/privacy-functions.php
@@ -12,7 +12,22 @@
 // Exit if accessed directly
 if ( ! defined( 'ABSPATH' ) ) exit;
 
+/**
+ * Given a string, mask it with the * character.
+ *
+ * First and last character will remain with the filling characters being changed to *. One Character will
+ * be left in tact as is. Two character strings will has the first character remain and the second be a *.
+ *
+ * @since 2.9.2
+ * @param string $string
+ *
+ * @return string
+ */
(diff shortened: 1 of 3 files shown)


includes/privacy-functions.php
```diff
@@ -0,0 +1,118 @@
...
+
+
+	return $masked_string;
+
+}
+
+/**
+ * Given a domain, mask it with the * character.
+ *
+ * TLD parts will remain in tact (.com, .co.uk, etc). All subdomains will be masked t**t.e*****e.co.uk.
```
pippinsplugins: `in tact` should be `intact`

PR #6533, opened 2018-04-30
issue/6290 - (WIP) - Handle PHP versions lower than 5.3

JJJ:
Work in progress.

See #6290, #6311.

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -31,107 +31,33 @@
 // Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) exit;
 
-if ( ! class_exists( 'Easy_Digital_Downloads' ) ) :
-
 /**
- * Main Easy_Digital_Downloads Class.
+ * The main plugin requirements checker
  *
- * @since 1.4
+ * @since 3.0
  */
-final class Easy_Digital_Downloads {
-	/** Singleton *************************************************************/
+final class EDD_Requirements_Check {
 
 	/**
-	 * @var Easy_Digital_Downloads The one true Easy_Digital_Downloads
-	 * @since 1.4
-	 */
-	private static $instance;
-
-	/**
-	 * EDD Roles Object.
+	 * Plugin file
 	 *
-	 * @var object|EDD_Roles
-	 * @since 1.5
-	 */
-	public $roles;
-
-	/**
-	 * EDD Cart Fees Object.
-	 *
-	 * @var object|EDD_Fees
-	 * @since 1.5
-	 */
-	public $fees;
-
-	/**
-	 * EDD API Object.
-	 *
-	 * @var object|EDD_API
-	 * @since 1.5
-	 */
-	public $api;
-
-	/**
-	 * EDD Utilities Object.
-	 *
-	 * @var object|EDD_Utilities
 	 * @since 3.0
+	 * @var string
 	 */
-	public $utils;
+	private $file = '';
 
 	/**
-	 * EDD HTML Session Object.
+	 * Plugin basename
 	 *
-	 * This holds cart items, purchase sessions, and anything else stored in the session.
-	 *
-	 * @var object|EDD_Session
-	 * @since 1.5
-	 */
-	public $session;
-
-	/**
-	 * EDD HTML Element Helper Object.
-	 *
-	 * @var object|EDD_HTML_Elements
-	 * @since 1.5
-	 */
-	public $html;
-
-	/**
-	 * EDD Emails Object.
-	 *
-	 * @var object|EDD_Emails
-	 * @since 2.1
-	 */
-	public $emails;
-
-	/**
-	 * EDD Email Template Tags Object.
-	 *
-	 * @var object|EDD_Email_Template_Tags
-	 * @since 1.9
-	 */
-	public $email_tags;
-
-	/**
-	 * EDD Cart Object
-	 *
-	 * @var object|EDD_Cart
-	 * @since 2.7
-	 */
-	public $cart;
-
-	/**
-	 * EDD Components array
-	 *
-	 * @var array
 	 * @since 3.0
+	 * @var string
 	 */
-	public $components = array();
+	private $base = '';
 
 	/**
-	 * EDD Requirements array
+	 * Requirements array
 	 *
+	 * @todo Extend WP_Dependencies
 	 * @var array
 	 * @since 3.0
 	 */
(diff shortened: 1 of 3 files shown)


cklosowski:
This appears to be killing the existence of the `Easy_Digital_Downloads` class? That is going to cause some problems as many add-ons, 3rd party and official alike, use a `class_exists( 'Easy_Digital_Downloads' )` check to avoid throwing Fatal Errors when EDD isn't active.

Is that the case @JJJ?

JJJ:
Here's what it looks like when requirements are not met:

screen shot 2018-05-04 at 1 10 24 pm

EDD has requirements. Today, those are minimum versions of PHP and WordPress. Tomorrow, who knows. The purpose of this code is to provide a simple pre-loader to check that all requirements are met before proceeding with the bootstrap process.

This means, in short, _not_ loading any files, functions, classes, constants, or firing off any actions, that add-ons might use to load themselves, as doing so is misleading, and may result in fatal errors.

JJJ:
Related: our add-ons are perpetuating a bad `admin_notice` habit:

screen shot 2018-05-05 at 4 06 21 pm

JJJ:
I've tested 10 add-ons so far, and only found 1 logic this pull request will reveal, in All Access Pass.

Created this issue & PR: https://github.com/easydigitaldownloads/edd-all-access/issues/203

This proves a few theories:

* The approach in this PR is sound
* Add-ons use a wild-west approach to looking for EDD
* Plugins that do weird things without EDD will still do weird things
* Plugins that check for EDD normally will continue to function normally

This PR is OK to merge into `release/3.0` IMO, so I'll likely do so this imminently.

mintplugins:
@jjj Can you clarify the bad admin-notice habit our add-ons have/are perpetuating?

JJJ:
> Can you clarify the bad admin-notice habit our add-ons have/are perpetuating, and why it is a problem?

notices

It's a problem, because everyone is using `admin_notices` to notify users about everything everywhere, which causes users to completely ignore them.

> I'm not sure if you're saying they should be removed entirely or if they should be changed in some way.

If we _can_ remove them, we should. Otherwise, we should generally improve them.

More specifically, in the case of All Access Pass, activating it without having EDD active previously resulted in 2 notices, one of which was erroneously saying that EDD was too old to use, even though it wasn't even active. My PR fixes that error, without modifying the notices themselves.

mintplugins:
For sure - I think we should still have the admin_notices in place, and in the case of All Access, make sure that it shows an "All Access requires Easy Digital Downloads" notice if EDD is not active. 

I don't feel that abuse of admin_notices by certain plugins means that we should avoid notifying users of important or required things. I agree they are annoying, but only when used for the wrong purpose. I feel this type of notice is the right purpose, while [advertising your unrelated blog or sale](https://wordpress.org/support/topic/random-advertisement-in-dashboard/) is the wrong purpose.

mintplugins:
I think we should do both (that looks dope by the way). 

This type of admin notice is technically dismissible, but only if you deactivate the plugin - which IMHO is how it should be for a dependency that isn't active. The user-intention when activating the extension is that the extension works. In my experience with WP support it's surprising how often this type of thing generate(s) support tickets without the admin notice in place.

nosegraze:
I agree with @mintplugins . The UI change on the plugin screen looks very nice, but you'd only see it on the plugins page. It would be very easy to activate the plugin and navigate to a different page without ever seeing a notice. As Phil said, this kind of stuff comes up in support a lot. Admin notices may be annoying if there's a lot, but the benefit is that they will show on all pages and at least then we can say we've done everything we can to communicate messages/error to the user. It's not our fault if the user doesn't see/read them.

JJJ:
Y’all aren’t alone. The Gutenberg team is remaking the same mistakes with their new notices:

https://github.com/WordPress/gutenberg/issues/6340

> It’s not our fault

It is our fault. And it’s WordPress’s fault. Admin notices aren’t for what we are using them for.

They are for “You did it!” type snippets of feedback to users; they aren’t for permanent “Something is wrong” hints.

I get that there are support concerns, so we should invent some new approach that addresses those without using an admin notice.

http://www.safetysignsupplies.co.uk/blog/overcoming-safety-signs-blindness~130.html

We are contributing to both different kinds of sign blindness: cluster, and familiarity.

* Because there are so many, people miss them
* Because they look like notices, people miss them

Not only are these notices ineffective, they actually have the potential to hurt users & support teams further by making users feel stupid for having missed them anyways.

In the real world, people miss signs all the time. Speed limits. Do not enter. Push vs. Pull. One person says “well, we have signs” and another will say “It’s impossible to notice unless you already know it’s there or what it says.”

When signs aren’t properly displayed, the municipality is liable. That means when tree branches grow over a stop sign, if the branches aren’t trimmed and you run that sign because you can’t see it, that isn’t your fault.

Put another way, this is something of an accessibility concern, which is why it’s elusive – folks that don’t see the problem will never see the problem.

I’m not saying my exact approach here is the best one ever, but it’s a decent start. We’ve all trained users to look the wrong direction for the wrong information, so we have a responsibility to fix that error for them and ourselves.

mindctrl:
There is a lot of merit in what everyone has said here.

Having said that, I don't think the notice on the plugin row is anywhere near sufficient to communicate a problem state considering the context in which we operate.

If a site runs 30 plugins, the chances of someone scrolling all the way down and seeing the notice is low. **Especially because WordPress shows an admin notice saying `Plugin activated`.** Why would anything think to scroll all the way down the list to look for an error? They wouldn't, because WordPress reassured them that it was activated with a nice green notification.

If there's a problem a week after activation, because something/anything changed, why would they navigate to the plugins list to look for message? They wouldn't, they'd just assume everything was humming along as it was before. Should they check their plugins list daily, just to be sure? I don't think so.

And having said that, this notice and admin notices are not mutually exclusive. Both are helpful for different reasons in different contexts. Maybe we can use this functionality in multiple ways, to communicate important state information in the most appropriate manner for that instance - whether that's plugin row, admin notice, email, or some new paradigm we haven't yet invented.

In the realm of ecommerce, where peoples' lives literally depend on their sites working, I think we should communicate very clearly in ways that cannot be missed that the site is in a broken state and possibly will not process transactions that put food on their tables. Because we wouldn't want to be the municipality that sets people up for failure by not trimming the trees that block the signs.

JJJ:
@mindctrl Those are other, separate use cases. Important as they may be, we should scrutinize them and decide what the appropriate communication method is. Instead, `admin_notices` is a dumping ground for everything, and I've offered up one solution to one problem here.

---

Separately, my reasoning still stands that traditional non-dismissible notices providing 1 direction for user intervention are not the correct solution. In almost all of these situations, we should just solve the problem automatically for them without asking for permission, especially when the only answer is "yes" or "do the thing." We should just do the thing, or do nothing.

Put another way, making users click links and buttons discretely shifts responsibility off of us and onto them. When something goes wrong, now they think it's their fault, and we don't think it's ours. We're throwing the grenade at them instead of falling on it for them.

I also think people's sites are important and we should communicate clearly, more clearly than we are currently. I can't tell you what exactly that is (I don't think plugins should _need to_ solve this problem independently anyways) but because WordPress core isn't interested in owning this, we've all been forced to.

----

Not to pick on All Access (sorry @mintplugins!) but here are a few reasons why we should have some unified front:

* EDD has never had an official way for add-ons to integrate with it. This means add-ons all do something different to see if EDD is active and up-to-date.
* Checking for `EDD_VERSION` isn't enough, because anyone could set that constant to anything
* Checking for `EDD()` isn't enough, because the function existing isn't an indication that EDD is running correctly, had its migrations run, etc...
* Checking for various classes isn't enough, because if EDD changes the load order or hooks things onto actions, possible race conditions crop up
* We don't make sure `current_user_can( 'install_plugins' );` before showing notices. So, _all users_ see a permanent admin notice with links to install or activate EDD that they can't follow.
* We don't make sure the installation isn't under version control (using `WP_Automatic_Updater::is_vcs_checkout()`.) The link we provide users to install EDD might be something their installation doesn't allow them to do.

----

Right now, as a Subscriber level user on a site with All Access active but without EDD, this is what I see:

screen shot 2018-05-08 at 11 37 17 am

Clicking "activate it!" obviously leads me to:

screen shot 2018-05-08 at 11 44 47 am

----

Hear me out (because I haven't typed this before and it might not make any sense) but my final reason that I feel `admin_notices` are the wrong place/way to notify users is because notices are for the system to communicate to the user, but we've completely stopped considering who the user actually is, what they're trying to do, or what they're even capable of.

So, my suggested solution isn't about `admin_notices` being right or wrong, it's that we all have been universally across the board using that approach incorrectly, and we will not break that decade-old habit without introducing a new approach with the proper checks in place already for us.

mintplugins:
@JJJ I definitely agree we need to have a standardized way for extensions to check for the existence of EDD core. And you're also correct that we should be checking user permissions before outputting a message into admin_notices. So I agree with those portions. 

The only part I still disagree with is that removing the notice for admins in admin_notices is a good idea. Admins need to be notified of this on _every_ page in order to avoid angry customers and angry tickets. If not through admin_notices, than in some other way.

pippinsplugins:
I think we can all agree that `admin_notices` suck. They've been used so much because they're easy and, frankly, they're one of the only seemingly/possibly effective options available to us.

If the proposal is to remove admin notices, which I do support at a fundamental level, I'd like to see what the coordinated proposal is for a replacement. As @mindctrl, @mintplugins, and @nosegraze have voiced, displaying an inline notice on the plugins, while _awesome_, is not sufficient.

@JJJ do you have a proposal for an `admin_notices` replacement that compliments the inline notice you've added?

JJJ:
@pippinsplugins I have an idea, but I'm not sure if it's a good one :)

For some context, when the "shiny updates" feature was rolled into core a few releases ago, inline notices were used to indicate that a plugin needs updating.

screen shot 2018-05-08 at 2 38 56 pm

* Clicking "Update Now" triggers an ajax call that updates a plugin in the background
* Clicking "Details" opens the ThickBox for the plugin from WordPress.org.

I think what actually makes this feature ultimately work, is the notification bubble on the "Plugins" admin menu, coupled with the admin-bar "1" that links to the global "Updates" page.

----

In my imagination, we'd add some kind of "Needs Attention" view/status to the Plugins page:

screen shot 2018-05-08 at 2 41 34 pm

And also an `awaiting-mod` bubble on the "Plugins" menu:

screen shot 2018-05-08 at 2 45 02 pm

And then _maybe_ a single admin notice (hidden behind the appropriate cap checks and what-not) with something like:

screen shot 2018-05-08 at 2 47 45 pm

----

I'd bake this into the `EDD_Requirements_Check` class, and we can rename & reuse that class in our add-ons with new strings for the new plugin.

Also... open to feedback, scrutiny, and suggestions. 🧡

pippinsplugins:
I like all of that except for one thing: the single admin notice is generic and so is even easier to skip or gloss over than the notices we already display. At least the existing notices are informative and specifically topical.

That might be able to be addressed by making the notice more clear it's about EDD pluginsn.

JJJ:
I'm not a fan of that part either. How about:

screen shot 2018-05-08 at 3 59 08 pm

mintplugins:
I agree with Pippin on all counts - I love it - and I wish I could see it working for our users. But the dashboard ignorance is pretty rampant out there. I myself am guilty of it - where I've been testing our extensions and it took me longer than it should have to realize I didn't have EDD core activated, because there was no admin notice. For people not well-versed in what EDD adds to the WP dashboard, it would take them even longer, and has resulted in tickets.

Issue #314, opened 2012-09-09
jQuery UI Style


shazahm1:
I see that jQuery UI is being used for the date picker... The default styles look out of place, no big deal, really. However, this does cause a weird display issue when you click the link button in TinyMCE on the "Email" settings admin page. Granted, this is not something one will do often but there is an easy "fix". Someone has already put in the effort to create a style that matches the WP admin and it also happens to be hosted on github, here, https://github.com/helenhousandi/wp-admin-jquery-ui

JJJ:
Should have worked on this on Pi Day.

I have some CSS that will address this, so I'll put together a PR for it.

JJJ:
<img width="790" alt="screen shot 2018-05-08 at 6 11 58 pm" src="https://user-images.githubusercontent.com/88951/39787525-631a6efa-52eb-11e8-8172-be3e429ff25b.png">
<img width="810" alt="screen shot 2018-05-08 at 6 12 16 pm" src="https://user-images.githubusercontent.com/88951/39787526-632b83a2-52eb-11e8-9fb2-7b1e83e588cb.png">

pippinsplugins:
@jjj would it be possible to update the CSS to include a wrapper class so that only our datepickers are affected? We've had compatibility problems with other plugins before when two or more plugins are defining UI styles.

JJJ:
> would it be possible to update the CSS to include a wrapper class so that only our datepickers are affected

I _think_ this is achievable with a bit of extra JavaScript and the `beforeShow` function:
http://api.jqueryui.com/datepicker/#option-beforeShow

I'll keep digging and report back.

JJJ:
Conflicts averted by removing the `ui-datepicker` class and adding `edd-datepicker` instead.

Now we can target our own datepickers without colliding with others.

JJJ:
Merged this into `release/3.0`.

Thanks everyone for being patient with this one.

PR #6551, opened 2018-05-08
Add jQuery UI Datepicker styling.

JJJ:
See #314.

Proposed Changes:
1. Add `edd-admin-datepicker.css`
2. Enqueue it only on EDD admin pages (to avoid CSS conflicts)
3. Enjoy nicer looking jQuery UI date pickers


JJJ:
This is done, and the results are pretty great, so I'm going to merge it imminently.

PR #6552, opened 2018-05-08
Remove PII and switch file download logs to use customer ID #6548

### LINKED ISSUE #6548: Improve File download Log accuracy and remove PII from log meta
Currently the file download log uses a meta key of `_edd_log_user_info` to store information like 
`a:3:{s:5:"email";s:15:"admin@local.dev";s:2:"id";i:1;s:4:"name";s:15:"Chris Klosowski";}`. This information is pulled from the customer associated with the purchase record.

While used to do some download count checks, we can be far more accurate, as well as remove PII from these log entry meta values by simply storing the customer ID. Since a customer can have more than one email address, and also set their primary email address from their profile editor, this could end up being in accurate when we do a download count using this function:
`edd_count_file_downloads_of_user( $customer_obj->email );`

My proposal will be to do the following:
1) Remove the `_edd_log_user_info` for file download logs _only_, with a migration routine.
2) Add in a new meta entry for file download logs of `_edd_log_customer_id`, with the above migration routine.
3) Add a new function of `edd_count_file_downloads_of_customer` which accepts a customer Id or email (since the customer object can find a customer with any email address associated with the customer)
4) Change any calls to get this count to the new method.

### Justification or use case
I realized this was possibly an issue while researching the implementation of #6499. This data isn't necessary and can be looked up elsewhere.

cklosowski:
Fixes #6548

Proposed Changes:
1. Adds a migration routine to kill the `_edd_log_user_info` for the file download logs and add `_edd_log_customer_id` for the file download logs
2. Adds `edd_count_file_downloads_of_customer` to return the file download count of a customer record.
3. Updates `edd_count_file_downloads_of_user` to switch to the `edd_count_file_downloads_of_customer` function if the string is an email (this is what it was doing previously anyway, now it's explicit)
4. Updates the list table to use the proper customer ID instead of always performing searches on users (users mean nothing to EDD for the most part, customer should be our canonical).
5. Updates the file download log list table to search customers, not users (as a canonical source)
6. Updates the Customers API endpoint to send the customer ID, directly.

Overall, this is far more accurate to count a customer's download count. The previous method was doing a `LIKE` search on the user info, but if a customer changed their primary email address, the logs would not be updated, so the search would be in accurate. By converting all searches for customer download counts to use the `$customer->id` the email address is irrelevant.

By removing the log user info for this type of log, we're removing PII which makes us easier to comply with GDPR.

### DIFF
--- a/includes/admin/upgrades/classes/class-file-download-log-migration.php
+++ b/includes/admin/upgrades/classes/class-file-download-log-migration.php
@@ -40,7 +40,7 @@ class EDD_File_Download_Log_Migration extends EDD_Batch_Export {
 	 * @since  2.9.2
 	 * @var integer
 	 */
-	public $per_step = 25;
+	public $per_step = 50;
 
 	/**
 	 * Get the Export Data
@@ -73,6 +73,11 @@ class EDD_File_Download_Log_Migration extends EDD_Batch_Export {
 
 				if ( $sanitized_log_id !== $log_id ) {
 					edd_debug_log( "Log ID mismatch, skipping log ID {$log_id}" );
+					continue;
+				}
+
+				$has_customer_id = get_post_meta( $log_id, '_edd_log_customer_id', true );
+				if ( ! empty( $has_customer_id ) ) {
 					continue;
 				}
 
--- a/includes/admin/upgrades/classes/class-file-download-log-migration.php
+++ b/includes/admin/upgrades/classes/class-file-download-log-migration.php
@@ -1,0 +1,286 @@
+<?php
+/**
+ * Migrate File Download Logs
+ *
+ * Removes some PII from the log meta, and adds the customer ID of the payment to allow more accurate
+ * file download counts for a customer.
+ *
+ * @subpackage  Admin/Classes/EDD_SL_License_Log_Migration
+ * @copyright   Copyright (c) 2018, Easy Digital Downloads, LLC
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       2.6.2
+ */
+
+// Exit if accessed directly
+if ( ! defined( 'ABSPATH' ) ) exit;
+
+/**
+ * EDD_File_Download_Log_Migration Class
+ *
+ * @since 2.9.2
+ */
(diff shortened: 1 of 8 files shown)


includes/admin/upgrades/classes/class-file-download-log-migration.php
```diff
@@ -0,0 +1,286 @@
...
+
+		if ( $this->step == 1 ) {
+			$this->delete_data( 'edd_file_download_log_ids' );
+		}
+
+		// Get any file download logs that have an entry for the user info
+		$file_download_log_ids = get_option( 'edd_file_download_log_ids', false );
+		if ( false === $file_download_log_ids ) {
+
+			$log_query = array(
```
pippinsplugins: I doubt this will be successful on really big sites that have hundreds of thousands of file download log entries. Can we move the query logic into the step processor?

cklosowski: @pippinsplugins Possibly, I'll have to play around with it to make sure the pagination still works.

pippinsplugins: If unable to make pagination work, let's at minimum test this on sites with x00,000 download logs.

SDavisMedia:
Will save my test/review for after Pippin's suggested changes are implemented.

cklosowski:
@SDavisMedia @pippinsplugins I'm going to bet something into `master` that is a WP CLI command that will generate file download logs for us, so we can stress test this in a local environment without taking forever to generate them.

cklosowski:
@SDavisMedia Ok, I've got a tool for `master` to generate a bunch of download logs in the existing format. If you don't mind could you test out this PR to make sure you're able to generate a ton of download logs? https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6558

cklosowski:
@pippinsplugins So, there is still an issue that we need to have a total count to estimate the length of the migration.

But I found a way to store all this data on prefetch. And with 101,254 log entires, I can get all the log IDs in `3.5-5.0ms` locally. I think that's a reasonable query time and size.

cklosowski:
this is running on the EDD staging site. I'm going to let it complete, audit it, and once it passes, merge.

Issue #6513, opened 2018-04-19
Move applying the WP offset out of the EDD\Utils\Date constructor


DrewAPicture:
The `EDD\Utils\Date` object (extending `Carbon` and therefore `DateTime`) is being introduced in 3.0. Currently, it's automagical in that it assumes whatever date/time being passed in is UTC and needs "conversion" to "WP time". The way this is facilitated is by instantiating the object with the WP timezone then manually applying the offset via passing a `DateInterval` object via `DateTime->add()`.

The main problem with this is that this effectively means that any incoming time with the offset already applied will have it applied twice. This has already come up in one place, which necessitated the adding of an optional `$apply_offset` parameter to the `EDD\Utils\date()` helper, which doesn't really solve the problem, it just reverses the manual `DateInterval` change after the fact.

This issue serves to directly address the problem in `Date` instead of bolting something on in the `EDD\Utils\date()` helper.

Date functionality was originally introduced via #6236, but later moved out to the `release/3.0` branch for visibility by other feature projects.

DrewAPicture:
By virtue of inheritance rules with constructors in PHP, we can't just add the `$apply_offset` method to the constructor as a third optional parameter – the parameters have to match upstream with `Carbon` and `DateTime`. So I'm thinking the best option to accomplish this would be to move applying the offset to a separate method and then instantiate in `EDD\Utils\date()` via a static method that accepts its own arguments. This new method would be considered the "best practice" for instantiating an `EDD\Utils\Date` object.

sunnyratilal:
Closing as @JJJ, @mindctrl and I have done a lot of work over the past few days to fix this.

PR #6555, opened 2018-05-10
(WIP) issue/6513 - Date/Time/Format/Zone improvements:

JJJ:
Work in Progress.

See #6513.

### DIFF
--- a/includes/admin/discounts/add-discount.php
+++ b/includes/admin/discounts/add-discount.php
@@ -141,8 +141,8 @@ defined( 'ABSPATH' ) || exit;
 					<label for="edd-start"><?php _e( 'Start date', 'easy-digital-downloads' ); ?></label>
 				</th>
 				<td>
-					<input name="start_date" id="edd-start" type="text" value="" class="edd_datepicker"/>
-					<p class="description"><?php _e( 'Enter the start date for this discount code in the format of mm/dd/yyyy. For no start date, leave blank. If entered, the discount can only be used after or on this date.', 'easy-digital-downloads' ); ?></p>
+					<input name="start_date" id="edd-start" type="text" value="" class="edd_datepicker" data-format="yyyy-mm-dd" placeholder="yyyy-mm-dd" />
+					<p class="description"><?php _e( 'Enter the start date for this discount code in the format of yyyy-mm-dd. For no start date, leave blank. If entered, the discount can only be used after or on this date.', 'easy-digital-downloads' ); ?></p>
 				</td>
 			</tr>
 
@@ -153,8 +153,8 @@ defined( 'ABSPATH' ) || exit;
 					<label for="edd-expiration"><?php _e( 'Expiration date', 'easy-digital-downloads' ); ?></label>
 				</th>
 				<td>
-					<input name="end_date" id="edd-expiration" type="text" class="edd_datepicker"/>
-					<p class="description"><?php _e( 'Enter the expiration date for this discount code in the format of mm/dd/yyyy. For no expiration, leave blank.', 'easy-digital-downloads' ); ?></p>
+					<input name="end_date" id="edd-expiration" type="text" class="edd_datepicker" data-format="yyyy-mm-dd" placeholder="yyyy-mm-dd" />
+					<p class="description"><?php _e( 'Enter the expiration date for this discount code in the format of yyyy-mm-dd. For no expiration, leave blank.', 'easy-digital-downloads' ); ?></p>
 				</td>
 			</tr>
 
--- a/includes/admin/discounts/add-discount.php
+++ b/includes/admin/discounts/add-discount.php
@@ -141,7 +141,7 @@ defined( 'ABSPATH' ) || exit;
 					<label for="edd-start"><?php _e( 'Start date', 'easy-digital-downloads' ); ?></label>
 				</th>
 				<td>
-					<input name="start_date" id="edd-start" type="text" value="" class="edd_datepicker" data-format="yyyy-mm-dd" placeholder="yyyy-mm-dd" />
+					<input name="start_date" id="edd-start" type="text" value="" class="edd_datepicker" data-format="<?php echo edd_get_date_picker_format(); ?>" placeholder="<?php echo edd_get_date_picker_format(); ?>" />
 					<p class="description"><?php _e( 'Enter the start date for this discount code in the format of yyyy-mm-dd. For no start date, leave blank. If entered, the discount can only be used after or on this date.', 'easy-digital-downloads' ); ?></p>
 				</td>
 			</tr>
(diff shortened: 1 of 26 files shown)


includes/admin/discounts/add-discount.php
```diff
@@ -141,8 +141,8 @@
 					<label for="edd-start"><?php _e( 'Start date', 'easy-digital-downloads' ); ?></label>
 				</th>
 				<td>
-					<input name="start_date" id="edd-start" type="text" value="" class="edd_datepicker"/>
-					<p class="description"><?php _e( 'Enter the start date for this discount code in the format of mm/dd/yyyy. For no start date, leave blank. If entered, the discount can only be used after or on this date.', 'easy-digital-downloads' ); ?></p>
+					<input name="start_date" id="edd-start" type="text" value="" class="edd_datepicker" data-format="<?php echo edd_get_date_picker_format(); ?>" placeholder="<?php echo edd_get_date_picker_format(); ?>" />
```
DrewAPicture: Escape?

includes/admin/payments/view-order-details.php
```diff
@@ -105,7 +105,7 @@
 										<div class="edd-admin-box-inside">
 											<p>
 												<span class="label"><?php _e( 'Date:', 'easy-digital-downloads' ); ?></span>&nbsp;
-												<input type="text" name="edd-payment-date" value="<?php echo esc_attr( date( 'm/d/Y', $payment_date ) ); ?>" class="medium-text edd_datepicker"/>
+												<input type="text" name="edd-payment-date" value="<?php echo esc_attr( date( 'Y-m-d', $payment_date ) ); ?>" class="medium-text edd_datepicker" placeholder="<?php echo edd_get_date_picker_format(); ?>" />
```
DrewAPicture: Why the change in entered format? The datepicker reformats to Y-m-d on `$_POST` anyway, doesn't it?

includes/admin/reporting/graphing.php
```diff
@@ -667,11 +667,19 @@ function edd_reports_graph_controls() {
...
+								'name'        => 'filter_from',
+								'placeholder' => 'yyyy-mm-dd',
+								'value'       => $from
+							) ); ?>
 
 							<span><?php _ex( 'To', 'date filter', 'easy-digital-downloads' ); ?></span>
-							<?php echo EDD()->html->date_field( array( 'name' => 'filter_to', 'placeholder' => __( 'mm/dd/yyyy', 'easy-digital-downloads' ), 'value' => $to ) ); ?>
+							<?php echo EDD()->html->date_field( array(
+								'name'        => 'filter_to',
+								'placeholder' => 'yyyy-mm-dd',
```
DrewAPicture: Same question on the entered format. I think it was previously mm/dd/yyyy because that's "more familiar".

includes/class-edd-utilities.php
```diff
@@ -126,44 +134,65 @@ public function get_registry( $name ) {
...
 
-		switch( $format ) {
-			case 'time':
-				$format = $this->get_time_format();
+		// What known format are we getting?
+		switch ( $format ) {
+
+			// jQuery UI Datepicker fields, placeholders, etc...
+			case 'date-attribute' :
+				$retval = 'yyyy-mm-dd';
```
DrewAPicture: Purely a style question, but ... why the variable name change from something self-documenting (`$format`) to something that's indicative only of being a return value?

includes/date-functions.php
```diff
@@ -80,6 +55,48 @@ function edd_get_timezone() {
  *
  * @return string date_format()-compatible date format string.
  */
-function edd_get_date_format( $format ) {
+function edd_get_date_format( $format = 'date' ) {
```
DrewAPicture: Documentation for `$format`?

includes/date-functions.php
```diff
@@ -80,6 +55,48 @@ function edd_get_timezone() {
...
+ */
+function edd_get_date_picker_format( $context = 'display' ) {
+
+	// What is the context that we are getting the picker format for?
+	switch ( $context ) {
+
+		// jQuery UI Datepicker does its own thing
+		case 'js' :
+		case 'javascript' :
+			$retval = EDD()->utils->get_date_format_string( 'date-js' );
```
DrewAPicture: Same question on `$retval` vs something self-documenting like `$format`.

includes/date-functions.php
```diff
@@ -80,6 +55,48 @@ function edd_get_timezone() {
...
+
+		// Used to display in an attribute, placeholder, etc...
+		case 'display' :
+		default :
+			$retval = EDD()->utils->get_date_format_string( 'date-attribute' );
+			break;
+	}
+
+	// Filter & return
+	return apply_filters( 'edd_get_date_picker_format', $retval, $context );
```
DrewAPicture: Missing hook doc.

includes/emails/class-edd-email-tags.php
```diff
@@ -833,11 +833,9 @@ function edd_email_tag_payment_method( $payment_id ) {
  * Email template tag: sitename
  * Your site name
  *
- * @param int $payment_id
- *
  * @return string sitename
  */
-function edd_email_tag_sitename( $payment_id ) {
```
DrewAPicture: If we're removing a parameter, we need to properly deprecate it.

JJJ: This parameter should never have been here. It never did anything, and was never used.

includes/date-functions.php
JJJ: Because it’s a return value based on the context.

Naming it something like `$retval` informs developers that that’s the variable that will inevitably be returned by the function.

Naming it `$format` does not indicate its the return value. It also possibly collides with other local variables or parameters.

`$retval` is like a reserved keyword, IMO. Use it only to define the variable used to filter and return something.

includes/admin/reporting/graphing.php
JJJ: The jQuery UI Datepicker got changed to `yy-mm-dd` so it no longer matches the format or pattern.

I did forget to use the new function here though, so I’ll note that.

includes/admin/payments/view-order-details.php
JJJ: This is so that the date matches the placeholder text and the returned format of jQuery UI Datepicker.

includes/admin/discounts/add-discount.php
JJJ: Good call.

I’ll probably end up with a new function to handle the escape and echo.

includes/date-functions.php
JJJ: Added!

includes/date-functions.php
JJJ: Done!

includes/admin/reporting/graphing.php
JJJ: Updated to use `edd_get_date_picker_format()` instead of `yyyy-mm-dd` or any other hard-coded string.

includes/admin/discounts/add-discount.php
JJJ: Done!

JJJ:
These changes here are not very huge, and the code itself is OK and purposeful, so I just merged it.

Can continue to iterate on this if/as needed.

Issue #6560, opened 2018-05-14
Another way to check if EDD uploads is public


JJJ:
## Bug Report

While working on `EDD_Notices` for 3.0 I noticed (ha!) that the `uploads/edd` directory check is different between Nginx and Apache. This is because Apache allows for `.htaccess` and Nginx requires rule changes. What was odd to me is once the notice is dismissed, it never checks for this again.

If the directory becomes public later, or the web server is changed from Apache to Nginx, the shop admin may never know to lock it back down using whatever method is required.

### Expected behavior

I guess I expected this to be hooked to a transient, or for the notice to be more directly linked to the problem it's calling out.

### Actual behavior

Directory is checked 1 time on activation. When the notice is dismissed, it never reappears.

### Steps to reproduce the behavior

1) Install and activate EDD for the first time
2) See the Apache or Nginx notice about `uploads/edd`
3) Dismiss the notice
4) The notice does not care what the state of the directory is

### Information (if a specific version is affected):

PHP Version: any modern
EDD Version (or branch): any modern
WordPress Version: any modern

### Justification or use case

It's easy to see this as an enhancement, but because the point is to hide the original items that are for sale away from public viewers, I think it's potentially catastrophic if this directory unknowingly becomes public.

PR #6561, opened 2018-05-15
Refactor `EDD_Notices` (and `EDD_Tracking` subsequently)

JJJ:
This change refactors `EDD_Notices` to work as a first-class object of the EDD singleton, exposing a new `add_notice()` method for adding EDD specific notices.

It also allows for additional logic when working with notices, specifically around dismissibles and classes.

See #6560 (for `release/3.0` but also `issue/6560` branch)


includes/admin/class-edd-notices.php
```diff
@@ -19,242 +19,557 @@
...
+	 */
+	private function add_data_notices() {
+
+		// Recount earnings
+		if ( class_exists( 'EDD_Recount_Earnings' ) ) {
+			$this->add_notice( array(
+				'id'             => 'edd-recount-earnings',
+				'class'          => 'error',
+				'is_dismissible' => false,
+				'message'        => sprintf( __( 'Easy Digital Downloads 2.5 contains a <a href="%s">built in recount tool</a>. Please <a href="%s">deactivate the Easy Digital Downloads - Recount Earnings plugin</a>', 'easy-digital-downloads' ), admin_url( 'edit.php?post_type=download&page=edd-tools&tab=general' ), admin_url( 'plugins.php' ) )
```
cklosowski: While we are here, let's update this to read "Since Easy Digital Downloads 2.5, a built in recount tool..."

includes/admin/class-edd-notices.php
```diff
@@ -19,242 +19,557 @@
...
+		// Dismiss notice
+		edd_update_option( $notice, 1 );
+		wp_redirect( remove_query_arg( array( 'edd_action', 'edd_notice' ) ) );
+		exit;
+	}
+
+	/**
+	 * Output all admin area notices
+	 *
+	 * @since 2.6.0 bbPress (r6771)
```
cklosowski: Looks like EDD has a case of the BBs ;)

includes/admin/class-edd-notices.php
```diff
@@ -19,242 +19,557 @@
...
+		}
 	}
 
 	/**
-	 * Dismiss admin notices when Dismiss links are clicked
+	 * Escape message string output
 	 *
-	 * @since 2.3
-	 * @return void
+	 * @since 2.6.0 bbPress (r6775)
```
cklosowski: Update this `@since`, and the `bbPress`

includes/admin/class-edd-notices.php
cklosowski: Also the `@since` needs updating

includes/admin/class-edd-notices.php
JJJ: Can you recommend the exact text? I can't make sense of how to combine what's here with your suggestion.

PR #6565, opened 2018-05-16
Anonymize customer records when user delete requests are executed #6498

### LINKED ISSUE #6498: Add helper functions for anonymizing customer data
## Enhancement Request
EDD needs some helper functions centered around customer data to facilitate anonymizing the unnecessary data.

### Justification or use case
This is a [GDPR](#6216) issue. While core hasn't finished their tools yet, we should start building our helper functions and making sure things don't break when used, so that when core is ready we have working functions/methods.

cklosowski:
Fixes #6498

Proposed Changes:
1. Adds `edd_anonymize_email`, which hashes an email address with `<hash>@site.invalid`
2. Adds `edd_anonymize_customer`, which does the following:
```
- The name is changed to 'Anonymized Customer'
- The email address is anonymized, but kept in a format that passes is_email checks
- The date created is set to the timestamp of 0 (January 1, 1970)
- Notes are fully cleared
- Any additional email addresses are removed
```
3. Adds the `edd_register_privacy_erasers` function to help us register our erasers/anonymizers
4. Adds `edd_privacy_customer_eraser` which executes on the eraser/anonymizer

### DIFF
--- a/includes/privacy-functions.php
+++ b/includes/privacy-functions.php
@@ -118,6 +118,38 @@ function edd_pseudo_mask_email( $email_address ) {
 }
 
 /**
+ * Return an anonymized email address.
+ *
+ * While WP Core supports anonymizing email addresses with the wp_privacy_anonymize_data function,
+ * it turns every email address into deleted@site.invalid, which does not work when some purchase/customer records
+ * are still needed for legal and regulatory reasons.
+ *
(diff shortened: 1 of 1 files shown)


cklosowski:
@pippinsplugins when you get a chance, take a look over this. I want two sets of eyes on it prior to merging, so I can move forward with the rest of these erasers.

PR #6568, opened 2018-05-17
Add payment anonymizer/eraser that integrates with WP Core

### LINKED ISSUE #6497: Add Helper functions for anonymizing Order (Payment) data
## Enhancement Request
EDD needs some helper functions centered around order (payment) data to facilitate anonymizing the unnecessary data, while keeping in mind that stores may need to retain certain data to remain compliant with local tax and financial laws.

### Justification or use case
This is a [GDPR](#6216 ) issue. While core hasn't finished their tools yet, we should start building our helper functions and making sure things don't break when used, so that when core is ready we have working functions/methods.

cklosowski:
Fixes #6497

Proposed Changes:
1. Adds support for payment records to be anonymized or deleted.
2. Adds `_edd_anonymize_payment` to handle anonymizing or deleting a payment record
3. Modified `edd_anonymize_customer` to `_edd_anonymize_customer` to signify it's not a 'normal' function.

There is a filter in place associated with the status of a payment. By default these are the actions taken:
```
case 'publish':
case 'refunded':
case 'revoked':
	$action = 'anonymize';
	break;

case 'failed':
case 'abandoned':
	$action = 'delete';
	break;

case 'pending':
case 'processing':
default:
	$action = 'none';
	break;
```
TL;DR - Published, refunded, or revoked get anonymized, failed and abandoned get deleted, and pending or processing have no action taken as they are still possibly needed. No action is also the default.

The filter `edd_privacy_payment_status_action_' . $action` exists so that extensions and developers with custom statuses can choose what to do with their status (thinking recurring and Stripe off hand).

The data that _is_ anonymized when that action is taken is:
```
- First Name is made blank
- Last  Name is made blank
- All email addresses are converted to the anonymized email address on the customer
- The IP address is run to only be the /24 IP Address (ending in .0) so it cannot be traced back to a user
- Address line 1 is made blank
- Address line 2 is made blank
```

### DIFF
--- a/includes/privacy-functions.php
+++ b/includes/privacy-functions.php
@@ -198,7 +198,12 @@ function _edd_anonymize_customer( $customer_id = 0 ) {
 
 	// Now we should look at payments this customer has associated, and if there are an payments that should not be modified,
 	// do not modify the customer.
-	$payments = edd_get_payments( array( 'post__in' => implode( ',', $customer->payment_ids ), 'output' => 'payments' ) );
+	$payments = edd_get_payments( array(
+		'customer' => $customer->id,
+		'output'   => 'payments',
+		'number'   => -1,
+	) );
+
 	foreach ( $payments as $payment ) {
 		$action = _edd_privacy_get_payment_action( $payment );
 		if ( 'none' === $action ) {
--- a/includes/privacy-functions.php
+++ b/includes/privacy-functions.php
@@ -118,6 +118,38 @@ function edd_pseudo_mask_email( $email_address ) {
 }
 
 /**
+ * Return an anonymized email address.
+ *
+ * While WP Core supports anonymizing email addresses with the wp_privacy_anonymize_data function,
+ * it turns every email address into deleted@site.invalid, which does not work when some purchase/customer records
+ * are still needed for legal and regulatory reasons.
+ *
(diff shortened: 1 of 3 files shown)


includes/privacy-functions.php
```diff
@@ -117,6 +119,266 @@ function edd_pseudo_mask_email( $email_address ) {
...
+	}
+
+	if ( $customer->user_id > 0 ) {
+		delete_user_meta( $customer->user_id, '_edd_user_address' );
+	}
+
+	$customer->update( array(
+		'name'         => __( 'Anonymized Customer', 'easy-digital-downloads' ),
+		'email'        => edd_anonymize_email( $customer->email ),
+		'date_created' => date( 'Y-m-d H:i:s', 0 ),
```
pippinsplugins: Why do we need to remove the date?

includes/privacy-functions.php
```diff
@@ -196,6 +196,24 @@ function edd_anonymize_customer( $customer_id = 0 ) {
 		return array( 'success' => false, 'message' => $should_anonymize_customer['message'] );
 	}
 
+	// Now we should look at payments this customer has associated, and if there are an payments that should not be modified,
```
SDavisMedia: `if there are an payments` -> `if there are any payments`

PR #6570, opened 2018-05-18
Use WP core’s privacy policy page setting

### LINKED ISSUE #6569: Make Privacy Agreement page use the WP core Privacy Policy page ID if it is set
## Enhancement Request
WP Core released a site wide option called `wp_page_for_privacy_policy` where people can identify the page they use for their Privacy Policy. If that has been chosen, we could automatically make it use that for the Privacy Policy link in the EDD settings under "Settings" > "Misc" > "Terms and Conditions" > "Privacy Agreement Page". 

### Justification or use case
Makes the Privacy Policy page experience more uniform and less prone to mistakes. Also, one less thing for the user to make a decision about.

mintplugins:
Fixes #6569

Proposed Changes:
1. Replace the edd setting for `page_for_privacy_policy` and use the WP core one instead.


SDavisMedia:
@mintplugins I personally like this. While we're already in these settings, let's squeeze some uniformity in there with the "privacy policy" text.

78098721517ab1f6bf0d065a865b42f8 _image 2018-05-19 at 5 29 03 pm

`privacy policy` -> `Privacy Policy`

And for the new link, let's add a period after the sentence since it's a full sentence. If that feels weird, the entire sentence doesn't have to be linked.

PR #6578, opened 2018-05-20
Pass download_id to download_history_files action

CryoutCreations:
The 'edd_download_history_files' action call passes the (probably leftover?) $id variable holding the current page id instead of a (more desirable) download_id.


JJJ:
Good spot!

Let's merge this in right away.

Thanks @CryoutCreations for the PR.

JJJ:
(Also merged this into `release/3.0`.)

PR #6593, opened 2018-05-24
Fixes a few items of feedback from 3rd party developers I had review the GDPR release.

cklosowski:
Proposed Changes:
1. Removes the `Privacy` sub-section from the `Misc` tab after we moved `Privacy` to its own tab.
2. Removes the `Terms and Privacy Agreements` header in the `Misc -> Terms of Agreement` tab.
3. Sets the `update_file_download_log_data` upgrade as complete on a fresh install.
4. Adds some missing filters and incorrectly documented filters.

### DIFF
--- a/includes/privacy-functions.php
+++ b/includes/privacy-functions.php
@@ -268,7 +268,7 @@ function _edd_anonymize_customer( $customer_id = 0 ) {
 	 */
 	$should_anonymize_customer = apply_filters( 'edd_should_anonymize_customer', array( 'should_anonymize' => true, 'message' => '' ), $customer );
 
-	if ( ! $should_anonymize_customer['should_anonymize'] ) {
+	if ( empty( $should_anonymize_customer['should_anonymize'] ) ) {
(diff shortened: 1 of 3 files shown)


includes/privacy-functions.php
```diff
@@ -362,7 +362,7 @@ function _edd_anonymize_payment( $payment_id = 0 ) {
 	 *     @type string $message          A message to display if the customer could not be anonymized.
 	 * }
 	 */
-	$should_anonymize_payment = apply_filters( 'edd_should_anonymize_payment', array( 'should_anonymize' => true, 'message' => array() ), $payment );
+	$should_anonymize_payment = apply_filters( 'edd_should_anonymize_payment', array( 'should_anonymize' => true, 'message' => '' ), $payment );
 
 	if ( ! $should_anonymize_payment['should_anonymize'] ) {
```
JJJ: Since this variable is now filterable, this should use `empty()` instead of `!` to avoid debug notices.

cklosowski: Good call.

PR #6597, opened 2018-05-25
Be more strignent with policy checkbox settings and variable naming #6596

### LINKED ISSUE #6596: Privacy Policy checkout error when Privacy Policy isn't enabled
## Bug Report

Customers receive checkout error for not agreeing to Privacy Policy, even when Privacy Policy is disabled.

### Expected behavior
If Privacy Policy is disabled, it should not be required at checkout.

### Actual behavior
Privacy Policy is required at checkout no matter what, regardless of its settings or any other settings (like Agree to Terms, which it has had conflicts with in the past).

1) Disable the Privacy Policy (all settings).
2. Attempt to complete checkout.

PHP Version: N/A

EDD Version (or branch): 2.9.2

WordPress Version: 4.9.6

cklosowski:
Fixes #6596

Proposed Changes:
1. Forces all checks for terms and privacy policy checkboxes to be explicitly strict type checked.
2. Updates the variable naming to make it more clear what is going on during the process ( as it was hard to debug due to naming conventions)
3. Fixes an issue with logging the privacy policy time on the customer meta, due to an incorrect array index
4. Fixes an issue causing some timeouts when no privacy policy page is selected in the Settings > Privacy of WP Core yet EDD is being told to show the privacy policy on checkout.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -892,7 +892,7 @@ function edd_get_registered_settings() {
 						'size' => 'regular',
 					),
 					'show_privacy_policy_on_checkout' => array(
-						'id'   => 'show_agree_to_privacy_policy_on_checkout',
+						'id'   => 'show_privacy_policy_on_checkout',
 						'name' => __( 'Show the Privacy Policy on checkout', 'easy-digital-downloads' ),
 						'desc' => __( 'Display your Privacy Policy on checkout.', 'easy-digital-downloads' ) . ' <a href="' . esc_attr( admin_url( 'privacy.php' ) ) . '">' . __( 'Set your Privacy Policy here', 'easy-digital-downloads' ) .'</a>.',
 						'type' => 'checkbox',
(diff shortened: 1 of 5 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -892,7 +892,7 @@ function edd_get_registered_settings() {
 						'size' => 'regular',
 					),
 					'show_privacy_policy_on_checkout' => array(
-						'id'   => 'show_agree_to_privacy_policy_on_checkout',
+						'id'   => 'show_privacy_policy_on_checkout',
```
pippinsplugins: We need an upgrade routine here. At least 7,000 sites have already updated so we have to ensure that those who have enabled this do not lose their setting.

cklosowski: @pippinsplugins ok we should just be able to do a quick swap that auto runs, no need for them to click anything.

cklosowski: @pippinsplugins fixed.

PR #6637, opened 2018-06-04
issue/6605 - Add sleep() if PayPal PDT is used.

### LINKED ISSUE #6605: Double PayPal payment processing
### Expected behavior
When a PayPal payment is submitted, I expect the user value to go up by the value of the transaction. I also expect the transaction to only send one notification email, and create only one license. That payment should also only have one set of log messages on the payment i.e. 'Status changed from Pending to Complete'.

### Actual behavior
At least half of the PayPal payments are being marked as complete twice. I have to watch for duplicate payment notification emails so I can delete the extra licenses generated ad recount the user totals. This may also affect report accuracy, but I haven't looked into that.
screen shot 2018-05-28 at 9 25 15 am

1. Submit a payment with the PayPal Standard gateway
2. After the IPN notification is received, the payment is processed twice.

### Possible fix
I've made this change on my site, but haven't yet verified if it solves the problem. This fix does not change the payment status unless it is different from the previous status.
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/payments/functions.php#L292

if( $payment && $payment->ID > 0 && $payment->status !== $new_status ) {

sunnyratilal:
Fixes #6605

Proposed Changes:
1. Call `sleep()` on IPN actions if PDT is enabled too. PayPal PDT and IPN both sending data back to the site can cause a race condition to develop whereby the status of a payment is changed to `complete` twice triggering all the completed purchase actions to be triggered twice: double the license keys, double referrals in AffiliateWP etc.

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -334,7 +334,7 @@ add_action( 'edd_gateway_paypal', 'edd_process_paypal_purchase' );
  */
 function edd_listen_for_paypal_ipn() {
 	// Regular PayPal IPN
-	if ( isset( $_GET['edd-listener'] ) && $_GET['edd-listener'] == 'IPN' ) {
+	if ( isset( $_GET['edd-listener'] ) && 'IPN' === $_GET['edd-listener'] ) {
 
 		edd_debug_log( 'PayPal IPN endpoint loaded' );
 
@@ -343,14 +343,14 @@ function edd_listen_for_paypal_ipn() {
 		/**
 		 * This is necessary to delay execution of PayPal PDT and to avoid a race condition causing the order status
 		 * updates to be triggered twice.
-         *
-         * @since 2.9.4
-         * @see https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6605
+		 *
+		 * @since 2.9.4
+		 * @see https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6605
 		 */
 		$token = edd_get_option( 'paypal_identity_token' );
 		if ( $token ) {
-		    sleep( 5 );
-        }
+			sleep( 5 );
+		}
 	}
 }
 add_action( 'init', 'edd_listen_for_paypal_ipn' );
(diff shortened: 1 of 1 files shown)


includes/gateways/paypal-standard.php
```diff
@@ -339,6 +339,18 @@ function edd_listen_for_paypal_ipn() {
 		edd_debug_log( 'PayPal IPN endpoint loaded' );
 
 		do_action( 'edd_verify_paypal_ipn' );
+
+		/**
+		 * This is necessary to delay execution of PayPal PDT and to avoid a race condition causing the order status
+		 * updates to be triggered twice.
+         *
+         * @since 2.9.4
```
pippinsplugins: Indention needs fixed here.

includes/gateways/paypal-standard.php
```diff
@@ -334,11 +334,23 @@ function edd_process_paypal_purchase( $purchase_data ) {
  */
 function edd_listen_for_paypal_ipn() {
 	// Regular PayPal IPN
-	if ( isset( $_GET['edd-listener'] ) && $_GET['edd-listener'] == 'IPN' ) {
+	if ( isset( $_GET['edd-listener'] ) && 'IPN' === $_GET['edd-listener'] ) {
```
pippinsplugins: This should be run through `strtolower` if running a strict comparison.

includes/gateways/paypal-standard.php
```diff
@@ -334,11 +334,23 @@ function edd_process_paypal_purchase( $purchase_data ) {
...
+		/**
+		 * This is necessary to delay execution of PayPal PDT and to avoid a race condition causing the order status
+		 * updates to be triggered twice.
+		 *
+		 * @since 2.9.4
+		 * @see https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6605
+		 */
+		$token = edd_get_option( 'paypal_identity_token' );
+		if ( $token ) {
+			sleep( 5 );
```
pippinsplugins: This needs to come before `do_action( 'edd_verify_paypal_ipn' );`

sunnyratilal: Whoops, fixed. :)

PR #6638, opened 2018-06-05
issue/6549 - Fix absolute Windows paths resulting in fatal errors

### LINKED ISSUE #6549: Reports API: fatal error on Windows when displaying table view
### Expected behavior
See reports table rendered.

### Actual behavior
There's a fatal error preventing the whole report from loading.
```
Fatal error: Uncaught Error: Class 'EDD_Gateawy_Reports_Table' not found
```

1) Install example report plugin: https://gist.github.com/DrewAPicture/69acc0a74af184f11443687802ac89a4
2) Be on Windows. :P 
3) Visit report page/tab
4) See fatal error about `EDD_Gateawy_Reports_Table` class table not existing.

Windows local environment.

This issue is due to the usage of `validate_file()` here in `EDD\Reports\Data\Table_Endpoint`:

```php
	private function set_class_file( $file ) {
		if ( 0 === validate_file( $file ) ) {
			$this->class_file = $file;
		}
	}
```

In Windows this function returns `2` - not `0`. This results in the class file never getting loaded, but the class gets used later on.

sunnyratilal:
Fixes #6549 

Proposed Changes:
1. Absolute Windows paths return `-2` when passed to `validate_file()` meaning Windows users get fatal errors when loading the reports.  The logic has been tweaked to allow for absolute paths to be loaded too.

### DIFF
--- a/includes/reports/data/class-table-endpoint.php
+++ b/includes/reports/data/class-table-endpoint.php
@@ -157,7 +157,7 @@ final class Table_Endpoint extends Endpoint {
 	 * @param string $file Class file.
 	 */
 	private function set_class_file( $file ) {
-		if ( 0 === validate_file( $file ) ) {
+		if ( false === strpos( $file, '..' ) && false === strpos( $file, './' ) ) {
 			$this->class_file = $file;
 		}
 	}


nosegraze:
This is working for me! The file is now getting included correctly and the display is as expected. No errors.

PR #6640, opened 2018-06-05
Update note selector to fix "No notes" issue. #6630

### LINKED ISSUE #6630: Incorrect "No notes." message shown after deleting a note
If you add two notes to a discount code and then remove one of them, an improper `No notes` message will be shown:
image
Refreshing the screen shows the notes properly without the message so it's only an issue in the JS processing.

mindctrl:
Fixes #6630 

Proposed Changes:
1. Update selector for counting notes.

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -539,7 +539,7 @@ jQuery(document).ready(function ($) {
 			$( document.body ).on('click', '.edd-delete-note', function(e) {
 				e.preventDefault();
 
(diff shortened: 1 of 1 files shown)


mindctrl:
Not sure this is the right fix, but I'm mobile at the moment. Will circle back later.

mindctrl:
Ah yeah, this works. I was misreading on mobile. It works when `edd_notes.length` is `1` because we do the DOM lookup before removal, instead of doing one afterwards.

PR #6675, opened 2018-06-20
Force cart->contents to be an array when requested #6674

### LINKED ISSUE #6674: EDD_Cart class should force contents to be an array
### Expected behavior
`$cart->contents` should always be an array.

### Actual behavior
Sometimes it is not an array, but we should always return at least an array instead of `null` or anything non-countable, to meet the stricter requirements of PHP 7.2

1) Run our unit tests on PHP 7.2
2) See lots of errors (not failing tests, just PHP warnings)
3)
etc, etc

PHP Version: 7.2

EDD Version (or branch): `master`

WordPress Version: All

Any other relevant information: The easiest way to recognize this is with Unit Tests

cklosowski:
Fixes #6674

Proposed Changes:
1. Changes the `EDD_Cart::is_empty` method to return an array.
2. Forces the `EDD_Cart::get_contents` method to return an array.

### DIFF
--- a/includes/cart/class-edd-cart.php
+++ b/includes/cart/class-edd-cart.php
@@ -198,7 +198,7 @@ class EDD_Cart {
 
 		do_action( 'edd_cart_contents_loaded' );
 
-		return $this->contents;
+		return (array) $this->contents;
 	}
 
 	/**
@@ -358,7 +358,8 @@ class EDD_Cart {
 	 * @return boolean
 	 */
 	public function is_empty() {
-		return 0 === sizeof( $this->contents );
+		$contents = array( $this->contents );
+		return 0 === sizeof( $contents );
 	}
 
 	/**


includes/cart/class-edd-cart.php
```diff
@@ -358,7 +358,8 @@ public function get_quantity() {
 	 * @return boolean
 	 */
 	public function is_empty() {
-		return 0 === sizeof( $this->contents );
+		$contents = array( $this->contents );
```
JJJ: I don't believe this code will do what it's intended to do. 

In EDD 3.0, we used:
```
return 0 === count( (array) $this->get_contents() )
```

This does a few different things:
* Casts to `array` rather than accidentally nesting it inside another array
`array( (array) $thing );
* Avoids the cost of aliasing `sizeof()` to `count()`
* Reads a bit more clearly to our EDD 3.0 eyes

cc @sunnyratilal for confirmation

cklosowski: @jjj, that makes sense, I can update it if @sunnyratilal confirms

cklosowski:
@JJJ @sunnyratilal we good here?

sunnyratilal:
Yep, all good.

PR #6683, opened 2018-06-22
Show ajax loading indicator while taxes are being recalculated. #6682

### LINKED ISSUE #6682: Show ajax indicator while taxes are being recalculated
Whenever the address fields on checkout are changed we recalculate the tax amount with an ajax request. Right now, however, this happens without any visible indicator that calculations are happening. The only indication of it happening is after it's complete and the purchase total is adjusted.

I'd like to propose that we show an loading indicator while the taxes are being recalculated.

pippinsplugins:
Fixes #6682

### DIFF
--- a/assets/js/edd-checkout-global.js
+++ b/assets/js/edd-checkout-global.js
@@ -355,6 +355,8 @@ function recalculate_taxes(state) {
 		card_zip: $edd_cc_address.find('input[name=card_zip]').val()
 	};
 
+	jQuery('#edd_purchase_submit').after('<span class="edd-loading-ajax edd-loading"></span>');
+
 	var current_ajax_count = ++ajax_tax_count;
 	jQuery.ajax({
 		type: "POST",
@@ -376,6 +378,7 @@ function recalculate_taxes(state) {
 				tax_data.response = tax_response;
 				jQuery('body').trigger('edd_taxes_recalculated', [ tax_data ]);
 			}
+			jQuery('.edd-loading-ajax').remove();
 		}
 	}).fail(function (data) {
 		if ( window.console && window.console.log ) {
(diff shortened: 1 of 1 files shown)


assets/js/edd-checkout-global.js
```diff
@@ -355,6 +355,8 @@ function recalculate_taxes(state) {
 		card_zip: $edd_cc_address.find('input[name=card_zip]').val()
 	};
 
+	jQuery('#edd_purchase_submit').after('<span class="edd-loading-ajax edd-loading"></span>');
```
SDavisMedia: Let's move this to display the loading icon after the submit button itself instead of after the fieldset, to avoid style inconsistencies.

So changing `jQuery('#edd_purchase_submit')` to `jQuery('#edd_purchase_submit [type=submit]')`.

This will match the loading icon placement for processing checkout. https://github.com/easydigitaldownloads/easy-digital-downloads/blob/07c3fe173290fa6bb72e7fbd2f06b7b49f5ad126/assets/js/edd-ajax.js#L384 Without this change, any theme adding a padding or margin to the bottom of the fieldset will create space above the loading icon when taxes are being recalculated, but not when the purchase is being processed.

SDavisMedia:
Cleared by @sunnyratilal to go ahead and do this myself. All goody here. 👍

cklosowski:
Looks like @sunnyratilal merged this as I was commenting. We need the `jQuery('.edd-loading-ajax').remove();` to be less greedy. In the event that multiple items on the page have the class `edd-loading-ajax` we should only remove the one associated with the tax recalculation.

SDavisMedia:
https://github.com/easydigitaldownloads/easy-digital-downloads/commit/b3b09b718d04ded5611b8b34194cbb77190a2c47

PR #6710, opened 2018-07-03
Fix alternate checkouts with better edd_is_checkout and edd_get_checkout_uri detection #6668

### LINKED ISSUE #6668: Redirect to checkout setting conflicts with multiple checkout pages
_EDD version 2.9.3_

The new multiple checkout page feature works well, but has an interesting edge case:

If using `?add_to_cart` in a URL going to an alternate checkout page, the user is redirected to the main checkout page if the `Redirect to checkout` setting is set to true.

For instance:

1. I navigate to `https://mylocalsite.test/alternate-checkout?edd_action=add_to_cart&download_id=481`
2. I expect to end up on `https://mylocalsite.test/alternate-checkout` with the download with ID 481 in my cart.
3. Instead, I end up on `https://mylocalsite.test/checkout` with download 481 in my cart.

If I uncheck `Redirect to checkout` on the Misc. tab of EDD settings, then number 2 above works as expected/desired.

I'm not sure what a good fix here is... it would be _nice_ if that setting could be turned on, but direct navigation to an alternate checkout page were allowed (and thus bypassed the redirect).

For this to happen, `edd_is_checkout()` would need to return true for any page with the `[download_checkout]` shortcode present. This is _probably_ a good idea anyway, though perhaps there are edge cases there to consider as well.

In `edd_is_checkou()`, `$wp_query->queried_object` appears to not be set in this case; if it were, I believe that function would return `true`, because [it does check for the presence of the `[download_checkout]` shortcode](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/39cd5c6098e80489adde7da4b72918478cf84deb/includes/checkout/functions.php#L40)... _if_ the `queried_object` property is set (and if `is_singular()` returns true).

cklosowski:
#6668

Proposed Changes:
1. ~Updates `edd_is_checkout` to use a global to cache for this thread, in order to reduce a number of calls to the function over and over.~
2. Adds a condition to `edd_get_checkout_uri` to detect if the current URL is a checkout, to return itself, instead of the default checkout URI only.
3. Adds new 'delayed hooks', so that things like `edd_add_to_cart` can run later than `init` in order to allow proper page detection, which allows determining what checkout page to redirect to.

### DIFF
--- a/includes/actions.php
+++ b/includes/actions.php
@@ -22,6 +22,13 @@ if ( ! defined( 'ABSPATH' ) ) exit;
 */
 function edd_get_actions() {
 	$key = ! empty( $_GET['edd_action'] ) ? sanitize_key( $_GET['edd_action'] ) : false;
+
+	$is_delayed_action = edd_is_delayed_action( $key );
+
+	if ( $is_delayed_action ) {
+		return;
+	}
+
 	if ( ! empty( $key ) ) {
 		do_action( "edd_{$key}" , $_GET );
 	}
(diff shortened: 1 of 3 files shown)


cklosowski:
I just removed the `global $is_checkout` it was causing issues for unit tests.

PR #6740, opened 2018-07-10
issue/6725 - remove incorrect sales stats function parameter

### LINKED ISSUE #6725: 3.0 Reports: Overview displays dollar amount for sales count
### Expected behavior
Reports Overview tab accurately display an overview of sales, earning, etc.

### Actual behavior
On the Reports Overview, the `Sales` tab (Last 30 Days) displays a dollar amount (which appears to be the dollar version of the actual sales count).

For example, I have a single purchase on a fresh site. This purchase was for a $10 product, plus 10% exclusive tax. So 1 sale, with $11 earnings. Here's the overview:

6aee098d-c3d6-4ad2-8ffe-39ffef40c61a

1) Complete at least one purchase.
2) View Reports Overview `Sales` tile.

PHP Version: `7.1.7`
WordPress Version: `4.9.7`

SDavisMedia:
Fixes #6725

Proposed Changes:
1. Remove incorrect parameter from `get_order_count()` method on Reports -> Overview - Sales tile

Note: I'm basing this PR solely off of what I learned from reading doc blocks on the `get_customer_count()` method. I believe the Sales tile in the Overview report is passing the method a value that is used for SQL functions, but SQL functions cannot be passed on that method.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -276,7 +276,6 @@ function edd_register_overview_report( $reports ) {
 						$stats = new EDD\Orders\Stats();
 						return apply_filters( 'edd_reports_overview_sales', $stats->get_order_count( array(
 							'range'    => $filter['range'],
-							'relative' => true,
 						) ) );
 					},
 					'display_args'  => array(


sunnyratilal:
The `relative` parameter is needed to display the increase or decrease in sales relative to the previous date range.  i.e. When the filter is set to `This Year`, it shows the sales figure relative to the previous year.

`relative` tells the Stats function to query against the previous data range. `function` overrides the SQL function (not possible on all methods).

SDavisMedia:
@sunnyratilal Thanks for the explanation!

PR #6744, opened 2018-07-10
Make file download log migrations batched

### LINKED ISSUE #6731: File download log migration failing (GDPR)
### Expected behavior
When you do an upgrade for https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6548, it should be batched and work for any number of download logs.

### Actual behavior
A few customers with stores containing a large number of file downloads attempt to do the upgrade routine, it fails with this error message:

```
PHP Fatal error: Allowed memory size of
> 134217728 bytes exhausted (tried to allocate 133883136 bytes)
```
1) Create a store with a large number of download logs (like 80,000) on version `2.9.1` or earlier. 
2) Attempt to do the upgrade routine from https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6548
3) It should fail with the error message above.

EDD Version (or branch): `2.9.1` being migrated to `2.9.2`

mintplugins:
Fixes #6731

Proposed Changes:
1. Make the file download log migration use batches of 50 instead of querying for all


mintplugins:
This is still a work in progress that I am testing. I just wanted to get it on github. I'll comment here when I've confirmed testing.

mintplugins:
This is ready for testing. However, it's a bit difficult if you've already run the upgrade routine. 

In order to test you'd have to do it in a store that hasn't done it yet, or that had the failure described in the issue:

https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6731

PR #6794, opened 2018-07-24
issue/6793 - Add `edd_insert_payment` hook.

### LINKED ISSUE #6793: 3.0: edd_insert_payment hook is missing
In latest `release/3.0`, the `edd_insert_payment` hook does not run. I think we need some backwards compatibility for this, as we use it in our own add-ons quite a bit and I know of a few custom integrations off the top of my head that rely on it.

sunnyratilal:
Fixes #6793.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -133,7 +133,7 @@ function edd_get_discount_by_code( $code = '' ) {
  *
  * @param string $field The field to retrieve the discount with.
  * @param mixed  $value The value for $field.
- * @return mixed object|bool EDD_Discount object or false if not found.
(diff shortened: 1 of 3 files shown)


cklosowski:
@sunnyratilal, this PR does way more than just reintroducing the hook, is that intentional?

mindctrl:
@sunnyratilal tested this with some custom code and with AffiliateWP and both worked as expected.

sunnyratilal:
@cklosowski It's just cleanup, formatting, and DocBlock changes.  PHPStorm ran auto format on the file before I committed and I only noticed the large diff once I pushed.

cklosowski:
👍 ok thanks @sunnyratilal

Issue #6820, opened 2018-08-01
3.0: Separate tax rates & revert adjustments back to discounts


sunnyratilal:
After discussion in Slack with @mindctrl and @cklosowski; a decision was made to create a new table for tax rates in order for the data structure to be best used.

With this change, it was also decided to rename `adjustments` back to `discounts`.

cklosowski:
@sunnyratilal Since we've been having people use this beta (in testing) we should probably kill the adjustments tables. Obviously a migration of the data isn't necessary since this is an 'in progress' release and shouldn't be on any production environments, but we should remove the adjustment and adjustmentmeta tables.

sunnyratilal:
Merged.

PR #6822, opened 2018-08-01
issue/6820 - Adjustments → Discounts & Tax Rates table

sunnyratilal:
See #6820.

Proposed Changes:
1. Change `adjustments` to `discounts` migrating all discounts.
2. Introduce a tax rates table.
3. Migrate tax rates to `edd_tax_rates` table.
4. Introduce `EDD\Tax_Rates\Tax_Rate` object along with helper functions.


cklosowski:
@sunnyratilal The default sort seems to be 'DESC', which is a weird experience when adding tax rates. As I add a tax rate to the bottom of the 'Rates' list in the UI, when I save, the order is reverse from when I add them. We should try and keep the order the same as the UI, which means it would default to an 'ASC' tax order, which makes sense from the perspective of applying tax rates anyway, You want to target more specific first, then get less specific.

Issue #6808, opened 2018-07-29
Reposition/refactor the pagination in the [downloads] shortcode


amdrew:
The pagination (`#edd_download_pagination`) in the `[downloads]` shortcode is currently a sibling of each download in the `.edd_downloads_list` div.

I propose: 

## We move the pagination so it appears after `.edd_downloads_list` in the HTML markup. 

Doing so will fix a fundamental problem when trying to use flexbox or CSS grid to lay out the downloads. Since the pagination is a sibling of the actual downloads, it takes on the properties on  `display: flex` or `display: grid` when they have been applied to the parent container (`.edd_downloads_list`). 

Here's an example where the pagination is positioned incorrectly (it should be below the downloads). The reason why it does this is because it has essentially become a flex item:

<img width="807" alt="screen shot 2018-07-29 at 7 14 23 pm" src="https://user-images.githubusercontent.com/52581/43363911-b31129fe-9363-11e8-824f-782bc9236cbf.png">

The above is a very basic example which can easily be solved by making the pagination div 100% wide however more CSS shouldn't be needed. Depending on the grid/flex properties used, more CSS may also be needed to fix the positioning/look of the pagination. 

The pagination is not a download so it should not exist within the div that wraps each download.

## We move the pagination to its own dedicated function.

The `downloads` shortcode is huge, moving the pagination to its own function makes things easier to see and manage. 

In the near future we should also build a dedicated function for outputting a grid of downloads. This is a small step towards that goal and would mean themes and other plugins can build download grids without having to call the shortcode. And of course the `[downloads]` shortcode would pass its attributes to this function to build its own download grid. 

The downloads block in EDD Blocks as an example currently needs to use the `[downloads]` shortcode as there is no other way to build a download grid: https://github.com/easydigitaldownloads/edd-blocks/blob/master/includes/blocks/downloads/index.php#L33-L38. 

## We load the dedicated function on the `edd_downloads_list_after` action hook.

This would allow themes/plugins to unhook the pagination and completely customize it (something I need to be able to do). The pagination's existing filter can only get you so far.

## Other notes

1. The current `edd_downloads_list_after` action hook doesn't actually work correctly. Because of where it's positioned in the code + object buffering, anything echoed to the hook will be shown ABOVE the download grid. The proposed PR fixes this.

[...]

PR #6853, opened 2018-08-14
Issue/6852 - Implementing CSS Grid (and keeping CSS floats

### LINKED ISSUE #6852: Implementing CSS Grid (and keeping CSS floats)
It's CSS Grid time 🎉 

PR: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6939

## Why we should implement CSS Grid

1. CSS Grid is well supported: https://caniuse.com/#feat=css-grid
1. It will eventually allow us to move away from our current implementation of CSS floats
1. It will cut down our CSS (once floats are removed), and allow us to do some interesting things in the future
1. We've been slowly working towards it for quite some time. https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6808, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3489, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5847
1. It will allow for better designed themes
1. EDD Blocks (https://github.com/easydigitaldownloads/edd-blocks) will have blocks for outputting a grid of downloads, download categories, and possibly download tags (or 1 block for all). Since the Gutenberg block(s) will eventually exist in EDD core, now's a good time to refactor the CSS and move ourselves into the future of layout.
1. It fixes an issue where the download grid is slightly indented and does not align with headings and other elements around it. This is caused by each download having padding and margin to space themselves apart. Usually a workaround (which we don't do in core) is to use negative margins on the main wrapping div. CSS Grid has a gap property for spacing, but the gaps are only in between each item in the grid meaning everything aligns nicely.
1. We'll be implementing a mobile-first approach. With the current implementation of floats we don't account for mobile devices. Downloads overlap each other at smaller device widths, making things illegible. This time around we stack the downloads, then expand into columns at larger device widths. Note, while we could also fix the floating issue in this PR (which would probably mean slightly more code), I think we should leave the current implementation exactly as-is, solely focusing on CSS grid going forward.

## Goals

1. Keep the existing HTML markup to avoid any disruption
1. Use CSS Grid for modern browsers while still using floats for crappy browsers (IE10 and below)
1. Get us in the position to drop floats completely when the time is right

## How it works

- An older browser sees the current implementation of CSS floats and uses that for the layout.
- Modern browsers will see floats _and_ CSS Grid, preferring the latter. When `display: grid` is set on the parent `div`, any `float` or `clear` property is completely ignored.
- All downloads are stacked (mobile-first approach) until the 768px breakpoint grid, where they are then displayed in a 2 column layout (except for when `columns="1"` is defined in the shortcode). Once the browser width reaches the 1200px breakpoint, the true column count is used.
- The `@supports` feature query is used to tell modern browsers to reset any widths, margin or padding. Widths interfere (since we're using the `fr` unit) with the download grid and there's no longer a need for margin and padding on each individual download.

## Other notes

- Tested with modern browsers
- Tested with IE10 (will use floats) and IE11 (will use CSS Grid)
- Vendd overrides widths based on the column class used. Because of the CSS specificity, it will override the `width: auto` rule added to reset the widths. All Vendd will need is to remove the widths and it's method of using `display: inline-block`. It can use CSS grid and floats will be provided as a fallback.
- This PR relies on what is happening in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6808. The pagination can't be a sibling of the downloads.
- This should be tested thoroughly with other themes just in case.

## Caveats

There is a _slight_ issue when specifying a 6 column layout in EDD, i.e . E.g.`[downloads columns="6"]`. 

With the current implementation of CSS floats and using the Twenty Sixteen theme, here's the 6 column layout:

screen shot 2018-08-14 at 6 34 01 pm

Everything is bunched up and doesn't look great at all. But it "stays" (well mostly) within the confines of the div.

With CSS grid, 6 columns looks like the following, slightly overflowing outside of the div since there's not enough space inside:

screen shot 2018-08-14 at 6 30 18 pm

This _can_ be fixed using a combination of CSS Grid's `minmax() function`, and either `auto-fit` or `auto-fill`, but not for our use-case (which is why I'm also using media queries at all). We need to specifically set the number of columns chosen by the user.

In any case, I see this as a non issue for 3 reasons:

1. Showing 6 columns is not exactly common (I don't think we support it in either Vendd or Themedd). Themes like Twenty Sixteen have narrow widths.
2. It looks better (even though it overlaps) than the existing float technique we use.
3. If you're really going to show 6 columns of downloads, the theme should provide ample space.

## How CSS Grid looks

Here's how each download grid with a column set looks with the Twenty Sixteen theme. Ignore the black line, that's caused by the full-page screenshot:

screenshot_2018-08-14 grid test twentysixteen

Sidenote: we should add some CSS in the future to reset any top margins on h3 headings within the download grid (as seen by the large space above each heading above).

And finally, Firefox has an amazing Grid inspector for testing/viewing grids:

screen recording 2018-08-14 at 06 47 pm

amdrew:
Fixes #6852.

This PR relies on https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6808. You need only worry about the `templates/edd.css` file.

### DIFF
--- a/templates/edd.css
+++ b/templates/edd.css
@@ -981,6 +981,11 @@ input.edd_submit_plain {
 
 /* =Downloads Shortcode
 -------------------------------------------------------------- */
+.edd_downloads_list {
+	display: grid;
+	grid-column-gap: 20px;
+	grid-row-gap: 40px;
+}
 .edd_downloads_list:after {
 	content: "";
 	display: table;
@@ -1013,7 +1018,29 @@ input.edd_submit_plain {
 .edd_download .edd_price {
 	margin-bottom: 10px;
 }
+@media (min-width: 768px) {
+	.edd_downloads_list:not(.edd_download_columns_1) {
+		grid-template-columns: repeat(2, 1fr);
+	}
+}
 
+@media (min-width: 1200px) {
+	.edd_downloads_list.edd_download_columns_2 { grid-template-columns: repeat(2, 1fr); }
+	.edd_downloads_list.edd_download_columns_3 { grid-template-columns: repeat(3, 1fr); }
+	.edd_downloads_list.edd_download_columns_4 { grid-template-columns: repeat(4, 1fr); }
+	.edd_downloads_list.edd_download_columns_5 { grid-template-columns: repeat(5, 1fr); }
+	.edd_downloads_list.edd_download_columns_6 { grid-template-columns: repeat(6, 1fr); }
+}
+
+@supports (display: grid) {
+	.edd_downloads_list .edd_download { 
+		width: auto;
+	}
+	.edd_download_inner {
+		padding: 0;
+		margin: 0;
+	}
+}
 /* =Misc styles
 -------------------------------------------------------------- */
 .edd-hide-on-empty.cart-empty {


amdrew:
@cklosowski The changes from #6852 are included in this PR so CSS grid can function properly. The only file in this PR that needs to be reviewed is really `edd.css`.

amdrew:
Updated PR. It's now based off master which has the pagination merged in from https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6808

amdrew:
Closing. Going to re-submit against release/3.0

PR #6862, opened 2018-08-19
Issue/6861 - Add order and orderby parameters

### LINKED ISSUE #6861: EDD API doesn't support "order" and "orderby" parameters for products endpoint
While using the EDD API to retrieve downloads I noticed I couldn't use `order` and `orderby` parameters to sort the results. 

### Expected behavior

The `order` and `orderby` parameters should be allowed to sort downloads as this is already possible with the core REST API.

### Actual behavior

When using `order` and `orderby`, the same results are returned, regardless of the parameters.

Make a request to the products endpoint and try to sort the products using `order` and `orderby`: 

Example (sorting products by the post name (aka slug) in ascending order: `https://yoursite.com/edd-api/products?order=ASC&orderby=name`

### Use case

In the EDD Blocks downloads block (https://github.com/easydigitaldownloads/edd-blocks) there's a select menu for `Order` and `Order By`:

screen shot 2018-08-19 at 5 36 56 pm

I was using the WP REST API to previously sort the downloads but switched to the EDD API. Hoping to get this fixed for 2.9.7 (and also `release/3.0` which I'm testing with) if possible 🙏

amdrew:
Fixes #6861

### DIFF
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -59,7 +59,7 @@ class EDD_API_V2 extends EDD_API_V1 {
 				case 'sales':
 					if ( user_can( $this->user_id, 'view_shop_sensitive_data' ) || current_user_can( 'view_shop_sensitive_data' ) || $this->override ) {
 						$query_args['meta_key'] = '_edd_download_sales';
-						$query_args['orderby']  = 'meta_value';
+						$query_args['orderby']  = 'meta_value_num';
 					}
 					break;
 
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -40,7 +40,7 @@ class EDD_API_V2 extends EDD_API_V1 {
 			$query_args = array(
 				'post_type'        => 'download',
 				'posts_per_page'   => $this->per_page(),
-				'suppress_filters'  => true,
+				'suppress_filters' => true,
 				'paged'            => $this->get_paged(),
 				'order'            => $args['order'],
 				'orderby'          => $args['orderby'],
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -56,6 +56,20 @@ class EDD_API_V2 extends EDD_API_V1 {
 					$query_args['orderby']  = 'meta_value_num';
 					break;
 
+				case 'sales':
+					if ( user_can( $this->user_id, 'view_shop_sensitive_data' ) || current_user_can( 'view_shop_sensitive_data' ) || $this->override ) {
+						$query_args['meta_key'] = '_edd_download_sales';
+						$query_args['orderby']  = 'meta_value';
+					}
+					break;
+
+				case 'earnings':
+					if ( user_can( $this->user_id, 'view_shop_sensitive_data' ) || current_user_can( 'view_shop_sensitive_data' ) || $this->override ) {
+						$query_args['meta_key'] = '_edd_download_earnings';
+						$query_args['orderby']  = 'meta_value_num';
+					}
+					break;
+
 				default:
 					$query_args['orderby'] = 'date';
 					break;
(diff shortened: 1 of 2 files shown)


sunnyratilal:
PR looks good to me. 👍

amdrew:
Made a small addition to the PR. I think we should allow ordering by `price`, `sales` and `earnings`, like the downloads shortcode: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/shortcodes.php#L364-L381).

amdrew:
I've added back the ability to order by `sales` and `earnings` but with some permission checks. I'm not 100% I've done it correctly so would appreciate another set of 👀's  (or two) 

Here's what I'm hoping to achieve:

1. Allow the downloads to be ordered by `sales` and `earnings` when making an _external_ request to the EDD API. This will need authentication with a token and key. I don't need this personally, but I see it as being useful.

2. Allow the downloads to be ordered by `sales` and `earnings` when making an _internal_ request to the EDD API. By that I mean adding a downloads block (via EDD Blocks) which uses the EDD API and then changing the `Order By` select menu to sales or earnings. This shouldn't need a token or key since it's just a preview of the downloads in the admin. So for this I've added `current_user_can( 'view_shop_sensitive_data' )` to the conditional since `$this->user_id` in `user_can()` doesn't relate to the currently logged in user viewing/editing the block.

I've tested the EDD API externally with and without a token and key and I can confirm the downloads are **not** ordered by sales and earnings when **not** using a token and key (correct result) 👍 

Also tested with EDD Blocks and as a user (admin) with the `view_shop_sensitive_data` capability, I can correctly order the downloads by sales or earnings (both ASC and DESC).

amdrew:
After doing some more testing I discovered the switch statement's default case was overriding normal order and orderby queries. E.g. ordering by `title` or `name`, oops.

1. I removed the default case from the switch statement. It will fallback to `date`.
2. I removed the `order()` and `orderby()` methods from the base class. I now see it's better to add them to the `process_query()` method that already exists (and since they are only for the products endpoint).
3. `order` and `orderby` are now passed into the `$query_args` array. If they aren't specified, they fallback to `DESC` and `date`, which are the defaults.
4. I re-tested all parameters with `ASC` and `DESC` to make sure the results were correct.

    `title`
    `price`
    `date`
    `name`
    `ID`
    `rand`

    And also these two with and without authentication:

    `earnings`
    `sales`

includes/api/class-edd-api-v2.php
```diff
@@ -40,14 +40,38 @@ public function get_products( $args = array() ) {
...
+			switch ( $query_args['orderby'] ) {
+				case 'price':
+					$query_args['meta_key'] = 'edd_price';
+					$query_args['orderby']  = 'meta_value_num';
+					break;
+
+				case 'sales':
+					if ( user_can( $this->user_id, 'view_shop_sensitive_data' ) || current_user_can( 'view_shop_sensitive_data' ) || $this->override ) {
+						$query_args['meta_key'] = '_edd_download_sales';
+						$query_args['orderby']  = 'meta_value';
```
cklosowski: @amdrew shouldn't this also be `meta_value_num` so we can treat them as numeric values?

amdrew: @cklosowski Oops, yes it should. I see how that happened, I copied it over from the downloads shortcode but the shortcode uses `meta_value` for `$atts['orderby']` and `meta_value_num` for the $query['orderby'].

```
case 'sales':
	$atts['orderby']   = 'meta_value';
	$query['meta_key'] = '_edd_download_sales';
	$query['orderby']  = 'meta_value_num';
break;
```

amdrew:
Whitespace is bizarre, I fixed that space above because this is how I see it in VS code 😂 

screen shot 2018-08-21 at 8 58 32 am

includes/api/class-edd-api-v2.php
amdrew: Fixed and tested again for peace of mind.

cklosowski:
Before we merge this, I'll see if I can work up some unit tests for it.

PR #6866, opened 2018-08-22
Issue/6865 - Use SVG images for payment icons at checkout

### LINKED ISSUE #6865: Use SVG images for payment icons at checkout
We should switch to SVG images for the payment icons at checkout, rather than the existing `.png` and `.gif` versions. These are the images that appear either above "Select Payment Method" (.png) or the image that appear when validating a credit card number (.gif)

## Reasons why:

1. The payment icons at checkout look blurry on retina displays. They should look crisp on any display.  We nearly had retina versions of the existing icons back in 2013 but they were never merged in: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/1212

    screen shot 2018-08-21 at 5 20 57 pm

    And zoomed in a bit for clarity:

    screen shot 2018-08-21 at 5 25 35 pm

2. I've researched each credit card provider and all our existing logos are now incorrect or outdated. We should be careful to use the assets provided by each provider, just as we would want others to use our brand guidelines correctly.

3. We can't just double the size of the existing .png images. Screens are becoming more pixel dense, SVGs will future-proof the issue.

4. SVGs would allow someone to use the images at whichever size they like. Perhaps the site owner/theme developer would like to place more emphasis on the icons to build trust.

5. It will allow for crisp images on the `Payment Gateways` admin screen for the `Accepted Payment Method Icons` and also when a credit card number is being validated via JS.

6. It would allow us to remove a decent chunk of CSS from our EDD stylesheet and 12 images from the icons folder.

7. SVGs are well supported: https://caniuse.com/#search=svg. 

## Proposed icons:

I've hunted down each provider's brand guidelines and logos they prefer others to use. Some companies have multiple versions of their logo but the one we want in each instance is called the "Acceptance mark".

> An acceptance mark, in respect of credit cards, is a logo or design that indicates which card schemes an automated teller machine or merchant accepts

We can't use any ol' credit card icon pack from the internet, from what I've seen they are all wrong. Here are the proposed SVG icons I've created:

screen shot 2018-08-21 at 5 21 19 pm

Zoomed in:

screen shot 2018-08-21 at 6 30 40 pm

## Notes

1. Each SVG image has a viewbox height of 32 pixels, so they are the same height as the existing icons
1. Some images are not as wide as the rest (e.g. American Express). This is how the provider would like their brand portrayed, we're not going to stretch the widths to suit us. IMO they all look perfectly fine.
1. The SVG images will be combined into 1 `svg-icons.svg` file (cutting down image requests) and loaded into the footer of the checkout page. This allows us to reference each SVG like so:

    ```
    <svg width="50" height="32" class="payment-icon icon-mastercard" aria-hidden="true" role="img"> <use href="#icon-mastercard" xlink:href="#icon-mastercard"></use> </svg>
    ```

1. This image mentioned above will exist in EDD's `templates/images` folder so a child theme can replace it with their own icons if need be.

1. We could create a function for handling the loading of not only payment icons, but any icon within EDD (and subsequently switch them to SVGs also). E.g the iPhone icon for the iOS app, the tick when you add a download to the cart, the X for removing downloads/discounts etc. I do this in Themedd 1.1 with `themedd_get_svg()`: https://github.com/easydigitaldownloads/themedd/blob/release/1.1/includes/icon-functions.php#L123

Here's a list of each credit card provider and their relevant guidelines/downloads for reference:

### Visa
https://www.visaeurope.com/receiving-payments/pos_branding

### American Express
https://www.amexpop.com/Asset/Index/5
https://merchant-supplies.americanexpress.com/?locale=en_CA#/catalog/All

### Mastercard
https://brand.mastercard.com/brandcenter/mastercard-brand-mark.html
https://brand.mastercard.com/brandcenter/mastercard-brand-mark/downloads.html

### Discover
https://www.discovernetwork.com/en-us/business-resources/free-signage-logos

### PayPal
https://www.paypal.com/us/webapps/mpp/logo-center

### Amazon
https://pay.amazon.com/us/merchant/tools

PR coming at some point.

amdrew:
WIP PR

Fixes #6865


includes/checkout/template.php
```diff
@@ -660,53 +660,59 @@ function edd_payment_mode_select() {
...
-
-				$image = edd_locate_template( 'images' . DIRECTORY_SEPARATOR . 'icons' . DIRECTORY_SEPARATOR . $card . '.png', false );
-
-				// Replaces backslashes with forward slashes for Windows systems
-				$plugin_dir  = wp_normalize_path( WP_PLUGIN_DIR );
-				$content_dir = wp_normalize_path( WP_CONTENT_DIR );
-				$image       = wp_normalize_path( $image );
-
-				$image = str_replace( $plugin_dir, WP_PLUGIN_URL, $image );
-				$image = str_replace( $content_dir, WP_CONTENT_URL, $image );
```
amdrew: Note: Unless I'm mistaken, the removed code above is no longer needed since there aren't any paths present when loading the new SVG images.

For example, the SVG markup to load the mastercard icon looks like this:

```
<svg width="50" height="32" class="payment-icon icon-mastercard" aria-labelledby="title-5b7cf8716cc9a" role="img"><title id="title-5b7cf8716cc9a">Mastercard</title> <use href="#icon-mastercard" xlink:href="#icon-mastercard"></use> </svg>
```

includes/misc-functions.php
```diff
@@ -998,3 +998,136 @@ function edd_doing_cron() {
...
+ *
+ * @return array $sizes Sizes array (width and height) of the icon requested.
+ */
+function edd_svg_dimensions( $icon ) {
+	
+	if ( empty( $icon ) ) {
+		return false;
+	}
+	
+	$sizes = apply_filters( 'edd_svg_dimensions',
```
amdrew: Note: Each SVG icon has its dimensions added to this array. This is so:

1. We don't need to set icon dimensions via CSS. Someone could add their own SVG icon, add their dimensions and be done with it.
2. Setting a width and height directly on the SVG is better IMO. If the browser has CSS disabled the icons will still be the same size. Set the size in CSS and when disabled the icons will take over the whole browser.

includes/checkout/functions.php
```diff
@@ -608,3 +608,94 @@ function edd_purchase_form_validate_cc_exp_date( $exp_month, $exp_year ) {
...
+
+	if ( empty( $payment_methods ) ) {
+		return;
+	}
+
+	?>
+	<svg style="position: absolute; width: 0; height: 0; overflow: hidden;" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+		<defs>
+		<?php if ( array_key_exists( 'mastercard', $payment_methods ) ) : ?>
+			<symbol id="icon-mastercard" viewBox="0 0 50 32">
```
amdrew: Note: SVGs have been minified as much as possible, we ain't squeezing any more outta these 😉

amdrew: Once I get to it, this function will also be used to load the SVG icons into the payment settings page.

includes/checkout/functions.php
```diff
@@ -608,3 +608,94 @@ function edd_purchase_form_validate_cc_exp_date( $exp_month, $exp_year ) {
...
+				<rect x="20.264" y="4.552" width="9.47" height="17.019" fill="#ff5f00"/>
+				<path d="m20.865 13.063c-2e-3 -3.319 1.524-6.46 4.134-8.508-1.906-1.499-4.262-2.314-6.687-2.314-5.938 0-10.823 4.886-10.823 10.823 0 5.938 4.885 10.823 10.823 10.823 2.425 0 4.781-0.815 6.687-2.313-2.611-2.05-4.137-5.192-4.134-8.511z" fill="#eb001b" fill-rule="nonzero"/>
+				<path d="m41.486 19.77v-0.349h0.142v-0.072h-0.358v0.072h0.141v0.349h0.075zm0.695 0v-0.421h-0.109l-0.126 0.301-0.126-0.301h-0.108v0.421h0.075v-0.319l0.117 0.274h0.081l0.118-0.274v0.319h0.078z" fill="#f79e1b" fill-rule="nonzero"/>
+				<path d="m42.511 13.063c0 5.937-4.885 10.823-10.823 10.823-2.425 0-4.782-0.816-6.689-2.315 2.609-2.051 4.136-5.191 4.136-8.51 0-3.318-1.527-6.459-4.136-8.509 1.907-1.5 4.264-2.315 6.689-2.315 5.938 0 10.823 4.886 10.823 10.823v3e-3z" fill="#f79e1b" fill-rule="nonzero"/>
+			</symbol>
+		<?php endif; ?>
+		<?php if ( array_key_exists( 'americanexpress', $payment_methods ) ) : ?>
```
amdrew: Each SVG symbol will only be loaded is active in the admin

includes/checkout/template.php
```diff
@@ -660,53 +660,59 @@ function edd_payment_mode_select() {
...
+				// Get SVG dimensions.
+				$dimensions = edd_svg_dimensions( $key );
+
+				// Get SVG markup.
+				$image = edd_get_svg( 
+					array( 
+						'icon'    => $key,
+						'width'   => $dimensions['width'],
+						'height'  => $dimensions['height'],
+						'title'   => $card,
```
amdrew: Renamed `$card` to `$card_lowercase` above since we need to pass the original `$card` variable here as the title to the SVG.

includes/checkout/template.php
```diff
@@ -660,53 +660,59 @@ function edd_payment_mode_select() {
...
+						'classes' => array( 'payment-icon' )
+					)
+				);
 
 			}
 
-			if( edd_is_ssl_enforced() || is_ssl() ) {
-
+			if ( edd_is_ssl_enforced() || is_ssl() ) {
 				$image = edd_enforced_ssl_asset_filter( $image );
```
amdrew: SVG icons don't have a URL present but I've tested this to make sure icons passed in with URLs still work.

includes/checkout/template.php
amdrew: Works fine in my testing

includes/admin/admin-actions.php
```diff
@@ -83,4 +83,17 @@ function edd_display_product_tabs() {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Loads SVG icons into the admin footer.
+ * 
+ * @since 2.9.x
+ */
+function edd_svg_admin_icons() {
```
amdrew: This function could one day be responsible for outputting any SVG icon into the admin (should we need to), hence why it's in this file. Let me know if there's a better place. 

Also, am I correct in saying we don't have a helper function that holds the default payment methods? We have `edd_get_option( 'accepted_cards', array() );` but that's only the methods saved. Ideally I'd like to pass in only the default payment method keys to `edd_svg_icons()` and not anything else.

includes/checkout/functions.php
```diff
@@ -608,3 +608,32 @@ function edd_purchase_form_validate_cc_exp_date( $exp_month, $exp_year ) {
...
+	if ( ! empty( $payment_methods ) ) {
+		foreach ( $payment_methods as $key => $method ) {
+			// Add each payment method's key to $icons array. I.e "mastercard".
+			$icons[] = $key;
+		}
+	}
+
+	return edd_svg_icons( $icons );
+}
+add_action( 'wp_footer', 'edd_svg_checkout_icons', 9999 );
```
amdrew: This function will only load the SVG icons at checkout, based on the payment methods selected in the admin.

JJJ:
Created a 3.0 specific issue branch for this PR.

https://github.com/easydigitaldownloads/easy-digital-downloads/tree/issue/6865-3.0

It merges `issue/6865` into `release/3.0` so when the dust settles here, we can (theoretically) merge that into `issue/6865-3.0` with fewer conflicts.

screen shot 2018-08-23 at 3 35 03 pm

amdrew:
Just finished the remaining tasks outlined in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6865#issuecomment-414916968

The credit card validator now uses the same SVG icons instead of the .png alternatives (which have now been removed).  Not sure why we had different .gif versions, I ran a search through the codebase but couldn't pinpoint why they are there.

Since the credit card validator needs to use/show an SVG icon regardless of a payment gateway being disabled, I've had to allow all SVG icons to be injected into the HTML. No biggie, but it would have been nice to avoid that.

Tested back to IE10 on Windows 7 and the the SVG icons are correctly shown while typing the various credit card numbers (see our `jquery.creditCardValidator.js` file for examples or the screenshot below).

@JJJ Could you give the credit card validation a spin?

👍 to merging into 3.0 sooner rather than later, would be nice to get some more eyes on it.

Twenty Sixteen theme:

screen shot 2018-09-14 at 3 55 44 pm

screen shot 2018-09-14 at 3 55 38 pm

screen shot 2018-09-14 at 3 55 28 pm

___

Notes:

1. Our `jquery.creditCardValidator.js` file validates cards that we don't have images for (this was the case previously also):

- diners_club_carte_blanche
- diners_club_international
- jcb
- laser
- visa_electron
- maestro

If a card is validated, the SVG HTML will be present, but since the symbol doesn't exist, it will simply show nothing - it won't show a "broken image". Quite handy, since a theme developer could technically add the required SVG code and it would just work.

JJJ:
> Could you give the credit card validation a spin?

Works for me!

I'm going to rename the `_svg_` functions to make them technology agnostic. In 5 years we may not want SVGs anymore, and it'd be nice to not need new functions then.

I'm going to push that to the `issue/6865-3.0` branch, then merge that into `release/3.0`.

JJJ:
Merged.

SVG images are now in the `release/3.0` branch, looking beautiful.

I think this PR is OK to close, and we can work on iterating things with issues as needed?

Issue #6874, opened 2018-08-25
Move "Upgrades" screen to "Tools > Upgrades"


JJJ:
Related to #6275.

The `page=edd-upgrades` hidden Upgrades page is an anomaly, and is our last remaining page to consolidate into another area.

Reasons to move it are:
* Consistency
* Simplify `edd_is_admin_page()` and friends
* Feels more inline with other EDD3 changes

pippinsplugins:
👍

JJJ:
Related: #6880.

pippinsplugins:
This is a nice to have and has the potential to break any add-on that has registered a custom upgrade screen. In the interest of lessening potential break points in 3.0, I'm going to punt this to 3.1.

PR #6875, opened 2018-08-25
issue/6874 - Move "Upgrades" page to "Tools"

JJJ:
WIP - See #6874.


spencerfinnell:
@JJJ FYI `edd-admin-tools-export` will need to be enqueued on this screen as well. Did not realize the export functionality contained the batch processing as well.

JJJ:
It looks like this may no longer be necessary, as of 3.0?

PR #6883, opened 2018-08-27
Fix ajax for adding item to order not finding variable prices #6882

### LINKED ISSUE #6882: 3.0 - Add New Order: Variable priced products do not insert correctly
### Expected behavior
When I add a new variably priced item to the 'Add New' order screen, the variable price name should show and the price should be found for the price ID.

### Actual behavior
When I add a new variably priced item to the `Add New` order screen, the product is shown with `Product Name -` and the price is $0.00

image

Be sure to use a product that has not had their prices re-ordered, it would be best to create a new product. For further testing, choose a product that has variable pricing and re-order all the variable prices to a different display position, so that the price ID no longer matches the 'order index'.
1) Add New Order
2) Click 'Add Item'
3) Select a variably priced item.

PHP Version: All

WordPress Version: All

cklosowski:
Fixes #6882

Proposed Changes:
1.Updates the logic for adding an item to an order via admin, to properly find the price ID, since `index` is not the value determined by the price ID, but the array key is.

### DIFF
--- a/includes/ajax-functions.php
+++ b/includes/ajax-functions.php
@@ -898,6 +898,8 @@ add_action( 'wp_ajax_edd_search_users', 'edd_ajax_search_users' );
 /**
  * Search for download, build, and return HTML.
  *
+ * This is used in the Admin for Adding items to an order.
+ * 
  * @since 3.0
  */
 function edd_ajax_add_order_item() {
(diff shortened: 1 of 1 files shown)


includes/ajax-functions.php
```diff
@@ -944,9 +946,10 @@ function edd_ajax_add_order_item() {
 		if ( false === $download['price_id'] ) {
 			$amount = floatval( $d->get_price() );
 		} else {
-			$prices = wp_filter_object_list( $d->get_prices(), array( 'index' => $download['price_id'] ) );
```
cklosowski: As a note here, since `index` is not a properly within the array, the `wp_filter_object_list` does not find a `price_id`. The `index` is used when re-ordering the price IDs to display in a different order. The Array keys are used to determine the price ID.

cklosowski: To be clear here, `index` is only used if you reorder the prices, in order to maintain their original state.

Basically `index` is not the `price_id` it is the order they should be displayed in. The `price_id` is the array key, which is what we want here, not the `display order` value.

image

pippinsplugins:
Replicated and confirmed the fix!

PR #6897, opened 2018-08-30
Improve Drag and Drop UI for edd-sortable-list.

### LINKED ISSUE #6872: 3.0: Payment Gateway Icons - Improve display
I can't capture it in the screenshot but the cursor is set to `move` when you hover over each `label` in either the `Payment Gateways` section, or the `Payment Method Icons` section. This is confusing since the primary interaction should be to _select_ a checkbox when clicking on a label, not to move it.

I propose we only make the cursor show the `move` icon...

screen shot 2018-08-24 at 4 04 18 pm

When you interact with the draggable icons to the far right.

screen shot 2018-08-24 at 4 00 39 pm

___

The icon used for drag and drop is `'\f228'`:

screen shot 2018-08-24 at 4 12 11 pm

This is typically an icon used for menus, and is used already in the WP admin menu bar at mobile sizes:

screen shot 2018-08-24 at 4 09 09 pm

I think we should replace the icon we are using with something more suitable, perhaps dashicons-move?

screen shot 2018-08-24 at 4 17 26 pm

spencerfinnell:
Fixes #6872

Proposed Changes:
- Only show "move" cursor when not hovering a label or checkbox.
- Only show "move" icon when hovering a list item.
- Change "move" Dashicon to match the cursor.
- Add a tooltip to let users know this can be rearranged.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -626,6 +626,8 @@ function edd_get_registered_settings() {
 						'desc'    => __( 'Choose the payment gateways you want to enable.', 'easy-digital-downloads' ),
 						'type'    => 'gateways',
 						'options' => $gateways,
+						'tooltip_title' => __( 'Adjust the Order', 'easy-digital-downloads' ),
+						'tooltip_desc'  => __( 'Drag and drop the payment gateways to determine the order they appear in the checkout. Default gateway always appears first.', 'easy-digital-downloads' ),
 					),
 					'default_gateway' => array(
 						'id'      => 'default_gateway',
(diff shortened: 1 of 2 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -647,6 +649,8 @@ function edd_get_registered_settings() {
 							'discover'        => 'Discover',
 							'paypal'          => 'PayPal'
 						) ),
+						'tooltip_title' => __( 'Adjust the Order', 'easy-digital-downloads' ),
+						'tooltip_desc'  => __( 'Drag and drop the payment method icons to determine the order they appear in the checkout.', 'easy-digital-downloads' ),
```
cklosowski: This tooltip seems redundant, as we have it above as well on the `gateways` setting as well.

JJJ:
I commented on the original issue. I do not prefer the design changes in this PR.

I don't think any of these proposed tweaks look or feel better than what is already in `release/3.0`.

I'd motion to table this PR until after 3.0 ships, and if user-complaints come in about these sortables, we revisit the work that's been done here, then.

PR #6900, opened 2018-08-30
Allow [downloads] shortcode inside a single download's content. 

### LINKED ISSUE #6072: Fatal error when using [downloads] shortcode
I did something really dumb and caused a fatal error. As silly as it was, it should probably fail silently.

I placed the `[downloads]` shortcode on a single download page...

```
PHP Fatal error:  Maximum function nesting level of '256' reached, aborting! in /wp-includes/cache.php on line 724
```

spencerfinnell:
Fixes #6072

- Updates shortcode-content-excerpt.php template file.
- Updates shortcode-content-full.php template file.
- Does not pollute the global excerpt_length  filter.
- Ensures no download buttons are appended to excerpts.
- Ensures full content can still be output and non-[downloads] shortcodes rendered.

Proposed Changes:
1. Allow [downloads] shortcode output in single download content.


includes/template-functions.php
```diff
@@ -1139,3 +1139,66 @@ function edd_download_shortcode_item( $atts, $i ) {
 	edd_get_template_part( 'shortcode', 'download' );
 }
 add_action( 'edd_download_shortcode_item', 'edd_download_shortcode_item', 10, 2 );
+
+/**
+ * Output full content for a download item in the [downloads] shortcode.
+ *
+ * Strips the [downloads] shortcode to avoid recurssion.
```
SDavisMedia: `recurssion` -> `recursion`

spencerfinnell:
@cklosowski I fixed merge conflicts. The comments mark `@since 3.0` -- I can update those if you want to get this out sooner rather than later but the issue has been around forever so there is no rush.

cklosowski:
Since I moved this to the `release/3.0` merge it obviously caused some conflicts, I'll fix these up @spencerfinnell otherwise it looks solid to me.

spencerfinnell:
@cklosowski Addressed feedback and seems to be working more consistently now. Thanks for the testing.

PR #6902, opened 2018-08-31
Tax Rate UI Updates

spencerfinnell:
- Surfaces the fact that tax rates cannot be deleted by showing them as deactivated.
- Enables simplified rate saving because all of the data will exist in the page. (this needs to be implemented)


spencerfinnell:
Pushed some updates. Now ensures the rows are all loaded on the page (hidden if deactivated, instead of not being rendered) so all rates can be passed through the sanitization method. This can be removed in the future when individual rates are updated via AJAX when there is an action.

I tried implementing the simplified saving but for some reason am still ending up with duplicate entries when trying to change the status of an existing rate.

spencerfinnell:
Saving seems to be working well now.

spencerfinnell:
@JJJ I agree about the UI points for sure; but not sure what to change. Experimented with it as a separate form instead of attached to the table. Having the form at the top doesn’t fix the scrolling problem because once you add one it will be out of view of there are many. I think a future iteration that filters by countries by default would help with this problem. 

Definitely think more auto AJAX updates in the next iteration would be nice as well. 

(P.S pretty sure official WP JS coding standards enforces trailing commas)

pippinsplugins:
I'm okay with the potential for a scrolling issue. It's always been that way and it's never been a problem yet.

There are two main things I'd like to see adjusted:

1. Do not permit navigating away from page without confirmation if any changes have been made
2. `Apply to whole country` and the state/province dropdown need to be placed next to each other instead of on top of one another.

Love it

JJJ:
1 is easily achieved.
2 is less easy, simply due to lack of horizontal real estate, and width issues with showing/hiding Chosen.

Excuse my linking to a Wikipedia page, but it seems like countrywide taxes are a decent default: https://en.m.wikipedia.org/wiki/List_of_countries_by_tax_rates

With that in mind, I don’t _love_ the region being on a new line, but it’s not terrible.

spencerfinnell:
Added a confirmation when rows are added. Doesn't mark individual changes to rows which is again something that will be improved with auto AJAX saving in the future.

The horizontal space is the issue for the side by side and really the whole UI falls apart on small devices (tables are no good for this). I'll try to branch off this and experiment with a more standard form that allows more flexibility if I can. But otherwise I think this is good to merge back in to @cklosowski's branch.

JJJ:
The AYS message incorrectly triggers when clicking "Save Changes" directly.

## Steps to duplicate:
1. add a tax rate
2. click "Save Changes"
3. see the AYS

spencerfinnell:
@JJJ Try that ^

JJJ:
@spencerfinnell now it's only working intermittently, specifically in Safari.

I tried bumping the version number to bust the asset cache, but it didn't make a difference.

Oddly it does seem to work consistently in Chrome.

JJJ:
Confirmed that other AYS messages are broken in Safari (version `12.0 (14606.1.36.1.9)`).
Working OK in Chrome. ✅ 

Sorry for the tangent.

JJJ:
@pippinsplugins do we want to stick with country/region codes, or do we want full-text labels, or some combination of the 2?

Right now, we only have codes:
screen shot 2018-09-03 at 11 04 06 am

JJJ:
WordPress list-table striped CSS targeting stinks when removing/hiding rows. Ignore that here for now. We can't really fix it without extra JavaScript to juggle some `even|odd` class implementation, which I'd prefer to avoid.

I think we should add styling to the `no-items` row, and the `tfoot` mock form area to separate them out, and not rely on the `striped` class.

JJJ:
I just noticed the Backbone files are stored in `includes/templates/admin/`. I don't think that's correct. I think files in the `templates` directory are front-end only, as part of a theme-compatibility API.

JJJ:
screen shot 2018-09-03 at 11 31 37 am

spencerfinnell:
@JJJ it was mainly so I could use template locator helper functions, but I can see why we would want to stop overrides. Feel free to move them somewhere and do a standard ‘include’

spencerfinnell:
Moved the template files.

@JJJ @pippinsplugins Showing the full country name is easy, but in my opinion makes the table harder to scan. And showing the full region name would be more complicated and again causes the same visual issue.

New inactive state looks great.

JJJ:
Pushed more styling improvements:

screen shot 2018-09-03 at 6 25 56 pm

This is ready for final review & merge IMO.

JJJ:
Just noticed a final quirk related to bulk-actions.

When choosing to hide deactivated tax rates, it's possible to have all rows hidden, which leaves the list-table looking a little weird:

screen shot 2018-09-03 at 6 28 14 pm

spencerfinnell:
Another annoying one solved by AJAX in the future. Will see what I can do.

JJJ:
Killed 2 birds with 1 toggle by creating an `edd-toggle` UI out of a checkbox. Now we can use this in our settings and meta-boxes in places where a toggle makes more sense than a checkbox.

screen shot 2018-09-03 at 9 31 54 pm

JJJ:
RTL support added

screen shot 2018-09-03 at 9 54 31 pm

sunnyratilal:
I’d suggest having a toggle to display long names. In the UK, we don’t refer to regions using the abbreviations that EDD uses. E.g. Surrey in the United Kingdom is displayed as `GB-SRY` which isn’t a “recognised” abbreviation.

spencerfinnell:
@JJJ Looks cool! If you haven't modeled it after https://github.com/WordPress/gutenberg/tree/master/packages/components/src/form-toggle I would take a look at that to see how they've handled some accessibility things.

@sunnyratilal I can definitely see the advantage there. The issue is that the full name isn't stored anywhere for each rate (as far as I can tell) so it would be a matter of printing each countries state list in to the page and then mapping based on the code. Not a huge deal but something to consider. Did the original implementation use the full region name?

mintplugins:
Would it be possible to add a column for "Orders", which would show a link to a list of order which used that tax rate? Here's a quick/dirty mockup of what I'm thinking:

screen shot 2018-09-04 at 10 31 13 am

JJJ:
> I’d suggest having a toggle to display long names

I strongly prefer sticking to only having 1 toggle in this UI. More than 1 starts to make this gnarly to render correctly, juggling the various states.

We _could_ maybe modify the label text and behavior to be a "Show Extras" instead? That would show/hide inactive rates _and_ the full-text labels of each country/region?

> take a look at that to see how they've handled some accessibility things.

@spencerfinnell _hopefully_ we don't have many of those same issues, thanks to us continuing to use an `input` element, and not relying on an SVG to draw our mock control. I have no doubts my design here could be improved (re: sizes, borders, contrast, etc...) but I'm also comfortable iterating on those things as feedback comes in

> Would it be possible to add a column for "Orders"

I'm all for connecting things together, so here's my thoughts on how we could do this.

We eliminated the "View Order Details" link from the Order History page, mostly because it's wordy, and also because it's out-of-place in the list-table UI how it's conventionally used. We also have some width constraints with this living in Settings, as we lose a lot of screen real-estate to the table label.

I think we _could_ go vertical more easily, without smooshing the table cells together more, though I also like the simplicity of this UI as it exists today. We also could add another link to Actions, or we could make actions more like traditional `row-actions` in list-tables, where they are hidden by default, and appear on `:hover`. That would free up a table column, too.

JJJ:
Here's a question: does this setting still need a description, or is it obvious enough we can omit it?

screen shot 2018-09-04 at 10 44 15 pm

pippinsplugins:
I prefer the description being shown.

JJJ:
This is feature complete and ready to merge.

@pippinsplugins can we get a review?

jz

spencerfinnell:
Updated. This is based off @cklosowski branch originally but I think @JJJ has been keeping it up to date with 3.0 so it should be able to be merged in there directly if needed.

PR #6919, opened 2018-09-09
only apply discounts to non-free cart items #6901

### LINKED ISSUE #6901: Division by zero error on checkout if discount + $0 product in cart
### Expected behavior
When I have a $0 product in the cart with a discount code, I expect the price to be $0. 

### Actual behavior
The page fails with a warning
`Warning: Division by zero in /app/public/wp-content/plugins/easy-digital-downloads/includes/cart/class-edd-cart.php on line 688`

1) Add a $0 product to the cart
2) Add a non-free product to the cart
3) Add a discount code to the cart
4) Remove the non-free product from the cart
5) You should see there error here

PHP Version: 7.2

WordPress Version: 5.0-alpha-43556

SDavisMedia:
Fixes #6901 

Proposed Changes:
1. Do not attempt to apply discounts to free cart items

### DIFF
--- a/includes/cart/class-edd-cart.php
+++ b/includes/cart/class-edd-cart.php
@@ -647,7 +647,8 @@ class EDD_Cart {
 
 		$discounts = false === $discount ? $this->get_discounts() : array( $discount );
 
-		if ( ! empty( $discounts ) ) {
+		// If discounts exist, only apply them to non-free cart items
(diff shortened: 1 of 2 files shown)


cklosowski:
I added some unit tests, which will trigger this line of code to be executed.

When running these tests on `master` we get the error of `division by 0`, on the `issue/6901` branch, these are not executed and all tests pass.

Once CI passes, I'll merge.

PR #6933, opened 2018-09-14
Issue/6931

### LINKED ISSUE #6931: 3.0 - View Order: Variably priced products do not list the price name in the purchased items.
### Expected behavior
When a variably priced product is purchased, the price name should show in the view order details.

### Actual behavior
The variably priced name is not showing.

1) Purchase a variably priced product
2) View the order details in admin
3) The Price ID name is not shown.

PHP Version: Any

WordPress Version: All

screenshot_2018-09-14 edit download 3dd easy digital downloads wordpress
screenshot_2018-09-14 edit order orders 3dd easy digital downloads wordpress

cklosowski:
Fixes #6931

Proposed Changes:
1. Ass a new `get_full_product_name` method to the `Order_Item` class
2. Updates the Order Items list table to use this new method.

### DIFF
--- a/includes/admin/payments/class-order-items-table.php
+++ b/includes/admin/payments/class-order-items-table.php
@@ -270,7 +270,7 @@ class Order_Items_Table extends List_Table {
 		$order_item_title = '<strong><a class="row-title" href="' . add_query_arg( array(
 				'edd-action' => 'edit_order_item',
 				'order_item' => $order_item->id,
-			), $base ) . '">' . stripslashes( $order_item->product_name ) . '</a>' . $state . '</strong>';
+			), $base ) . '">' . stripslashes( $order_item->get_full_prodcut_name() ) . '</a>' . $state . '</strong>';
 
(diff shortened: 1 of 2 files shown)


JJJ:
The PR fixes the bug, but I think instead we should just save the full product name when the sale is completed.

That maintains the integrity of the original order, and also sticks to the "snapshot" concept for order item data. That way if something gets renamed or deleted, the original (correct) name is maintained.

JJJ:
On second thought, I think this PR has to be the way it is.

Previous versions of EDD only stored the original item name and tacked the price variation on the end like this PR suggests. But if we start storing the name and price together now, old order items won't match – we'd need to also update the upgrade routine, and it risks add-ons double-adding price variation names.

I don't love it, but I think this PR is the way to go.

----

Tangent: we don't use "full name" anywhere in EDD. I think we should stick with "price name" or use something else instead.

cklosowski:
@jjj :+1:

PR #6954, opened 2018-09-26
Issue/6953 - Remove background/border on CC fields

### LINKED ISSUE #6953: 3.0 - Remove background color from credit card fields
I propose we remove the grey background color (`#f9f9f9`) from the credit card field at checkout Since we're making other checkout improvements with v3.0 I feel this is worth the discussion.

I believe it was originally added to make the credit card field "stand out" and attempt to improve conversion rates, though I cannot find any evidence to support this.

By removing the background color we would:

1. Simplify the design of the checkout, making it cleaner out of the box. A lot of checkouts I've seen don't add a background color for the credit card fields (see: http://nicelydone.club/patterns/checkout/)
2. Simplify the checkout CSS.
3. Increase theme compatibility, reducing the chance of theme conflicts. By setting the background color to `#f9f9f9` we're making a very specific color choice. I'd imagine many themes would change this anyway to match their theme's "grey" color or perhaps something different altogether. The fieldset still has a CSS classname so any theme that has changed it will still see their own color.

We should strive to make EDD look as good as possible on default WordPress themes, since a lot of customers use them. Here's the Twenty Sixteen theme out of the box with EDD. Note the grey input fields on the grey background which makes the fields difficult to see:

screen shot 2018-09-26 at 8 48 12 am

Proposed changes:

1. Remove the background property and value from `#edd_checkout_form_wrap fieldset#edd_cc_fields`

2. Remove the `border-color`property and value (`1px solid #f0f0f0`) from `#edd_checkout_form_wrap fieldset#edd_cc_fields`. This isn't needed since it will fallback to the border color we use for the other fieldsets.

3. Remove `position: relative` from `#edd_checkout_form_wrap fieldset#edd_cc_fields`. I see no reason for this.

Any thoughts on keeping it? Removing it?

amdrew:
Fixes #6953


JJJ:
Merged into `release/3.0`. Thanks @amdrew and everyone else!

PR #6955, opened 2018-09-27
Pass all available contextual arguments to edd_requested_file filter.

clifgriffin:
Proposed Changes:
1. Provides `$args` array to edd_requested_file filter, providing additional context 

This is necessary for plugins that need to make changes to the download delivery that uses information from the purchase or purchaser.


cklosowski:
I can't think of a reason this would cause any issues.

@SDavisMedia If you're good with it, drop an approval and merge.

PR #6963, opened 2018-10-02
Strip slashes from names when updated by profile form

zackkatz:
Customer names like `O'Connor` currently get saved as `O\'Connor`.

Proposed Changes:
1. Strip slashes from customer names when updating via the profile form


SDavisMedia:
#6964

Issue #4708, opened 2016-06-22
average earnings data doesn't account for fees


SDavisMedia:
"`edd_get_average_monthly_download_earnings` incorrect when fees present due to `EDD_Download()` earnings property" ~ @cklosowski 

As shown here:

![2016-06-22 at 10 01 am](https://cloud.githubusercontent.com/assets/2359131/16275946/3e70c140-3861-11e6-880c-a4fe3a12ee1d.png)

That's 2 purchases of a $10 item with a $1 fee. $11 \* 2 for $22. The average earnings shows $20, though.

chriscct7:
Adding to 3.0 release as related ticket #6859 caused by the same root issue, has a pr (#6969) which also fixes this issue (once totals are re-done using the tool). We're running the recalculate functions anyway in the 3DD migration routine, so merging of the PR will also resolve this issue out as well.

chriscct7:
Closed with #6969 as well.

PR #6969, opened 2018-10-03
Account for Fees of Cart Item Price

### LINKED ISSUE #6859: Cart contents price is incorrect when using fees
# Description:
When you attach a fee in EDD to an item, the $item['price'] is set incorrectly. This causes:
- Order report totals to be wildly incorrect
- Extensions that utilize this feature get the wrong numbers (which if fed into accounting systems or used in tax audit software (to save an independent-of-edd-record) or to back calculate taxes paid (to say generate the taxes to pay to a state)) will be completely wrong

# Instructions (to replicate):
To make this easy, I'm going to use Discounts Pro, but this affects *all EDD Fees* it does not require Discounts Pro.

1. Install Discounts Pro
2. Make a download with a price of $20
3. Give this download a $20 discount in Discounts Pro
4. Hook into edd_payment_saved (picked for simplicity in testing, any hook works, even non-hooks, ie new EDD_Payment also works).
5. Dump $payment_meta = edd_get_payment_meta( $payment_id );
 in this data, $payment_meta['cart_details'] is the cart details, which contains the prices of each product. For each product in the cart details $item['price'] is supposed to contain the "net" (post-tax, post-discount) number for the product ($item['item_price'] contains the pre-number)

Notice, that despite the order being free, and the Fee being directly applied to the product, and the cart details item information having that fee there as -20, it still shows it as $20.

image

Now, repeat this, but instead of using a fee (turn discounts pro off) use a discount code. Note how this adjusted to the correct ($0).

I've tested EDD versions between 2.9.4 and 2.6.0. They all seem to have this issue.

This is not a big deal to fix, but it needs to be fixed. And ideally there needs to be a routine that goes through previous orders (we could do this on the 3.0 since we're already migrating them, we could fix them as we insert them), and fixes this.

dancameron:
Fixes #6859

Proposed Changes:
1. Loop through the fees and add them to the total
2. Negative fees will decrease total.

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -1148,7 +1148,7 @@ class EDD_Payment {
 		}
 
 		$fees = 0;
-		if ( isset( $args['fees'] ) && ! empty( $args['fees'] ) ) {
+		if ( ! empty( $args['fees'] ) && is_array( $args['fees'] ) ) {
 			foreach ( $args['fees'] as $feekey => $fee ) {
 				$fees += $fee['amount'];
 			}
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -1147,7 +1147,16 @@ class EDD_Payment {
 			$subtotal -= round( $tax, edd_currency_decimal_filter() );
 		}
 
-		$total      = $subtotal - $discount + $tax;
+		$fees = 0;
+		if ( isset( $args['fees'] ) && ! empty( $args['fees'] ) ) {
+			foreach ( $args['fees'] as $feekey => $fee ) {
+				$fees += $fee['amount'];
+			}
+
+			$fees = round( $fees, edd_currency_decimal_filter() );
+		}
+
+		$total      = $subtotal - $discount + $tax + $fees;
 
 		// Do not allow totals to go negative
 		if( $total < 0 ) {


chriscct7:
This also resolves #4708

includes/payments/class-edd-payment.php
```diff
@@ -1147,7 +1147,16 @@ public function add_download( $download_id = 0, $args = array(), $options = arra
 			$subtotal -= round( $tax, edd_currency_decimal_filter() );
 		}
 
-		$total      = $subtotal - $discount + $tax;
+		$fees = 0;
+		if ( isset( $args['fees'] ) && ! empty( $args['fees'] ) ) {
```
chriscct7: Ideally this should be
`if ( ! empty( $args['fees'] ) && is_array( $args['fees'] ) ) {`

since empty = isset() && !== false, therefore a separate isset isn't needed, though we do need it to be an array to use foreach

dancameron: agreed

chriscct7:
Restarted the travis build because on one of the tests Travis died, but all others completed and passed just fine

chriscct7:
The build is failing because on the  PHP: 5.5 WP_VERSION=4.4.2 WP_MULTISITE=0 combo the Travis verification fails. All other 50+ ones pass. This fail can be ignored.

JJJ:
Tested, and resolves the bug in #6859.

@cklosowski - OK to merge?

cklosowski:
Well crap that was for `master` but the current issue attached is for `release/3.0` it's a long standing bug so we can release it as soon the next EDD Core release is ready and port it to `release/3.0` @JJJ

JJJ:
Yep! I will merge. No problemo!

PR #6972, opened 2018-10-04
Isolating fees by ID when doing a cart recovery #6971

### LINKED ISSUE #6971: Fees and Cart recovery can end up with duplicate fees due to key naming
### Expected behavior
When resuming a payment, all fees should match what they previously were and only show the proper number of times.

### Actual behavior
In an edge case with Discounts Pro (where a specific ID is used), you can end up with the same Fee showing twice (only applying once however) due to the fact that the item based fees don't have the ID defined in the array, where as the global fees array does.

1) Setup a Discounts Pro Discount
2) Add item to cart so Discounts Pro discount applies
3) Checkout using PayPal but do not complete purchase
4) Go to Purchase History and click Complete Purchase for the abandoned order
5) Cart should show duplicate discounts

PHP Version: All

EDD Version (or branch): `master`

WordPress Version: All

This was discovered and reported with Discounts Pro here: https://github.com/easydigitaldownloads/edd-discounts-pro/issues/147

cklosowski:
Fixes #6971

Proposed Changes:
1. Isolates fee IDs when doing a cart recovery to avoid duplicating their output

### DIFF
--- a/includes/payments/actions.php
+++ b/includes/payments/actions.php
@@ -472,14 +472,16 @@ function edd_recover_payment() {
 
 		// Recover any item specific fees.
 		if ( ! empty( $download['fees'] ) ) {
-			foreach ( $download['fees'] as $fee ) {
+			foreach ( $download['fees'] as $key => $fee ) {
+				$fee['id'] = ! empty( $fee['id'] ) ? $fee['id'] : $key;
 				EDD()->fees->add_fee( $fee );
 			}
 		}
 	}
 
 	// Recover any global fees.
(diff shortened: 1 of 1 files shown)


cklosowski:
@mintplugins I know that @SDavisMedia can't replciate the issue to begin with, but the list of people who replicated and confirmed the fix are:
1) You (@mintplugins)
2) @dgoldak 
3) @misulicus 

Being that this does affect checkout, that's 3 confirmations of the replication and fix, so if you could approve/request changes on this PR that'd be :+1: in my book :)

PR #6975, opened 2018-10-08
Reset the attachment ID to 0 if the file URL is changed after blur #6713

### LINKED ISSUE #6713: Download File URL inconsistency when manually changing
### Expected behavior
When manually changing a download`s file name and File URL by copy/pasting the full file link it should download the new uploaded file. File name changes correctly

### Actual behavior
The new file name changes but when clicking a download link it still serves the previous file url.

1) Create a Download and Upload a file and give it a file name
2) Upload another download file via WP media manager and copy the full file path.
3) Edit the download and change the file name to a new name and manually select the old full file url and paste in the newly uploaded file. Save the download
4) try to redownload from the purchase history page, the new file name will be visible but the download will serve the originally uploaded file and not the new one uploaded via the media manager

PHP Version: 7+

WordPress Version: latest

cklosowski:
Fixes #6713 

Proposed Changes:
1. Adds JS to the download Edit screen that sets the hidden attachment ID field to 0 if the file url field is modified manually.


spencerfinnell:
@cklosowski This should be added to https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/js/admin/downloads/index.js now.

JJJ:
Merged `release/3.0` in (which includes `test/webpack`) and resolved conflicts, including what @spencerfinnell guided us to above.

Going to confirm this fix, and will merge after review.

JJJ:
This PR doesn't address the problem as I understand it. I think this is a two-way issue:

* Editing a file URL directly does not disconnect the attachment ID (reported above)
* Manually editing a file URL to something from the media library does not reassign the new attachment ID

I'm reluctant to add more client-side JavaScript code before addressing the issue that it's possible for the attachment ID and the URL string to be mismatched.

I think the attachment IDs should be verified and set on product "Update", and that attachment IDs via hidden form fields are not to be trusted until validated by `attachment_url_to_postid()`.

If I'm correct, that's quite a behavioral change, and it should be accompanied by unit tests to make sure it doesn't break other things.

spencerfinnell:
@JJJ I think that added check would be helpful.

So when a download is saved, each download file row that matches:

- local URL (`edd_is_local_file()`) 
- attachment ID of 0 (set by the client when a change is made manually)

Should search using `attachment_url_to_postid()` in an attempt to restore the attachment ID.

cklosowski:
@JJJ  we allow external file URLs, Dropbox URLs, Amazon S3 URLs. Therefore making sure the file exists within the media library isn't always necessary as the file might not be in the library if you are changing to an external source.

spencerfinnell:
@cklosowski shouldn't a quick check against `edd_is_local_file()` let it know to try and find it or not? Not sure the performance implications of using that function though.

spencerfinnell:
@cklosowski @pippinsplugins is current functionality enough of an enhancement for 3.0? Seems like a nice little improvement on original functionality and leaves room to expand further as discussed above after 3.0.

pippinsplugins:
I think simply removing the attachment ID when the URL changes is sufficient.

JJJ:
This seems good to go!

PR #6977, opened 2018-10-09
Modularize Javascript and Introduce Webpack

### LINKED ISSUE #3535: Modularize JS
Right now, all of EDD's JS is loaded on every EDD page. This makes things like the reports JS and flot among others loaded on pages that don't use it. 

We should break the JS into their own pieces so they can be included correctly:
1. Take edd_is_admin_page and only return true on EDD pages (not all post new/edit screens)
2. Introduce a new function that just does is post new/edit screen
3. Take the JS and CSS for "Insert Download" modal and put them in their own files. Then enqueue just them if the new function for 2 is true.
4. Take all of the other JS and CSS we enqueue and split them so that we can include them more specifically using the new edd_is_admin_page improvements we're doing for 2.4. It would be cool if each module was its own EDD function that enqueued the css/js for that module. For example edd_enqueue_admin_reporting() would enqueue the css and js for the reporting stuff. It should be designed so that it can be called by itself. That way extensions could simply call that function and EDD core would take care of all of the work if they wanted to use our reporting js/css. This would let extensions include EDD js and CSS without having to include all of EDD's js and css.
5. All of the EDD JS and CSS would be modularized into new JS and CSS files, so that core doesn't use any of the old ones. For backwards compatibility, the old JS and CSS files would then be converted into just including the new ones using http://api.jquery.com/jQuery.getScript/ and import respectively, and deprecated. We'd remove them if we want far into the future. Alternatively we could hook into the enqueue css and js functions and change the paths and what they are requesting.
6. Introduce a filter by which the presence of the "Insert Download" modal can be turned off (turning off the HTML output, as well as the CSS and JS output.
7. Change our jQuery UI to use the prefix as mentioned in #3033

This would solve probably all of our issues with backend plugin conflicts.

spencerfinnell:
Fixes #3535 

Proposed Changes:
1. Only load relevant javascript on admin pages.

To test:

It should be ensured that all scripts listed in https://github.com/easydigitaldownloads/easy-digital-downloads/tree/test/webpack/assets/js are loaded on their necessary pages.

---

This is my proposal for updating EDD's Javascript infrastructure. I consider this "Step 1" since this implementation does not change any of the code's base architecture, and only changes how it is loaded.

The major component of this proposal is using Webpack to help bundle the Javascript. My reason for introducing Webpack is eventually I would like to see it be used for handling all of EDD's assets, and the removal of Grunt.

Webpack is currently the bundler of choice for Gutenberg (all hail) and the majority of the principles for how it is being used in this PR are based off that. Webpack works with standard Node packages which means any of the existing Grunt-wrapper packages can easily be converted to their native libraries in the future (removing dependencies).

Webpack also allows trans/compilers like Babel (which is implemented in this proposal) to use "next generation" Javascript which is not currently supported in all browsers. This is the basis for implementing this PR as it allows for easy import/export functionality between Javascript files.

This PR also introduces ESLint and the standard WordPress coding standards for Javascript (again mainly taken from Gutenberg). Linting can be run by the `npm run lint` command and automatic fixes applied with `npm run lint:fix`

---

To make this implementation easier to test and to avoid changing the development process of the repository further, minified scripts and sourcemaps have been committed to this branch. This means if a user downloads the repository they will have access to a working plugin because all scripts are included. The committed scripts are automatically in development mode and their sourcemaps will allow for easy user debugging.

In my opinion it would be more ideal in the future to remove these scripts from the repository and require developers who work directly from the repository to run the necessary build scripts. This will ensure a better developer experience (cleaner diffs, up to date scripts, and locked dependencies) while ensuring the repository remains a development-only resource. Users will still be able to download built packages from tagged releases.

---

tl;dr Everything should work exactly how it was before. This creates no functional or visual changes.


spencerfinnell:
Here is an example of the types of simple refactors that comes much easier when only working with a single component: https://github.com/easydigitaldownloads/easy-digital-downloads/compare/test/webpack...js/dashboard

JJJ:
Merged `release/3.0` changes into `test/webpack` to prepare this for merging back the other direction.

Issue #4200, opened 2016-01-16
Add support for fees that include taxes


felixarntz:
Currently one can only add fees that exclude taxes. The plugin does not support fees which include taxes, it basically overrides the global setting by adding a filter in `edd_get_cart_fee_tax()` for example.

I'm proposing to add support for fees that include taxes, just like we have for regular download products. I'm currently using a pretty hacky solution to (kind of) fix this problem (which I can share if desired), but I would appreciate if EDD actually supported it in a solid way.

Before we approach this issue, we'd have to think about whether we would take the global setting into account for those fees or whether we made it an additional argument for every individual fee itself. I prefer the latter approach since fees can be of any kind and the way that they are handled may differ. Also, if we just took the global setting, we'd break backwards compatibility with people who have the "fees included" setting active but define their fees without tax.

Maybe we could add a new property to a fee called `tax_mode` and then allow values like 'none', 'include' or 'exclude'. 'exclude' would be the default (for compatibility), and we could migrate the `no_tax` property to this new one by setting the `tax_mode` to 'none' for those.
Of course this enhancement would go beyond that in terms of how taxes, subtotal and total are calculated, so we'd need to work on these processes as well.

Opinions on this?

pippinsplugins:
@felixarntz I'm not clear what you mean by we don't support taxes on fees. When `no_tax` is set to `false`, tax gets added to the fees.

felixarntz:
Yes, taxes on fees are supported, but they are always _added_ on top of the fee amount. I'm talking about the possibility to add fees where their taxes are automatically _included_ in the fee amount you specify. Similar like it is possible for regular products as well. Basically adding fee prices inclusive of tax. Right now we can only add fee prices exclusive of tax.

pippinsplugins:
Ah! That makes more sense, thanks :)

mintplugins:
This is causing incorrect tax amounts when using EDD Recurring sign-up fees if you have taxes set to be included in prices.

**Set prices to include tax:**
<img width="831" alt="screen shot 2016-11-11 at 10 35 43 am" src="https://cloud.githubusercontent.com/assets/1638128/20220221/a5e2af2e-a7fa-11e6-92ce-17de7534d407.png">

**Create recurring product with a signup fee:**
[...]

PR #6978, opened 2018-10-12
Remove override for exclusive of tax for fees

mintplugins:
#4200


mintplugins:
This is a work in progress. Do not merge. Needs major testing with all fee related plugins

PR #6980, opened 2018-10-17
Use ES6 Imports for recalculate_taxes()

spencerfinnell:
Previously the `edd-ajax` file was able to use the `recalculate_taxes()` function defined in another file because it was loaded before it.

With the introduction of Webpack and the subsequent self contained modules the function is no longer accessible directly.

This PR does two things:

1. Converts `recalculate_taxes` to use ES6 imports.
2. Assigns other previously exposed functions to the window namespace for backwards compatibility.


JJJ:
Confirmed. ES6 imports are fine by me. Merged! Thanks @spencerfinnell.

PR #7022, opened 2018-11-07
Move Report Chart Generation to a Separate Script

spencerfinnell:
Properly enqueuing ChartJS means the position it is loaded cannot be guaranteed. If it is used by another script that outputs in the footer it will be moved to the footer as well. This means the inline Javascript printed in the page to render charts will be run immediately and without the proper dependency (charts are broken in `release/3.0` currently).

This PR fixes that by building charts in a JS file that is dependent on the main reporting javascript, which can be properly enqueued.

I need a bit of input on some potential abstraction that could be done to facilitate the extra data that was added to the `build_config()` method in the chart manifest class. @DrewAPicture I think that's you?

To test:

1. `npm run dev` (I can commit the built scripts if needed)
2. View the reports and use the charts.

To do:

- [x] Pie charts aren't working. See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7023


pippinsplugins:
@spencerfinnell still working on this or ready for testing?

spencerfinnell:
Good to test! The compiled JS is included in the branch so you shouldn't need to run anything.

JJJ:
Looks good!

PR #7047, opened 2018-11-20
add missing nonce data attribute to country select - issue/7035

### LINKED ISSUE #7035: Edit Payment Screen - Address state cannot be updated when country changed
### Expected behavior

If you try to edit a payment that has no address fields pre=populated and change the `country` select field, the `state` select does not appear, so it's impossible to change the state properly.

### Actual behavior

As per the above.

1) Select any existing payment record. In my case, taxes were disabled.
2) Change the payment country select field
3) The state select does not appear.

PHP Version: 7.2

EDD Version (or branch): 2.9.8

WordPress Version: 4.9.8

SDavisMedia:
Fixes #7035

EDD 2.9.4 added nonces to AJAX actions. See: https://easydigitaldownloads.com/development/2018/07/09/important-update-to-ajax-requests-in-easy-digital-downloads-2-9-4/

Proposed Changes:
1. Add missing nonce attribute to the Country select field (chosen) on the Edit Payment screen. The corresponding nonce support already exists in the JS for editing the payment.

### DIFF
--- a/includes/admin/payments/view-order-details.php
+++ b/includes/admin/payments/view-order-details.php
@@ -733,6 +733,7 @@ $customer       = new EDD_Customer( $payment->customer_id );
 															'data'             => array(
 																'search-type'        => 'no_ajax',
 																'search-placeholder' => __( 'Type to search all Countries', 'easy-digital-downloads' ),
+																'nonce'              => wp_create_nonce( 'edd-country-field-nonce' )
 															),
 														) );
 														?>


SDavisMedia:
@JJJ Requesting your review here because you handled a similar issue back on EDD 2.9.5 (and 3.0), so you should be familiar. Thanks.

JJJ:
Works for me. Nice work. 🥇

Issue #6760, opened 2018-07-12
Bundle product selector: show only product variations if the product is variably priced. 


mintplugins:
### Expected behavior
When I create a bundle and choose child products within it, if the child product is variably priced, I expect to _only_ have the option of choosing variable prices. 

### Actual behavior
When I create a bundle and choose child products within it, if the child product is variably priced, I have the option of choosing variable prices (good) _and_ a non-variable priced version (bad). The non-variable price does not actually exist, and therefore should not be in the list. 

<img width="980" alt="screen shot 2018-07-12 at 1 11 45 pm" src="https://user-images.githubusercontent.com/1638128/42648810-e9b88e66-85d5-11e8-8251-248571dfdbd4.png">

### Steps to reproduce the behavior
1) Create a variably priced product. 
2) Create a bundle. 
3) In the "Bundled Download" area, notice how you can choose a version of your product without choosing a variable price. If chosen, it's unclear what the expectation should be, and really doesn't make sense in this context. 

### Information (if a specific version is affected):
PHP Version: Any

EDD Version (or branch): Any

WordPress Version: Any

Any other relevant information:
This has confused numerous customers. Here's one related ticket:
https://secure.helpscout.net/conversation/590847803/83208?folderId=749231

It's important to note that this could affect backward compatibility for some stores. Likely the best approach here would be to migrate any of these that are already-selected in bundles to default to the smallest variable available. For example:

If `Variable Product A` is selected in a bundle, automatically update it so that `Variable Product A: First Variation` is selected.

cklosowski:
We'll just have to update our Product selector where we allow 'show variations' to 'disable main product' as a flag. This should do the trick.

There are cases we'll want to include all variations of a product (like a discount code), but in the case of the bundle price selector we'll for sure want to allow that to only be to a price Id.

We could also use this in the 'files list' which would solve some support pain points when you have a mix of variable priced items in a bundle that you want to deliver different files for each variable price point.

mintplugins:
I would disagree about the discount code. If you have variable products, selecting a non-variable version of that product could easily be mistaken as meaningless, since a non-variable version of the product does not actually exist. It's an assumption on our part that the site owner would assume that selecting a non-variable version of the product means they are including all variations. 

I believe we should remove non-variable versions of variable products from all select fields, since those products do not actually exist. 

[...]

PR #7055, opened 2018-11-29
3.0 Fix Chosen Search

### LINKED ISSUE #7054: 3:0 Search Hidden on Chosen
### Expected behavior

Chosen should allow searching on certain inputs.

### Actual behavior

Search input is always disabled.

1) Try to add a download to a bundle. (#6760)

---

The search was effectively removed on the majority of Chosen instances with this commit: https://github.com/easydigitaldownloads/easy-digital-downloads/commit/61f1059c439af79ef5ee95394e77437de3169ba1#diff-4e07975920906985a5fbbe3afd720629R10

@JJJ I assume this was for things like settings that only have a few options?

Not sure the best way this. Perhaps another argument in the HTML class to disable search for that specific field?

spencerfinnell:
Fixes #7054 

- Ensures search field shows on Chosen inputs that require searching.
- Removes unchecked `console.log()` call.
- Uses Underscore's `debounce()` function instead of a timer. https://underscorejs.org/#debounce

Accidentally built off the PR for #6760 but merging this will make testing for that easier.


cklosowski:
Going to close this issue and merge it in with #6760

PR #7064, opened 2018-12-07
Fix JS errors when no charts are loaded in reports (only tiles)

### LINKED ISSUE #7063: 3.0 - Reports: Viewing Download Report throws JS error
### Expected behavior
No JS errors should be thrown when viewing the Downloads Report

### Actual behavior
Currently we're getting the following JS error on the Downloads reports page:
```
ReferenceError: eddAdminReportsCharts is not defined
 /wp-content/plugins/easy-digital-downloads/assets/js/edd-admin-reports.js?ver=1544204626:97
```

1) Visit `Reports > Downloads`
2) See JS Error

PHP Version: All

WordPress Version: All

Any other relevant information: Looks like we just need to fine tune the scripts loaded and called from the report view.

cklosowski:
Fixes #7063 

Proposed Changes:
1. Conditionally defines `_eddAdminReportsChart` and `charts` only if `eddAdminReportsCharts` is defined (which means we are loading a page with charts).

There are a lot of changes listed here, but it's mostly just because webpack rebuilt all the JS as I was doing the dev work.

The major change is this in `js/edd-admin-reports.js`
From:
```
var _eddAdminReportsChart = eddAdminReportsCharts,
  charts = _eddAdminReportsChart.charts;
```

To:
```
var charts = {};
if ( typeof eddAdminReportsCharts !== 'undefined' ) {
  var _eddAdminReportsChart = eddAdminReportsCharts;
  
  charts = _eddAdminReportsChart.charts;
}
```


spencerfinnell:
@cklosowski this needs to be fixed in the source file, not the built file.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/js/admin/reports/charts/index.js

It should look something like:

https://gist.github.com/spencerfinnell/99daa6d4514d5ac11232f91b4adc6b1d

Wraps the calls in an immediately-invoking function so it can safely be returned/terminated if the variables do not exist.

spencerfinnell:
Alternate branch: https://github.com/easydigitaldownloads/easy-digital-downloads/tree/issue/7063-2

cklosowski:
@spencerfinnell ah ok, I'll kill this PR and make one off yours.

PR #7066, opened 2018-12-10
[3.0] Fix currency type taking over all tooltip labels.

### LINKED ISSUE #7048: 3.0: Sales and Earnings report has a currency sign on the sale count label
The `Sales` label has a currency sign on the sale count when it clearly shouldn't.

screen shot 2018-11-21 at 12 51 58 pm

spencerfinnell:
Fixes #7048 

Proposed Changes:
1. Ensure tooltips do not show currency symbols on numeric values.

---

This fix relies on the registered chart's options to be in the same order as the chart's datasets. It looks like this is the case for existing charts but it should be made clear in any docs.

For example:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports.php#L523-L526

`sales` then `earnings`

and

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports.php#L530-L554

`sales` then `earnings` matches the same order in the array.

Chart.js doesn't use the array's keys so the actual `datasets` used to chart are numerically keyed, which is why the order now matters.


cklosowski:
@spencerfinnell going to fix up the merge conflicts and then merge this.

PR #7069, opened 2018-12-10
Use consistent return values in AJAX edd_ajax_customer_addresses

### LINKED ISSUE #7068: 3.0 - Add Order: JS error thrown when selecting an existing customer
### Expected behavior
When selecting an existing customer while adding a new payment, no JS errors should be thrown.

### Actual behavior
The following error is being triggered in JS when selecting an existing customer from the Add Order screen:
```
TypeError: Cannot read property 'html' of null
 /wp-content/plugins/easy-digital-downloads/assets/js/edd-admin-orders.js?ver=1544466215:263
```

1) Orders > Add New
2) Click 'Assign' under the customer section, and pick a customer from the chosen dropdown.

PHP Version: All

WordPress Version: All

spencerfinnell:
Also properly hides the address HTML if an invalid (or none) user is selected.

Fixes #7068


cklosowski:
This fixes #7068 as well.

Issue #7072, opened 2018-12-10
[3.0] Tax Calculation Fails on Order Creation With Quantities Disabled


spencerfinnell:
Related #7052 

### Expected behavior

When setting an address or doing something that affects taxes, the order items should be adjusted accordingly.

### Actual behavior

The value is incorrect (`NaN`)

### Steps to reproduce the behavior
1) Disable cart item quantities
2) Create an order, add an item.
3) Set the address to a region with a tax rate.

---

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/js/admin/orders/index.js#L324

If the quantity field is not rendered (quantities disabled) then this will be undefined and create invalid math.

spencerfinnell:
New PR attached.

PR #7073, opened 2018-12-10
[3.0] More consitent AJAX action returns

### LINKED ISSUE #7070: Add Consistent Return Values to AJAX Actions
### Expected behavior

When making an AJAX request the callbacks should be returning a valid JSON response.

### Actual behavior

When something goes wrong a lot of things call `edd_die( -1 )` which will just 500 error on a lot of servers, or display something random. This will cause a lot of these AJAX requests to fail without the ability to catch the error (like in #7068)

---

I can go through the functions with a `@since 3.0` tag and update these and the relevant JS callbacks.

spencerfinnell:
Fixes #7070, #7072 

Proposed Changes:
1. For new 3.0 AJAX actions (that are only tied to AJAX) remove `edd_die()` usage in favor of consistent JSON responses.


spencerfinnell:
Things that need to be tested:

- Recalculating taxes when editing/creating an order.
- Adding an adjustment to an order.
- Adding an item to an order

cklosowski:
@spencerfinnell Tax recalculation on adding orders seems to be broken in `release/3.0` anyway. I'm giong to get this merged and then we can move forward with the fixes for other things.

Your PR fixes a few other areas for the tax recalculation that are broken for what's in `release/3.0` so it's a step in the right direction, and i'll log a new issue.

PR #7075, opened 2018-12-10
Auto fill the postal code when an address is selected on the add order #7074

### LINKED ISSUE #7074: 3.0 - Add Order: Customer Postal Code not being populated when Address is Selected from dropdown.
### Expected behavior
When I select an address on the Add Order Screen, I expect all available fields to be populated

### Actual behavior
Not all fields are populated. For example I have two addresses for a user:
`Arizona, 85142, United States`
`85142, United States`

In both cases, only the State and Country are populated (where relevant), but the postal code is not populated.

1) Have a customer with addresses that contain a State, Zip, and Country
2) Click 'Add Order'
3) Select the customer from step 1 in the Customer tab
4) Select the addresses for the customer.

PHP Version: All

WordPress Version: All

cklosowski:
Fixes #7074

Proposed Changes:
1. Adds the postal_code as an auto filled field when changing the address while adding an order


spencerfinnell:
Does the response ensure all these values exist? Otherwise they should have checks that the object property is available.

cklosowski:
@spencerfinnell the response is normalized, to have all address fields in the response, empty or not:
```
{"success":true,"data":{"addresses":{"2":{"id":"2","customer_id":"1","type":"billing","status":"active","address":"","address2":"","city":"","region":"AZ","postal_code":"85142","country":"US","date_created":"2018-12-10 18:23:26","date_modified":"2018-12-10 18:23:26","uuid":"urn:uuid:201f82b2-0f05-48e1-a139-d19af024ef3c"},"1":{"id":"1","customer_id":"1","type":"billing","status":"active","address":"","address2":"","city":"","region":"","postal_code":"85142","country":"US","date_created":"2018-09-11 22:32:05","date_modified":"2018-09-11 22:32:05","uuid":"urn:uuid:4b0ba0a2-17a3-4ca5-9abf-6662ef99f4ce"}},"formatted":{"2":"Arizona, 85142, United States","1":"85142, United States"},"html":"<select data-nonce=\"7bc351e5a7\" data-placeholder=\"Select a previously used address\" class=\"edd-select-chosen  add-order-customer-address-select\"><option><\/option><option data-key=\"2\" value=\"2\">Arizona, 85142, United States<\/option><option data-key=\"1\" value=\"1\">85142, United States<\/option><\/select>"}}
```

PR #7086, opened 2018-12-14
[3.0] Remove Gray Box on Vertical Sections

### LINKED ISSUE #7080: 3.0 - Reports: Gray box appears on larger screens in the bottom-left corner
On Reports tabs, there is a gray box at the bottom-left of each content area that grows and shrinks with the window size:

Little gray box!

Here are my browser details: https://whatsmybrowser.org/b/YPS6QPF

Here's a video showing it expanding: http://share.gravityview.co/TdCFkn

spencerfinnell:
Fixes #7080

Proposed Changes:
1. Remove Gray Box on Vertical Sections 3.0 UI component.

---

A more ideal solution is to just use `flexbox`, but I think IE9 is still supposed to be supported? https://caniuse.com/#search=flexbox -- instead it uses some semi-hacky `table` solution to the columns are equal height.

This also (what I assume) fixes the active state for items based on the default admin color scheme.

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -2810,7 +2810,6 @@ body.download_page_edd-reports {
 		font-size: 18px;
 	}
 	.edd-item-has-tabs  .edd-sections-wrap .section-wrap {
-		border-top: 1px solid #e5e5e5;
 		border-left: 0;
 		margin-top: -1px;
 	}
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -2518,6 +2518,7 @@ body.download_page_edd-reports {
 	height: 0;
 }
 .edd-sections-wrap {
+	display: table;
 	clear: both;
 }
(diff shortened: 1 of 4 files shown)


spencerfinnell:
Eh this solution breaks pretty easily. `flexbox` is the real solution. Only `4.5%` globally don't have support for it.

spencerfinnell:
Okay pushed a fix that seems to work better.

JJJ:
This has broken something.

When resizing the browser window, the `section-wrap` area no longer expands to fit.

screen shot 2018-12-18 at 10 50 09 am
screen shot 2018-12-18 at 10 50 14 am
screen shot 2018-12-18 at 10 50 53 am

JJJ:
In addition, the vertical tabs no longer resize. They are fixed at `200px` width, which is not part of the original design.

JJJ:
Here is a screenshot from `release/3.0` for comparison.

screen shot 2018-12-18 at 10 56 20 am

Issue #7093, opened 2018-12-19
PayPal Standard: add PayPal email to customer during IPN processing


pippinsplugins:
## Enhancement Request

When processing PayPal IPN requests, I'd like to see the PayPal account email, if different than the purchasing customer's email, recorded on the `EDD_Customer` object as an additional email.

### Justification or use case

It's really common for the PayPal email address to be different than the email entered on checkout. By recording the PayPal email on the customer record, we can make it a lot easier to locate customers and purchase records when given a PayPal email.

For example, when a payment dispute is opened on PayPal, PayPal provides us with the email address associated with the person/business that is disputing the payment. This email rarely matches the email that was entered into the EDD checkout. This results in a cumbersome lookup process where we have to frequently search for and locate the relevant EDD purchase record through the recurring profile ID or transaction ID (which has issues when using Recurring Payments). Having the PayPal email already on the customer record would make this process a lot simpler and will make our data more complete.

PR #7094, opened 2018-12-19
Record the payer email on the EDD_Customer record during PayPal IPN if it is different than the email entered on checkout - issue/7093

pippinsplugins:
#7093

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -616,9 +616,9 @@ function edd_process_paypal_web_accept_and_cart( $data, $payment_id ) {
 	}
 
 	// Record the payer email on the EDD_Customer record if it is different than the email entered on checkout
-	if( ! empty( $data['payer_email'] ) && $data['payer_email'] != $customer->email ) {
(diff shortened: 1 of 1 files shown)


includes/gateways/paypal-standard.php
```diff
@@ -609,6 +609,19 @@ function edd_process_paypal_web_accept_and_cart( $data, $payment_id ) {
...
 	}
 
+	if( empty( $customer ) ) {
+
+		$customer = new EDD_Customer( $payment->customer_id );
+
+	}
+
+	// Record the payer email on the EDD_Customer record if it is different than the email entered on checkout
+	if( ! empty( $data['payer_email'] ) && $data['payer_email'] != $customer->email ) {
```
cklosowski: This needs to be slightly changed:
1) We should be doing a `strtolower` on the email from paypal.
2) We need to search all emails for the customer, not just the primary, so we aren't trying to add an email address that we might have previously added

Keeping in line with the first item, we'll have to do an `array_map` on the emails for the customer, and do a `strtolower` and then also `strtolower` the email from paypal, that way we can be sure that we're comparing the same strings and not adding emails when they possible have different case characters and we can finish off the check with a simple `in_array` check of all the now matched-case email addresses.

pippinsplugins: @cklosowski changes pushed.

PR #7106, opened 2019-01-06
Add multi-step determination of payment gross in PDT validation. Resolves #7105

### LINKED ISSUE #7105: PDT amount validation sometimes fails due to some accounts not sending payment_gross
We currently use the `payment_gross` field during PDT for validating the payment amount. In some cases, this can result in the validation failing because the `payment_gross` field is not present.

Example report: https://wordpress.org/support/topic/pdt-validation-problem/#post-11057229

PayPal's documentation is inconsistent and unclear on which of these three variables should be used:

- amt
- payment_gross
- mc_gross

In some places they say `payment_gross` is deprecated; in some places they show PDT coded examples using `payment_gross`; and in some places they show a `amt` field that is specific to PDT.

What @cklosowski and I have found:

- `mc_gross` appears to always be set
- `payment_gross` is deprecated and only sometimes set
- `amt` doesn't seem to exist at all in PDT data

We have decided to do a four step check to be certain:

1. Check if `mc_gross` is present and use it if so
2. If `mc_gross` is not present, check for `payment_gross` and use it
3. If `payment_gross` is not present, check for `amt` and use it
4. If `amt` is not set, bail and leave payment as pending

Issue #4008 corroborates the theory that `mc_gross` is always set, even though PayPal documentation possibly indicates otherwise.

pippinsplugins:
#7105

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -947,13 +947,31 @@ function edd_paypal_process_pdt_on_return() {
 					$data[ urldecode( $parsed_line[0] ) ] = urldecode( $parsed_line[1] );
 				}
 
-				if ( ! isset( $data['payment_gross'] ) ) {
+				if ( isset( $data['mc_gross'] ) ) {
+
+					$total = $data['mc_gross'];
+
+				} else if ( isset( $data['payment_gross'] ) ) {
+
+					$total = $data['payment_gross'];
+
+				} else if ( isset( $_REQUEST['amt'] ) ) {
+
+					$total = $_REQUEST['amt'];
+
+				} else {
+
+					$total = null;
+
+				}
+
+				if ( is_null( $total ) ) {
 
 					edd_debug_log( 'Attempt to verify PayPal payment with PDT failed due to payment total missing' );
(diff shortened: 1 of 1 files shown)


includes/gateways/paypal-standard.php
```diff
@@ -947,13 +947,31 @@ function edd_paypal_process_pdt_on_return() {
...
+
+				}
+
+				if ( is_null( $total ) ) {
 
 					edd_debug_log( 'Attempt to verify PayPal payment with PDT failed due to payment total missing' );
-					$payment->add_note( __( 'Payment failed while validating PayPal PDT. Missing payment_gross.', 'easy-digital-downloads' ) );
-					$payment->status = 'failed';
+					$payment->add_note( __( 'Payment could not be verified while validating PayPal PDT. Missing payment total fields.', 'easy-digital-downloads' ) );
+					$payment->status = 'pending';
```
cklosowski: We need a `$payment->save()` here.

pippinsplugins: No we don't. That gets triggered on line 1002.

cklosowski: ah, right, ok sorry misread the indendation here in the PR.

cklosowski:
Test failures are due to a misconfiguration in the Travis Config. Code passes.

PR #7110, opened 2019-01-07
Check for selector's existence vs. just a jQuery object. 

### LINKED ISSUE #7072: [3.0] Tax Calculation Fails on Order Creation With Quantities Disabled

spencerfinnell:
Fixes #7072

Proposed Changes:
1. Ensure tax calculation works when quantities are disabled.


spencerfinnell:
@cklosowski anything else that needs to be done here?

PR #7112, opened 2019-01-08
Create a better UX when managing tax settings

### LINKED ISSUE #7109: Revert Hide Tax Rates Until Enabled, and add notice to tax rates when taxes are disabled
### Expected behavior

I should not be able to add tax rates when taxes are not enabled.

### Actual behavior

I can add a rate without enabling global taxes.

1) Visit "Settings > Taxes" 
2) Visit "Rates"
3) Add a rate.
4) Use `edd_use_taxes()` function and it returns false even though other functions check directly for rates (vs global usage).

----

Another option would be to force-enable this to be true if any active rates are detected.

spencerfinnell:
Fixes #7109

~- Force enable taxes to be enabled if a rate was created (and is active) without previously globally enabling taxes.~
- Hide "Rates" subtab until global taxes are enabled

---

Someone may want to add a test case for the `edd_use_taxes()` being enabled via the option filter. I don't have a test environment setup on this machine currently.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -778,26 +778,6 @@ function edd_get_registered_settings() {
 						'tooltip_desc'  => __( 'This option will determine whether the product price displays with or without tax on checkout.', 'easy-digital-downloads' ),
 					),
 				),
-				'rates' => array(
-					'tax_rate' => array(
-						'id'            => 'tax_rate',
-						'name'          => __( 'Default Rate', 'easy-digital-downloads' ),
-						'desc'          => __( 'Customers not in a region below will be charged this tax rate instead. Enter <code>6.5</code> for 6.5%. ', 'easy-digital-downloads' ),
-						'type'          => 'number',
-						'size'          => 'small',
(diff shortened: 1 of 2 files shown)


cklosowski:
I like the idea of this but it's causing some weird UI, that probably in the real world, makes sense, where if you have any active tax rates you cannot disable taxes. The even more impactful situation however, is if someone has tax rates defined after the migration, but have taxes disabled on purpose, they will upgrade to EDD 3.0 and their taxes will be force enabled, which could cause some accounting problems, since they didn't want to enable them.

For now @spencerfinnell I think it's best if we don't force the `use_taxes` setting if there are active tax rates, just hide the tab. We should be using the `edd_use_taxes()` boolean check anytime we go to apply tax rates anyway.

spencerfinnell:
Yeah that makes sense. What about just toggling the other subtab based on the global setting?

I just think if I was able to mess something up myself that means someone using it for the first time most certainly could fall in the same situation.

spencerfinnell:
Sorry... re-read your comment and that's exactly what you suggested. PR Updated.

PR #7115, opened 2019-01-09
Fix incorrect param type in method documentation.

rvdsteege:
Proposed Changes:
1. Fix incorrect `edd_get_cart_item_name( $item = array() )` param documentation from **int** to **array**.


cklosowski:
Thanks @rvdsteege

PR #7117, opened 2019-01-11
Changes the publish status to complete and fixes counts for orders.

### LINKED ISSUE #7088: 3.0 - Orders: Refunds list view does not show refunds (counts are correct)
### Expected behavior
When I view Orders > Refunds, I should see a list of refunds.

### Actual behavior
No refunds are shown, however the counts are correct.

1) Have refunded orders.
2) View the Orders > Refunds (Tab).

PHP Version: All

WordPress Version: All

Any other relevant information: All

### LINKED ISSUE #7089: 3.0 - Orders: Finalize 'complete' status designation
We need to finalize the status terminology for orders/refunds. In the past we've always set a 'completed' order (previously a payment) to a status of `publish`.

We even went as far as catching people trying to set the status of `complete` or `completed` to `publish` in the code. With the introduction of refunds, they have a completed status of `complete`. This causes some weird issues with the query and stats classes, as they are 'completed' but don't have a status of `publish` like most of our code is looking for.

We should finalize this and choose one 'final' state for order types.

My vote is try move everything to `complete`, which would signify that an order or refund has reached it's end state of either collecting funds or sending funds back to the gateway.

I want to do this in a single branch and I believe it will solve things like #7088. Things we'll have to do:
1) Update our migration to change orders to 'complete' that are publish.
2) Update our order status update method to catch publish and 'completed' and convert it to complete.
3) Update query classes and stats classes to include 'complete' instead of 'publish'.
4) Update any UI where a dropdown sets the status to 'publish' when 'Completed' selected.

cklosowski:
Fixes #7089 
Fixes #7088 

Proposed Changes:
1. All references to `publish` for orders is now changed to `complete`
2. Attempts to catch `publish` for a status and convert to `complete`
3. Updates count functions/methods to make sure they are using a `type` check for the orders table.
4. Fixes up some unit tests (Not usually something I would do here, but I wanted to clear those up as much as possible in order to make sure the status change code wasn't the culprit)


cklosowski:
@mintplugins I know you've looked this over and done some testing, If you don't mind at your next availability give it an approval on the PR here, for record keeping.

PR #7121, opened 2019-01-15
issue/7120 – Introduce an endpoint view registry for reports

### LINKED ISSUE #7120: 3.0: Shift get_endpoint_views() to leverage a (largely immutable) registry
Per a discussion with @JJJ today, I'm proposing that we shift away from using the static array of endpoint views and their corresponding attributes returned from `Reports\get_endpoint_views()` function into a registry.

The original thinking behind making endpoint views immutable in this way was to deliberately prevent registering new endpoint views for the first phase of the new reports API. This is largely because there's very little safety net in place for custom endpoint views. It was also done this way to prevent core endpoint view attributes from being overridden.

As it happens the SellBird team actually needs to override some attributes associated with the core endpoint views, namely in the realm of display callbacks. Because we're unable to leverage the admin UI on the front end, we need to be able to override the display callbacks. We're not looking to registering custom endpoint views or even mess with core endpoint views in any other way.

The attached PR (WIP) introduces an `Endpoint_View_Registry` class that core itself will leverage to "register" endpoint views. The bulk of the attributes are hardcoded into a private property and immutable. Because the registry allows for overwriting records, I've been protections in place to allow overwriting certain _attributes_ and only in core endpoint views.

If/when in the future EDD opens up the endpoint views API, the registry will already be there, and y'all should be able to take the chains off as needed.

DrewAPicture:
Fixes #7120


DrewAPicture:
Added tests for `Endpoint_View_Registry`.

cklosowski:
This doesn't appear to have broken any other tests that weren't already breaking, going to merge and do a tests fix up to try and clear up as many of these as possible.

PR #7128, opened 2019-01-22
Issue/7127 – Standardize data retrieval for reports table endpoints

### LINKED ISSUE #7127: Add a dedicated data getter to EDD\Admin\List_Table
As part of the revamping of the Reports API for 3.0, we introduced an `EDD\Admin\List_Table` middleware class to live between EDD list tables and `WP_List_Table`.

Due to the nature of how core's implementations of `WP_List_Table` have tended to work, there's never really been a good example set by core for making a standardized method for data retrieval. Typically the way retrieval and population of the data works in a `WP_List_Table` subclass is that the `$items` property gets arbitrarily initialized somewhere in the `prepare_items()` method. Our endpoint table subclasses have likewise followed suit with this pattern.

All that to say, as we're beginning to extend the reports API on the front end in SellBird, we're starting to run up against certain limitations, including but not limited to not having core WP-admin functionality available to us in the realm of `WP_List_Table` and all of the CSS and JS that goes with it.

What we're asking for here is the introduction of an abstract `get_data()` method in the `EDD\Admin\List_Table` class. This would make data retrieval from the table endpoints consistently possible and would break anyone extending the API free from having to reinvent the wheel for any list-table-based endpoint views if not displaying in the admin.

The biggest change of course would be needing to have all of the table endpoint classes implement the new method, but this is actually a fairly trivial change. All of the table endpoint classes are already leveraging their own unique getter methods, this would actually just standardize it all.

DrewAPicture:
Fixes #7127

Proposed Changes:
1. Adds an abstract `get_data()` method to `EDD\Admin\List_Table`
2. Adjusts all reports endpoint list table sub-classes to extend the new `get_data()` method


cklosowski:
Looks good @DrewAPicture merged!

Issue #7130, opened 2019-01-22
3.0 - Utilities: DateTimeZone fails on PHP 5.4


cklosowski:
### Expected behavior
When attempting to use the utilities class with PHP 5.3/4, the `DateTimeZone` should work correctly.

### Actual behavior
Per the unit tests, we're seeing this error when running against PHP 5.4.
`Exception: DateTimeZone::__construct(): Unknown or bad timezone (GMT-5)`

It appears as though sending in a `GMT+/-X` format isn't supported until PHP 5.6. We'll need another way to kick up a `\DateTimeZone` object that is both 5.4 and 5.6+ compatible.

### Steps to reproduce the behavior
You'll need a way to monitor your error logs. I use the `WP_DEBUG_LOG` constant and tail the `wp-content/debug.log` file.
1) Run a site in PHP 5.3 or 5.4
2) Open Debug Bar Console
3) Type the following and execute `$dt = new DateTimeZone('GMT-5');`
4) Look at the error log for the fatal error that occurs.

### Information (if a specific version is affected):
PHP Version: 5.3/4

EDD Version (or branch): `release/3.0`

WordPress Version: All

PR #7131, opened 2019-01-22
A PR to shore up the unit tests.

cklosowski:


cklosowski:
So this looks to fix all tests except for PHP 5.4 tests, which are failing due to #7130. So once this is merged we can fix up that issue, with unit tests that are accurate for it.

cklosowski:
Correction it fixes up most of the unit tests, will dig in once more tests have completed.

cklosowski:
Ok, so we've decided 3 major versions back for WP core, so that's excluding the WP 4.4.x - 4.6.x, so we should just be down to a few PHP 5.4 fixes and other minor changes.

cklosowski:
Going to merge this so our other branches can get quicker unit tests as well as be more accurate in the pass/fail going forward.

PR #7137, opened 2019-01-23
[3.0] Decouple Chart Javascript from Reports

spencerfinnell:
After discussion with @DrewAPicture it became clear that the previous movement of the Javascript from the `render()` method of the chart manifest (https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7022) caused unnecessary coupling of *charts* and *reports*. A chart doesn't need a report. A report has the option to utilize charts to display its data.

This PR allows `$chart->display()` to be called anywhere and ensures that something like 

```php
$chart = new Chart_Endpoint( array( ... ) );
$chart->display();
```

will enqueue the necessary scripts and output a rendered chart.

---

This was discovered with the intention of outputting charts on the frontend of a website, so it's worth considering there are other aspects of this still somewhat closely to the admin environment (such as registering the script on `admin_init`) and the name of the script itself.


DrewAPicture:
Throwing my support behind the merge of this PR in EDD 3.0. In all of the effort we're putting into reusing EDD reports on the front end, this adjustment to how the reports JS gets enqueued is a much-needed improvement.

Case in point, with this PR, we're now able to render charts on the front end by simply registering the edd-admin-reports script ourselves. That's pretty awesome and this really does make charts portable!

This is even possible in lieu of #7127, though we still need that too for being able to effectively display tile and table endpoints.

cklosowski:
I don't see any issues with this, and it inherently fixes an issue we coded around in #7063 I believe, allowing us to remove some of the conditional logic from the JS in reporting.

Tested reports and nothing seems to be broken, and based off @DrewAPicture's PRs, it seems to be what's necessary here. Merging.

DrewAPicture:
Thanks @cklosowski!

PR #7142, opened 2019-01-26
Fix Conditional Check Typo at edd_is_admin_page()

lushkant:
**Fixes a conditonal check typo at `edd_is_admin_page();`**

Switch case for 'settings' submenu page has a conditional check for 'general' tab which is spelled 'genera'.


cklosowski:
Thanks @lushkant. Appreciate your first code-contribution to EDD!

Issue #7156, opened 2019-02-06
Discount is_started should be checking in UTC


mintplugins:
## Bug Report

Discount codes are saved in UTC, which is good. They are entered in UTC as well, which is not good, as it can be confusing for site owners. Additionally, logic performed on that date is done as though it was entered and saved in the local WordPress timezone. 

### Expected behavior
When I have a discount code set to start at 03:00:00 UTC, I expect it to start at 03:00:00 UTC, regardless of my WordPress timezone.

### Actual behavior
When I have a discount code set to start at 03:00:00 UTC, it only starts working after 03:00:00 UTC if my WordPress Timezone is also UTC. Otherwise it starts working at 03:00:00 in my local timezone. 

The problem with this approach is that if I move to another timezone, changing my WP timezone will adjust the start-time of _all_ my discount codes as well. Certain start-times are customer specific, such as discount codes created by EDD Reviews. Changing the start times to match the WP timezone throws it out of sync from the customer's timezone. 

Therefore, the date should always be in-sync with UTC. 

### Steps to reproduce the behavior
1) Set your WP timezone to be your local timezone. I'm in the EST timezone, which is UTC-5 at this time of year (it's February, taking into account daylight savings).

2) Create a discount code, setting the start time to today's date, starting at the current time in your local timezone. For me, that's 2:38PM EST, which is 14:38:00 EST in 24 hour time.

<img width="564" alt="screen shot 2019-02-06 at 2 41 47 pm" src="https://user-images.githubusercontent.com/1638128/52368853-66e1c100-2a1d-11e9-811f-df32578205f3.png">

2) Notice the discount works
<img width="748" alt="screen shot 2019-02-06 at 2 44 18 pm" src="https://user-images.githubusercontent.com/1638128/52368999-b3c59780-2a1d-11e9-81d6-02466325fb1f.png">

3) Pretend you moved west, and change your WP Timezone to be something at least an hour earlier than your timezone. For me that's UTC-6. 

4) Notice the discount does not work now. 
<img width="734" alt="screen shot 2019-02-06 at 2 45 55 pm" src="https://user-images.githubusercontent.com/1638128/52369109-eec7cb00-2a1d-11e9-8d13-fbb6d8da2bdb.png">

5) My expectation is that the discount start time is not affected by the WordPress timezone, but rather, that all logic that is executed is based on the time in UTC. 

[...]

PR #7157, opened 2019-02-06
Use UTC when checking start and expiration

mintplugins:
#7156


cklosowski:
Going to fix up the unit tests here (nothing is failing due to the changes) but maybe we should get some checks in place for that to be sure? I'll see what I can come up with.

cklosowski:
Ok I've pushed up some changes to the unit tests, including a new unit test that failed on `master` and passes here in this issue branch to verify that the UTC checking works.

PR #7171, opened 2019-02-26
Initial pass at migrating refunded orders with a propr refund order #7170

### LINKED ISSUE #7170: 3.0 - Migration: Refunded orders are being marked as 'refunds' not as 'orders' instead of a separate 'refund' order being created.
### Expected behavior
When I migrate refunded orders, I expect them to be listed in the order list as a sale (much like any order created and refunded on 3.0). I also expect a 'refund' order to be created, which references this refunded sale order.

### Actual behavior
The refunded 'sale order' is migrated as a 'refund order', which makes it not show up in the sale order list, and no matching refund order is created.

1) On 2.9.x, purchase and refund an order.
2) Migrate this refunded order to 3.0
3) Verify that this order does not exist in the sales log, but is listed as a `refund` type.

PHP Version: 7.0

WordPress Version: 5.1

Any other relevant information: Another relevant note, order items are not listed as 'refunded' either, leaving them in a 'complete' state after they are migrated. Order items that are on a refunded `sale` should be listed as `refunded`

cklosowski:
Fixes #7170 

Proposed Changes:
1. At all points where an order is created, also creates an order with a type of `refund` if the main `sale` order is marked as `refunded`.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -398,10 +398,17 @@ class Data_Migrator {
 
 		if ( 'refunded' === $order_status ) {
 			$refund_data = $order_data;
-			$refund_data['parent'] = $order_id;
-			$refund_data['number'] = $order_id . apply_filters( 'edd_order_refund_suffix', '-R-' ) . '1';
-			$refund_data['type']   = 'refund';
-			$refund_data['status'] = 'complete';
+			$refund_data['parent']       = $order_id;
+			$refund_data['order_number'] = $order_id . apply_filters( 'edd_order_refund_suffix', '-R-' ) . '1';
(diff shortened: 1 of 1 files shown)


cklosowski:
Ok, this is ready for some testing. To test quickly I've been using these steps after you have 3 products created (I use the WordPress Importer and the .xml file in the `assets` directory)
1) On `master` create 5 orders via CLI `wp edd payments create --number=5 --date=random --range=365 --generate_users=1`
2) Go in and refund one of the created orders, that was not created today
3) Switch to `issue/7170`
4) Run the upgrade routine
5) Verify that:
5a) The original order was marked as refunded after migration
5b) The order items on the original order are marked as `refunded`
5c) A refund order was created and its items were marked as refunded
5d) The refund order number is in the format of `<original order ID>-R-1`

PR #7176, opened 2019-03-05
Fix strpos warning when using product API

tobeyadr:
When passing an array of categories a warning is thrown due to the fact that there is no check if categories is a string first.

Fixes #7176

Proposed Changes:
1. Add category string check to product api to avoid PHP warnings

### DIFF
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -73,7 +73,7 @@ class EDD_API_V2 extends EDD_API_V1 {
 			}
 
 			if( ! empty( $args['category'] ) ) {
-				if ( is_string( $args[ 'categrory' ] ) && strpos( $args['category'], ',' ) ) {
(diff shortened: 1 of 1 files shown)


includes/api/class-edd-api-v2.php
```diff
@@ -73,7 +73,7 @@ public function get_products( $args = array() ) {
 			}
 
 			if( ! empty( $args['category'] ) ) {
-				if ( strpos( $args['category'], ',' ) ) {
+				if ( is_string( $args[ 'categrory' ] ) ) {
 					$args['category'] = explode( ',', $args['category'] );
 				}
```
cklosowski: Because we're always getting a `string` to `array` conversion here...the `is_numeric` check will always fail and is no longer necessary, essentially making it unused code. Since the `foreach` loop in the `is_array` check does check for `is_numeric` so that'll catch things.

PR #7180, opened 2019-03-07
Migration updates for hardening, improved compatiblity and catching some data errors

### LINKED ISSUE #7173: 3.0 - Migration: Collection of improvements
This is a catch-all issue as I work through the migration routine to capture any necessary issues to fix. They will be included in a single issue to in order to allow us to test this easier and more quickly.

- [x] Pending orders are having a 'completed' date added to them.
- [x] WPCLI count was incorrect in the results, however, in my admin I have 101898 orders. I suspect that this is due to the new process of creating a 'refund' order for each refunded 'sale'.
```
Old Records: 101898
New Records: 104962
```
image
- [x] Remove any unnecessary calls to instantiate objects.
- [x] Multiple PHP notices and warnings related to array types, null items, etc.
- [x] Disable caching of the new DB class while migrating. - There is a real chance we could overwhelm a caching layer with all the new writes when migrating, so we should avoid writing every new item to the cache, knowing full well, the next item is just going to overwrite it again.
- [x] `v30_migration` method in the WPCLI class does not exist (as suggested by the admin notice about the upgrade)
- [x] Fix whitespace issue in order migrations for when refunds are present.
- [x] Fix SQL query in CLI to get all log entries (performance/failure)
- [x] Update child payment IDs with new parent order IDs

A few of these should help improve the performance of the migration as well, but if we need to we can do a deeper dive into the profiling of the migration to reduce the time to migrate.

cklosowski:
Fixes #7173

### DIFF
--- a/includes/admin/payments/class-payments-table.php
+++ b/includes/admin/payments/class-payments-table.php
@@ -370,7 +370,7 @@ class EDD_Payment_History_Table extends List_Table {
 			'customer' => __( 'Customer',  'easy-digital-downloads' ),
 			'gateway'  => __( 'Gateway',   'easy-digital-downloads' ),
 			'amount'   => __( 'Amount',    'easy-digital-downloads' ),
-			'date'     => __( 'Completed', 'easy-digital-downloads' )
+			'date'     => __( 'Date', 'easy-digital-downloads' )
 		) );
 	}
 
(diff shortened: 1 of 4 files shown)


mintplugins:
@cklosowski Would it be possible to add code comments to the if statements in this file to help make sure I'm understanding the intention of everything better?

cklosowski:
@mintplugins sure, I can do that.

cklosowski:
@JJJ created the method to fix serilization.

PR #7191, opened 2019-03-26
PDT should only run when on the same purchase session #7190

### LINKED ISSUE #7190: PayPal PDT should limit when it is sent.
### Expected behavior
PayPal PDT verification should only be run when the purchase session matches that of the PDT payment trying to be verified.

### Actual behavior
If you have a valid purchase session, but revisit (via bookmark) the purchase confirmation page that you visited while using PDT, the purchase confirmation will be for your valid purchase session, however, PDT will still fire.

** Ensure Debug Logging is enabled **
1) Make a purchase with PayPal Standard with PDT enabled. (Take note of the payment ID)
2) When complete, bookmark the purchase confirmation page that should contain `payment-id` and `txn` (at a minimum)
3) Make another purchase (I used stripe to prove the differences)
4) Once at the purchase confirmation page, make note of the new payment ID
4a) Clear your EDD Debug Log to make this easier.
5) Click on the bookmark you created from step 2
6) Notice that the purchase confirmation page is loaded for the payment in step 4
7) Check your debug log, and notice a PDT verification was sent.
etc, etc

PHP Version: All

EDD Version (or branch): 2.8.13 or greater

WordPress Version: All

cklosowski:
Fixes #7190 

Proposed Changes:
1. Only fires PDT when on the same purchase session.

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -909,7 +909,18 @@ function edd_paypal_process_pdt_on_return() {
 		return;
 	}
 
-	$payment = new EDD_Payment( $payment_id );
+	$purchase_session = edd_get_purchase_session();
+	$payment          = new EDD_Payment( $payment_id );
+
+	// If there is no purchase session, don't try and fire PDT.
+	if ( empty( $purchase_session ) ) {
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


cklosowski:
@mintplugins if you don't mind I'd like 2 approvals on this one just to be sure.

Issue #7210, opened 2019-04-15
3.0 - Allow order and discount IDs to be directly migrated, without the need for legacy IDs using auto_increment.


mintplugins:
I am opening this as a place to have a conversation and record of the thoughts behind how the `id` column will operate in EDD3, what its purpose is, and what the expectations are for it going forward from EDD3. 

Previously to EDD3.0, payment IDs were automatically assigned by WordPress, as they were stored in the wp_posts table. Any blog post or revision (or any other post_type) that would happen "in the middle" of payments, would cause their IDs to become non sequential.

For example, imagine this series of events:
1. Payment comes in (ID is 1)
2. Blog post is published (ID is 2)
3. Another blog post is published (ID is 3)
4) Another payment comes in (ID is 4)

You now have 2 payments:
- Payment 1 
- Payment 4

EDD 3.0 will migrate those payments to their own unique table, and also change their terminology from "payments" into "orders". That new table has an ID column which is sequential. So after migration, you will have 2 payments:
- Order 1 (previously was Payment 1)
- Order 2 (previously was Payment 4)

The problem here is that the ID of Payment/Order `4` has changed to `2`. Since the ID has been used for the ID of the payment, the invoice, and receipts, customers have records of their payment with the id `4` attached. We do not know the legality of changing an order number in all countries and regions. Aside from legality, we do not know if this will cause issues in 3rd party book-keeping integrations like Xero and others. 

For this reason, a second ID column is being proposed which would keep the "legacy" IDs exactly the same after migration as before. While the `id` column would still be incremental (1,2,3), the `order_id` column would be considered as the "official" id of the order. After migration, using the example above, the next payment that comes in would sequentially start where the payments were left off. So it would look like this:

- Order 1 (previously was Payment 1)
- Order 4 (previously was Payment 4)
- Order 5 (new payment after migration to EDD3.0)
- Order 6 (new payment after migration to EDD3.0)
- Order 7 (new payment after migration to EDD3.0)
- Order 8 (new payment after migration to EDD3.0)

The `id` column would be ignored by any code that intends to query post the ID of the payment.

mintplugins:
A few other issues that could arise if IDs change (props @chriscct7):

- Software licensing generates the license based on the payment ID 
- Google Analytics integrations could be thrown off
- Rest API calls using the ID will be affected
- Bookkeeping integrations could be affected
- Admin URLs will be affected (#7209)
[...]

PR #7216, opened 2019-04-17
[WIP] Issue/7210 - Retain unique item ID's when migrating

JJJ:
Addressing #7210.

Proposed Changes:
- [x] The database API Query class requires a very small change to support this
- [x] The database API Table class should get auto-incrementor manipulation methods, in the event someone wants to (or needs to) manually adjust them
- [x] The migration API will require more tweaks and more testing to make sure it correctly addresses all of our above concerns


cklosowski:
@mintplugins this seemed to work fine for me. all things migrated, the auto_increment was pre-defined to be 1 greater than the max edd_payment post type.

When you get a chance, can you please try this again?

mintplugins:
Running another migration now...

mintplugins:
Here's the results of my last 2 migration tests. I used the same database for both, and wiped out all tables completely before each migration:

Screen Shot 2019-04-30 at 5 47 37 PM

The second migration is the correct one, as the auto_increment ID used to create the new `refund` orders was set to a higher number than the highest payment ID pre-migration. 

It's possible that in the first migration something went wonky with my tables. I can't see any other reason for it. So I am okay to merge this now, but we should keep an eye on this happening when we do further migration tests in the future.

mintplugins:
I wonder if we should migrate tax rates after discount codes so that they don't begin at 1:

Screen Shot 2019-05-01 at 3 29 32 PM

cklosowski:
@mintplugins by setting the auto increment, it should take care of that. I'll have to re-test to verify what you're seeing.

mintplugins:
After using wp cli to export my database things were correct after migration. My issues above must have been due to some sort of strange issue with the table migrated through Migrate DB Pro

PR #7222, opened 2019-04-24
Unserialize extra payment meta

### LINKED ISSUE #7221: 3.0 Arrays in payment meta become strings after migration
### Expected behavior
After migrating, I expect all values to remain in the same format. 

### Actual behavior
Payment meta (maybe other data as well) wraps serialized arrays as a string during migration. For example:

Value in DB before migration:
`a:1:{s:7:"52622-1";b:1;}`

Value in DB after migration:
`s:24:"a:1:{s:7:"52622-1";b:1;}";`

1) Have some payment meta
2) Migrate to EDD3

EDD Version (or branch): 3.0

mintplugins:
Fixes #7221

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -857,7 +857,7 @@ class Data_Migrator {
 
 		// Migrate additional payment meta.
 		foreach ( $remaining_meta as $meta_key => $meta_value ) {
-			$meta_value = $meta_value[0];
+			$meta_value = maybe_unserialize( $meta_value[0] );
 
 			edd_add_order_meta( $order_id, $meta_key, $meta_value );
 		}


cklosowski:
@JJJ I don't believe so. This appears to be additional 3rd party meta stored in that master-meta key of `_edd_payment_meta` that plenty of extensions injected data into.

mintplugins:
I wonder if anyone ever added custom meta to a discount. Should we migrate all meta attached to a discount post type just to be safe?

cklosowski:
We already are migrating discount meta: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L179

mintplugins:
Ah right. I think we do have to run maybe_unserialize on that though, unless I'm reading it wrong.

mintplugins:
I might have been reading it wrong. I believe that happens here:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L160

cklosowski:
I think it's unlikely as we run it up here:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L160

The main reason this happened with the `_edd_payment_meta` is because it
was a dumping ground for data.

PR #7232, opened 2019-04-29
Move taxes back to always showing and add notice if they are disabled #7109

### LINKED ISSUE #7109: Revert Hide Tax Rates Until Enabled, and add notice to tax rates when taxes are disabled

cklosowski:
Fixes #7109 

Proposed Changes:
1. Moves tax rates tab and general settings back into always showing
2. Adds a notice to the top of the tax rates section if taxes are disabled.

### DIFF
--- a/includes/admin/settings/display-settings.php
+++ b/includes/admin/settings/display-settings.php
@@ -227,3 +227,25 @@ function edd_options_page() {
 	// Output the current buffer
 	echo ob_get_clean();
 }
+
+/**
+ * Conditionally shows a notice on the Tax Rates screen if taxes are disabled, to inform users that while they are adding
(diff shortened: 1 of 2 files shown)


cklosowski:
Thanks @spencerfinnell . Merging.

PR #7237, opened 2019-04-30
Fix reset tool to use truncate, add filters to allow truncating other tables, and fix JS loading #6831

### LINKED ISSUE #6831: 3.0: Reset Tool is broken
### Expected behavior
Using the reset tool in `Tools -> Recount Stats` should remove all articles of customers and orders from the EDD site. 

### Actual behavior
Customers and orders are all still present.

1) In Tools > General from the `Recount Stats` section choose 'Reset Store'
2) Check the box and run the reset.

PHP Version: All

WordPress Version: All

cklosowski:
Fixes #6831 

Proposed Changes:
1. Enqueues proper JS to load for the tools page.
2. Swaps ID based deletion for table based truncation
3. Added filter `edd_reset_tables_to_truncate` so extensions can register their own tables to be truncated when stores are reset.


cklosowski:
@jjj @mintplugins @pippinsplugins thoughts on this approach?

JJJ:
Hmm. The commit I see isn't functionally equivalent to what was here.

It no longer removes downloads, their meta & comments, or any other custom post type via the `edd_reset_add_queries_%` filter (I did not look to see if anything uses it.)

If that's by design, then great! If it's unintentional, then I think this PR will need to be reworked to only truncate our custom tables. If you're able to reach into the Component table itself, you can call it's `truncate()` method directly instead of needing to touch `$wpdb` yourself.

Truncating the custom tables will be _loads_ faster than any Delete statements would ever dream of being. The only caveat (which I do not think is a problem for us) is that `TRUNCATE` requires the MySQL user to have `DROP` permission, which they should have – if they don't, that's probably on purpose courtesy of the system admin, at which point it's on that person to manually perform these routines anyways.

cklosowski:
@jjj, the reset tool never removed downloads, just reset their store stats for sales and earnings. We'll just have to update the meta for the download IDs. All that was removed prior was sales based data.

mintplugins:
Seems like a pretty good approach to me

mintplugins:
Just tested this and it worked for me.

PR #7242, opened 2019-05-03
add/add-states/province-and-currency-to-mozambique

Zebedeu:
Subdivisions
1 city (en) / ville (fr) / cidade (pt)
10 province (en) / province (fr) / província (pt)

https://www.iso.org/obp/ui/fr/#iso:code:3166:MZ

MZM (Mozambique Metical) is the national currency the African nation of the Republic of Mozambique.

https://www.investopedia.com/terms/forex/m/mzm-mozambique-metical.asp

### DIFF
--- a/includes/country-functions.php
+++ b/includes/country-functions.php
@@ -105,6 +105,9 @@ function edd_get_shop_states( $country = null ) {
 		case 'MY' :
 			$states = edd_get_malaysian_states_list();
 			break;
+		case 'MZ' :
+			$states = edd_get_mozambique_states_list();
+			break;
 		case 'NP' :
 			$states = edd_get_nepalese_states_list();
 			break;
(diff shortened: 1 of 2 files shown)


cklosowski:
@jchristopher Since the translation fix is not part of the addition of the `mozambique` data, it should not be part of this PR.

Also, this is a core string, so it's not necessary to be translated within EDD, since WP Core will have the string `Add New` in their translations. If you'd like to submit that as a new PR via a different branch that would be fine.

Zebedeu:
I am considering doing the same for all countries that EDD does not have their respective cities. What do you tell me about it? Can I take on this task?

Zebedeu:
> @jchristopher Como a correção da tradução não faz parte da indexação dos `mozambique`dados, ela não deve fazer parte desse PR.
> 
> Além disso, é uma string central, portanto, o WP Core tem uma string `Add New`em suas traduções. Se você gostaria de enviar isso de um lado para o outro.

I'll arrange this. Thank you

PR #7244, opened 2019-05-03
Use symbol when formatting currency for the default.

### LINKED ISSUE #7241: Some countries like Angola, When selecting the type of currency is not displayed the symbol but rather the iso code.
When selecting the currency expect an output of the currency symbol and not the ISO 4217 code.

Captura de ecrã 2019-05-03, às 03 30 03
1) Select Angola as the base country
2) Select Currency
3) add a product and a price
4) Expected output kz 25.00 and not AOA 25.00
PHP Version:
PHP Version 7.2

EDD Version (or branch):
EDD version 2.9.7
WordPress Version:

Wordpress Version 5.2-RC2-45280

mintplugins:
Fixes #7241 The edd_currency_symbol function returns the currency if no symbol is found. Therefore we can use $symbol here

### DIFF
--- a/includes/formatting.php
+++ b/includes/formatting.php
@@ -152,7 +152,7 @@ function edd_currency_filter( $price = '', $currency = '' ) {
 				$formatted = $symbol . $price;
 				break;
 			default :
-				$formatted = $currency . ' ' . $price;
+				$formatted = $symbol . ' ' . $price;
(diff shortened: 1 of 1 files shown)


Zebedeu:
I tested it here in my test environment and it's working as expected.

PR #7250, opened 2019-05-08
Issue/7249

### LINKED ISSUE #7249: You do not have permission to download this file after upgrading to 2.9.13
# Instructions (START):
After updating EDD to 2.9.3 users are facing an error like "You do not have permission to download this file".

I have debug this this issue and found that function `edd_local_file_location_is_allowed` return false while checking "ABSPATH" and "WP_CONTENT_DIR" path

When we have attachment `$requested_file` argument is coming with DIR path like `E:\wamp\www\wprelease\wp\wp-content\uploads\edd\2019\05\Tulips-300x225.jpg`

ABSPATH becomes like this : `E:\wamp\www\wprelease\wp/`
WP_CONTENT_DIR path becomes like this : `E:\wamp\www\wprelease\wp/wp-content`

Condition is checking strpos but due to backward slash problem `\` it returns false. Same with WP_CONTENT_DIR condition.

1) After updating EDD 2.9.13
2) Navigate to purchase history.
3) Click on "View Details and Downloads"
4) Try to download the file.

PHP Version: 7.2.14
EDD Version (or branch): 2.9.13
WordPress Version: 5.2
OS : windows

pippinsplugins:
Fixes #7249


cklosowski:
I had to push some updates, after testing on a Windows server this wasn't quite fixed.

PR #7266, opened 2019-05-17
Sendwp

pippinsplugins:
Introduces an integration for SendWP for simpler, better email delivery.


assets/js/admin-scripts.js
```diff
@@ -2158,3 +2167,38 @@ function edd_attach_tooltips( selector ) {
...
+
+function easy_digital_downloads_sendwp_remote_install() {
+	var data = {
+		'action': 'easy_digital_downloads_sendwp_remote_install',
+	};
+
+	// since 2.8 ajaxurl is always defined in the admin header and points to admin-ajax.php
+	jQuery.post(ajaxurl, data, function(response) {
+		var data = JSON.parse(response);
+		easy_digital_downloads_sendwp_register_client(data.register_url, data.client_name, data.client_secret, data.client_redirect, data.partner_id);
```
spencerfinnell: This doesn't check for failures in `wp_ajax_easy_digital_downloads_sendwp_remote_install` response and can cause the registration redirection to be invalid/incorrect.

includes/scripts.php
```diff
@@ -234,6 +234,7 @@ function edd_load_admin_scripts( $hook ) {
 		'show_advanced_settings'      => __( 'Show advanced settings', 'easy-digital-downloads' ),
 		'hide_advanced_settings'      => __( 'Hide advanced settings', 'easy-digital-downloads' ),
 		'no_downloads_error'          => __( 'There are no downloads attached to this payment', 'easy-digital-downloads' ),
+		'wait'                        => __( 'Please wait . . .', 'easy-digital-downloads' ),
```
cklosowski: We should switch this to `&hellip;` instead of `. . .`

includes/emails/functions.php
```diff
@@ -281,3 +281,71 @@ function edd_get_email_names( $user_info, $payment = false ) {
...
+		exit;
+	}
+
+	echo json_encode( array(
+		'partner_id' => 81,
+		'register_url' => sendwp_get_server_url() . '_/signup',
+		'client_name' => sendwp_get_client_name(),
+		'client_secret' => sendwp_get_client_secret(),
+		'client_redirect' => sendwp_get_client_redirect(),
+		//'client_redirect' => admin_url( '/edit.php?post_type=download&page=edd-settings&tab=emails' ), // Custom redirects aren't supported yet
```
cklosowski: Remove comment.

includes/emails/functions.php
```diff
@@ -281,3 +281,71 @@ function edd_get_email_names( $user_info, $payment = false ) {
...
+		$activated = activate_plugin( 'sendwp/sendwp.php' );
+
+	}
+
+	/*
+	* Final check to see if SendWP is available.
+	*/
+	if( ! function_exists('sendwp_get_server_url') ) {
+		ob_end_clean();
+		echo json_encode( array( 'error' => __( 'Something went wrong. SendWP was not installed correctly.', 'easy-digital-downloads' ) ) );
```
cklosowski: This should become `wp_send_json`

includes/emails/functions.php
```diff
@@ -281,3 +281,71 @@ function edd_get_email_names( $user_info, $payment = false ) {
...
+		exit;
+	}
+
+	echo json_encode( array(
```
cklosowski: `wp_send_json` should be used here again.

cklosowski:
Any place we use `echo json_encode( <data> );` we should shift to `wp_send_json`

assets/js/admin-scripts.js
pippinsplugins: Added simple handling for it.

includes/scripts.php
pippinsplugins: Done.

spencerfinnell:
Small, but the `target="_blank"` links need `rel="noopener noreferrer"` 

https://support.performancefoundry.com/article/186-noopener-noreferrer-on-my-links

pippinsplugins:
@spencerfinnell added.

spencerfinnell:
It would be nice if we could filter this URL to send it back to the EDD Emails tab: https://plugins.trac.wordpress.org/browser/sendwp/trunk/includes/admin/views/disabled-notice.html.php

(also they are missing a text domain on that string)

Made a i18n adjustment to split up the string/remove some HTML. The rest looks good!

cklosowski:
@spencerfinnell I made a small commit, could you process your approval again?

Issue #7259, opened 2019-05-10
Add filter to conditionally show nginx warning


mintplugins:
## Enhancement Request
The nginx warning about file protections shows for every site running nginx. However, some webhosts (like Kinsta) already have the required/specific protections in place for this, and this warning is not needed. A filter to conditionally remove that admin_notice can be used to prevent that. 

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b3b09b718d04ded5611b8b34194cbb77190a2c47/includes/admin/class-edd-notices.php#L59

cklosowski:
We have a host detection method that maybe we could use to do this. If you notice sometimes we have the system info detect the host. If Kinsta has a specific type of header or something on the server level we could detect, they could be added to that.

mintplugins:
Do you mean here?
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b3b09b718d04ded5611b8b34194cbb77190a2c47/includes/misc-functions.php#L208

mintplugins:
To test this, create a new site on an nginx server, and make sure the notice shows properly. 

I have confirmed it is for me.

<img width="795" alt="Screen Shot 2019-05-24 at 3 33 02 PM" src="https://user-images.githubusercontent.com/1638128/58352373-55d88680-7e39-11e9-8e59-b7399518aab4.png">

Related ticket:
https://secure.helpscout.net/conversation/846146047/119766?folderId=749231

PR #7275, opened 2019-05-24
Add filter around nginx redirect warning

mintplugins:
#7259

### DIFF
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -56,7 +56,10 @@ class EDD_Notices {
 			$notices['updated']['edd-payment-history-test-mode'] = sprintf( __( 'Note: Test Mode is enabled. While in test mode no live transactions are processed. <a href="%s">Settings</a>.', 'easy-digital-downloads' ), admin_url( 'edit.php?post_type=download&page=edd-settings&tab=gateways' ) );
 		}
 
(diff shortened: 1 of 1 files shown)


spencerfinnell:
@mintplugins What about checking for an environment variable that hosts could add to the `wp-config.php` as well? Similar to how core allows the PHP upgrade button to be taken over: https://github.com/WordPress/WordPress/blob/9785381a8c7bdf668f49d2606aee8224515da447/wp-includes/functions.php#L6990

mintplugins:
@spencerfinnell I haven't used PHP environment vars before. Would you need to use `putenv` in the wp-config file to set it? If so, why use that as opposed to a constant, similar to WP_DEBUG?

spencerfinnell:
@mintplugins Sorry you're right, it wouldn't be added to `wp-config.php`, it would be set through something like `putenv` or `$_ENV`

mintplugins:
Ah interesting. The environment var can be set at the php-fpm level, avoiding the need for any MU plugins or wp-config.php changes. Hosts will probably prefer that. I'll add it.

mintplugins:
@spencerfinnell I am having a bit of trouble figuring out how to make it default to true if it is not set. 

I am thinking we would have to use "no" instead of an actual boolean to accomplish this, as an empty or unset value comes through as falsey.

PR #7284, opened 2019-06-07
Allow logs to be added without an object ID

### LINKED ISSUE #7283: 3.0 - Payment Gateway errors are not being logged.
### Expected behavior
When I attempt to log a payment gateway error using `edd_record_gateway_error`, I expect that error to be logged in the edd_logs table. 

### Actual behavior
No log is added.

1) Add this code:
```
function my_gateway_log_tester() {
	edd_record_gateway_error( 'Test gateway error title', 'Test gateway error content' );
}
add_action( 'init', 'my_gateway_log_tester' );
```
2) Notice that no log is added to the edd_logs table. 

EDD Version (or branch): 3.0

mintplugins:
Fixes #7283

Also fixed the reference to the "content" column, which prevented the "content" from being added to the database. It was "message" previously.


includes/class-edd-logging.php
```diff
@@ -151,13 +151,15 @@ public function insert_log( $log_data = array(), $log_meta = array() ) {
...
-			'object_type' => isset( $args['object_type'] )
-				? $args['object_type']
-				: 'download',
+			'content'     => $args['post_content'],
+			'object_id'   => isset( $args['post_parent'] )
+				? $args['post_parent']
+				: 0,
+			'object_type' => isset( $args['log_type'] )
+				? $args['log_type']
+				: 'unknown',
```
cklosowski: I think I'd rather prefer a `null` or something, that way we can filter later and do queries against things as `NULL` or `NOT NULL`.

mintplugins: I updated that, but just wanted to mention that it needs to be changed to `null` in several places, as 0 was the standard for "none" previously to EDD 3.0. See changes here:

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7284/commits/9b08893c31d692de68a0e582936f77e499dfa710

cklosowski: @mintplugins I'm fine with the object ID being 0, I was referencing the `object_type`. I'd like that to be `null` instead of `unknown`.

mintplugins: Ah I see. One of the problems there is that a type is required. So we can't pass null, unless we decide that a type is not required. 

See type being required here:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7284/files#diff-22827a25e9cea91ff2d49d62f066894eR51

mintplugins: (PS I [reverted](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7284/commits/ff5c1a8b47dddfdfa0f6da00a02358a053d4aa51) https://github.com/easydigitaldownloads/easy-digital-downloads/commit/9b08893c31d692de68a0e582936f77e499dfa710

cklosowski: @mintplugins I'm fine with updating that to be allowed to be `null` only, so maybe we just do a check for empty and not is_null. Also, we should only allow a null type so long as the object ID is `0`.

mintplugins: @cklosowski See those changes as requested here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7284/commits/f2250700ab971d68f8e9db560dfdc06aa3a3202a

mintplugins:
I'd still like to add unit tests for these changes.

cklosowski:
@mintplugins yes, let's get some unit tests built for this before we confirm to merge it.

mintplugins:
@cklosowski Unit tests added in #7283

PR #7290, opened 2019-06-10
Issue/7197

### LINKED ISSUE #7197: 3.0 - Order item meta needs to be created for order item options
### Expected behavior
When migrating over orders, the `options` array could contain information that should be migrated over as meta like license upgrades/renewals, etc.

### Actual behavior
The options are simply dropped other than `price_id` and `quantity`

cklosowski:
Fixes #7197 

Proposed Changes:
1. Migrates order item `options` over to the `order_itemmeta` table.
2. Any new order item options are stored in the `order_itemmeta` table.


cklosowski:
@mintplugins this was the code you tested for teh order item options, just didn't have a PR for it.

PR #7295, opened 2019-06-12
Fix adding notes to an existing order via the admin UI #7294

### LINKED ISSUE #7294: 3.0 - Clicking "Add Note" for an order does not work
### Expected behavior
When I try to add a note to an order, I expect the note to be saved and added.

### Actual behavior
The "Add Note" button appears to be doing nothing. 

1) attempt to add a note to an order
2) notice the "Add note" button has no affect.

Screen Shot 2019-06-11 at 4 43 07 PM

EDD Version (or branch): 3.0

cklosowski:
Fixes #7294

Proposed Changes:
1. Re-introduces the JS and AJAX listener to add notes from the admin.


cklosowski:
@mintplugins can you re-test this please, now that I fixed it using the 3.0 methods.

PR #7316, opened 2019-06-24
Refactor how edd_get_payment_status finds the status, with unit tests  #7292

### LINKED ISSUE #7292: edd_get_payment_status() always return false if status label is translated.
the code below in function edd_get_payment_status() is search value instead of key, which will cause always return false if status label is translated (Not English).

`return array_search( strtolower( $post_status ), array_map( 'strtolower', $statuses ) );`

cklosowski:
Fixes #7292

Proposed Changes:
1. Changes all references to possibly using post_status and WP_Post objects to using an EDD_Payment object.
2. Swaps use of literal status labels for the keys, which are not translated.

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -711,28 +711,26 @@ function edd_get_payment_status( $payment, $return_label = false ) {
 
 	if( is_numeric( $payment ) ) {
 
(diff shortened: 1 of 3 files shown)


cklosowski:
Need to revert to make room for 2.9.17, this will go to 2.9.18

PR #7318, opened 2019-06-25
Fix edd_order_item table to accept negative integers for quantity #7307

### LINKED ISSUE #7307: 3.0 - Quantity is not set on refund order items
### Expected behavior
When I refund an order, I expect the "quantity" column to be correct in the order_items table. 

### Actual behavior
It is set to 0.

1) Buy an item.
2) Refund the order.
3) Look at the order_items table
4) Notice the quantity is set to `1` for the original order item, but is set to `0` for the refund order item.

Screen Shot 2019-06-19 at 3 26 30 PM

EDD Version (or branch): 3.0

cklosowski:
Fixes #7307

Proposed Changes:
1. Updates `edd_order_items` table to use `signed int` for the `quantity` column.
2. Adds new `edd_negate_int` function and implements for negating integers instead of assuming floats.
3. Unit tests to verify negative values for refunded quantity

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -709,7 +709,7 @@ class Data_Migrator {
 					$refund_item_args['status']   = 'refunded';
 
 					// Negate the amounts
-					$refund_item_args['quantity'] = edd_negate_amount( $cart_item['quantity'] );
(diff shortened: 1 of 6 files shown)


mintplugins:
Tested and confirmed it fixes #7307

cklosowski:
@mintplugins you ok with that last change I made? Just had a capital letter in the function for the unit test.

PR #7325, opened 2019-06-25
Move to using single 'name' column for customer/order addresses

### LINKED ISSUE #7302: 3.0 - Handle "name" column in the Customer/Order Addresses tables
## Enhancement Request
The Order Address table includes columns for the first_name and last_name. 

The Customer Addresses table does not. 

In the Order Addresses table, we should replace those 2 columns with 1 (proposed by @jjj on basecamp). Perhaps replacing `first_name` and `last_name` with a column called `name`. 

To the Customer Addresses table, we should also add a `name` column, as a "Customer" could actually be multiple people, and would also be consistent with the understanding of a complete "Address" as seen in the Order Addresses table.

cklosowski:
Fixes #7302 

Proposed Changes:
1. Converts the `first_name` and `last_name` columns in the `edd_order_addresses` table to a single `name` column, including upgrade routine to add the column, concat the strings to the new column, and removing the old column.
2. Adds the `name` column to the `edd_customer_addresses` table, including upgrade routine.
3. Updates the data migrator to use the new columns.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -55,6 +55,7 @@ class Data_Migrator {
 			edd_add_customer_address( array(
 				'customer_id' => $customer->id,
 				'type'        => 'primary',
+				'name'        => $customer->name,
 				'address'     => $address['line1'],
 				'address2'    => $address['line2'],
 				'city'        => $address['city'],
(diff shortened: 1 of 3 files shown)


mintplugins:
Looks Good. Can we also add the customer's name to the customer addresses table during this upgrade?

Screen Shot 2019-06-26 at 1 32 06 PM

cklosowski:
@mintplugins Not really b/c there is no way to know if that _was_ the name used for the address at that time. We'd be assuming a name which I'm not 100% positive is the right choice.

mintplugins:
Yeah either approach has it's problems. Sure leave it as is.

PR #7326, opened 2019-06-25
Add price name to order_items table #7308

### LINKED ISSUE #7308: 3.0 - Store the name of the price ID in the product_name column of the order_items table
The order_items table has a column for `product_name`. This is good for immutability because you can always know what the name of the product was at the time of purchase. 

However, the price_id does not have its name stored in this table. If the site owner changes the name of a price_id at any time, any purchases of that item will change, including on things like the purchase receipt. 

@cklosowski recommended storing the Price Name after the Product Name in a format like this:

`Product Name - Price ID Name`

cklosowski:
Fixes #7308

Proposed Changes:
1. Adds logic to `edd_build_order` to add the price ID name to an order_item when one exists.
2. Adds logic to the EDD order migrator to add the price ID name to an order_item when one exists.

### DIFF
--- a/includes/admin/payments/class-order-items-table.php
+++ b/includes/admin/payments/class-order-items-table.php
@@ -260,6 +260,14 @@ class Order_Items_Table extends List_Table {
 				'post'  => $order_item->product_id,
 			), admin_url( 'post.php' )  ) . '">' . $order_item->get_order_item_name() . '</a>' . $state . '</strong>';
 
+		// See what the current name of the product is:
+		$current_product_name = edd_get_download_name( $order_item->product_id, $order_item->price_id );
+		if ( strtolower( htmlentities( $current_product_name ) ) !== strtolower( htmlentities( $order_item->get_order_item_name() ) ) ) {
+			$status_help = '<p>' . sprintf( __( 'This product has been renamed since this purchase. It is now %s.', 'easy-digital-downloads' ), $current_product_name );
+
+			$order_item_title .= ' <span alt="f223" class="edd-help-tip dashicons dashicons-backup" title="' . $status_help . '"></span>';
+		}
+
 		// Return order_item title & row actions
 		return apply_filters( 'edd_order_item_title_and_actions', $order_item_title . $this->row_actions( $row_actions ), $order_item );
 	}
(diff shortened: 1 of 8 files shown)


mintplugins:
Look good. We should also change the way orders are shown here so it pulls from the order items table instead of from the download itself:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-order-item.php#L250

cklosowski:
@mintplugins I'm not 100% sure I agree fully here...

I need to play with that more and see if it creates a confusion situation. I think I'd prefer we do both if the name/price ID have changed. This would give the administrators a bit more insight as to what's happening.

For instance, I help out an EDD customer locally and they changed all their price ID names at some point. However, their new support person is having lots of trouble determining what each plan is for products purchased prior to the name change.

I'd propose we have a way to identify the 'current' name and the 'legacy' name, to make it clearer for shop workers (and maybe customers)

mintplugins:
I suppose it all comes down to immutability. Since this is an order record, it should be immutable. But then again that's also the philosophical issue with a digital product. It's not an unchanging/immutable "thing", but rather is more like an idea that is constantly in flux (with software updates, etc).

But whatever way we go, we need to be consistent. Currently if it's non-variable we [pull from order_items](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-order-item.php#L246), and if it's variable we [pull from the download data](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-order-item.php#L250).

cklosowski:
@mintplugins Yeah we should be consistent, one or the other...

I think I like the approach of signifying what it used to be honestly. I think that could solve some unique issues for people who rename products and price IDs (as we know people like to do).

spencerfinnell:
Thinking about similar things in WPSP: https://github.com/wpsimplepay/WP-Simple-Pay-Pro-3/issues/859

We currently reference the stored form data (post data) instead what it was at the actual time of payment. I personally like the idea ensuring the customer's purchase receipt reflects what they originally intended to purchase.

cklosowski:
> Thinking about similar things in WPSP: [wpsimplepay/WP-Simple-Pay-Pro-3#859](https://github.com/wpsimplepay/WP-Simple-Pay-Pro-3/issues/859)
> 
> We currently reference the stored form data (post data) instead what it was at the actual time of payment. I personally like the idea ensuring the customer's purchase receipt reflects what they originally intended to purchase.

I think this is the approach we'll take. Use what's in the receipt, but I would really like to reflect (for admins) what the 'current' product name is set as if it's different. This would solve so many questions I get about that process.

PR #7328, opened 2019-06-26
Enable paypal as a safe redirect before redirecting to PayPal.

### LINKED ISSUE #7312: 3.0 - If paying with PayPal, you get redirected to wp-admin instead of PayPal
### Expected behavior
When I choose that I want to pay with PayPal, I expect to be redirected to PayPal upon clicking Complete, on the checkout page.

### Actual behavior
I get redirected to wp-admin.

1) Enable PayPal as a gateway
2) Add an item to the cart and checkout using PayPal
3) You get redirected to wp-admin instead of PayPal. If logged out you hit the login screen.

EDD Version (or branch): 3.0

mintplugins:
Fixes #7312

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -313,11 +313,29 @@ function edd_process_paypal_purchase( $purchase_data ) {
 		// Fix for some sites that encode the entities
 		$paypal_redirect = str_replace( '&amp;', '&', $paypal_redirect );
 
-		// Redirect to PayPal
+		// Allow paypal as a redirect destination.
+		add_filter( 'allowed_redirect_hosts', 'edd_allow_redirect_to_paypal', 10 );
+
+		// Redirect to PayPal.
 		edd_redirect( $paypal_redirect );
 	}
 }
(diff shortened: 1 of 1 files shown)


cklosowski:
While I think this fixes the issue for PayPal, it will cause issues or make implementations for any other off-site gateway required to do the same thing.

I think the more _whole_ approach would be to create a generic method for allowed external redirects that can be used by _any_ gateway or 3rd party solution that wants to use `edd_redirect`, and therefore they only need to use a single filter for adding their external URL to the array of allowed URLs.

mintplugins:
I feel like other gateways will probably still be using wp_redirect for the time being. If they are following best practices they'll switch to wp_safe_redirect and add their own allowed hosts prior to using it. I don't see a need for them to use edd_redirect.

cklosowski:
@mintplugins ok, as long as we've discussed it I think I can live with that. We might want to add some lines to the comments of `edd_redirect` and add the `@uses wp_safe_redirect` to the docs around it so that if people do run into this later, they can at least know where to start looking to fix it on their end.

PR #7330, opened 2019-06-28
Add type column to order addresses.

### LINKED ISSUE #7300: 3.0 - Add a "type" column to the Order Addresses table.
While the Customer Addresses table has a "type" column to signify whether the address is "billing" or "shipping", the Order Addresses table does not. 

If Simple Shipping should/will be updated to store the Shipping Address for an order in the Order Addresses table, currently it cannot indicate that the address is for "shipping". 

Keeping in line with what we decided in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7254, we should add a `type` column to the Order Addresses table, which will contain the purpose of the address (billing/shipping/etc).

mintplugins:
Fixes #7300


includes/database/tables/class-order-addresses.php
```diff
@@ -105,4 +107,27 @@ protected function __201807270003() {
 		// Return success/fail
 		return $this->is_success( $result );
 	}
-}
\ No newline at end of file
+	/**
+	 * Upgrade to version 201906280001
+	 * - Add the `uuid` varchar column
```
cklosowski: This should read that it's adding the `type varchar column`

mintplugins:
@cklosowski After the merge conflict fix, I am seeing an issue happening with the upgrade function. I'm not totally sure whether it's a result of the merge conflict, or if there's something funky with the upgrade routines happening (relating to https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7302). 

Ultimately this is what happens:
Screen Shot 2019-07-05 at 2 48 27 PM

So I don't know if this is a result of it running multiple upgrades at once, or if something isn't firing here:

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7325/files#diff-fa076d0dfda157d61c569ac39ba9d13aR134

cklosowski:
@mintplugins looks like it's just an issue with us working on the same table at the same time. I'm merging and will take care of things.

PR #7340, opened 2019-07-09
Fix fees not being applied to orders

### LINKED ISSUE #7243: 3.0 - Fees not being applied to orders
### Expected behavior
When I have a fee set up through Discounts Pro or Simple Shipping, I expect that fee to be applied to the order upon completion.

### Actual behavior
While a row is added to the order adjustments table, and the amount seems to be correct, it is not reflected on the order itself. 

1) Set up a 20% discount with Discounts Pro.
2) Purchase a $5 product. With the discount it should come to $4
3) Upon purchase completion, note that you paid $5, despite the cart showing $4. 

Here's a screenshot of the order adjustment in the table:
Screen Shot 2019-05-03 at 9 19 06 AM

However, notice it was not applied to the order:
Screen Shot 2019-05-03 at 9 20 04 AM

Side question: Why does "adjustments" show on the order screen alongside "discounts"? What is the expectation for the "adjustments label? Does/Should it show discounts?

mintplugins:
Fixes #7243 

I am not sure why this was being prevented. The furthest back I was able to trace it was here:

https://github.com/easydigitaldownloads/easy-digital-downloads/blame/release/3.0/includes/orders/functions/orders.php#L1003

Any ideas why that line might have been added @JJJ? Do you think it is safe to remove it?

Additionally, this issue uncovered an issue with the flow to the gateway. The gateways were recording this properly, while EDD was not (and thus the customer's receipt was for the wrong amount). 

That is because [send_to_gateway](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/7243/includes/process-purchase.php#L195) is passed data from the session, as opposed to data from the custom tables.

To ensure that the same data is used for everything (and to help avoid these types of data-sync issues), I wonder if `edd_process_purchase_form` should call `edd_build_order`. Then, in `edd_build_order` we can call `send_to_gateway` and pass the data directly from the tables.

This would prevent the customer ever being charged something their receipt does not show. That might be a separate issue to tackle though. I just noticed it here and thought it was important to make a note of.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,18 +21,12 @@ env:
 
 matrix:
   include:
-    - php: 7.1
-      env: WP_VERSION=latest WP_MULTISITE=0
-    - php: 7.1
-      env: WP_VERSION=4.7.12 WP_MULTISITE=0
     - php: 5.6
       env: WP_VERSION=latest WP_MULTISITE=1
     - php: 5.3
-      env: WP_VERSION=4.7.12 WP_MULTISITE=0
+      env: WP_VERSION=4.7.13 WP_MULTISITE=0
       dist: precise
   exclude:
-    - php: 5.3
-      env: WP_VERSION=latest WP_MULTISITE=0
     - php: 5.4
       env: WP_VERSION=latest WP_MULTISITE=0
     - php: 5.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,20 @@ sudo: false
 dist: trusty
 
 php:
-- 5.4
-- 5.5
-- 5.6
-- 7.0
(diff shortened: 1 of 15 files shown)


mintplugins:
Going through my thoughts above again regarding order accuracy with gateways vs tables: I don't think we can use the tables as a "single source of truth" because of the way our extension system works. EDD Recurring isn't "known" by EDD core's table structure, and passing only data from Core's table structure to the gateways (as opposed to passing the cart session) would mean that recurring purchases don't work properly. So I think we can say that idea is scrapped.

mintplugins:
As it is, this issue resolves the originally-reported bug.

PR #7345, opened 2019-07-11
Consistently handle user ID changes on orders and customer

### LINKED ISSUE #7321: 3.0 - On the orders table, the user_id is inconsistently updated to reflect changes to the customer.
## Bug Report
When I update the user attached to a customer, all of that customer's orders get updated so that the `user_id` column reflects the new user. 

However, if I simply unlink a user from a customer, it does not set the user for those orders to `0`. It leaves them what they were before. 

### Expected behavior
When I unlink a user from a customer, I expect that customer's order's `user_id` columnb to be set to `0`. This would match what happens when you assign a new user to a customer.

### Actual behavior
The `user_id` column is not updated.

1) Have a few products purchased by a customer with a user attached.
2) Edit the customer and unlink the user by clicking the broken chain icon.
3) Look at the edd_orders table and notice that the user_id column for those orders was not updates to be `0`, but still matches what the user _id previously was.

EDD Version (or branch): 3.0

Screen Shot 2019-06-25 at 12 58 05 PM

cklosowski:
Fixes #7321

Proposed Changes:
1. Alters the change and remove User ID actions when altering a customer to alter the orders as well.

### DIFF
--- a/includes/class-edd-customer.php
+++ b/includes/class-edd-customer.php
@@ -334,11 +334,24 @@ class EDD_Customer extends \EDD\Database\Rows\Customer {
 
 		do_action( 'edd_customer_pre_update', $this->id, $data );
 
-		$updated = false;
+		$updated          = false;
+		$previous_user_id = $this->user_id;
 
 		if ( edd_update_customer( $this->id, $data ) ) {
 			$customer = edd_get_customer( $this->id );
 			$this->setup_customer( $customer );
+
+			// If the user ID changed, we need to update all the orders associated with this customer and change the user ID.
+			if ( intval( $previous_user_id ) !== intval( $this->user_id ) ) {
+
+				// Update some payment meta if we need to
+				$order_ids = edd_get_orders( array( 'customer_id' => $this->id, 'number' => 9999 ) );
+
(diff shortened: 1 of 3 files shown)


includes/class-edd-customer.php
```diff
@@ -334,12 +334,25 @@ public function update( $data = array() ) {
...
 
 		if ( edd_update_customer( $this->id, $data ) ) {
 			$customer = edd_get_customer( $this->id );
 			$this->setup_customer( $customer );
 
+			// If the user ID changed, we need to update all the orders associated with this customer and change the user ID.
+			if ( intval( $previous_user_id ) !== intval( $this->user_id ) ) {
+
+				// Update some payment meta if we need to
+				$order_ids = edd_get_orders( array( 'customer_id' => $this->id, 'number' => 9999 ) );
```
mintplugins: I assume this is a known fail-point for an edge case where a customer has more than 9999 payments. Is that accurate? (I'm not saying we need to do anything about that, just confirming)

cklosowski: @mintplugins yeah, it's a known failure point, I'm 100% ok with it at this time.

PR #7346, opened 2019-07-12
Remove the ability to 'reverse' a refund.

### LINKED ISSUE #7333: 3.0 - Remove the option to reverse a refund.
In EDD 3.0 the ability to reverse a refund was temporarily introduced as a possible concept. However, because of the incredibly complexity involved in making that actually work at the payment gateways, its removal is being proposed here.

mintplugins:
Fixes #7333


mintplugins:
As far as I can tell this is the only change that needs to be made here. The function which it called (edd_handle_order_item_change) is pretty generic as far as changing statuses goes, so I don't believe any other code is/was specific to this action.

PR #7356, opened 2019-07-15
Move the edd_payment_number filter to EDD\Orders\Order

### LINKED ISSUE #7309: 3.0 Fees do not get recorded properly when the quantity is greater than 1.
### Expected behavior
When I purchase a product with a fee, say a t-shirt with shipping, and set the quantity to 10, I expect the fee to be applied 10 times. 

### Actual behavior
The fee is applied 1 time.

1) Add a product with shipping to the cart (using Simple Shipping extension)
2) Set the quantity to 10
3) Complete your purchase and check the order_adjustments table. Notice it does not reflect the cost of the `fee * 10`, but only `fee * 1`.

Screen Shot 2019-06-20 at 4 22 38 PM

EDD Version (or branch): 3.0

### LINKED ISSUE #7315: 3.0 - Advanced Sequential Order Numbers are not overriding Sequential Order Numbers.
### Expected behavior
When I have [Advanced Sequential Order Numbers](https://easydigitaldownloads.com/downloads/advanced-sequential-order-numbers/) activated and enabled, I expect the `order_number` column in the `orders` table to be populated with the value from `Advanced Sequential Order Numbers`. 

### Actual behavior
The `order_number` column is populated with the value from EDD core's `Sequential Order Numbers` option, instead of the value from the `Advanced Sequential Order Numbers` extension.

The reason I am logging this here (on the EDD core repo) is to investigate whether this is a breakdown in backwards compatibility in EDD core or not. 

1) Enable both `Sequential Order Numbers` and `Advanced Sequential Order Numbers`. Set up a prefix for Advanced so that it's easy to determine whether it was used.
2) Make a purchase.
3) In the `orders` table, check the value in the order_number column.
4) Notice it does not have the prefix.

EDD Version (or branch): 3.0

mintplugins:
Fixes #7315 

To test this, use Advanced Sequential Order Numbers [issue/22](https://github.com/easydigitaldownloads/edd-advanced-sequential-order-numbers/issues/22)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,33 +4,33 @@ sudo: false
 dist: trusty
 
 php:
-- 5.4
-- 5.5
-- 5.6
-- 7.0
-- 7.1
-- 7.2
+  - 5.4
+  - 5.5
+  - 5.6
+  - 7.0
+  - 7.1
+  - 7.2
 
 env:
-- WP_VERSION=latest WP_MULTISITE=0
-- WP_VERSION=4.9.9 WP_MULTISITE=0
-- WP_VERSION=4.8.8 WP_MULTISITE=0
-- WP_VERSION=4.7.12 WP_MULTISITE=0
+  - WP_VERSION=latest WP_MULTISITE=0
+  - WP_VERSION=5.1.1 WP_MULTISITE=0
+  - WP_VERSION=5.0.4 WP_MULTISITE=0
+  - WP_VERSION=4.9.10 WP_MULTISITE=0
+  - WP_VERSION=4.8.9 WP_MULTISITE=0
+  - WP_VERSION=4.7.13 WP_MULTISITE=0
 
 matrix:
   include:
-    - php: 7.1
-      env: WP_VERSION=latest WP_MULTISITE=0
-    - php: 7.1
-      env: WP_VERSION=4.7.12 WP_MULTISITE=0
     - php: 5.6
       env: WP_VERSION=latest WP_MULTISITE=1
     - php: 5.3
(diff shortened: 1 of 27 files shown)


cklosowski:
@mintplugins my main concern with removing it from `EDD_Payment` is that we may have people still using `EDD_Payment` directly while we transition people to using the order system, and therefore the filter would no longer run. Is there any negative side affect to leaving it there?

mintplugins:
@cklosowski Just to clarify after digging in, EDD_Payment gets its order_number value from EDD\Orders\Order, so the filter does get run for EDD_Payment.

mintplugins:
This actually completely broke sequential orders, despite fixing Advanced Sequential Orders. 

The EDD_Order class is somehow using the same order_number for every order. It's almost like the class isn't actually being instantiated in the EDD_Payment call...

Screen Shot 2019-07-24 at 6 11 19 PM

So the static var is returning a cached result for every single order afterwards. Digging in...

mintplugins:
Confirmed that this is caused by the static variable we added to prevent the infinite loop issue. removing the static var allows sequential order to work properly again.

mintplugins:
@cklosowski This is working for me now, and the unit tests are passing. This actually follows the pattern of all other values set up in EDD_Payment->setup_payment, and probably should have been like this all along. 

Basically, for all other values in EDD_Payment->setup_payment, no filters are run upon setup. They are only run when the value is fetched later through the magic getter method callback.

cklosowski:
All the failing checks are a PHP bug with the date query that isn't related to this issue. Approved and merging.

PR #7380, opened 2019-08-20
Issue/7240 - Fix rounding issues with tax migrations, etc...

### LINKED ISSUE #7240: 3.0 - Migration: Tax rates on orders are rounded incorrectly, empty tax rates being migrated into order adjustments.
A few things to fix for migrating tax rates:
1) While the tax rates themselves are migrated correctly, the order adjustment tax rates are being rounded incorrectly. I migrated a tax rate of 6.8% but on the order adjustment it was 7%.
2) When no tax rate applies to an order, we still have a 0% tax rate added into the order adjustments. I would state that unless a tax rate is applied to an order, we shouldn't be migrating or inserting one.
3) @mintplugins  suspects that the 0 rates are always applying even though a tax rate for an order was specified as it's looking at the use_taxes setting. Need to investigate this further.

JJJ:
Fixes #7240.

Proposed Changes:
1. Add `validate_decimal()` to BerlinDB
2. Remove `edd_sanitize_amount()` callbacks from custom database columns
3. Normalize `edd_sanitize|format_amount()` functions internally
4. Add a bunch of clarifying documentation to sanitize/format amount functions

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -865,23 +865,26 @@ class Data_Migrator {
 			? (float) $meta['_edd_payment_tax_rate'][0]
 			: 0.00;
 
-		// Tax rate is no longer stored in meta.
-		edd_add_order_adjustment( array(
-			'object_id'   => $order_id,
-			'object_type' => 'order',
-			'type_id'     => 0,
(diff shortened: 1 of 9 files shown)


cklosowski:
This seems to be working for the migration, and for new purchases just fine. Going to go ahead and merge it in. Nice work @JJJ

PR #7408, opened 2019-10-03
Refactor how edd_get_payment_status finds the status, with unit tests #7292

### LINKED ISSUE #7292: edd_get_payment_status() always return false if status label is translated.

cklosowski:
Fixes #7292

### DIFF
--- a/includes/payments/functions.php
+++ b/includes/payments/functions.php
@@ -711,28 +711,26 @@ function edd_get_payment_status( $payment, $return_label = false ) {
 
 	if( is_numeric( $payment ) ) {
 
-		$payment = new EDD_Payment( $payment );
+		$payment = edd_get_payment( $payment );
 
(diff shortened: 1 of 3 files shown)


cklosowski:
Resubmitting this PR into the `release/2.9.18` branch since we had to punt it out of `master` for 2.9.17 to be released

mintplugins:
@cklosowski I just added one more test to ensure that the translation is actually in effect:

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7408/commits/617d60075e142cf0fbecbc9d33058a4bb8669559

PR #7429, opened 2019-10-28
Introduce convert_file_string_to_array() for product import that ignores / for delimiters. #7199

### LINKED ISSUE #7199: Download import tool Amazon S3 path issue
### Expected behavior
Within my CSV file I have my download path set like this for each product:

`bucketname/BEPE_EnglishDaisy_20180518_2.jpg`

However, when the import tool is used it breaks this up into two separate files within the "download", based on the /. I would expect this to be treated like an Amazon S3 path and a single file created.

### Actual behavior
We end up with two files assigned with the incorrect names and paths. For a single path, we should only see one file.

1) Create a basic CSV file with a single download, give the "download files" field this data: bucketname/BEPE_EnglishDaisy_20180518_2.jpg
2) Map and import the contents of the file.
3) Check the download and you will see this:

Screenshot 2019-04-03 21 45 48

PHP Version: 7.2

EDD Version (or branch): 2.9.12

WordPress Version: 5.1.1

pippinsplugins:
Fixes #7199

Proposed Changes:
1. Introduce a `convert_file_string_to_array()` method that ignores `/`s in the file value

This resolves the issue completely without introducing significant complications nor extension-specific compatibility.

### DIFF
--- a/includes/admin/import/class-batch-import-downloads.php
+++ b/includes/admin/import/class-batch-import-downloads.php
@@ -192,7 +192,7 @@ class EDD_Batch_Downloads_Import extends EDD_Batch_Import {
 				// setup files
 				if( ! empty( $this->field_mapping['files'] ) && ! empty( $row[ $this->field_mapping['files'] ] ) ) {
 
-					$files = $this->str_to_array( $row[ $this->field_mapping['files'] ] );
+					$files = $this->convert_file_string_to_array( $row[ $this->field_mapping['files'] ] );
 
 					$this->set_files( $download_id, $files );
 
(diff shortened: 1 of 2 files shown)


includes/admin/import/class-batch-import-downloads.php
```diff
@@ -310,6 +310,49 @@ private function set_price( $download_id = 0, $price = '' ) {
...
 	}
 
+	/**
+	 * Convert a files string containing delimiters to an array
+	 *
+	 * @since 2.9.20
+	 * @param $str Input string to convert to an array
+	 * @return array
+	 */
+	public function convert_file_string_to_array( $str = '' ) {
```
mintplugins: @pippinsplugins Did you already consider putting this in the parent class EDD_Batch_Import?

mintplugins: I'm just thinking it might be useful for other extenders as well in the future.

pippinsplugins: I placed it in the download import class because it was specific to products. I don't think core has a use for it in the base class, but someone else that extends it certainly might.

I'll move it.

PR #7433, opened 2019-10-30
Better pre-checkout tax handling.

tobeyadr:
Proposed Changes:
1. Allow for a filter to decide whether the store location should be used to show taxes in the absence of location data from the user.
This is useful because if you only collect taxes in Canada (like me) US customers get confused by the tax rate and abandon the checkout process.

2. Allow for a filter to modify the customer address.
Useful because a developer can hook into this process and perhaps provide geolocation data based on IP to give better tax estimates and improve the checkout experience.


cklosowski:
@tobeyadr Can you please create this PR against the `release/3.0` branch? We’re trying to finalize the release branch in order to prepare it for an initial beta, and I’d like to try and get any fixes like filters in to 3.0 instead of the `2.9.x` release cycle.

Thank you!

includes/user-functions.php
```diff
@@ -549,7 +549,7 @@ function edd_get_customer_address( $user_id = 0 ) {
 		'state'   => '',
 	) );
 
-	return $address;
+	return apply_filters( 'edd_get_customer_address', $address, $user_id );
```
spencerfinnell: This new filter should be documented: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

spencerfinnell (changes requested):

PR #7439, opened 2019-11-01
Added TLS info to system report using cURL

### LINKED ISSUE #7409: Add TLS Check to System Info

The System Report currently reports if `cURL` is enabled, but it would also be nice to show if `TLS` is enabled as well.

### Justification or use case

Many payment gateways (Stripe especially) require `TLS` to be enabled.

---

Originally brought up in https://github.com/easydigitaldownloads/edd-stripe/issues/318 but after discussion there it was determined having the information available for all extensions would be the most beneficial.

SimonChaw:
Fixes #7409 

Proposed Changes:
1. Added function from referenced SO page to display TLS info on the system report page.

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -1421,6 +1421,7 @@ function edd_tools_sysinfo_get() {
 	$return .= 'PHP Version:              ' . PHP_VERSION . "\n";
 	$return .= 'MySQL Version:            ' . $wpdb->db_version() . "\n";
 	$return .= 'Webserver Info:           ' . $_SERVER['SERVER_SOFTWARE'] . "\n";
+	$return .= 'TLS Info:                 ' . apply_filters( 'edd_get_tlsinfo', $return ) . "\n";
 
 	$return  = apply_filters( 'edd_sysinfo_after_webserver_config', $return );
 
@@ -1467,6 +1468,26 @@ function edd_tools_sysinfo_get() {
 	return $return;
 }
 
+/**
+ * Get TLS Info
+ *
+ * @since       2.9
+ * @return      string
+ */
+function edd_tools_tlsinfo() {
+
+	if( ! function_exists( 'curl_init' ) ) {
+		return 'TLS info unavaible while cURL is disabled';
+	}
+
+	$ch = curl_init( 'https://www.howsmyssl.com/a/check' );
+	curl_setopt( $ch, CURLOPT_RETURNTRANSFER, true );
+	$data = curl_exec( $ch );
+
+	$json = json_decode( $data );
+	return $json->tls_version;
+}
+add_filter( 'edd_get_tlsinfo', 'edd_tools_tlsinfo' );
 
 /**
  * Generates a System Info download file


includes/admin/tools.php
```diff
@@ -1467,6 +1468,26 @@ function edd_tools_sysinfo_get() {
...
+/**
+ * Get TLS Info
+ *
+ * @since       2.9
+ * @return      string
+ */
+function edd_tools_tlsinfo() {
+
+	if( ! function_exists( 'curl_init' ) ) {
+		return 'TLS info unavaible while cURL is disabled';
```
mintplugins: Noting a spelling mistake on "unavaible", and also this should be wrapped in a transalation function.

includes/admin/tools.php
```diff
@@ -1467,6 +1468,26 @@ function edd_tools_sysinfo_get() {
...
+	}
+
+	$ch = curl_init( 'https://www.howsmyssl.com/a/check' );
```
mintplugins: WordPress [recommends](https://github.com/WordPress/WordPress-Coding-Standards/blob/fab1ed525e057c19d723bf7c93feaf0a48512abe/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php#L87) using wp_remote_get over cURL.

includes/admin/tools.php
```diff
@@ -1467,6 +1468,26 @@ function edd_tools_sysinfo_get() {
...
+	curl_setopt( $ch, CURLOPT_RETURNTRANSFER, true );
+	$data = curl_exec( $ch );
+
+	$json = json_decode( $data );
+	return $json->tls_version;
```
mintplugins: This should likely be sanitized using something like sanitize_text_field

includes/admin/tools.php
SimonChaw: Yikes. Sorry about that.

mintplugins: All good!

includes/admin/tools.php
```diff
@@ -1421,6 +1421,7 @@ function edd_tools_sysinfo_get() {
 	$return .= 'PHP Version:              ' . PHP_VERSION . "\n";
 	$return .= 'MySQL Version:            ' . $wpdb->db_version() . "\n";
 	$return .= 'Webserver Info:           ' . $_SERVER['SERVER_SOFTWARE'] . "\n";
+	$return .= 'TLS Info:                 ' . apply_filters( 'edd_get_tlsinfo', $return ) . "\n";
```
spencerfinnell: I think this filter should perhaps be filtering a default `Unavailable` string that is transformed if it is determined TLS is available.

```php
apply_filters( 'edd_get_tlsinfo', esc_html__( 'Unavailable', 'easy-digital-downloads' ) );
```

includes/admin/tools.php
SimonChaw: For some reason when I use wp_remote_get it tells me "TLS Info:                 Err 0 The website calling howsmyssl.com's API has been making many calls and does not have a subscription. See https://subscriptions.howsmyssl.com/signup for how to get one."

But when I use cURL it's fine. Any idea why this might be?

mintplugins: It appears GiveWP went through a similar discussion here: 
https://github.com/impress-org/givewp/issues/2998

SimonChaw: Oh that's great, thank you.

includes/admin/tools.php
SimonChaw: Gotcha, will do!

includes/admin/tools.php
SimonChaw: So it looks like GiveWP's solution was to use file_get_contents.

includes/admin/tools.php
```diff
@@ -1467,6 +1468,29 @@ function edd_tools_sysinfo_get() {
...
+ *
+ * @since       2.9
+ * @return      string
+ */
+function edd_tools_tlsinfo( $fail_response ) {
+
+	$tls_info = false;
+
+	if ( ini_get( 'allow_url_fopen' ) ) {
+		$tls_info = file_get_contents( 'https://www.howsmyssl.com/a/check' );
```
mintplugins: @SimonChaw Just noting that file_get_contents is also discouraged by WPCS. They still recommend wp_remote_get here:

https://github.com/WordPress/WordPress-Coding-Standards/blob/fab1ed525e057c19d723bf7c93feaf0a48512abe/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php#L114

PR #7445, opened 2019-11-05
Fix wrong return type from search_box method

sultann:
There was a type `search_box` method was `@return svoid` instead of `@return void`


cklosowski:
Thanks @sultann im going to close this and fix it in 3.0 release branch. I’m trying to get limited changes into the 2.9.x branches.

PR #7465, opened 2019-11-14
Trigger update payment status hook a second time with 'publish' status. #7399

### LINKED ISSUE #7399: 3.0 - Stripe refunds not working
The new refunds API in 3.0 needs to be integrated and tested with Stripe ( and ultimately all gateways) extensively. 

This issue is to ensure that refunds work with Stripe, and that subscription cancellation functionality from Recurring Payments also works with Stripe.

nosegraze:
Fixes #7399

Proposed Changes:
1. When triggering `edd_update_payment_status`, run it a second time for the `publish` status. This accounts for add-ons listening for status changes from `publish` instead of `complete`.


cklosowski:
@nosegraze This here is an oldie, but quick glance at it, is it ready to go, or do we have more work to do here based off the conversation on the issue itself?

nosegraze:
@cklosowski This is read to go. The original issue expanded beyond what we see here, but then I decided to split off everything else into separate issues since they weren't actually related. So I'd like this PR to just stick to the status change and this is ready to do as-is.

cklosowski:
Merging in the `release/3.0` changes to get unit tests passing.

Issue #7467, opened 2019-11-15
Adding Jilt section to Emails page


mintplugins:
A button to make people aware of Jilt is being added to the Emails page. That will be tracked in this issue.

spencerfinnell:
Email UI with `issue/7467` applied now:

<img width="974" alt="Screen Shot 2019-11-22 at 2 50 47 PM" src="https://user-images.githubusercontent.com/491124/69455958-87be9d00-0d37-11ea-89df-b4ae882d25e4.png">

A few suggestions @mintplugins:

- The action being above the description (for both SendWP and Jilt) goes against most conventions. I think it makes sense to read about the setting before taking action on it.

- Having the two buttons use the same icon doesn't make them stand out as different actions to me. If adding the respective logos isn't possible it might be clearer if there were no icons: 

<img width="981" alt="Screen Shot 2019-11-22 at 2 57 55 PM" src="https://user-images.githubusercontent.com/491124/69456381-80e45a00-0d38-11ea-8042-b58e0b64b127.png">

mintplugins:
@spencerfinnell I agree on both of those. Those changes are good for me here. Do you already have them in a branch you can push?

spencerfinnell:
@mintplugins Nope, but I can open a PR against yours.

spencerfinnell:
@mintplugins PR https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7478

PR #7468, opened 2019-11-15
Add Jilt link in settings

mintplugins:
#7467


jdeeburke:
👍 Looks good from the Jilt side of things!

mintplugins:
@spencerfinnell If you get a moment to review this, that would be great. Once we get an approval it can be merged.

PR #7474, opened 2019-11-23
Add action before purchase receipt table

Spreeuw:
Lacking an alternative, I've seen several plugins (Jilt, Recurring Payments) (ab)use the `edd_payment_receipt_before` action to insert information before the table. For proper HTML this would need to be in a table row (the action bein in the `<thead>` element), but layout wise this is not always desirable. Both the aforementioned plugins output a `<div>` and while the browser will do a best attempt to fix it, a simple additional action would be much better (if not for 2.9, then at least for 3.0)


cklosowski:
@Spreeuw Would you mind making this PR against the `release/3.0` branch? We’re trying to keep things in the `2.9.x` to bugs at this point so we can try and get a beta ready for 3.0. Minimizing ‘features’ like adding filters/hooks is something we’re trying to keep in the upcoming major release as we close in on the beta.

Thanks!

Spreeuw:
done!

templates/shortcode-receipt.php
```diff
@@ -21,6 +21,8 @@
 $cart    = edd_get_payment_meta_cart_details( $order->id, true );
 $user    = edd_get_payment_meta_user_info( $order->id );
 $status  = edd_get_payment_status( $order, true );
+
+do_action( 'edd_payment_receipt_before_table', $payment, $edd_receipt_args );
```
spencerfinnell: I like the idea of this hook to promote more freeform (and valid) HTML. 👍 

This should be documented to ensure it is used, and used properly: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

Spreeuw: I'm all for documentation, but in this case since none of the other actions in this template have documentation it would seem strange to me to require it for this action. That may be better handled as a project/pull request in itself. I was just seeing plugins (including your own) abusing another hook for getting the desired result, producing invalid HTML and this is a proper alternative.

spencerfinnell: Existing hooks and filters will eventually need inline documentation (likely as a separate project like you mentioned) so it makes sense to me to go ahead and document any newly introduced hooks and filters added before that project can be started.

I think it is especially important to ensure new hooks and filters in overridable template files have documentation to help theme and plugin authors from using the wrong one or needlessly overriding the entire file.

Spreeuw:
@spencerfinnell I have implemented the requested changed and marked as resolved, not sure if there's anything else I should do here?

templates/shortcode-receipt.php
```diff
@@ -21,6 +21,16 @@
...
 $user    = edd_get_payment_meta_user_info( $order->id );
 $status  = edd_get_payment_status( $order, true );
+
+/**
+ * Output HTML/perform actions before displaying the receipt table
+ *
+ * @since 3.0.0
+ *
+ * @param EDD_Payment  $payment           EDD_Payment object
+ * @param array        $edd_receipt_args  arguments for the receipt shortcode
```
spencerfinnell:
```suggestion
 * Allows additional output before displaying the receipt table.
 *
 * @since 3.0
 *
 * @param \EDD_Payment $payment          Current payment.
 * @param array        $edd_receipt_args [edd_receipt] shortcode arguments.
```

spencerfinnell:
Thank you for your contribution @Spreeuw!

PR #7498, opened 2019-12-18
Check if status key is set before using #7497

### LINKED ISSUE #7497: 3.0: Notice: Undefined index: status
## Bug Report

On a brand new install, I'm seeing the following notice displayed on the order details screen after completing a purchase:

Notice: Undefined index: status in /Users/pippin/Sites/edd/wp-content/plugins/easy-digital-downloads/includes/misc-functions.php on line 1589

Screen Shot 2019-12-17 at 2 15 56 PM

1) On `release/3.0` make a purchase
2) Go to the Edit Order screen
3) The notice is shown just above the Order Adjustments section

PHP Version: 7.0.15

WordPress Version: latest

nosegraze:
Fixes #7497

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1586,7 +1586,7 @@ function edd_format_counts( $counts = array(), $groupby = '' ) {
 			$c[ $count[ $groupby ] ] = absint( $count['count'] );
 
 			// We don't want to include trashed orders in the counts.
-			if ( 'trash' !== $count['status'] ) {
+			if ( ! isset( $count['status'] ) || 'trash' !== $count['status'] ) {
 				$c['total'] += $count['count'];
 			}
 		}


mintplugins:
The fact that the status is intermittently in existence indicates a possible deeper issue to me here.

mintplugins:
@pippinsplugins In the orders table, did your order not have a status?

mintplugins:
If it did have a status, I would have to assume this is not formatting counts for orders (probably adjustments), in which case, having a conditional check for "trash" seems like it could cause problems for things where you _do_ want to count the number of item in the trash. The inline comment indicates that this is for orders only, but no check is being done to ensure that it is actually an "order" and not something else.

pippinsplugins:
@mintplugins it was on my other machine so I can't see right now.

nosegraze:
This function is a super generic function that's used across all database tables. So yes I agree that hard-coding `trash` inside this function instead of somewhere in something only related to orders is kind of strange.

This error was triggered when the adjustments table was running it, which has no status and was grouping by "type".

PR #7508, opened 2019-12-31
Adjust item earnings logic to only subtract fees if purchase date is …

### LINKED ISSUE #7507: Download Product Earnings report incorrect when item-specific fees are present
### Expected behavior
When viewing the earnings report for a single product, the revenue numbers should represent the money earned after fees are collected/deducted.

For example, if a $10 product is sold with a negative 50% fee (a discount), the earnings should be reported as $5 on the graph.

### Actual behavior
The current behavior is that fees are getting deducted from the actual earnings _twice_ when viewing a report.

For example, a $10 product with a negative 50% fee is getting reported as $0.00.

This is happening due to the changes made for #6859. Previously, the `$item['price']` key was the _pre fees_ amount, but after #6859 was merged, `$item['price']` became (correctly) the _post fees_ amount, but we forgot to adjust reports accordingly.

1) Set up a $10 product
2) Using the Discounts Pro extension, set up a 50% cart discount
3) Purchase the product
4) Go to Downloads, locate your product, click on the Earnings column value
5). The earnings will show as $0.00 but should show $5.00

To resolve this, we have a few options:

1. In the `get_earnings()` method of `EDD_Payment_Stats`, we can adjust the base amount off of `item_price` instead of `price`. This corrects the problem but has the unfortunate side effect of making the earnings data incorrect for any products sold with item fees prior to the #6859 fix being applied.

2. We can remove the item-specific fees from the calculation in the `get_earnings()` method of `EDD_Payment_Stats`. This has a similar side effect as option 1 because the `price` key is incorrect for products sold prior to that fix being applied.

3. We could do one of the above options combined with a date check to adjust the logic used when payments were placed before `DECEMBER 3, 2018`, which is the date that our fix for #6859 was released.

I have not yet verified it, but this may only be a problem with _negative_ fees, per the information in #7281.

Related: #3763, #6859, #7281

pippinsplugins:
Fixes #7507

### DIFF
--- a/includes/admin/tools/class-edd-tools-recount-all-stats.php
+++ b/includes/admin/tools/class-edd-tools-recount-all-stats.php
@@ -131,7 +131,7 @@ class EDD_Tools_Recount_All_Stats extends EDD_Batch_Export {
 					 *
 					 * See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7507
 					 */
-					if( $payment->date > '2018-12-03' ) {
+					if ( $payment->date > '2018-12-03' ) {
 
 						$amount = $item['price'];
 
--- a/includes/admin/tools/class-edd-tools-recount-all-stats.php
+++ b/includes/admin/tools/class-edd-tools-recount-all-stats.php
@@ -94,6 +94,8 @@ class EDD_Tools_Recount_All_Stats extends EDD_Batch_Export {
 
 			foreach ( $payments as $payment ) {
 
+				$payment = new EDD_Payment( $payment->ID );
+
 				// Prevent payments that have all ready been retrieved from a previous sales log from counting again.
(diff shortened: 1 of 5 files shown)


cklosowski:
I don't super love the use of a hard coded release date for the 2.9.9 release date however I think that's the only way around this specifically. We can't know for sure that someone updated on that date, but again, it's the closest we can get to the actual date without knowing what version of the plugin that the payment was made on. There isn't any other changes in that release that would affect payment data that we can key in on either.

cklosowski:
@mintplugins I think I'm good with this change, but if you have a few moments can you look it over for me and give your yay or nay on it?

mintplugins:
The code looks fine. I agree that hardcoding a date doesn't seem like a good thing, but not sure there's any other way around it. Testing it now.

PR #7509, opened 2020-01-03
Issue/6840

### LINKED ISSUE #6840: 3.0 - Reports: Tax report does not display any data related to 'global' tax rates, just tax regions.
### Expected behavior
When I have a global tax rate (not a region specific one), I expect the report to show how much tax revenue was collected for it.

### Actual behavior
Only region based tax rates are ported specifically.

1) Enable a fallback tax rate
2) Make a purchase using address information that does not use a specified tax rate, but falls back to the global tax rate.
3) View tax reports, only region specific taxes will be shown in the table.

PHP Version: All

WordPress Version: All

Another way way to replicate is to use `WPCLI` to generate a large number of orders when you have a global tax rate enabled. Since the CLI integration does not supply address information, it will use the fallback.

pippinsplugins:
WIP PR for Fixes #6840


SimonChaw:
@pippinsplugins apologies for the bugs! I missed that when I was testing 😬

PR #7512, opened 2020-01-07
Issue/7511

### LINKED ISSUE #7511: 3.0: Remove default tax rate
Per the conversation [here](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6840#issuecomment-570452997), we are going to remove the default tax rate and then show a notice to site admins when we detect that a default tax rate has been previously configured.

pippinsplugins:
Fixes #7511

### DIFF
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -319,8 +319,7 @@ class EDD_Notices {
 		if ( edd_get_option( 'tax_rate' ) ) {
 
 			// URL to fix this
-			$url = add_query_arg( array(
-				'post_type' => 'download',
+			$url = edd_get_admin_url( array(
 				'page'      => 'edd-settings',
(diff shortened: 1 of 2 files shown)


includes/admin/class-edd-notices.php
```diff
@@ -355,6 +355,29 @@ private function add_settings_notices() {
 					'is_dismissible' => false
 				) );
 			}
+
+			// Default tax rate detected
+			if ( edd_get_option( 'tax_rate' ) && edd_is_test_mode() ) {
```
cklosowski: Are we only showing this notice if we're in `test_mode`? It's likely that our customers will not be in test mode so would never see this notice, if I'm reading this correctly?

pippinsplugins: That was an oops left over from a copy-paste.

PR #7513, opened 2020-01-15
Allow rows/cols attributes on textarea setting fields to be customised

leewillis77:
Currently, if you add a textarea setting that uses edd_textarea_callback() then the rows/cols attributes on the output textarea are hardcoded to 5/50 respectively. This PR leaves those as defaults, but also allows them to be overriden when adding the setting field, e.g.

```php
add_settings_field(
	'edd_settings[my_field]',
        'field description'
	'edd_textarea_callback',
	'edd_settings_my_fieldset_main',
	'edd_settings_my_fieldset_main',
	[
		'id'   => 'my_field',
		'name' => 'field description',
		'desc' => 'field long description'
		'type' => 'textarea',
		'rows' => 16,
                'cols' => 80,
	]
);
```

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -1863,8 +1863,10 @@ function edd_textarea_callback( $args ) {
 	}
 
 	$class = edd_sanitize_html_class( $args['field_class'] );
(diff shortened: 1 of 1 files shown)


leewillis77:
As far as I can tell, the sporadic test failures aren't anything related to these changes?

Issue #7514, opened 2020-01-16
Include minimum PHP version required in readme.txt


mintplugins:
## Enhancement Request
The readme.txt file can indicate the minimum PHP version required, and WordPress core now uses that as part of its health checks and other helpers. 

EDD does not currently include that, but it would be simple to add. 

It also shows on the WP plugin repo (screenshot from woocommerce):
<img width="338" alt="Screen Shot 2020-01-16 at 9 34 36 AM" src="https://user-images.githubusercontent.com/1638128/72533721-9ad78000-3843-11ea-9132-7069123d1e4a.png">

PR #7515, opened 2020-01-16
PR for #7514: Add min php version to readme.txt

mintplugins:
#7514


cklosowski:
We technically support PHP 5.3 for sure as it is in our unit tests at Travis. I know we strive for 5.2 support, however, it isn't testable.

mintplugins:
@cklosowski I bumped it back to 5.3. I'm not sure if it matters, but I believe WP core itself requires 5.6.20 now. 

https://wordpress.org/news/2019/04/minimum-php-version-update/

cklosowski:
They do, but if we bump that in our readme, people can't install updates who are still on legacy versions. We can't deprecate a supported version without a major, and some announcement.

PR #7517, opened 2020-01-18
Remove no downloadable files message when product is service #7486

### LINKED ISSUE #7486: Wrap "No downloadable files found" in filter called edd_receipt_no_files_found_text
### Expected behavior
When I use the Downloads as Services add-on and make a product a "Service", I expect the message "No downloadable files found" to be not shown in the [download_history] shortcode. 

### Actual behavior
The message is shown in the [download_history] shortcode.

1) Install Downloads as Services. 
2) Make a product a "Service". 
3) Purchase the product. 
4) View a page with the [download_history] shortcode and navigate to the files view. 
5) Notice the message is still shown.

This happens because the filter is missing here:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b2ce5ed02d62737dd17ac665a71a493d9efe5898/templates/history-downloads.php#L74

In other places, it is wrapped in a filter called `edd_receipt_no_files_found_text`:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b3b09b718d04ded5611b8b34194cbb77190a2c47/templates/shortcode-receipt.php#L207
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/dc24a2a8a325dc87000ba4492cd0e5dc2006740c/includes/emails/class-edd-email-tags.php#L466

A filter should also be added for bundled items here:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b3b09b718d04ded5611b8b34194cbb77190a2c47/templates/shortcode-receipt.php#L198

Jo-IE:
Fixes #7486 

Used the `edd_receipt_no_files_found_text` function to wrap the `No downloadable files` text.
Fix works locally, text no longer shows up for products that are services.

screenshot

### DIFF
--- a/templates/history-downloads.php
+++ b/templates/history-downloads.php
@@ -71,7 +71,7 @@ if ( $purchases ) :
 										endforeach;
 
 									else :
-										_e( 'No downloadable files found.', 'easy-digital-downloads' );
+										apply_filters('edd_receipt_no_files_found_text', __( 'No downloadable files found.', 'easy-digital-downloads' ), $download['id']);
 									endif; // End if payment complete
 
 								else : ?>
@@ -94,7 +94,7 @@ if ( $purchases ) :
 		?>
 	</table>
 	<?php
-		echo edd_pagination( 
+		echo edd_pagination(
 			array(
 				'type'  => 'download_history',
 				'total' => ceil( edd_count_purchases_of_customer() / 20 ) // 20 items per page
--- a/templates/shortcode-receipt.php
+++ b/templates/shortcode-receipt.php
@@ -195,7 +195,7 @@ $status    = edd_get_payment_status( $payment, true );
 													do_action( 'edd_receipt_bundle_files', $filekey, $file, $item['id'], $bundle_item, $payment->ID, $meta );
 												endforeach;
 											else :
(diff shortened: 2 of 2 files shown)


templates/history-downloads.php
```diff
@@ -71,7 +71,7 @@
 										endforeach;
 
 									else :
-										_e( 'No downloadable files found.', 'easy-digital-downloads' );
+										apply_filters('edd_receipt_no_files_found_text', __( 'No downloadable files found.', 'easy-digital-downloads' ), $download['id']);
```
mintplugins: @Jo-IE It looks like this won't get output to the screen because the `_e` function has been removed. You'll need to replace it with an `echo` command so that it gets output to the screen once the filter has been applied.

Jo-IE: ok, I see. I will get on it right away. Thank you @mintplugins !

Jo-IE:
I'm not sure if the build is failing because of my PR but I see some tests failing. Do I need to rectify those?

cklosowski:
@Jo-IE Don't worry about those. Those failures are due to a change in WordPress core that we just need to write some conditions for. The failures are not for your fix and you don't need to rectify those at this time.

Jo-IE:
ok, thank you @cklosowski

PR #7555, opened 2020-02-05
issue/7545 - Fix payment meta / order meta routing

### LINKED ISSUE #7545: 3.0 - Manual SL renewals not working due to cart_details => cart/cart item meta routing
### Expected behavior

When I manually renew a SL license, I expect the existing license to be renewed.

### Actual behavior

When I manually renew an existing license key, I get a whole new license key instead of the existing one being renewed.

1) Purchase a license key (no recurring).
2) Manually renew that license key.
3) You will see that you've now been issued an entirely new license key as if this was a new purchase instead of a renewal.

PHP Version:

WordPress Version:

Split off from this issue here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7399#issuecomment-556055500

This is an issue with `$payment->cart_details` not correctly routing to the new order meta table, where the flags are being set.

ashleyfae:
Closes #7545

Changes:

1. Items saved in cart details `options` array were previously converted into order item meta. This still happens, but now we handle re-populating that meta in `cart_details` so that plugins looking for their values there can still find them. This doesn't change the _storage_ it just ensures that values that were saved into cart_details can still be retrieved via cart_details.
2. Introduce `edd_30_post_meta_key_whitelist` filter. Post meta keys added to this filter are permitted to be rerouted to order meta.


ashleyfae:
Updated to add the Software Licensing keys to the whitelist: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7555/commits/ca5c5f31c4ac148bb8b50e99e23bc38384fa6145

includes/compat/class-payment.php
```diff
@@ -150,8 +150,21 @@ public function get_post_metadata( $value, $object_id, $meta_key, $single ) {
 			'_edd_payment_total',
 			'_edd_payment_tax',
 			'_edd_payment_number',
+			'_edd_sl_upgraded_payment_id', // EDD SL
+			'_edd_sl_is_renewal', // EDD SL
+			'_edds_stripe_customer_id', // EDD Stripe
```
spencerfinnell, line 155: @ashleyfae If these whitelists are going to be the same (`get_` and `update_`), should we add a private method to this class so we only have to maintain one list, and document one filter?

ashleyfae: Good idea @spencerfinnell ! I'll add that tomorrow.

Issue #7016, opened 2018-11-07
3.0: Reports\get_filter_value( 'dates' ) returns incorrect date range


pippinsplugins:
## Bug Report

The `Reports\get_filter_value( 'dates' )` method returns an incorrect date range when passed a value of `today`.

### Expected behavior
Date range returned should be 00:00:00 of the current day until 23:59:59 of the current day.

### Actual behavior
A multi-day range is returned.

Example return: Array ( [from] => 2018-11-06 [to] => 2018-11-07 [range] => today

pippinsplugins:
Seems this actually affects all ranges.

For example:

`Last Month` returns a range that includes the last day of the month before the previous month (I.E September 30 - October 31).

`This Month` returns  October 31 to November 30.

`Today` returns November 5 to November 7.

`Yesterday` returns November 4 to November 6.

`Custom` with `2018-11-05` and `2018-11-05` selected actually returns November 4 to November 7.

pippinsplugins:
I believe this has to do with the `setTimezone()` method that used a lot:
```php
'start' => $date->copy()->setTimezone( edd_get_timezone_id() )->startOfDay()->setTimezone( 'UTC' ),
'end'   => $date->copy()->setTimezone( edd_get_timezone_id() )->endOfDay()->setTimezone( 'UTC' ),
```

We if remove the usage of `setTimezone( edd_get_timezone_id() )` the dates are closer to what is expected.

@DrewAPicture can you weigh in on this please?

sunnyratilal:
This is actually correct behaviour. @JJJ and I spent a _lot_ of time with the issue of time zones.

As of 3.0, all dates will be stored in UTC and as part of that we need to apply offsets to dates/times to ensure the correct offsets get passed into the query classes.

For example, if your server offset is UTC - 2 and we are using the “today” range, we need to query the previous day. The database will return the correct records and then we apply further offset calculations to ensure the rows have your server time displayed.

If  the offset calculations are removed it’ll cause all the reports to break and subsequently report inaccurate data so these steps had to be taken.

The offset calculations are applied all across Core now, even on the list tables to ensure the correct data is returned from the database.

FWIW, @JJJ and I have extensively tested this to ensure that it does indeed work and I wrote a plethora of unit tests to verify that this is indeed the correct behaviour.

Happy to elaborate further.

DrewAPicture:
[...]

PR #7559, opened 2020-02-06
issue/7528 - Create Separate UI for Viewing Refunds

### LINKED ISSUE #7528: 3.0 - Orders - Create Separate Refund UI

Hide the Refunds order type UI.

### Justification or use case

Exposing the `refund` order type UI causes confusion between partially refunded orders, fully refunded orders (via refunding all line items), and the actual `refund` record itself without adding any additional information that cannot be easily found on the main order. As it stands almost everything in the UI is irrelevant to the record itself (logs, details, attributes, adjustments, etc).

Instead, just the date of refund could be added to the line item on the original order:

Screen Shot 2020-01-22 at 5 59 07 PM

Perhaps there were other intentions for the separate record UI but as far as I can tell it's not needed currently. cc @JJJ @cklosowski maybe you have more insight?

spencerfinnell:
Fixes #7528 

Proposed Changes:
1. Create separate UI for viewing refunds.

Screen Shot 2020-04-24 at 10 39 58 AM
Screen Shot 2020-04-24 at 10 42 36 AM


pippinsplugins:
@spencerfinnell I got a lot of `Parse error: syntax error, unexpected ')'` errors due to trailing commas. I'm on PHP 7.0.15.

I'm pushing a commit up in a moment to fix them.

includes/admin/payments/class-payments-table.php
```diff
@@ -528,7 +528,9 @@ public function column_number( $order ) {
 		// View URL
 		$view_url = edd_get_admin_url( array(
 			'page' => 'edd-payment-history',
-			'view' => 'view-order-details',
+			'view' => 'order' === $order->type
```
pippinsplugins: This is breaking for me. All completed orders are using `view-refund-details`.

My orders all habe a `type` of `sale` not `order`.

spencerfinnell: Oops, nice catch. Fixed.

spencerfinnell:
This will need to be refreshed after #7599 is merged.

includes/admin/payments/refunds.php
```diff
@@ -0,0 +1,200 @@
...
+		)
+	);
+?>
+
+<div class="edit-post-editor-regions__header">
+	<div class="edit-post-header notice-warning" style="border-left-width: 4px; border-left-style: solid;">
+
+		<div class="edit-post-header__settings">
+			<a href="<?php echo esc_url( $order_url ); ?>" class="button button-secondary" autofocus>
+				<?php esc_html_e( 'View Order', 'easy-digital-downloads' ); ?>
```
robincornett, line 39: What would you think about changing this to "View Original Order" instead? (or "Related" but I think original makes more sense in this context).

includes/admin/payments/refunds.php
```diff
@@ -0,0 +1,200 @@
+<?php
+/**
+ * Refund Details Sections
+ *
+ * @package     EDD
+ * @subpackage  Admin/Orders
+ * @copyright   Copyright (c) 2018, Easy Digital Downloads, LLC
```
robincornett, line 7: Since this is a new file, should be 2020. Wondering if it should be Sandhills instead of EDD?

includes/admin/payments/refunds.php
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * @since 3.0
+ *
+ * @param \EDD\Orders\Order $refund Current Refund.
+ */
+function edd_refund_details_notes( $refund ) {
+?>
+
+<div id="edd-order-items" class="postbox edd-edit-purchase-element">
+	<h3 class="hndle">
```
robincornett, line 156: This should be an `h2` to be the correct heading level.

I would remove the class, as `hndle` appears to have a special JS relationship/meaning in WP (for movable boxes, which this is not). Removing it just removes the border underneath, nothing else, so maybe we can implement an edd class if the border is important and there isn't another one we could appropriate.

includes/admin/payments/refunds.php
```diff
@@ -0,0 +1,200 @@
...
+		<span><?php esc_html_e( 'Refund Notes', 'easy-digital-downloads' ); ?></span>
```
robincornett, line 157: Why the `<span>`? Feels unnecessary.

includes/admin/payments/refunds.php
```diff
@@ -0,0 +1,200 @@
...
+ * @since 3.0
+ *
+ * @param \EDD\Orders\Order $refund Current Refund.
+ */
+function edd_refund_details_attributes( $refund ) {
+	$refund_date = edd_get_edd_timezone_equivalent_date_from_utc( EDD()->utils->date( $refund->date_created, 'utc', true ) );
+?>
+
+<div class="postbox">
+	<h3 class="hndle">
```
robincornett, line 180: See the heading/class/span notes above.

Also not sure about the wording, "Refund Attributes", but not sure I have a better suggestion. Maybe "Details"?

spencerfinnell:
@robincornett Addressed most of the feedback, thanks! I'm not sure about 

> this applies to viewing orders as well, but I need to write it here while I'm thinking about it. I do not like that the product titles are linked in the view orders screens (including here). If I'm reviewing an order, I do not necessarily want a primary link to take me off to edit the product.

I'm not convinced one way or the other about if the links should remain, or where they should go, so let's see if there is any other feedback regarding that. Another possibility would be linking them to their relevant Reports page, which may have more of a relation to Sales/Refunds.

includes/admin/payments/refunds.php
```diff
@@ -2,11 +2,11 @@
...
  * Refund Details Sections
  *
- * @package     EDD
- * @subpackage  Admin/Orders
- * @copyright   Copyright (c) 2018, Easy Digital Downloads, LLC
- * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
- * @since       3.0
+ * @package    EDD
+ * @subpackage Admin/Orders
+ * @copyright  Copyright (c) 2020, Easy Digital Downloads, LLC
```
robincornett, line 7: I believe this should be Sandhills Development, LLC now.

includes/admin/payments/refunds.php
```diff
@@ -166,20 +166,20 @@ function edd_refund_details_notes( $refund ) {
...
 /**
- * Outputs Refund Attribuets box.
+ * Outputs "Refund Details" box.
  *
  * @since 3.0
  *
  * @param \EDD\Orders\Order $refund Current Refund.
  */
-function edd_refund_details_attributes( $refund ) {
+function edd_refund_details_details( $refund ) {
```
robincornett, line 175: There are currently not many details here, which is fine if none are needed.

However, one I would request is a line item referencing the original order **number** (and possibly date). I know the button on the screen takes me to the original order, but in my install, the refund order numbers are quite different from the original order numbers.

It would also be nice if the details could show whether this represents a full or partial refund of the original order.

spencerfinnell: > It would also be nice if the details could show whether this represents a full or partial refund of the original order.

Maybe this Refund screen would be good place to have the "Related Refunds" box that was originally shown on the original Order.

LisaCee:
> this applies to viewing orders as well, but I need to write it here while I'm thinking about it. I do not like that the product titles are linked in the view orders screens (including here). If I'm reviewing an order, I do not necessarily want a primary link to take me off to edit the product.

I like Robin's idea of having it link to the storefront view.  This is what I expect when viewing a refund elsewhere, like on Amazon.  I can't think of a case where the store owner would want to be linked to the edit page.

robincornett:
I can see how a link to the specific download report might be handy--make it easy for a shop owner to see if there is a trend with a certain download.

I'm not sure if I care if the product name in these tables is linked at all--I think potentially adding an action link (or a few) might make more sense, but I know I've inadvertently ended up on the download edit screen more than once from an order screen, and I am not sure it's ever been what I wanted to do. But I'm very willing to concede this could be just me!

LisaCee:
Would it be helpful to have the refund id / link be with the line item, maybe as a hover detail?  In this example I have 2 separate refunds in one order and cannot easily tell which refund is which (except for price in this example, but just go with me here).

image

spencerfinnell:
I like the idea of action links under the Download name (especially ensuring the Refund is clearly linked to the relevant line item), but we just need to be careful it's not overloaded. There is already:

Tax: $1.00 | Discount: $1.00 | Copy Download Links

Which could get large: 

Tax: $1.00 | Discount: $1.00 | Copy Download Links | View Download | View Refund

includes/admin/payments/refunds.php
```diff
@@ -192,21 +202,33 @@ function edd_refund_details_attributes( $refund ) {
 	</h2>
 
 	<div class="edd-admin-box-inside edd-admin-box-inside--row">
-		<span>
-			<?php esc_html_e( 'Created', 'easy-digital-downloads' ); ?>
-		</span>
-		<time datetime="<?php echo esc_attr( EDD()->utils->date( $refund->date_created, null, true )->toDateTimeString() ); ?>">
-			<?php echo edd_date_i18n( $refund->date_created, 'M. d, Y' ) . '<br />' . edd_date_i18n( strtotime( $refund->date_created ), 'H:i' ); ?> <?php echo esc_html( edd_get_timezone_abbr() ); ?>
-		</time>
+		<div class="edd-form-group">
```
robincornett, line 205: I feel like my notifications pulled up a different version of the changes and I was seeing the order of the Created/Original Order output reversed, and then was super confused when that was not what I was seeing in the admin. My preference would be to keep the refund creation date/time at the top of the box, and then have the original order number/link underneath (partly because with the "Created" not saying "Refund Created" it would be possible to confuse that data with the original order since that's immediately above it).

includes/admin/payments/refunds.php
```diff
@@ -192,21 +202,33 @@ function edd_refund_details_attributes( $refund ) {
...
-		<span>
-			<a href="<?php echo esc_url( $order_url ); ?>"><?php echo esc_html( $order->number ); ?></a>
-		</span>
+	<div class="edd-admin-box-inside">
+		<div class="edd-form-group">
+			<span class="edd-form-group__label">
+				<?php esc_html_e( 'Created', 'easy-digital-downloads' ); ?>
+			</span>
+			<div class="edd-form-group__control">
+				<time datetime="<?php echo esc_attr( EDD()->utils->date( $refund->date_created, null, true )->toDateTimeString() ); ?>" style="line-height: normal">
```
robincornett, line 221: I've noticed this in other places where EDD is displaying the date and/or time, but is it possible for the date/time to use the format set in General > Settings? I've changed mine a couple different ways to experiment but this appears to be hard coded instead of using my preferences.

spencerfinnell: @robincornett I took the same formatting that is used on the main "Orders" table for the "Date" column. There is some discussion around timezone and dates here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7016

The actual format could be changed to your site preference, but for the time being I'm using the Orders table as the source of truth for how to output the date and time.

robincornett: Yes, I asked at one point about creating an issue with the tables not honoring the user settings for dates and times...I think we should honor the settings, with regional differences in how dates are displayed, and 24 hour times potentially confusing some (at least me).

Issue #7239, opened 2019-05-01
3.0 - Improve UI/UX on the order list table.


SDavisMedia:
This issue branches off from: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6642

## Problem

The issue being addressed is that it's hard to click on an Order number from the Orders table, especially when the Order number is only one or two characters. See below:

![Screenshot_2019-05-01 Orders ‹ Easy Digital Downloads 3 0 — WordPress](https://user-images.githubusercontent.com/2359131/57003121-51da8100-6b79-11e9-987b-a384afd3440d.png)

The Order numbers would be easier to click if they were longer. One suggested solution was to make them `display: block;`. However, that isn't recommended because non-complete Orders will show what "State" the Order is in. For example:

![Screenshot_2019-04-30 Orders ‹ Easy Digital Downloads 3 0 — WordPress](https://user-images.githubusercontent.com/2359131/57003168-b269be00-6b79-11e9-9fc7-fd726d7881a3.png)

## Suggested solution

I think a simple way to solve this is to prepend all Order numbers with `View Order: `. This approach provides several benefits, such as:

1. It's familiar, considering the payment history table in previous versions of EDD will show `View Order Details` links for each row.
2. It provides a little bit more consistency from the table itself, to the edit screen for an Order. The edit screen has a heading of `Edit Order: 2`. Clicking into it from a `View Order: 2` link is pretty consistent.
3. The fact that it's currently _just_ a number makes it even more mysterious than normal. Sure, the context of the page should tell you that it's an Order number. But considering the user never _chose_ that Order number because it's created automatically, it just kind of leaves the inexperienced user assuming. `View Order: 2` makes it pretty clear that it's an Order number.
4. It adds decent length to the link, even for Order number `1`, while leaving plenty of room for "states."

Here's how it would look (read comments below the image):

![Screenshot_2019-04-30 Orders ‹ Easy Digital Downloads 3 0 — WordPress(1)](https://user-images.githubusercontent.com/2359131/57003290-594e5a00-6b7a-11e9-96ab-451ec5225d41.png)

As you can see, this solves the original problem but still has the potential to cause wrapping. That's not a deal breaker, but it's something we should address. The number of columns, and the column widths have gotten out of hand in the past. Perhaps certain columns, like Number, should have more defined widths while others have loose but effective restrictions.

Comment by @cklosowski:

[...]

PR #7561, opened 2020-02-06
issue/7239 - Make Clickable Area Larger on "Number" Column

spencerfinnell:
Partially addresses #7239 

Proposed Changes:
1. Ensure clicking anywhere in the "Number" `<td>` or `<strong>` (status) follows the main row action.


spencerfinnell:
This is no longer needed with a separate column: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7239#issuecomment-584199426

PR #7590, opened 2020-02-14
Issue/7578 - Use null or CURRENT_TIMESTAMP for datetime defaults

### LINKED ISSUE #7578: 3.0 - Default datetime value `0000-00-00 00:00:00` causes errors in MySQL 5.7+
## Enhancement Request

All the datetime default values are set to `0000-00-00 00:00:00`. I propose we use `null` instead (or in some cases, like `date_created` we could use `CURRENT_TIMESTAMP`).

### Justification or use case

MySQL version 5.7 defaults to disallowing `0000-00-00 00:00:00` as a default value unless you override the settings to explicitly allow it.

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-strict

> Strict mode affects whether the server permits '0000-00-00' as a valid date:
> 
> If strict mode is not enabled, '0000-00-00' is permitted and inserts produce no warning.
> 
> If strict mode is enabled, '0000-00-00' is not permitted and inserts produce an error, unless IGNORE is given as well. For INSERT IGNORE and UPDATE IGNORE, '0000-00-00' is permitted and inserts produce a warning.

RCP has been using `0000-00-00 00:00:00` for default values too. Recently Mihai was trying to drop a column from inside PhpMyAdmin to do some RCP testing, and he wasn't able to do so because a different column in the same table had a default value of `0000-00-00 00:00:00` https://github.com/restrictcontentpro/restrict-content-pro/issues/2698#issuecomment-584580987

```
SQL query:
ALTER TABLE `wp_rcp_discounts` DROP `one_time`
MySQL said: Documentation

#1067 - Invalid default value for 'expiration'
```

ashleyfae:
Fixes #7578

Proposed Changes:
1. Sets default value for all `date_created` and `date_modified` columns to `CURRENT_TIMESTAMP`.
2. Sets default value for all other date columns to `null`.
3. Updates any functions that check for `0000-00-00 00:00:00` date values to check `is_null()` instead.
4. Patches BerlinDB to allow `null` default datetime values and `null` values to be saved.


nosegraze:
I've edited the unit test `test_edd_add_log_with_null_type_and_no_id()`.

It was failing because it was setting `object_type` to `null`, then asserting that it came out as an empty string. However, the table schema was actually set up to default to null:

```mysql
object_type varchar(20) DEFAULT NULL
```

Before my BerlinDB edits, `null` was being converted to an empty string, which is why that test was previously passing. Now it's actually allowed to be set to a `null` value and the schema supports it, so the test was failing.

Screenshot from 2020-02-14 14-59-25

So I've adjusted the test to expect a null value to come out the other end. Alternatively we can adjust the table schema to not allow null.

Issue #7593, opened 2020-02-20
3.0 - Discounts - Cart Not Respecting Discount Product Requirements


spencerfinnell:
### Expected behavior

When a Discount is created with product requirements, those requirements should be respected when applying in the Checkout.

### Actual behavior

Setting "Apply discount only to selected Downloads" is not honored.

### Steps to reproduce the behavior
1) Create "Product A" for $100.
2) Create "Product B" for $100.
2) Create a 10% Discount that requires "Product A" and only applies to that product.
3) Create a 10% Discount that applies to all products.

### Information (if a specific version is affected):
EDD Version (or branch): `release/3.0`

Any other relevant information:

### `master`:

<img width="826" alt="Screen Shot 2020-02-20 at 3 18 08 PM" src="https://user-images.githubusercontent.com/491124/74975093-70ee1c00-53f4-11ea-8e1f-328ab3b1ec78.png">

```
Download B: ( $100 * .10 ) + ( $100 * .10 ) = $20
Download A: ( $100 * .10 ) = $10

$200 - $30 = $170
$170 * .10 = $17.00
✅$170 + $17 = $187
```

### `release/3.0`:

<img width="826" alt="Screen Shot 2020-02-20 at 3 19 36 PM" src="https://user-images.githubusercontent.com/491124/74975357-efe35480-53f4-11ea-919c-e659112ee914.png">

```
❌Download B: ( $100 * .10 ) + ( $100 * .10 ) = $20
Download A: ( $100 * .10 ) + ( $100 * .10 ) = $20

❌$200 - $40 = $160
❌$160 * .10 = $16.00
❌$160 + $16 = $176
```

These incorrect values follow through to the database.

cklosowski:
@spencerfinnell wouldn't Download B's calculation be the one that's wrong? If we have a discount for All and a discount for Product A, then it should have 2 discounts, and Download B should only have one discount (the all discount)

spencerfinnell:
@cklosowski Oopps, yes. Moved the ❌

spencerfinnell:
PR #7594

PR #7594, opened 2020-02-21
Issue/7593

jonnps:
Fixes [#7593](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7593)

Proposed Changes:
1. Change value in comparison inside of ```edd_is_discount_not_global()```

The function was doing the inverse of what was intend.


jonnps:
The same function is used in EDD API, but with the purpose of check if is global.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/d6bc5fffabe5f50d49e9bca478b6060df7d9bbb1/includes/api/class-edd-api.php#L1607

and

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/d6bc5fffabe5f50d49e9bca478b6060df7d9bbb1/includes/api/class-edd-api.php#L1628

Just putting an "!" in front of function would fix, but i'm not sure if it is the best way to handle it.

spencerfinnell:
@jonnps Those should be changed to check the value of `edd_get_discount_scope()` I think. 

```php
'global' === edd_get_discount_scope( $discount );
```

Issue #7576, opened 2020-02-10
3.0 - Discounts - Redeeming a Discount Resets Some Discount Settings


spencerfinnell:
### Expected behavior

When a Customer redeems a Discount the Discount settings should remain the same.

### Actual behavior

Certain settings (such as Download requirements) are reset/removed.

### Steps to reproduce the behavior
1) Create a Discount with Download Requirements
2) Purchase a Download with the Discount applied.

### Information (if a specific version is affected):
EDD Version (or branch): `release/3.0`

spencerfinnell:
This is likely related to another issue where "Apply discount only to selected Downloads." is checked but the discount is applying to all Order Items/Downloads.

nosegraze:
PR: #7608

Issue #7572, opened 2020-02-10
3.0 - Orders - Renamed Product Name Not Accessible


spencerfinnell:
### Expected behavior

When a Download has been renamed the Order details should make the new name of the Download available to screen readers.

### Actual behavior

The information about the new name is under an EDD tooltip, which is not announced by screen readers.

### Steps to reproduce the behavior
1) Purchase a download.
2) Rename the download.
3) View the Order record.
4) Tab through the Order Items table and listen for announcements.

### Information (if a specific version is affected):
EDD Version (or branch): `release/3.0`

spencerfinnell:
PR #7573 ensures a screen reader will announce the name change.

It's still not 100% keyboard accessible because EDD tooltips are not focusable and therefore cannot be toggled by the keyboard alone. This will need to be dealt with separately project-wide.

Issue #7539, opened 2020-01-23
3.0 - Orders - Amounts Don't Respect Currency Symbol Position


spencerfinnell:
### Expected behavior

When viewing the Order Amounts the amounts should respect the sitewide currency settings.

### Actual behavior

The symbol is forced to the left.

### Steps to reproduce the behavior
1) Create an order with an order item.

### Information (if a specific version is affected):
EDD Version (or branch): `release/3.0`

spencerfinnell:
This was likely setup initially so the actual value (excluding the symbol) could be used to calculate subtotal, total, etc. However these calculations currently don't take in to account different thousand separators so the whole calculation needs to be updated anyway.

We'll need to remove all formatting from the text values before doing any calculations, then format the value before updating the UI.

In the future keeping track of true values as items are added would reduce the amount of back and forth unformatting/formatting and allow for faster calculations.

spencerfinnell:
PR #7541 fixes some additional calculation bugs as well.

Issue #7538, opened 2020-01-23
3.0 - Orders - Can't Enter Comma Decimal in Adjustments


spencerfinnell:
### Expected behavior

When I add an adjustment to an Order I should be able to enter an amount in my chosen currency format.

### Actual behavior

The `<input type="number" />` restricts the value to a number with a `.` decimal.

### Steps to reproduce the behavior
1) Attempt to add `14,40` as a credit amount to an order.

### Information (if a specific version is affected):
EDD Version (or branch): `release/3.0`

spencerfinnell:
A fix is included in https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7541

spencerfinnell:
Fix is not in #7599

PR #7599, opened 2020-02-25
issue/7582 - Orders - Combine Items, Adjustments, and Totals UI

### LINKED ISSUE #7582: 3.0 - Orders - Combine Items, Adjustments, and Amounts UI

Create a unified UI to present Order Items, Adjustments, and Amounts.

### Justification or use case

Currently when viewing or creating an order in EDD 3.0 there are 3 separate areas containing amounts related to the Order: Items, Adjustments, and Amounts.

As it stands the Items and Adjustments are implemented with `WP_List_Table`, separate from the Amounts area:

Screen Shot 2020-02-12 at 11 14 36 AM

The "Order Adjustments" area is likely causing the most disconnect in the UI. Grouping applied Discounts inside of "Order Adjustments" but then displaying "Adjustments" and "Discount" amounts separately in other areas causes confusion as to what an Adjustment actually is.

While the new database has provided tons of great new functionality I think the current UI presents the information too much in a _database-y_ way versus a more traditional invoice. The user does not need to know that we store or classify tax, credit, fee, or discount as an `adjustment`.

Another potentially confusing concept is adjusting or displaying a "Refunded" status for individual line items. To help further enforce the concept of "order amounts cannot change" not adjusting the line item will help show that in this order nothing has changed.

Screen Shot 2020-02-12 at 11 25 21 AM

The above prototype (amounts are not accurate) displays a way to potentially resolve these disconnects.

By not "classifying" adjustments but instead simply listing how they affect the Order subtotal the user has fewer dots to connect.

Beyond that "refunded" line items remain the same after a Refund has been issued and instead a new entry is added to the amount summary after the existing Order total is listed. This helps enforce that the current order has not changed -- but there is a new record that affects the net. The Net amount could be shown below as well.

The line-item Discount and Tax amounts were moved below the Download name as secondary information because having them in their own column disrupts the Discount and Tax total amounts in the summary area. They can be shown when hovering a row if the specific information is needed. 

Clicking "Issue Refund" will work similarly to what is setup now: listing the available order items that can be "refunded". Ideally you should be able to create a refund for a specific item quantity as well as any additional fees.

Creating an Order would present a similar UI, with "Add" buttons replacing the Refund button at the bottom. They would trigger a modal similar to the Refund modal allowing options to be selected (and the ability to override amounts). 

---

Some previous discussion in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7341

spencerfinnell:
Fixes #7582 

Also closes/supersedes open PRs for: #7539, #7538, #7341

This also has https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7463 merged in to it so the WordPress Jest testing environment could be used as well as updated linting rules.

Proposed Changes:
1. Combine Order Items, Order Adjustments, and Order Amounts in to a unified UI.

Screen Shot 2020-02-25 at 9 50 29 AM
Screen Shot 2020-02-25 at 9 50 45 AM
Screen Shot 2020-02-25 at 9 51 02 AM
Screen Shot 2020-02-25 at 10 13 00 AM
Screen Shot 2020-02-25 at 10 36 26 AM
Screen Shot 2020-02-25 at 10 37 50 AM
Screen Shot 2020-02-25 at 10 38 05 AM

----

## Why?

The first iteration of the Add/View Order screen created disconnect between how an Order Item, Order Adjustment, and their respective amounts related to each other. A lot of the numbers presented overlapped each other and made it difficult to determine how calculations were being done.

By combining these components in to a single area the numbers can be presented in a much more linear fashion that make it easy to quickly walk through each Item or Adjustment and see how the numbers are derived.

The final amounts that ultimately get added to subtracted to find the Order's net total are all displayed along a single edge for easy consumption. Individual Discount and Tax amounts are available on Order Items if further specificity is needed.

The "Actions" bar along the bottom of the Order Overview allow these separate flows to contain rich forms without cluttering the initial interface. Without needing to save space we can be verbose with our options and ensure full flexibility when creating an Order while guiding users to make choices that will ultimately keep their records and reporting in tip top shape.

## Backbone?

Isn't that... old?

Yes.

Easy Digital Downloads supports versions of WordPress back to 4.9. This means there is not the luxury of utilizing many of the newly bundled libraries such as React. While these assets could certainly be bundled and managed individually by the plugin itself (and many other plugins do this), I personally felt that utilizing Backbone was an appropriate half-step in the right direction.

Moving to Backbone at least helps the plugin start consuming and manipulating _data_ on the client versus consuming and manipulating the _DOM_. It also starts exposing where the server side of the plugin can improve to help facilitate more rich javascript-based UIs in the future (such as a more robust REST API). 

Using Backbone also aligns with the updated Tax Rate UI in `release/3.0`.

Again, these are half-steps that bring us within a more reasonable distance to eventually making a switch to match more of how WordPress core is beginning to do things. 

## Improvements

Beyond modifying the UI/UX, there are many functional improvements in this version:

- Order Item amounts are automatically calculated.
- Discounts respect their usage restrictions and only apply to the correct Order Items
- Fees can be added
- Fees can be added to a specific Order Item
- Tax rate is calculated based on the selected Address

## Future

This iteration wires in the Refund flow previously done by @cklosowski. In the future this will be redone to utilize similar techniques introduced in this PR as well as hopefully provide more overall refund flexibility (include ability to refund fees, specific quantity amounts, etc).

## Known Things

- There are also a few "hacky" ways that Discounts are tracked at an OrderItem-level to help more easily calculate totals. I believe the we should investigate creating `order_item` Adjustments on the server for Discounts that have product requirements. But for now I worked around it to avoid too much server-side fiddling.
- For some reason calling `edd_update_order_status()` https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/7582/includes/orders/functions/actions.php#L352 is resetting the `wp_edd_orders` record's `discount` to `0.00`
- ~~Be careful when testing Discounts: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7576~~
- The "Download Name Has Changed" Icon has been removed for the time being. It has confused people and the current implementation is not accessible (leaving a random icon #7572). We can add it back if people are confused, but I don't think it will be an issue.
- ~~Setting a tax rate might be confusing to some. You need to click "Address" in the "Order Details" section and assign an address that has a rate. Might need to look at linking the text under the Order Item manual adjustment toggle to take you to that area.~~
- Actual refund view will need some tweaking. This can be done after this is merged in #7559 

There might be more, but opening this for initial testing/feedback.


nosegraze:
Not sure if I'm doing something wrong but tax rates aren't working for me.

I have a tax rate set up:

Screenshot from 2020-02-26 16-55-23

And taxes do get applied when purchasing via the front-end with my saved customer/address.

When creating an order, I've selected this customer with this billing address:

Screenshot from 2020-02-26 16-56-23

But tax is not being applied:

Screenshot from 2020-02-26 16-57-03

---

Also some feedback.. It's probably not related to any of your new work here, but relevant to this UI still.. I really don't like the "Search for a download" dropdown for two reasons:

1. I don't like how there aren't a few downloads already loaded by default. I think I recently saw you tweaked the customer selection dropdown to already load a few customers by default, and I can indeed see that working. It's super nice. Right now on my test site I have like 3 products added and it would be nice if they were just all loaded already because there's only 3.
2. When I start typing, it immediately says `no results found`, then after a second it updates to show the results. So there ARE results, it just shows `no results found` for a second, then corrects itself. It can be confusing/misleading and make you feel like your search didn't work.

---

Overall just want to say this is looking and feeling awesome though. :)

spencerfinnell:
> Also some feedback.. It's probably not related to any of your new work here, but relevant to this UI still.. I really don't like the "Search for a download" dropdown for two reasons:

@nosegraze Agreed for sure. Unfortunately Chosen is not the easiest to work with at the start and utilizing it with Backbone makes it even more annoying. I'll see what I can do to address those issues though.

> Not sure if I'm doing something wrong but tax rates aren't working for me.

Can you confirm a tax rate is applied when creating an Order via the frontend checkout?

pippinsplugins:
This is great work @spencerfinnell!

There are a few issues that stand out to me on the Refund details screen.
Screen Shot 2020-02-26 at 11 41 15 AM

1. The `?` tooltip after `Order items cannot be modified. ` says to issue a refund to adjust amounts. This _is_ a refund so that tooltip is confusing.

2. The `Issue Refund` button is still shown even those we're on a refund order.

3. The `Completed` status is really confusing. Are we using that to indicate that the refund process is done? 

4. The page title is `Edit Order: 12`. What if we changed that to `Edit Refund: 12`?

5. It should not be possible to send a new receipt for a refund order, at least not until such time as we have email notifications for refunds
Screen Shot 2020-02-26 at 11 44 03 AM

6. The order total should be shown in red, not green, since it is a negative amount.

And when issuing a refund:

a. Can we make the product name clickable for easier selection?
Screen Shot 2020-02-26 at 11 46 17 AM

b. Refund orders are generating sale logs for individual products and improperly adjusting the earnings/sales stats. In the screenshot below, order 12 was a refund
Screen Shot 2020-02-26 at 11 48 47 AM

spencerfinnell:
@pippinsplugins Sorry for the confusion -- those Refund issues will be followed up in a refreshed version of https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7559 after the Add/View Order screens are finalized. 

> Refund orders are generating sale logs for individual products and improperly adjusting the earnings/sales stats. In the screenshot below, order 12 was a refund

I'll confirm this exists on `release/3.0` and open a separate issue to address this.

spencerfinnell:
@nosegraze I see the Tax issue. Selecting a saved address is not triggering a `change` event on the Country and Region fields so the tax rate isn't updated. To continue testing you can manually choose a Country and Region in the address fields. Working on a fix to make it happen automatically.

nosegraze:
@spencerfinnell Tax rate is working well now when selecting an address, then adding the download. However, tax rates do not get applied if you do it in the reverse:

- Add a new download.
- Then select the customer.
- Then choose the address.

nosegraze:
Responsive design could also use some work:

Screenshot from 2020-02-27 12-05-22
Screenshot from 2020-02-27 12-05-36

spencerfinnell:
> However, tax rates do not get applied if you do it in the reverse:

@nosegraze There wasn't a conscious decision to not implement this direction, and it would certainly be possible. My concern is that adjusting the address and changing the amounts _after_ an Order Item has been added might not be clear. You can always remove the Order Item and add it again with the correct tax rate automatically calculated but adjusting all Order Items when the address changes might be heavy handed.

The link between the Address determining the Tax Rate is already a little abstract just based on how the two areas are linked on the page. I'm not sure that the first guess to "fix" an Order Item that doesn't have tax would be to adjust the Address again rather than removing the item an adding it again.

nosegraze:
@spencerfinnell I think my concern is with the order of items on the page. Downloads appear first and customer is at the bottom. Does that encourage people to do it in that order? (Add downloads, then add customer.) Not sure if that's true in majority of cases, but it's also not an unreasonable thought.

Perhaps another option is to show some kind of notice if the tax rate changes and include a button/link to "recalculate tax". Clicking that would recalculate tax on all added items. This provides clarity that a tax rate was set at all and leaves it up to the user if they want to recalculate tax on existing items or not.

spencerfinnell:
@nosegraze I had a note somewhere (maybe just internally) about moving "Order Details" first. Since an Order requires a Customer to be created I think that could make sense.

For the same of keeping the scope of this PR somewhat focused I think a button to manually recalculate Order Item amounts that _haven't been previously added with manual amount adjustments_ could work.

cklosowski:
> For the same of keeping the scope of this PR somewhat focused I think a button to manually recalculate Order Item amounts that haven't been previously added with manual amount adjustments could work.

I would agree, a button would be just fine.

robincornett:
Some thoughts/suggestions:

- the output for the main `postbox` elements do not match the visual appearance/flow, so keyboard users will end up having to go through the smaller "sidebar" boxes before they can get to the "main" one for adding and viewing downloads to an order. Changing the order of these in the code will not affect anything visually, but will improve the [keyboard] navigation experience significantly.
- It's disconcerting for a keyboard user to hit the "Delete order" (edit screen) or "Add order" (add screen) as the first actionable element when using the tab key to navigate. I think the flow for this could be improved to move the `edit-post-header` bar to the end of the dom, and then use CSS to position it at the top of the screen so that the visual appearance is the same (this may not be best practice, though, as the visual flow will be different than the physical flow).
- on the edit screen, even if the "save" button is on the right of the top bar visually, it may make sense to have it be the first/primary button reached with the keyboard, instead of the "delete" button

spencerfinnell:
@robincornett Implemented those suggestions.

LisaCee:
This looks fantastic.  I especially like the `select` menus in the popups, like Add Discount.  They populate the data really well.

robincornett:
When I start the process to add a new order and try to add an expensive item to the order (not sure if this is the issue, but it seems to be how I am able to trigger it), I get this error:

```
A non-numeric value encountered in \plugins\easy-digital-downloads.30\includes\ajax-functions.php on line 1494
```
([line 1494](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/7582/includes/ajax-functions.php#L1494))

I have a variable product and am attempting to add the highest price variation to the order (`10000`, but the `1000` option has the same issue; `100` is fine). The item is added to the order, but with a `0` amount, and subsequent items cannot be added to the cart as the subtotal becomes `NaN` (line 1491).

cklosowski:
The generic message of 'Unable to process refund.` is a bit generic. We should attempt to supply a reason?

@spencerfinnell While testing this, I found that all my orders could not be refunded (existing orders) because they are past the refund window. I could not set my refund window to `0`, so I opened up a PR for #7774 

However it makes me wonder if we should find a way to either hide the `Issue Refund` button if we detect that an order is not refundable. There is a case however, that should we allow store admins the ability to override that setting? While a `shop_worker` shouldn't be able to submit a refund past the refund period, a `shop_owner` should be able to override that (much like what happens in a store when a cashier needs a manager key code).

That makes me think this button should either be grayed out and have a helper tooltip to tell them that it cannot be refunded because it is past the refund window, or we find a way to allow someone with the proper permissions to override that (which is a much deeper issue I think).

spencerfinnell:
@cklosowski This PR doesn't address anything specifically related to the Refund modal -- the current implementation is a drop-in of what is in `release/3.0`. Adjustments up to that UI/UX was going to happen separately. 

> While a shop_worker shouldn't be able to submit a refund past the refund period, a shop_owner should be able to override that (much like what happens in a store when a cashier needs a manager key code).

> That makes me think this button should either be grayed out and have a helper tooltip to tell them that it cannot be refunded because it is past the refund window, or we find a way to allow someone with the proper permissions to override that (which is a much deeper issue I think).

I think that makes a lot of sense. I can disable the button and update the tooltip based on the Refund window and current user role.

cklosowski:
@spencerfinnell Yeah let's go ahead and do that. We'll fix up forcing the refund modal in another issue.

I think disabling if out of the refund window with a tool tip when the user cannot refund it is ideal.
If the user is an admin, and it is outside of the refund window, we should somehow notify them that it is out of the window as well, but still allow them to open the modal.

We can address allowing forcing the refund if an admin, as well as the messaging around it being outside of the refund window within the modal in a different process.

spencerfinnell:
@cklosowski How about this:

**Default:**
Screen Shot 2020-04-22 at 10 25 50 AM

**Refund Window Passed**
Screen Shot 2020-04-22 at 10 02 03 AM
Screen Shot 2020-04-22 at 10 02 09 AM

**Current User Override**
Screen Shot 2020-04-22 at 9 59 59 AM
Screen Shot 2020-04-22 at 10 00 07 AM

robincornett:
@spencerfinnell grammar correction, please:

"The refund window for this Order has passed; however, you have the ability to override this."

spencerfinnell:
@cklosowski Here's a branch that implements a way to override refundability: 

https://github.com/easydigitaldownloads/easy-digital-downloads/compare/issue/7582...issue/7582-refund-lock?expand=1

Default overrides if the user has `edit_shop_products`, with a filter to add capabilities, and a final filter for complete control.

cklosowski:
@spencerfinnell That is exactly what I was envisioning. thanks.

cklosowski:
@spencerfinnell as far as the permission, I think you got it mixed up in the comment. the code has `edit_shop_payments` but the comment says `edit_shop_products` I agree `payments is correct. So guessing it is just a mix up in the comment?

spencerfinnell:
@cklosowski Yep, my mistake. Want me to merge that in to this main branch, or review separately?

cklosowski:
@spencerfinnell lets go ahead and do a separate review since it affects more than just this interface.

spencerfinnell:
@cklosowski https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7776

robincornett:
Ran into this with an admin notice showing on the add new order screen:
image
Not sure about the top bar getting pushed down like that, but I do prefer the existing sticky positioning to the alternatives.

spencerfinnell:
@robincornett Any JS errors on the page? The notice should be positioned the same as when you update an Order:

Screen Shot 2020-04-23 at 9 58 42 AM

That is WordPress core behavior.

spencerfinnell:
@robincornett Pushed a fix to ensure `.update-nag` (separate from `.notice`) are moved below the fixed header.

robincornett:
This particular notice doesn't have a class--`update-nag` is the ID. That's accounted for in the WP core CSS, so maybe it used to be standard practice? I wonder if the JS will need to check for the class and/or the ID?

spencerfinnell:
@robincornett Yep, looks like that updater might be using an old practice, which core still supports: 

https://github.com/WordPress/WordPress/search?q=update-nag&unscoped_q=update-nag
https://github.com/WordPress/WordPress/blob/984d1c60b5f4633fa09ecb387d5f35a622313f46/wp-admin/css/common.css#L1559

Updated for that use-case as well.

robincornett:
@spencerfinnell yes, that works perfectly now--I'll see about letting that team know they should consider updating.

robincornett:
Running into some JS issues, I think:

- if I click the "Issue Refund" button, a modal pops up. If I hit `esc` or click the X to close the modal, the order screen reloads, even though I haven't made any changes. If I have made a different change to the order, such as changing the linked customer, that change is lost.
- if I click the `disabled` "Issue Refund" button in the modal, without selecting any order items, the refund is processed and I get a PHP warning:
```
PHP Warning:  array_map(): Expected parameter 2 to be an array, null given in C:\Users\robin\Documents\Websites\plugins\easy-digital-downloads.30\includes\admin\payments\actions.php on line 675
```
- should a refund process before I click the "Save Order" button? I'm sure it would be more complicated, but it might be nice to select the items for the refund, and then somehow a notice could show somewhere that a refund is pending, so there is an additional confirmation step?
- when I click on the overlay, the modal does not close--the add media overlay does and I'm used to being able to close the modal with a sloppy click
- if the focus is somehow outside of the modal, the "esc" key no longer works to close it
- I'm not sure, but I think it would be preferable for the initial focus in the modal to be one of the inputs, not the X (but tabbing gets me straight to the checkboxes so maybe less of an issue)
- on the add order screen, I have some of the same general issues with the modals there, but the page does not reload when I close the modal
- I noticed that I could not click the "submit" button on a discount modal if I had not selected one, but that appeared to be an HTML validation catch, not JS. But preventing the modal closing if the button is disabled would be helpful
- the `edd-submit-refund-submit` button in the modal should be a `<button>`, not an `<a>` please (this may help with the disabled issue but I am not sure)
- minor detail, but if the empty `<td>` in the submit button row had a `colspan` of 4 instead of five, and then the button wrapper td had its at two, and CSS to `text-align:right;`, it would look correct even on smaller screens (or with the text alignment, the empty cell might not be necessary)

I know there is a Refund UI issue open, but I think the work is going on here, so if I should report this elsewhere, please let me know.

spencerfinnell:
@robincornett Can you please open a new issue regarding the UI/UX issues for the Refund flow? They are things I would like to address, but have purposely been left out of this PR which focuses on the Add/Edit flow explicitly. 

Once this PR is merged it will unblock a lot of things that the Refund aspect of managing Orders requires. 

https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7528 is for the Refund once it's been created. We'll want a separate issue to address the actual Refund creation flow (the modal).

PR #7638, opened 2020-03-10
Issue/7637 Correct refund report queries

### LINKED ISSUE #7637: 3.0 - Refunds report queries missing "type = 'refund'" condition, resulting in incorrect counts
### Expected behavior

I expect the refunds report to show me accurate counts for number of refunded orders, refund amount, etc.

### Actual behavior

Screenshot from 2020-03-10 15-48-11

This shows I had 44 refunds last year. But in reality, I had 44 total orders and 1 refund. You can see the discrepancy in the numbers at the top versus the graph below.

Here's one of the refund queries:

```mysql
SELECT COUNT(id) AS total
FROM wp_edd_orders
WHERE 1=1 AND wp_edd_orders.status IN ('complete', 'partially_refunded')   AND wp_edd_orders.date_created >= '2019-01-01 00:00:00' AND wp_edd_orders.date_created <= '2019-12-31 23:59:59'
```

It's missing a condition for `AND type = 'refund'`

1) Make two sales.
2) Refund one of the sales.
3) Check refund report. Sale count should be 2, refund count should be 1, but refund count will be showing as 2.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #7637 

Proposed Changes:
1. Include type SQL in `get_order_count()`. This ensures `type = 'refund'` condition is picked up in refund count queries, and that refund records are not picked up in order queries.
2. Include type SQL `get_order_earnings()`. This ensures `type = 'refund'` condition is picked up refund amounts queries, and that refund records are not picked up in order queries.
3. `get_order_earnings()` now sets default query vars for `type = 'sale'` and `status = array( 'complete', 'revoked', 'refunded', 'partially_refunded' );`. These can be overridden. But these changes ensure that the `Sale / Earnings` report shows gross earnings, before factoring in refunds. See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7637#issuecomment-597617049
4. `get_order_item_refund_count()` now joins with the orders table to ensure it only picks up the order item records from `refund` objects. See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7637#issuecomment-597171748
5. Add more unit tests to cover `get_order_refund_count()`, `get_order_refund_amount()`, and `get_order_item_refund_count()` stat methods.
6. Adjust `get_order_refund_amount()` so that it saves and uses the originally requested `output`. Previously it was impossible to get a `raw` value out; it was only and always returning a formatted value. Now it can handle either. I needed/wanted `raw` for unit tests.
7. Rework refund rate query to count the number of records, as opposed to number of earnings. See: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7637#issuecomment-597692996
8. Remove `partially_refunded` status from refund object queries. This is under the assumption that `partially_refund` status is meant to be associated with _orders_ and not _refunds_.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -1079,6 +1079,7 @@ function edd_register_refunds_report( $reports ) {
 						return apply_filters( 'edd_reports_refunds_refund_rate', $stats->get_refund_rate( array(
 							'range'  => $filter['range'],
 							'output' => 'formatted',
+							'status' => array( 'complete', 'revoked', 'refunded', 'partially_refunded' )
 						) ) );
 					},
 					'display_args'  => array(
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -160,6 +160,13 @@ class Stats {
 		$this->query_vars['column']            = 'total';
 		$this->query_vars['date_query_column'] = 'date_created';
 
+		/*
+		 * By default we're checking sales only and excluding refunds. This gives us gross order earnings.
+		 * This may be overridden in $query parameters that get passed through.
+		 */
+		$this->query_vars['type']   = 'sale';
+		$this->query_vars['status'] = array( 'complete', 'revoked', 'refunded', 'partially_refunded' );
+
 		// Run pre-query checks and maybe generate SQL.
 		$this->pre_query( $query );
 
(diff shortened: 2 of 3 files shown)


cklosowski:
@nosegraze Can we get some unit tests around those queries along side this?

nosegraze:
@cklosowski Yep :+1:

PR #7654, opened 2020-03-18
issue/7652 - Restore `edd_payment_stats_post_statuses` Filter

### LINKED ISSUE #7652: 3.0: Overview report not counting renewal payments
### Expected behavior
The `Overview` report should display the total earnings from all payment records.

### Actual behavior
Renewal payments (status of `edd_subscription`) are not being included. It's likely that no custom payment statuses are getting included but I have not been able to confirm that.

1) View the `Overview` report with a store that has renewal payments from the Recurring Payments extension
2) Compare sale counts and earnings to see discrepancy

spencerfinnell:
Fixes #7652 

Proposed Changes:
1. Restore `edd_payment_stats_post_statuses` filter and ensure all stats use a consistent set of statuses.

### DIFF
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -171,6 +171,15 @@ class Stats {
 		$this->query_vars['type']   = 'sale';
 		$this->query_vars['status'] = array( 'complete', 'revoked', 'refunded', 'partially_refunded' );
 
+		/**
+		 * Filters Order statuses that should be included when calculating stats.
+		 *
+		 * @since 2.7
+		 *
+		 * @param array $statuses Order statuses to include when generating stats.
+		 */
+		$this->query_vars['status'] = apply_filters( 'edd_payment_stats_post_statuses', $this->query_vars['status'] );
+
 		// Run pre-query checks and maybe generate SQL.
 		$this->pre_query( $query );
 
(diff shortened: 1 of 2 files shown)


pippinsplugins:
@spencerfinnell `refunded` sales should not be included in sale nor earnings counts. Looks like they will be based on the changes here.

pippinsplugins:
@spencerfinnell Otherwise this PR fixes the issue.

spencerfinnell:
@pippinsplugins I updated the list of statuses in `class-payment-stats.php` to match what is currently in `class-stats.php`: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-stats.php#L172

I think that implements what was described here: https://eddwp.slack.com/archives/C8F4B00RK/p1583932226015100

> We need a GROSS earnings and and NET. Gross would be everything, NET would be after refunds
>  A refund being processed does not negate that the original transaction occurred

This PR at least brings status parity in the queries and restores the filter functionality. I think if there are additional concerns to what determines a sale or earnings we should follow up with that separately.

pippinsplugins:
@spencerfinnell ok yes, I agree with that.

PR #7658, opened 2020-03-18
issue/7640 - Set Orders to "Partially Refunded" When Refunding an Order Item

### LINKED ISSUE #7640: 3.0 - Partial refunds not changing original order status to "partially_refunded"
### Expected behavior

When I refund part of an order, I expect the order's status to change to `partially_refunded` (both in the database and in the UI dropdown).

### Actual behavior

When I refund part of an order, the status in the database and UI remains `complete`. Also, there is no "Partially Refunded" dropdown option. I assume there is supposed to be one?

1) Have an order with at least two order items.
2) Refund just one of the order items.
3) The refund record will be created, but the original order status remains as "Completed".

PHP Version:

WordPress Version:

spencerfinnell:
Fixes #7640 

Proposed Changes:
1. Set an Order status to "Partially Refunded" when one or more, but not all, Order Items are refunded.

### DIFF
--- a/includes/orders/functions/refunds.php
+++ b/includes/orders/functions/refunds.php
@@ -36,7 +36,7 @@ function edd_is_order_refundable( $order_id = 0 ) {
 	}
 
 	// Only completed orders can be refunded.
-	if ( 'complete' !== $order->status ) {
+	if ( ! in_array( $order->status, array( 'complete', 'publish', 'partially_refunded', true ) ) ) {
 		return false;
 	}
 
--- a/includes/orders/functions/refunds.php
+++ b/includes/orders/functions/refunds.php
@@ -289,9 +289,14 @@ function edd_refund_order( $order_id = 0, $status = 'complete', $order_items = a
 			break;
 		}
 	}
+
 	if ( 'complete' === $status && $all_refunded ) {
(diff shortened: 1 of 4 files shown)


includes/orders/functions/refunds.php
```diff
@@ -36,7 +36,7 @@ function edd_is_order_refundable( $order_id = 0 ) {
 	}
 
 	// Only completed orders can be refunded.
-	if ( 'complete' !== $order->status ) {
+	if ( ! in_array( $order->status, array( 'complete', 'publish', 'partially_refunded', true ) ) ) {
```
ashleyfae: What's the `true` for? Or was that meant to be outside the `array()` ?

spencerfinnell: It tells `in_array()` to a strict type check recommended by WPCS: https://github.com/WordPress/WordPress-Coding-Standards/blob/d2ef2b5370dca0a11c7735bd9e14d711f8e18f32/WordPress/Sniffs/PHP/StrictInArraySniff.php

spencerfinnell: Edit: Oh, yes. ☕️

PR #7666, opened 2020-03-19
Issue/7665

### LINKED ISSUE #7665: 3.0: File Downloads report duplicates products
### Expected behavior

The File Downloads report shows the top 5 most downloaded products. These five products should all be unique products.

### Actual behavior
A product with multiple price options but a single file can result in it being shown multiple times in the top 5 downloads:
image

pippinsplugins:
Fixes #7665 

Proposed Changes:
1. Pull file_id from file downloads log instead of price_id
2. Add support for file_id arg in method to get average file downloads


spencerfinnell:
This doesn't seem to work for my single Download where each Price Option has a different file associated with it. It works when you only group by `product_id` though.

pippinsplugins:
@spencerfinnell updated.

PR #7672, opened 2020-03-19
issue/7660 - Invert Refund Amounts in Report Chart

### LINKED ISSUE #7660: 3.0 - Reports - Invert Refund Chart Amounts

Use positive amounts when showing Refund amounts in charts.

### Justification or use case

The "Refunds" chart currently uses the negative refund amounts to create the chart. This results in the number of refunds vs. the refund amounts being on opposite sides of the axis.

The tiles also use a positive number for the overall refund amounts.

Before:

Screen Shot 2020-03-18 at 2 59 41 PM

After:

Screen Shot 2020-03-18 at 2 59 52 PM

spencerfinnell:
Fixes #7660 

Proposed Changes:
1. Ensure amounts for Refunds are positive when constructing the Report chart.

### DIFF
--- a/includes/admin/reporting/reports-callbacks.php
+++ b/includes/admin/reporting/reports-callbacks.php
@@ -179,21 +179,21 @@ function edd_overview_refunds_chart() {
 				// If the date of this db value matches the date on this line graph/chart, set the y axis value for the chart to the number in the DB result.
 				if ( $date_of_db_value->format( 'Y-m-d H' ) === $date_on_chart->format( 'Y-m-d H' ) ) {
 					$number[ $timestamp ][1]    = $result->number;
-					$amount[ $timestamp ][1] = edd_format_amount( $result->amount );
+					$amount[ $timestamp ][1] = edd_format_amount( $result->amount * -1 );
 				}
 				// Add any refunds that happened during this day.
 			} elseif ( $day_by_day ) {
 				// If the date of this db value matches the date on this line graph/chart, set the y axis value for the chart to the number in the DB result.
 				if ( $date_of_db_value->format( 'Y-m-d' ) === $date_on_chart->format( 'Y-m-d' ) ) {
 					$number[ $timestamp ][1]    = $result->number;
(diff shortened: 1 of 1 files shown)


spencerfinnell:
@cklosowski I looked at that, but it uses `absint()` first which stops it from actually inverting it. A separate `edd_invert_amount()` helper maybe?

cklosowski:
@spencerfinnell it uses `abs`, not `absint`:

```
function edd_negate_amount( $value = 0.00 ) {
	return abs( floatval( $value ) ) * -1;
}
```

```
	$order_data = array(
		'parent'       => $order_id,
		'order_number' => $number,
		'status'       => $status,
		'type'         => 'refund',
		'user_id'      => $order->user_id,
		'customer_id'  => $order->customer_id,
		'email'        => $order->email,
		'ip'           => $order->ip,
		'gateway'      => $order->gateway,
		'mode'         => $order->mode,
		'currency'     => $order->currency,
		'payment_key'  => strtolower( md5( uniqid() ) ),
		'subtotal'     => edd_negate_amount( $subtotal ),
		'discount'     => edd_negate_amount( $discount ),
		'tax'          => edd_negate_amount( $tax ),
		'total'        => edd_negate_amount( $total ),
	);
```

`The absolute value of number. If the argument number is of type float, the return type is also float, otherwise it is integer (as float usually has a bigger value range than integer).`

We specifically use this on amounts already for when we generate a refund payment here:

spencerfinnell:
@cklosowski It's getting the absolute value of `-10.00`, which is `10.00`, then multiplying by `-1`, which keeps it negative.

```php
wp> var_dump( abs( -10 ) * -1 );
float(-10)

wp> var_dump( abs( -10.00 ) * -1 );
float(-10)
```

So we could just call `abs()` directly on these amounts to avoid the multiplication.

cklosowski:
I think that might be easier to read then interpreting the math. Let’s go the `abs` route.

PR #7675, opened 2020-03-19
Fixing reporting warnings of undefined variables #7673

### LINKED ISSUE #7673: 3.0 - Download related reports throw warnings due to missing filters variable.
### Expected behavior
When viewing the Downloads report, no errors should be present.

### Actual behavior
the downloads report throws warnings due to the fact that the `$filter` variable is never defined.\

PHP Version: All

WordPress Version: All

cklosowski:
Fixes #7673 

Proposed Changes:
1. Correct references to the wrong `$filter` variable when getting the `$dates` ranges.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -486,6 +486,7 @@ function edd_register_downloads_report( $reports ) {
 		$dates         = Reports\get_filter_value( 'dates' );
 		$exclude_taxes = Reports\get_taxes_excluded_filter();
 		$label         = $options[ $dates['range'] ];
+		$filter        = Reports\get_filters();
 
 		$download_data = Reports\get_filter_value( 'products' );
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/reports.php
```diff
@@ -486,6 +486,7 @@ function edd_register_downloads_report( $reports ) {
 		$dates         = Reports\get_filter_value( 'dates' );
 		$exclude_taxes = Reports\get_taxes_excluded_filter();
 		$label         = $options[ $dates['range'] ];
+		$filter        = Reports\get_filters();
```
spencerfinnell: We should remove `$filter` from the `use()` statements instead since it's not being used inside of the callbacks.

cklosowski: I'll go ahead and verify that in all the registered ones. Will update this shortly.

cklosowski: All set @spencerfinnell, let me know if this is better.

PR #7696, opened 2020-03-31
Webpack - Always Compile Scripts in a "production" Environment

spencerfinnell:
This PR ensures that running `npm run dev` to make script changes will compile the scripts in an optimized-for-production manner. 

Because we have opted to track/commit build files this will keep the files small, and in turn keep diffs small.

To verify, with this branch checked out, run `npm install && npm run dev` and ensure there are no untracked changes.


JJJ:
I am a big +1 to this, because it reduces the number of file variations from 3 to 2.

* sources, which we use for development
* optimized, which we use for production
* compiled but not optimized, goes away entirely

PR #7698, opened 2020-04-01
Fallback to site currency when payment currency meta is empty.

### LINKED ISSUE #7689: 3.0 - Migration - Undefined index "currency"
### Expected behavior

When I do the migration to 3.0, I expect to not have any errors/notices logged.

### Actual behavior

While running the 3.0 migration, I got the following notice a couple times:

```
PHP Notice:  Undefined index: currency in /var/www/shop/wp-content/plugins/easy-digital-downloads/includes/admin/upgrades/v3/class-data-migrator.php on line 478
```

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L478

I ran this migration on a clone of a live site that had orders dating back to 2013. So I'm not sure if this is a problem with old data or if my site just had stuff missing... But either way, it's easy for us to just check if the key exists before using. If it doesn't, then perhaps we can fall back to the global site currency?

Not including steps here because I don't think it is reproduceable, as I'm pretty sure this is caused by data from 2013 / 2014 or just my bad database. :woman_shrugging: 

PHP Version:

WordPress Version:

jeffreyvr:
Fixes #7689

Uses fallback `edd_get_currency()` when the payment meta currency is empty (as suggested in #7689).

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -475,7 +475,7 @@ class Data_Migrator {
 			'ip'             => $ip,
 			'gateway'        => $gateway,
 			'mode'           => $mode,
-			'currency'       => $payment_meta['currency'],
+			'currency'       => ! empty( $payment_meta['currency'] ) ? $payment_meta['currency'] : edd_get_curreny(),
 			'payment_key'    => $purchase_key,
 			'subtotal'       => $subtotal,
 			'tax'            => $tax,
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -475,7 +475,7 @@ class Data_Migrator {
 			'ip'             => $ip,
 			'gateway'        => $gateway,
 			'mode'           => $mode,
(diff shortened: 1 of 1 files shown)


includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -475,7 +475,7 @@ public static function orders( $data = null ) {
 			'ip'             => $ip,
 			'gateway'        => $gateway,
 			'mode'           => $mode,
-			'currency'       => $payment_meta['currency'],
+			'currency'       => ! empty( $payment_meta['currency'] ) ? $payment_meta['currency'] : edd_get_curreny(),
```
ashleyfae, line 478:
```suggestion
			'currency'       => ! empty( $payment_meta['currency'] ) ? $payment_meta['currency'] : edd_get_currency(),
```

Just a typo in the function name.

jeffreyvr: I can't believe I missed that @ashleyfae, thanks!

cklosowski:
Thanks @jeffreyvr

PR #7700, opened 2020-04-01
Fix illegal string offset in Customers report

### LINKED ISSUE #7699: 3.0 - Illegal string offset in Customers report
### Expected behavior
When I view the Customers report, I want to see the report without errors.

### Actual behavior
When I view the Customers report, I get this warning:

```
PHP Warning:  Illegal string offset 'exclude_taxes' in \plugins\easy-digital-downloads.30\includes\reports\data\customers\class-top-five-customers-list-table.php on line 41
```

1) in my 3.0, I do not have taxes enabled at all. I do not know if it's a factor
2) in 3.0, visit the Reports screen and then click on Customers

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4

The conditional on that line checks if the value of the array key is false, but in my case, the key isn't set at all, so it looks like we could just add an `isset` check to it.

robincornett:
Fixes #7699

Proposed Changes:
Check if the `exclude_taxes` key is set before checking if it's false.


includes/reports/data/customers/class-top-five-customers-list-table.php
```diff
@@ -36,9 +36,9 @@ class Top_Five_Customers_List_Table extends \EDD_Customer_Reports_Table {
 	 */
 	public function get_data() {
 		$data  = array();
-		$taxes = Reports\get_filter_value( 'taxes' );
+		$taxes = Reports\get_taxes_excluded_filter();
 
-		if ( false === $taxes['exclude_taxes'] ) {
+		if ( ! $taxes ) {
```
spencerfinnell, line 41: Lets keep the strict type check with:

```suggestion
		if ( false === $taxes ) {
```

PR #7703, opened 2020-04-02
Fix redundant action text on repeatable rows

### LINKED ISSUE #7702: Repeatable rows repeat link labels
### Expected behavior
Links which include additional screen reader text should not repeat the text.

### Actual behavior
The "Remove" links in mextabox.php (https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/admin/downloads/metabox.php#L443) properly add screen reader text to indicate what item is being removed, but both the visual link and the hidden text include the word "Remove", so it's read twice for screen reader users. Here's the output to clarify:

```html
<a class="edd-remove-row edd-delete" data-type="price">Remove<span class="screen-reader-text">Remove price option 1</span></a>
```

1) Inspect the "Remove" link for any repeatable row
2) If you don't have a screen reader handy, use dev tools to inspect the accessibility properties of the link. Firefox shows:
image

PHP Version: 7.3.5
EDD Version (or branch): master, 3.0
WordPress Version: 5.4

This will also apply to any addons which implement repeater rows in their own metaboxes and use core EDD as an example.

The correct output would be:
```html
<a class="edd-remove-row edd-delete" data-type="price">Remove<span class="screen-reader-text"> price option 1</span></a>
```
which would keep the visual appearance consistent, and remove the repetition.

robincornett:
Removes extra "remove" from the screen-reader-text in repeatable row actions.

Fixes #7702

### DIFF
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -440,7 +440,12 @@ function edd_render_price_row( $key, $args = array(), $post_id, $index ) {
 			$actions['show_advanced'] = '<a href="#" class="toggle-custom-price-option-section">' . __( 'Show advanced settings', 'easy-digital-downloads' ) . '</a>';
 		}
 
-		$actions['remove'] = '<a class="edd-remove-row edd-delete" data-type="price">' . sprintf( __( 'Remove', 'easy-digital-downloads' ), $key ) . '<span class="screen-reader-text"> ' . sprintf( __( 'price option %s', 'easy-digital-downloads' ), $key ) . '</span></a>';
+		$actions['remove'] = sprintf(
+			'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="price">%s</a>',
+			/* translators: %s: download price option */
+			sprintf( __( 'Remove price option %s', 'easy-digital-downloads' ), $key ),
+			__( 'Remove', 'easy-digital-downloads' )
+		);
 		?>
 		<span class="edd-repeatable-row-actions">
 			<?php echo implode( '&nbsp;&#124;&nbsp;', $actions ); ?>
@@ -646,7 +651,14 @@ function edd_render_products_field( $post_id ) {
 										?>
 									</div>
 									<span class="edd-bundled-product-actions">
-										<a class="edd-remove-row edd-delete" data-type="file"><?php printf( __( 'Remove', 'easy-digital-downloads' ), $index ); ?><span class="screen-reader-text"> <?php printf( __( 'bundle option %s', 'easy-digital-downloads' ), $index ); ?></span></a>
+									<?php
+									printf(
+										'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="file">%s</a>',
+										/* translators: %s: download bundle option */
+										sprintf( esc_html__( 'Remove bundle option %s', 'easy-digital-downloads' ), $key ),
+										esc_html__( 'Remove', 'easy-digital-downloads' )
+									);
+									?>
 									</span>
 									<?php do_action( 'edd_download_products_table_row', $post_id ); ?>
 								</div>
@@ -702,7 +714,13 @@ function edd_render_products_field( $post_id ) {
 									?>
 								</div>
 								<span class="edd-bundled-product-actions">
-									<a class="edd-remove-row edd-delete" data-type="file" ><?php printf( __( 'Remove', 'easy-digital-downloads' ) ); ?><span class="screen-reader-text"> <?php __( 'bundle option 1', 'easy-digital-downloads' ); ?></span></a>
+								<?php
+								printf(
+									'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="file">%s</a>',
+									esc_html__( 'Remove bundle option 1', 'easy-digital-downloads' ),
+									esc_html__( 'Remove', 'easy-digital-downloads' )
+								);
+								?>
 								</span>
 								<?php do_action( 'edd_download_products_table_row', $post_id ); ?>
 							</div>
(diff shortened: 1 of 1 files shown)


includes/admin/downloads/metabox.php
```diff
@@ -827,8 +845,14 @@ function edd_render_file_row( $key = '', $args = array(), $post_id, $index ) {
 			<input type="hidden" name="edd_download_files[<?php echo $key; ?>][index]" class="edd_repeatable_index" value="<?php echo $index; ?>"/>
 		</span>
 		<span class="edd-repeatable-row-actions">
-			<a class="edd-remove-row edd-delete" data-type="file"><?php printf( __( 'Remove', 'easy-digital-downloads' ), $key ); ?><span class="screen-reader-text"><?php printf( __( 'Remove file %s', 'easy-digital-downloads' ), $key ); ?></span>
-			</a>
+		<?php
+		printf(
+			'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="file">%s</a>',
+			/* translators: %s: download file */
+			sprintf( esc_html__( 'Remove file %s', 'easy-digital-downloads' ), $key ),
```
spencerfinnell, line 852:
```suggestion
			esc_attr( sprintf( __( 'Remove file %s', 'easy-digital-downloads' ), $key ) ),
```

We need to ensure we escape the replaced value as well.

includes/admin/downloads/metabox.php
```diff
@@ -702,7 +714,13 @@ function edd_render_products_field( $post_id ) {
 									?>
 								</div>
 								<span class="edd-bundled-product-actions">
-									<a class="edd-remove-row edd-delete" data-type="file" ><?php printf( __( 'Remove', 'easy-digital-downloads' ) ); ?><span class="screen-reader-text"><?php __( 'Remove bundle option 1', 'easy-digital-downloads' ); ?></span></a>
+								<?php
+								printf(
+									'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="file">%s</a>',
+									esc_html__( 'Remove bundle option 1', 'easy-digital-downloads' ),
```
spencerfinnell, line 720:
```suggestion
									esc_attr__( 'Remove bundle option 1', 'easy-digital-downloads' ),
```

includes/admin/downloads/metabox.php
```diff
@@ -646,7 +651,14 @@ function edd_render_products_field( $post_id ) {
 										?>
 									</div>
 									<span class="edd-bundled-product-actions">
-										<a class="edd-remove-row edd-delete" data-type="file"><?php printf( __( 'Remove', 'easy-digital-downloads' ), $index ); ?><span class="screen-reader-text"><?php printf( __( 'Remove bundle option %s', 'easy-digital-downloads' ), $index ); ?></span></a>
+									<?php
+									printf(
+										'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="file">%s</a>',
+										/* translators: %s: download bundle option */
+										sprintf( esc_html__( 'Remove bundle option %s', 'easy-digital-downloads' ), $key ),
```
spencerfinnell, line 658: See other note about escaping the replaced value.

includes/admin/downloads/metabox.php
```diff
@@ -440,7 +440,12 @@ function edd_render_price_row( $key, $args = array(), $post_id, $index ) {
 			$actions['show_advanced'] = '<a href="#" class="toggle-custom-price-option-section">' . __( 'Show advanced settings', 'easy-digital-downloads' ) . '</a>';
 		}
 
-		$actions['remove'] = '<a class="edd-remove-row edd-delete" data-type="price">' . sprintf( __( 'Remove', 'easy-digital-downloads' ), $key ) . '<span class="screen-reader-text">' . sprintf( __( 'Remove price option %s', 'easy-digital-downloads' ), $key ) . '</span></a>';
+		$actions['remove'] = sprintf(
+			'<a aria-label="%s" class="edd-remove-row edd-delete" data-type="price">%s</a>',
+			/* translators: %s: download price option */
+			sprintf( __( 'Remove price option %s', 'easy-digital-downloads' ), $key ),
+			__( 'Remove', 'easy-digital-downloads' )
```
spencerfinnell, line 447: See other note about escaping the replaced value.

We should update these to include `esc_html()` and `esc_attr()` as well.

spencerfinnell:
@robincornett Looking at this again, I think we might need to scale back the change some. The current Javascript used to manage these repeatable items is pretty fragile, and inflexible. It manually updates the necessary input names:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/e45245800ef8a8f46a769ce4518597df668b6652/assets/js/admin/downloads/index.js#L39-L55

But this won't look for things like "Remove price option 1". Each item will have the same extended `aria-label`. 

For the time being, instead of trying to fix that JS that needs to be completely rewritten, we should just simplify these to "Remove price option" (and similar), or omit the extended string in `aria-label`.

robincornett:
> For the time being, instead of trying to fix that JS that needs to be completely rewritten, we should just simplify these to "Remove price option" (and similar), or omit the extended string in `aria-label`.

Based on your original suggestions, @spencerfinnell, I think the `aria-label` needs to stay? Because of the translation string issues. Originally the output was:

```html
<a class="edd-remove-row edd-delete" data-type="price">Remove<span class="screen-reader-text">Remove price option 3</span></a>
```

and the goal with this PR was to remove the extra "remove". Because of the possible differences in word order for translations, we changed to an `aria-label`. Removing the number still leaves the word order as a potential issue, right? So even if we remove the reference to the specific option, leaving the "price option" in seems like it's worthwhile.

spencerfinnell:
@robincornett Yes, if we want to keep the full "Remove price option" as the screen reader text we need to keep it in `aria-label`. I was just saying we could possibly remove that since "price option" doesn't necessarily provide that much extra context. 

Keeping "Remove price option", "Remove bundle option", and "Remove file" via `aria-label` with just "Remove" visible in the link works.

PR #7712, opened 2020-04-06
Fix "Recent Purchases" dashboard widget

### LINKED ISSUE #7708: 3.0 - Dashboard widget recent purchases are missing the order ID
### Expected behavior
The "Recent Purchases" on the dashboard widget should show complete information, such as:

> 93 —  Synthia.Lemons@edd.local

### Actual behavior
Currently, the dashboard widget is not showing the order number, only the dash and customer email address:

>  — Synthia.Lemons@edd.local 

1) Make sure you have some recent orders in an EDD 3.0 store.
2) Visit the Dashboard

PHP Version: 7.3.5
WordPress Version: 5.4

It looks like the output here uses [get_the_title](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/dashboard-widgets.php#L171) on the payment ID, and it should maybe just output it as `esc_attr( $payment->ID );` instead.

robincornett:
Proposed Changes:
1. Replace `edd_get_payments` with `edd_get_orders` to get access to order information
2. Replace `get_the_title` with `$payment->get_number()` to properly return the order number
3. Move en dash out of "Recent Purchases" link

Fixes #7708


includes/admin/dashboard-widgets.php
```diff
@@ -149,16 +149,16 @@ function edd_load_dashboard_sales_widget( ) {
 		<div style="clear: both"></div>
 		<?php do_action( 'edd_sales_summary_widget_after_stats', $stats ); ?>
 		<?php
-		$payments = edd_get_payments( array( 'number' => 5, 'status' => 'complete' ) );
+		$payments = edd_get_orders( array( 'number' => 5, 'status' => 'complete' ) );
```
spencerfinnell, line 152: This needs to have `'type' => 'sale',` in order to not return refunds.

spencerfinnell, line 152: I think we should update this to `$orders` to avoid any possible confusion for what is returned.

spencerfinnell:
I'm wondering if we should take this slightly further and match what is output on the Orders history screen by using the Customer name instead of the email address + login name.

Perhaps even changing it to a sentence because we don't really use `&mdash;` like that much anymore.

"44 by Johnny Appleseed"

Screen Shot 2020-04-06 at 2 24 03 PM

This will make the "View All" link to Order history more of a continuation of what you are seeing in the widget vs. new data.

robincornett:
@spencerfinnell if we change that to the sentence, as you suggest, should the output be tweaked so that the translation has context? Replacing just the dash would result in `__( 'by', 'easy-digital-downloads );`, but it seems like the order number and customer name should be included as placeholders with a note to provide context.

spencerfinnell:
@robincornett Yep, we would need to do something like:

```php
sprintf(
  /* translators: Recent order list. %1$s Order number. %2$s Customer name. */
  __( '%1$s by %2$s', 'easy-digital-downloads' ),
  '',
  ''
);
```

includes/admin/dashboard-widgets.php
```diff
@@ -153,42 +153,57 @@ function edd_load_dashboard_sales_widget( ) {
...
+				?>
+				<li class="edd_order_label">
+					<a href="<?php echo esc_url( $link ); ?>">
 						<?php
-					} // End foreach ?>
-				</tbody>
-			</table>
+						$customer      = edd_get_customer( $order->customer_id );
+						$customer_name = $customer->name ? $customer->name : __( 'No Name', 'easy-digital-downloads' );
+						$item_counts   = edd_get_order_item_counts( array( 'order_id' => $order->ID ) );
```
spencerfinnell, line 175: My fault, wrong function suggestion. This should be `edd_count_order_items()` which returns an `int`.

includes/admin/dashboard-widgets.php
```diff
@@ -153,42 +153,57 @@ function edd_load_dashboard_sales_widget( ) {
...
+						printf(
+							wp_kses_post(
```
spencerfinnell, line 177: These should be switched to avoid code standard notices. `wp_kses_post()` needs to sanitize the entire output; right now it's not sanitizing any of the placeholder replacements.

includes/admin/dashboard-widgets.php
```diff
@@ -153,42 +153,57 @@ function edd_load_dashboard_sales_widget( ) {
...
-								<a href="<?php echo add_query_arg( 'id', $order->ID, admin_url( 'edit.php?post_type=download&page=edd-payment-history&view=view-order-details' ) ); ?>">
-									<span class="edd_price_label"><?php echo edd_currency_filter( edd_format_amount( $order->total ), edd_get_payment_currency_code( $order->ID ) ); ?></span>
-								</a>
-							</td>
-						</tr>
+			<h3><?php esc_html_e( 'Recent Purchases', 'easy-digital-downloads' ); ?></h3>
+			<ul>
+			<?php
+			foreach ( $orders as $order ) {
+				$link = add_query_arg(
```
spencerfinnell, line 160: EDD 3.0 has a `edd_get_admin_url()` function which handles a few of these query arguments.

includes/admin/dashboard-widgets.php
```diff
@@ -153,42 +153,57 @@ function edd_load_dashboard_sales_widget( ) {
...
-									); ?>
-								</a>
-							</td>
-							<td class="edd_order_price">
-								<a href="<?php echo add_query_arg( 'id', $order->ID, admin_url( 'edit.php?post_type=download&page=edd-payment-history&view=view-order-details' ) ); ?>">
-									<span class="edd_price_label"><?php echo edd_currency_filter( edd_format_amount( $order->total ), edd_get_payment_currency_code( $order->ID ) ); ?></span>
-								</a>
-							</td>
-						</tr>
+			<h3><?php esc_html_e( 'Recent Purchases', 'easy-digital-downloads' ); ?></h3>
```
spencerfinnell, line 156: To keep nomenclature consistent lets change the title and button to use "Orders".

includes/admin/dashboard-widgets.php
robincornett: I am not getting notices with them in this order, but I can swap it to `echo wp_kses_post( sprintf(...` without error. With the `printf`, each item was being separately escaped.

includes/admin/dashboard-widgets.php
```diff
@@ -149,48 +149,60 @@ function edd_load_dashboard_sales_widget( ) {
...
+				array(
+					'page' => 'edd-payment-history',
+				),
+			);
+			?>
+		<a href="<?php echo esc_url( $all_orders_link ); ?>" class="button-secondary"><?php esc_html_e( 'View All Orders', 'easy-digital-downloads' ); ?></a>
 		</div>
 		<?php } // End if ?>
-		<?php do_action( 'edd_sales_summary_widget_after_purchases', $payments ); ?>
+		<?php do_action( 'edd_sales_summary_widget_after_purchases', $orders ); ?>
```
ashleyfae, line 205: I haven't actually checked, but are there any backwards compatibility concerns with changing this hook parameter? @spencerfinnell

spencerfinnell: I wondered the same thing, but _assumed_ we had similar changes in `release/3.0` already. Looking now, I don't think so! And it's an easy switch back.

Pushed an update.

includes/admin/dashboard-widgets.php
```diff
@@ -152,42 +152,54 @@ function edd_load_dashboard_sales_widget( ) {
...
+		<div class="table recent_orders">
+			<h3><?php esc_html_e( 'Recent Orders', 'easy-digital-downloads' ); ?></h3>
+			<ul>
+			<?php
+			foreach ( $payments as $payment ) {
+				$link = edd_get_admin_url(
+					array(
+						'page' => 'edd-payment-history',
+						'view' => 'view-order-details',
+						'id'   => $payment->ID,
```
ashleyfae, line 164:
```suggestion
						'id'   => urlencode( $payment->ID ),
```

robincornett: Thanks, @ashleyfae! I have two questions about your suggestion, please:
- why `urlencode` instead of something like `esc_attr` (or `rawurlencode` because PHPCS says that's preferable)?
- the entire URL is escaped on output with `esc_url` as well, so are both needed?

includes/admin/dashboard-widgets.php
```diff
@@ -170,7 +170,7 @@ function edd_load_dashboard_sales_widget( ) {
 					<a href="<?php echo esc_url( $link ); ?>">
 						<?php
 						$customer      = edd_get_customer( $payment->customer_id );
-						$customer_name = $customer->name ? $customer->name : __( 'No Name', 'easy-digital-downloads' );
+						$customer_name = isset( $customer->name ) ? $customer->name : __( 'No Name', 'easy-digital-downloads' );
```
spencerfinnell, line 173: @robincornett Maybe fallback to email here?

includes/admin/dashboard-widgets.php
ashleyfae: @robincornett `edd_get_admin_url()` uses `add_query_arg()`. You can see the docs on that here: https://developer.wordpress.org/reference/functions/add_query_arg/ Specifically:

> Values are expected to be encoded appropriately with urlencode() or rawurlencode().

So when we are putting together arguments to add to a URL, we should use `urlencode()` or `rawurlencode()` on anything that isn't a hard-coded value ( `'view' => 'view-order-details'` is a hard-coded value so we don't really need it there).

In this particular example, `$payment->ID` is a pretty safe value (it's just an integer), but it's generally good practise to treat everything like it may not be safe/URL-friendly. That's why we would encode it just in case.

As for why not `esc_attr()`, because this particular item at this point of time isn't being inserted into an HTML attribute. It's just not the right place to use it. If this were alt text or something, then we'd use `esc_attr()` around the entire alt text value.

As for why not `rawurlencode()`, we certainly could use that instead, but in this particular case we're working with an integer, so there shouldn't be a practical difference here. The main differences between the two functions are how spaces get encoded and we don't expect this value to contain spaces. There are also some differences with other characters like `~`, but again, we're not expecting any of those characters here. Either function is fine to use because we're dealing with an integer. If it were a string, may contain spaces, etc. then the difference would matter more.

And yes, we should use `urlencode()` in any values we insert into the URL while we're building it, then use `esc_url()` on the entire formed URL right as we display it.

includes/admin/dashboard-widgets.php
robincornett: I ended up adding the `isset` check because I managed to create/add an order which is apparently not associated with any customer at all, so `edd_get_customer()` does not return an object. Adding the check brings this more in line with a similar check in `class-payments-table.php`, which is where I originally found the "No Name" fallback.

spencerfinnell: Good call on the consistency. I'm not sure it stands out enough as _missing data_ in either place, but keeping them consistent is fine for now, and we can iterate both areas later if needed.

ashleyfae:
@spencerfinnell Feel free to merge if you're happy with this now.

PR #7716, opened 2020-04-07
Update discount amount/type inputs

### LINKED ISSUE #7713: EDD 3.0 - adjust styling, label on discount amount
I'm looking at the updated discounts page in EDD 3.0 and have encountered two issues:
* if I change the discount type to `%`, the `%` character is cut off visually
* the discount type has no accessible text or label. A screen reader user will only hear "percent sign" and "dollar sign" without any indication of what those are changing

Visually, my suggestion would be to replace the current CSS, which uses a lot of absolute positioning and thus requires padding to shim things out, which is not flexible, with something like flexbox, which could give the a very similar visual appearance but with more adaptability for different characters.

Not having a label on the select field is problematic for accessibility, but I understand the desire to visually unify the two inputs. The select field has no label. One option would be to add a visually hidden label, such as:

```html
<label class="screen-reader-text" for="edd-amount-type">Amount Type</label>
```
which would keep the visual appearance as it is, but provide context for screen reader users. It might also be helpful, if possible, to use words for screen reader users instead of the symbols in the select element.

I know historically, the amount type has come before the discount amount in the settings page, but with this visual approach, it might be worth putting the type after the amount, as I'm reading this in my mind as a discount of "ten dollars" or "ten percent".

### Justification or use case
When we're doing creative things visually, we need to make sure they work regardless of screen/browser zoom/selection, and absolute positioning doesn't really lend itself to that. And we do not want to reduce the accessibility of this feature with the new major version, please.

robincornett:
Proposed changes:
* Change CSS to use flexbox instead of absolute positioning
* Remove wrapper span which is no longer needed
* Add a label to the amount type select
* Reorder the amount and amount type inputs

Fixes #7713


assets/css/edd-admin.css
```diff
@@ -596,63 +596,21 @@ textarea[name="edd-note"] {
...
-	border-top-right-radius: 0;
-	border-bottom-right-radius: 0;
-	box-shadow: none;
-	position: relative;
-	z-index: 3;
-}
-
-.edd-amount-type-wrapper input {
-	margin: 0;
+	background-color: #fafafa;
```
spencerfinnell, line 603: I think we should remove this too so we aren't overriding any core input styles besides the border radius.

PR #7718, opened 2020-04-07
Update Sales Summary dashboard widget Earnings/Sales data

### LINKED ISSUE #7709: 3.0 - Dashboard widget "Today" shows total number of sales
### Expected behavior
The "Today" column of the dashboard widget should show the total number of sales made today.

### Actual behavior
The column is showing the total number of sales made through the store.

1) Make sure you have orders in your store covering a range of dates
2) Add an order with today's date
3) View the dashboard widget--the Earnings amount is correct, but the number of sales is the total number of sales in the shop

PHP Version: 7.3.5
WordPress Version: 5.4

The [`get_sales()`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-payment-stats.php#L42) method accepts `0` for the download ID, to query payments for all downloads, but the main conditional doesn't factor in the time parameters if the download ID is `empty`. My guess is that the conditional needs to be adjusted to always check for the date, and not necessarily include the download ID in the date query, if that is possible.

robincornett:
Fixes #7709

Proposed Changes:
1. Create a new `edd_get_dashboard_sales_widget_data` helper function to retrieve earnings/sales counts for the four dashboard widget ranges
2. The new function uses `EDD\Stats` to retrieve all data
3. Update the table output to use the new formatted data, both on initial ajax load and when the heartbeat is received

Props @spencerfinnell for all guidance on this.


includes/admin/dashboard-widgets.php
```diff
@@ -42,6 +42,33 @@ function edd_dashboard_sales_widget( ) {
 	echo '<p><img src=" ' . esc_attr( set_url_scheme( EDD_PLUGIN_URL . 'assets/images/loading.gif', 'relative' ) ) . '"/></p>';
 }
 
+/**
+ * Get the sales earnings/count data for the dashboard widget.
```
spencerfinnell, line 46: WordPress tries to stick with _third-person singular verbs_ for summaries: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#language

So something like `Gets` or `Retrieves` would be more appropriate here.

includes/admin/dashboard-widgets.php
```diff
@@ -42,6 +42,33 @@ function edd_dashboard_sales_widget( ) {
...
+	$ranges = array( 'this_month', 'last_month', 'today', 'total' );
+	foreach ( $ranges as $range ) {
+		$args = array(
+			'range'  => $range,
+			'output' => 'formatted',
+		);
+		if ( 'total' === $range ) {
+			unset( $args['range'] );
+		}
+		$sales          = new EDD\Stats( $args );
```
spencerfinnell, line 62: Minor, but let's call this `$stats` since we have access to the full `EDD\Stats` and not just sales data here.

includes/admin/dashboard-widgets.php
```diff
@@ -54,46 +81,40 @@ function edd_load_dashboard_sales_widget( ) {
...
-						<td colspan="2"><?php _e( 'Current Month', 'easy-digital-downloads' ) ?></td>
+						<td colspan="2"><?php esc_html_e( 'Current Month', 'easy-digital-downloads' ); ?></td>
 					</tr>
 				</thead>
 				<tbody>
 					<tr>
-						<td class="first t monthly_earnings"><?php _e( 'Earnings', 'easy-digital-downloads' ); ?></td>
-						<td class="b b-earnings"><?php echo edd_currency_filter( edd_format_amount( $stats->get_earnings( 0, 'this_month' ) ) ); ?></td>
+						<td class="first t monthly_earnings"><?php esc_html_e( 'Earnings', 'easy-digital-downloads' ); ?></td>
+						<td class="b b-earnings"><?php echo esc_attr( $data['this_month']['earnings'] ); ?></td>
```
spencerfinnell, line 97: These should all use `esc_html()` if they are inside of HTML tags versus HTML attributes.

```suggestion
						<td class="b b-earnings"><?php echo esc_html( $data['this_month']['earnings'] ); ?></td>
```

PR #7728, opened 2020-04-13
Update vertical section navigation to improve accessibility

### LINKED ISSUE #7726: 3.0 - Vertical Sections tabs have accessibility issues
### Expected behavior
Custom navigation elements, such as the vertical sections tabs, should have distinct visual focus styles and appropriate aria attributes.

### Actual behavior
I've been looking at the Customer Details screen but this may be a general `EDD\Admin\Sections` behavior issue (I do not know what other screens use the vertical tabs). This is about the vertical navigation tabs for Profile, Notes, Tools, and Delete.

- When using the keyboard to navigate the screen, there are no focus styles on the navigation tabs
- When the page loads, `aria-selected="true"` is added to the `<li>` corresponding to the current active section, but looking at the source code for the page, JS has added the same attribute to the `<a>` and `<i>` elements of the active tab and the inactive tabs as well. (Loading the page without JS leaves just the initial `<li>` marked this way.)

1) in EDD 3.0, visit a Customer Details screen
2) using the keyboard, tab to the main content, and then keep hitting tab. Your first several presses are going through the `<li>` elements but there is no corresponding visual indicator that you are doing so
3) open dev tools and inspect the `<li>` elements and see all of the incorrect `aria-selected="true"`

PHP Version: 7.3.5
WordPress Version: 5.4

The relevant JS is in `assets\js\admin\components\vertical-sections\index.js`.

robincornett:
Fixes #7726

Proposed Changes:
1. Fixes CSS selector in admin JS for marking vertical section `li` as selected
2. Adds focus/hover styles for vertical tabs
3. Change vertical tab links to use flexbox


assets/css/edd-admin.css
```diff
@@ -3224,17 +3223,25 @@ body.download_page_edd-reports {
...
 	position: relative;
+	align-items: center;
+}
+
+#edd-item-tab-wrapper li a:focus,
+#edd-item-tab-wrapper li a:hover,
+.edd-vertical-sections .section-nav li a:hover,
+.edd-vertical-sections .section-nav li a:focus {
+	box-shadow: inset 5px 0;
+	transition: all .5s;
```
spencerfinnell, line 3241: Let's speed this up some:

```suggestion
	transition: all 0.25s ease-in-out;
```

assets/css/edd-admin.css
```diff
@@ -3224,17 +3223,25 @@ body.download_page_edd-reports {
...
 	border-bottom: 1px solid #e5e5e5;
 	box-shadow: none;
 	position: relative;
+	align-items: center;
+}
+
+#edd-item-tab-wrapper li a:focus,
+#edd-item-tab-wrapper li a:hover,
+.edd-vertical-sections .section-nav li a:hover,
+.edd-vertical-sections .section-nav li a:focus {
```
spencerfinnell, line 3239: When I `:focus` there is briefly an extra outline. I think this needs an explicit `outline: 0;` to avoid temporarily seeing an outline.

assets/css/edd-admin.css
```diff
@@ -3265,68 +3272,97 @@ body.download_page_edd-reports {
...
 }
 
-.edd-vertical-sections .section-nav li[aria-selected="true"] a {
-	position: relative;
+.edd-vertical-sections .section-nav .section-title--is-active a {
 	font-weight: bold;
 	color: #555;
 	background-color: #fff;
 	border-right: none;
 	margin-right: -1px;
```
spencerfinnell, line 3280: I forgot to add this in the PR that implemented `box-shadow`, but we should offset the inner content by another `5px` when active, hovered, or focused, so the icon stays the same distance between the edge and the "border" edge.

robincornett:
> 
> 
> When I switch sections on a JS-enabled (no page refresh) tab list, the first tab does not become unselected.

I think this is because I committed only the main `index.js` file--I thought I was not supposed to commit compiled files? Once the JS has been compiled, the behavior on this should be correct.

assets/css/edd-admin.css
robincornett: Can I confess that I prefer the titles/content here not moving around on focus and hover? I know it results in a visual difference between the perceived edge, but I think I prefer the titles to line up regardless of active/focus/hover state. $.02

spencerfinnell:
@robincornett Thanks, forgot about that.

I see that I can go to a hash directly if I manually update the URL, but I think clicking a tab item should update the URL in the address bar automatically so when you copy/paste a URL it contains the correct tab.

robincornett:
@spencerfinnell to update the URL, would I look into `history.pushState()` JS?

spencerfinnell:
@robincornett Yep!

assets/css/edd-admin.css
robincornett: @spencerfinnell I realized that I forgot to tag you in that comment--have made the other changes you requested, but may I leave this one out so that the titles don't jump around on focus/hover?

assets/css/edd-admin.css
```diff
@@ -3616,6 +3616,10 @@ body.download_page_edd-reports {
 	background-color: #fcfcfc;
 }
 
+.edd-vertical-sections .section-nav:last-of-type {
```
spencerfinnell, line 3619: @robincornett This should be targeting `.section-title:last-of-type`. Right now it's putting margin on the whole list:

Screen Shot 2020-04-29 at 3 06 18 PM
Screen Shot 2020-04-29 at 3 06 24 PM

assets/css/edd-admin.css
spencerfinnell: @robincornett I'm fine with that. 👍

assets/js/admin/components/vertical-sections/index.js
```diff
@@ -1,15 +1,31 @@
...
+	var hash = window.location.hash;
+	if ( hash && hash.includes( 'edd_' ) ) {
+		// Show the section content related to the URL.
+		$( '.edd-vertical-sections.use-js' ).find( hash ).show();
+
+		// Set the aria-selected for section titles to be false
+		$( '.edd-vertical-sections.use-js .section-title' ).attr( 'aria-selected', 'false' );
+
+		// Set aria-selected true on the related link.
+		$( '.edd-vertical-sections.use-js' ).find( '.section-title a[href="' + hash + '"]' ).parents( '.section-title' ).attr( 'aria-selected', 'true' );
```
spencerfinnell, line 20: This `.section-title--is-active` class needs to be toggled alongside `aria-selected`.

robincornett: Thanks; I had missed it in another spot as well.

PR #7729, opened 2020-04-13
issue/7693 - 3.0 - Ensure last_month Is Parsed With No Overflow

### LINKED ISSUE #7693: 3.0 - Reports/Carbon - `last_month` date range not actually retrieving last month
### Expected behavior

When I change the report date range to Last Month, I expect to get results from the previous month (so, at the time of writing -- 30 March -- that would be February).

### Actual behavior

Last month is actually retrieving results from this month (March).

I think these steps will only work today and tomorrow (so until we tick over to April).

1) Go to Downloads > Reports.
2) Set date range to Last Month.
3) Notice how stats are exactly the same as This Month. If you print out the query you can also see that it's querying this month. You can also use `EDD\Reports\parse_dates_for_range( 'last_month' )` directly to check the dates.

Screenshot from 2020-03-30 14-06-26

Screenshot from 2020-03-30 14-06-36

PHP Version:

WordPress Version:

This looks related: https://github.com/briannesbitt/Carbon/issues/428

I think it's a problem with today being March 30th and there not being a February 30th. Seems like changing `subMonth` to `subMonthNoOverflow` fixes the issue.

spencerfinnell:
Fixes #7693

Proposed Changes:
1. Ensures parsing `last_month` on a date with more days than the previous still fetches the true "last month"

### DIFF
--- a/tests/reports/tests-reports-functions.php
+++ b/tests/reports/tests-reports-functions.php
@@ -408,8 +408,8 @@ class Reports_Functions_Tests extends \EDD_UnitTestCase {
 	 */
 	public function test_parse_dates_for_range_with_last_month_range_should_return_those_dates() {
 		$expected = array(
-			'start' => self::$date->copy()->subMonth( 1 )->startOfMonth()->toDateTimeString(),
-			'end'   => self::$date->copy()->subMonth( 1 )->endOfMonth()->toDateTimeString(),
+			'start' => self::$date->copy()->subMonthNoOverflow( 1 )->startOfMonth()->toDateTimeString(),
+			'end'   => self::$date->copy()->subMonthNoOverflow( 1 )->endOfMonth()->toDateTimeString(),
 			'range' => 'last_month',
 		);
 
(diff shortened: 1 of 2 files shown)


tests/reports/tests-reports-functions.php
```diff
@@ -422,6 +422,29 @@ public function test_parse_dates_for_range_with_last_month_range_should_return_t
...
+	 * @covers \EDD\Reports\parse_dates_for_range()
+	 * @group edd_dates
+	 */
+	public function test_parse_dates_for_range_with_overflow_last_month_range_should_return_those_dates() {
+		$overflow_day  = '2020-03-30';
+		$overflow_date = EDD()->utils->date( $overflow_day );
+
+		$expected = array(
+			'start' => $overflow_date->copy()->subMonthNoOverflow( 1 )->startOfMonth()->toDateTimeString(),
+			'end'   => $overflow_date->copy()->subMonthNoOverflow( 1 )->endOfMonth()->toDateTimeString(),
```
ashleyfae, line 435: Since we're passing in an explicit `$overflow_day`, would it make more sense and be possible for these `start` and `end` values to be the hard-coded dates we expect instead of using Carbon again?

PR #7734, opened 2020-04-13
Issue 7725, Add Hidden Label for Email Field

### LINKED ISSUE #7725: Email address field on Customer Details screen is missing a label
### Expected behavior
Form elements should have labels. For assistive technology users, a placeholder is an insufficient replacement for a `<label>` element.

### Actual behavior
The email address field on the Customer Details screen does not have any kind of label.

1) On the Customer Details screen, inspect the field for adding an email address to the customer's profile. There is no label.

PHP Version: 7.3.5
EDD Version (or branch): master, release/3.0
WordPress Version: 5.4

Adding a label element which is visually hidden will improve the accessibility of this input field.

Related: the "Primary" `span` (class `edd-chip`) has insufficient color contrast (1.91) and is very small text.

LisaCee:
Fixes #7725 

Proposed Changes:
1. Added hidden label to email field for screen reader accessibility

To Test:
1. Downloads -> Customers -> Customer Emails
2. No label visible, only placeholder text
3. Inspect field element and see hidden label

### DIFF
--- a/includes/admin/customers/customers.php
+++ b/includes/admin/customers/customers.php
@@ -800,7 +800,7 @@ function edd_customers_view( $customer = null ) {
 						<div class="add-customer-email-wrapper">
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
-							<label class="screen-reader-text" for="additional-email"><?php esc_html_e( 'Email Address', 'easy-ditigal-downloads' ); ?></label>
+							<label class="screen-reader-text" for="additional-email"><?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?></label>
 							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
 							<input type="checkbox" name="make-additional-primary" value="1" id="make-additional-primary" />&nbsp;<label for="make-additional-primary"><?php esc_html_e( 'Make Primary', 'easy-digital-downloads' ); ?></label>
 						</div>
--- a/includes/admin/customers/customers.php
+++ b/includes/admin/customers/customers.php
@@ -801,7 +801,7 @@ function edd_customers_view( $customer = null ) {
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
 							<label class="screen-reader-text" for="additional-email"><?php esc_html_e( 'Email Address', 'easy-ditigal-downloads' ); ?></label>
-							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html__( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
+							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
 							<input type="checkbox" name="make-additional-primary" value="1" id="make-additional-primary" />&nbsp;<label for="make-additional-primary"><?php esc_html_e( 'Make Primary', 'easy-digital-downloads' ); ?></label>
 						</div>
 					</td>
--- a/includes/admin/customers/customers.php
+++ b/includes/admin/customers/customers.php
@@ -800,7 +800,7 @@ function edd_customers_view( $customer = null ) {
 						<div class="add-customer-email-wrapper">
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
-							<label class = "hidden" for="additional-email">Email Address</label>
+							<label class = "hidden" for="additional-email"><?php_e( 'Email Address', 'easy-ditigal-downloads' ); ?></label>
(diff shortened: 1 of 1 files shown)


includes/admin/customers/customers.php
```diff
@@ -800,6 +800,7 @@ function edd_customers_view( $customer = null ) {
 						<div class="add-customer-email-wrapper">
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
+							<label class = "hidden" for="additional-email">Email Address</label>
 							<input type="email" name="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
```
spencerfinnell, line 804: @LisaCee A `label` is linked to an `input` via matching the label's `for` attribute to the input's `id` attribute. This is because an input can have match name values, such as a list of checkboxes:

```
<input name="foo[]" value="1" />
<input name="foo[]" value="2" />
<input name="foo[]" value="3" />
```

The input below this label needs to have an ID for the label's `for` attribute to be correct.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

includes/admin/customers/customers.php
```diff
@@ -800,6 +800,7 @@ function edd_customers_view( $customer = null ) {
 						<div class="add-customer-email-wrapper">
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
+							<label class = "hidden" for="additional-email">Email Address</label>
```
spencerfinnell, line 803: The label value needs to be wrapped in a WordPress translation function: https://developer.wordpress.org/themes/functionality/internationalization/

For EDD the text domain is `easy-digital-downloads`. 

This ensures people using WordPress in another language see/hear the correctly translated value.

These translation functions can (and should be) combined with WordPress sanitization functions: https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/#escaping-with-localization

spencerfinnell, line 803:
```suggestion
							<label class="screen-reader-text" for="additional-email">Email Address</label>
```

While `.hidden` is correctly hiding the label from showing, the `.hidden` class applies `display: hidden` to the element. Elements in the DOM that are hidden are not announced by assistive technology such as screen readers (in their mind they don't exist, so they are not read). 

The specific `.screen-reader-text` class ensures the text is hidden without changing the `display` property by moving the label off the screen, but without stopping the browser from rendering it. See the notes here: https://make.wordpress.org/accessibility/handbook/markup/the-css-class-screen-reader-text/#explanation-of-some-of-these-css-properties

Just like `.hidden` is a class WordPress includes by default, so is `.screen-reader-text`, so the only thing that needs to change here is which class is used.

(And a few extra spaces snuck in around the `=` that aren't needed)

LisaCee:
I thought that was too easy.  I think I've got it now, though I'm still not sure why phpcs didn't catch the spacing issue.

includes/admin/customers/customers.php
```diff
@@ -801,7 +801,7 @@ function edd_customers_view( $customer = null ) {
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
 							<label class="screen-reader-text" for="additional-email"><?php esc_html_e( 'Email Address', 'easy-ditigal-downloads' ); ?></label>
```
spencerfinnell, line 803:
```suggestion
							<label class="screen-reader-text" for="additional-email"><?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?></label>
```

includes/admin/customers/customers.php
```diff
@@ -801,7 +801,7 @@ function edd_customers_view( $customer = null ) {
...
-							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
+							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html__( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
```
spencerfinnell, line 804:
```suggestion
							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
```

This was correct before. We want the `input` placeholder to still be visible to give a hint of the field type. The `input` `placeholder` is acting as the "label" for non-vision impaired users and the actual `<label>` value will be read via screen reader for vision impaired users.

includes/admin/customers/customers.php
```diff
@@ -800,7 +800,8 @@ function edd_customers_view( $customer = null ) {
 						<div class="add-customer-email-wrapper">
 							<input type="hidden" name="customer-id" value="<?php echo esc_attr( $customer->id ); ?>" />
 							<?php wp_nonce_field( 'edd-add-customer-email', 'add_email_nonce', false, true ); ?>
-							<input type="email" name="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
+							<label class="screen-reader-text" for="additional-email"><?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?></label>
+							<input type="email" name="additional-email" id="additional-email" value="" placeholder="<?php esc_html_e( 'Email Address', 'easy-digital-downloads' ); ?>" />&nbsp;
```
ashleyfae, line 804: Can you prefix the ID with `edd-`? Or maybe `edd-customer-`. Leave the name the same though.

I know there are plenty of other places in the file that don't have this prefix, but really we should. So let's add it for any new IDs we're creating.

Issue #444, opened 2012-10-25
Show download link - view download


samikeijonen:
Hello.

Is it possible to have **View download** link, when you create new download, or update one?

I mean of course it's possible, but is it coming to EDD using that same `post_updated_messages` filter in https://github.com/pippinsplugins/Easy-Digital-Downloads/blob/master/includes/post-types.php?

PR #7735, opened 2020-04-13
changed background and font on chip, issue #7731

### LINKED ISSUE #7731: 3.0 - "Chip" Does Not Have Sufficient Contrast
### Expected behavior

When using the "Chip" component (`<span class="edd-chip"></span>`) the rendered chip should have sufficient contrast between the text and the background color.

### Actual behavior

There is not enough contrast.

1) View "Downloads > Customers > Edit Customer > Customer Emails"

Screen Shot 2020-04-13 at 11 57 09 AM

Originally reported in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7725#issue-597387841 with follow up via https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7725#issuecomment-611623947

LisaCee:
Fixes #7731 

Proposed Changes:
1. Increased contrast for `.edd-chip` to `#444`, the core text color
2. Removed border from `.edd-chip` to gain more space for text
3. Increased text on `.edd-chip` to 11px to meet WAVE recommendations

image

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -1924,7 +1924,7 @@ body.download_page_edd-reports {
 	font-size: 10px;
 	text-transform: uppercase;
 	font-weight: bold;
-	line-height: 8px;
+	line-height: 1;
 	vertical-align: middle;
 	padding: 2px 2px 3px 3px;
 	margin: -1px 0 0 2px;
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -1921,7 +1921,7 @@ body.download_page_edd-reports {
 }
 
 .edd-chip {
-	font-size: 7px;
+	font-size: 10px;
 	text-transform: uppercase;
 	font-weight: bold;
 	line-height: 8px;
@@ -1930,8 +1930,7 @@ body.download_page_edd-reports {
 	margin: -1px 0 0 2px;
 	border-radius: 3px;
 	color: #fff;
-	background-color: #bbb;
-	border: 1px solid #999;
+	background-color: #444;
 	opacity: 0.8;
 }
 
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -1927,7 +1927,6 @@ body.download_page_edd-reports {
 	line-height: 1;
 	vertical-align: middle;
 	padding: 2px 2px 3px 3px;
-	margin: -1px 0 0 2px;
 	border-radius: 3px;
 	color: #fff;
 	background-color: #444;
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -1923,14 +1923,11 @@ body.download_page_edd-reports {
 .edd-chip {
 	font-size: 10px;
 	text-transform: uppercase;
-	font-weight: bold;
 	line-height: 1;
-	vertical-align: middle;
-	padding: 2px 2px 3px 3px;
+	padding: 3px;
 	border-radius: 3px;
 	color: #fff;
 	background-color: #444;
-	opacity: 0.8;
 }
 
 .edd-vertical-sections .edd-legacy-label {
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -1922,6 +1922,7 @@ body.download_page_edd-reports {
 
 .edd-chip {
 	font-size: 10px;
+	font-weight: bold;
 	text-transform: uppercase;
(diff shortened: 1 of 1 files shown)


assets/css/edd-admin.css
```diff
@@ -1921,7 +1921,7 @@ body.download_page_edd-reports {
 }
 
 .edd-chip {
-	font-size: 7px;
+	font-size: 10px;
 	text-transform: uppercase;
 	font-weight: bold;
 	line-height: 8px;
```
spencerfinnell, line 1927: Let's set this to `line-height: 1;` so it always matches the current font size, which will fix ensure it stays snug to the text but also stop any cascading line heights from applying.

assets/css/edd-admin.css
```diff
@@ -1921,17 +1921,16 @@ body.download_page_edd-reports {
...
 .edd-chip {
-	font-size: 7px;
+	font-size: 10px;
 	text-transform: uppercase;
 	font-weight: bold;
-	line-height: 8px;
+	line-height: 1;
 	vertical-align: middle;
 	padding: 2px 2px 3px 3px;
 	margin: -1px 0 0 2px;
```
spencerfinnell, line 1930: Looking at other uses of this component, I think we should remove the margin from the base and let anything implementing it to define its own margin.

The existing uses continue to look fine with the margin removed:

Screen Shot 2020-04-14 at 9 25 59 AM
Screen Shot 2020-04-14 at 9 26 19 AM

But could be tweaked some in a follow up issue.

cklosowski:
Here is a closer look, for details:
image

assets/css/edd-admin.css
```diff
@@ -1921,17 +1921,15 @@ body.download_page_edd-reports {
...
 
 .edd-chip {
-	font-size: 7px;
+	font-size: 10px;
 	text-transform: uppercase;
 	font-weight: bold;
-	line-height: 8px;
+	line-height: 1;
 	vertical-align: middle;
 	padding: 2px 2px 3px 3px;
```
robincornett, line 1929: Changing the padding to be consistent on all sides may resolve the issue @cklosowski mentioned; I think the `vertical-align` may also be contributing to the issue.

assets/css/edd-admin.css
```diff
@@ -1921,17 +1921,15 @@ body.download_page_edd-reports {
...
-	margin: -1px 0 0 2px;
 	border-radius: 3px;
 	color: #fff;
-	background-color: #bbb;
-	border: 1px solid #999;
+	background-color: #444;
 	opacity: 0.8;
```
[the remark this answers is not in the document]
spencerfinnell: @robincornett Just noticed the opacity when looking again this morning. Yes, let's remove that.

spencerfinnell:
@robincornett Can you branch off `issue/7731` and implement your suggested changes so we can compare?

robincornett:
Sure, try this: [issue/7731-1](https://github.com/easydigitaldownloads/easy-digital-downloads/tree/issue/7731-1)

image

If the user's browser zoom is set fairly high (150%, for example), the extra styling (absolute positioning) on the chip for the reports screen causes potential issues:

image

But changing that would impact the height of those tabs, as well--something to consider.

spencerfinnell:
@robincornett I like those changes, but I think it needs the bold. It's not appearing blurry for me. 

I'm personally not worried about the alignment in this particular context, as long as this component itself is consistent. We'll always get better alignment using `flexbox` but that shouldn't be a worry of this particular class.

assets/css/edd-admin.css
LisaCee: With `vertical-align` (current):
image

Without `vertical-align`: 
image

Thoughts @robincornett and @cklosowski?

spencerfinnell:
Merged in @robincornett's changes, minus removing `bold`.

assets/css/edd-admin.css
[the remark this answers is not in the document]
LisaCee: Pushed a commit removing opacity

robincornett:
> 
> 
> Merged in @robincornett's changes, minus removing `bold`.

Sorry @spencerfinnell I just pulled the latest of the original 7731 branch and only see that the opacity was removed, not the padding/vertical align updates?

LisaCee:
Sorry @robincornett, that was me.  I committed without seeing your your merge.

spencerfinnell:
Should be good now @robincornett @LisaCee

PR #7741, opened 2020-04-14
Deprecate the edd_tools_banned_emails_display function

### LINKED ISSUE #7740: 3.0 - deprecate edd_tools_banned_emails_display function
The `edd_tools_banned_emails_display` function should be deprecated in 3.0 as it's no longer used.

### Justification or use case
This functionality has been moved to the EDD settings page (under Moderation > Order Blocking) and the display is managed differently.

robincornett:
Fixes #7740


cklosowski:
@robincornett as I look at this now, I think we should create a new file in `includes/admin` named `admin-deprecated-functions.php` and add it to the load process.

By putting this function within the main `depreacted-functions.php` file, we've basically changed the scope of it's usage from an admin only function, to one that can be called from anywhere.

Issue #7739, opened 2020-04-14
3.0 - Orders - Remove Right Aligned Labels and Standardize Form Controls


spencerfinnell:
### Expected behavior

Labels should be able to be variable length.

### Actual behavior

If a label is too long it overflows incorrectly:

<img width="381" alt="Screen Shot 2020-04-14 at 10 10 26 AM" src="https://user-images.githubusercontent.com/491124/79234464-4ccd1b00-7e38-11ea-8269-2d27c2c32f99.png">

Other sections have a different fixed-width to account for longer strings, but will still break with translations:

<img width="721" alt="Screen Shot 2020-04-14 at 10 10 44 AM" src="https://user-images.githubusercontent.com/491124/79234453-48a0fd80-7e38-11ea-886b-17beba56feec.png">

### Steps to reproduce the behavior
1) Visit "Orders > Add New > Order Details"
2) Translate strings to something longer than they are now.

### Information (if a specific version is affected):

EDD Version (or branch): `release/3.0`

Any other relevant information:

Some items were previously fixed in: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7523

This would be a really good time to standardize this markup and CSS before we start spiraling out of control with different UI patterns for different areas.

Right now on the Orders screen we have:

```html
<div class="input-wrap">
	<label>Line 1:</label>
	<input type="text" name="edd_order_address[address]" value="">
</div>
```

```html
<div class="edd-order-payment-key edd-admin-box-inside">
	<span class="label">Key</span>
	<input type="text" name="payment_key" readonly="" value="">
</div>
```

```html
<p>
	<strong>What is your first name:</strong>
	<input type="text" name="edd-new-customer-first-name" value="" class="medium-text">
</p>
```

```html
<div class="column-container change-customer">
	<label for="customer_id" class="edd-order-details-label-mobile">Assign to an existing customer</label>
	<select name="customer-id" id="customer_id"></select>
</div>
```

All attempting to accomplish the same thing but needing specific CSS to do so.

We need a reusable and flexible way to setup EDD form controls that are not context-specific (nor screen-specific).

Related https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7701, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7737

---

[...]

PR #7744, opened 2020-04-16
Adding labels to all input fields, #7737

### LINKED ISSUE #7737: 3.0 - Orders - Inputs Missing Labels
### Expected behavior

When Adding or Viewing an Order, any `input`s should have a linked label.

### Actual behavior

Many inputs use a `span` instead of a true `label`.

1) Visit Orders > Add New
2) Click the "Key:" label.

Notice the input is not focused.

Changes should be made against [`try/edd3-orders`](https://github.com/easydigitaldownloads/easy-digital-downloads/tree/try/edd3-orders) which includes some layout changes.

LisaCee:
Fixes #7737

Proposed Changes:
1. Added labels to all input fields for accessibility

Note:
Chosen disables the ability to tab between `select` fields, though I still tried to label them as best as possible.


includes/admin/payments/orders.php
```diff
@@ -905,6 +905,7 @@ function edd_order_details_attributes( $order ) {
 					<?php
 					echo EDD()->html->select( array(
 						'name'             => 'edd-payment-time-min',
+						'id'               => 'edd-payment-time-min',
```
spencerfinnell, line 908: I wonder if we should change the current "Time" `label` to a `span` and then have a visually hidden `label` for "Hour" and "Minute" for these two selects. @robincornett?

includes/admin/payments/orders.php
```diff
@@ -351,27 +351,27 @@ function edd_order_details_addresses( $order ) {
...
 			<div class="input-wrap">
-				<label><?php echo esc_html_x( 'Zip / Postal Code:', 'Zip / Postal code of address', 'easy-digital-downloads' ); ?></label>
-				<input type="text" name="edd_order_address[postal_code]" value="<?php echo esc_attr( $address->postal_code ); ?>" class="med-text" />
+				<label for="edd_order_address[postal_code]"><?php echo esc_html_x( 'Zip / Postal Code:', 'Zip / Postal code of address', 'easy-digital-downloads' ); ?></label>
+				<input type="text" name="edd_order_address[postal_code]" id="edd_order_address[postal_code]" value="<?php echo esc_attr( $address->postal_code ); ?>" class="med-text" />
 			</div>
 
 			<div class="input-wrap">
-				<label><?php echo esc_html_x( 'Country:', 'Address country', 'easy-digital-downloads' ); ?></label>
+				<label for="edd-order-address-country"><?php echo esc_html_x( 'Country:', 'Address country', 'easy-digital-downloads' ); ?></label>
```
spencerfinnell, line 374: When using `EDD()->html->select()` to generate the fields, the `id` is automatically converted to [`snake_case`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-html-elements.php#L601). So Country and Region labels need to be updated accordingly.

includes/admin/payments/orders.php
LisaCee: > In the "Customer" tab of "Order Details" if you click "Create New Customer", these need to be switched to `label`:
> 
> Screen Shot 2020-04-16 at 4 08 32 PM
> 
> Let's add a visually hidden (`.screen-reader-text`) `label` to:
> 
> Screen Shot 2020-04-16 at 4 46 15 PM

I _think_ the directions for these are backwards.  I added `label`s to the "Customer" tab of "Order Details" -> "Create New Customer".  They lost their `strong` and some formatting.  Do you want me to deal with this now or after #7739?

image

LisaCee: And I added `.screen-reader-text` to `edd-payment-time-hour` and `edd-payment-time-min`, but I can't figure out where the `style="display: none"` is coming from.

spencerfinnell: @LisaCee The `style="display: none;"` on the actual `<select>` element is from the [Chosen](https://harvesthq.github.io/chosen/) script which renders an "enhanced" select box and hides the original. 

The hidden label changes look good, but I don't know if we should keep the `<label>` tag around "Time" now, since it duplicates the hidden "Hour" label now. I think it's safe to change that to `span`.

spencerfinnell:
> I think the directions for these are backwards. I added labels to the "Customer" tab of "Order Details" -> "Create New Customer". They lost their strong and some formatting. Do you want me to deal with this now or after #7739?

@LisaCee We can fix up styling as we implement the new component markup from #7739.

includes/admin/payments/orders.php
robincornett: @LisaCee the `display:none;` is being added by the `chosen` script to replace the standard dropdown.

Question for both of you (ping @spencerfinnell): how about we consider changing the hour and minute fields to `number` inputs, please? This seems like an ideal use for them because we can set a minimum and a maximum value and it's totally a value which can be changed incrementally.

Also, thinking through our discussions about radio inputs and grouped controls, I did some more digging and I think the hour/minute inputs would benefit from being "grouped" using a `<fieldset>` as they are interdependent.
- see especially the [SSN example](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA17) on the W3 site
- [address field](https://www.w3.org/WAI/WCAG21/Techniques/html/H71) as an example as well, so maybe we should consider doing the same with our address field groups
- consider using `aria-labelledby` to provide additional context for [fields in groups](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria)

spencerfinnell: @robincornett `number` does seem like a reasonable choice for these inputs. Can you open  a separate issue for that so we follow up with those adjustments at a later date?

includes/admin/payments/orders.php
```diff
@@ -761,14 +761,14 @@ function edd_order_details_extras( $order = false ) {
 				<?php endif; ?>
 
 				<div class="edd-order-payment-key edd-admin-box-inside">
-					<span class="label"><?php esc_html_e( 'Key', 'easy-digital-downloads' ); ?></span>
-					<input type="text" name="payment_key" <?php echo esc_attr( $readonly ); ?> value="<?php echo esc_attr( $order->payment_key ); ?>" />
+					<label for="payment_key" class="label"><?php esc_html_e( 'Key', 'easy-digital-downloads' ); ?></label>
+					<input type="text" name="payment_key" id="payment_key" <?php echo esc_attr( $readonly ); ?> value="<?php echo esc_attr( $order->payment_key ); ?>" />
```
ashleyfae, line 765: Prefix the `id` with at minimum `edd-`.

includes/admin/payments/orders.php
```diff
@@ -761,14 +761,14 @@ function edd_order_details_extras( $order = false ) {
...
 				</div>
 
 				<div class="edd-order-ip edd-admin-box-inside">
-					<span class="label"><?php esc_html_e( 'IP', 'easy-digital-downloads' ); ?></span>
+					<label for="ip" class="label"><?php esc_html_e( 'IP', 'easy-digital-downloads' ); ?></label>
 					<?php if ( edd_is_add_order_page() ) : ?>
-						<input type="text" name="ip" value="<?php echo esc_attr( edd_get_ip() ); ?>" />
+						<input type="text" name="ip" id="ip" value="<?php echo esc_attr( edd_get_ip() ); ?>" />
```
ashleyfae, line 771: Prefix the `id` with at minimum `edd-`.

includes/admin/payments/orders.php
```diff
@@ -783,8 +783,8 @@ function edd_order_details_extras( $order = false ) {
 
 				<?php if ( edd_is_add_order_page() ) : ?>
 					<div class="edd-order-tx-id edd-admin-box-inside">
-						<span class="label"><?php esc_html_e( 'Transaction ID', 'easy-digital-downloads' ); ?></span>
-						<input type="text" name="transaction_id" value="" />
+						<label for="transaction_id" class="label"><?php esc_html_e( 'Transaction ID', 'easy-digital-downloads' ); ?></label>
+						<input type="text" name="transaction_id" id="transaction_id" value="" />
```
ashleyfae, line 787: Prefix the `id` with at minimum `edd-`.

includes/admin/payments/orders.php
```diff
@@ -529,7 +529,7 @@ function edd_order_details_items( $order ) {
 			<div class="edd-add-download-to-purchase">
 				<ul>
 					<li class="download">
-						<span class="edd-order-details-label-mobile"><?php printf( esc_html_x( '%s To Add', 'order details select item to add - mobile', 'easy-digital-downloads' ), edd_get_label_singular() ); ?></span>
+						<label for="edd_order_add_download_select_chosen" class="edd-order-details-label-mobile"><?php printf( esc_html_x( '%s To Add', 'order details select item to add - mobile', 'easy-digital-downloads' ), edd_get_label_singular() ); ?></label>
```
robincornett, line 532: The select ID for this is actually `edd_order_add_download_select`, looking at the markup. It's weird because the ID being passed to the `product_dropdown` method is `edd-order-add-download-select`, but it looks like the method is overriding/replacing that. And then we have the `chosen` issue as well, but I think our label should match the `select` element we are actually creating.

includes/admin/payments/orders.php
```diff
@@ -889,6 +889,7 @@ function edd_order_details_attributes( $order ) {
 						<?php echo esc_html( __( 'Time', 'easy-digital-downloads' ) . ' (' . edd_get_timezone_abbr() . ')' ); ?>
 					</label>
 
+					<label for="edd_payment_time_hour_chosen" class="screen-reader-text">Hour</label>
```
robincornett, line 892: This has the same issue as the add download select--`chosen` should not be included in the label, as it's not the ID of the actual `select` element. Looks like this `for` should be `edd_payment_time_hour`, as the `id` parameter being passed to the `EDD()->html->select()` method is being disregarded.

LisaCee:
> In general, I would like to see consistency with the id/for attributes we are adding. Currently we have a mix of underscores, dashes, and brackets being used. Any element using the `EDD()->html` methods is likely to have issues when using `[]`, so I would stick with either `_` or `-`, but don't have a preference which. Just since we're thinking through everything, let's be consistent.
> 
> Based on the behavior of the `EDD()->html->select`, from what I can tell, we may need to use `_` as the select method intentionally replaces `-` with `_`.

Any issues with me changing `id`s to all use `_` rather than `-` and then change all of the corresponding `label for=` tags?

spencerfinnell:
> Any issues with me changing ids to all use _ rather than - and then change all of the corresponding label for= tags?

@LisaCee For changes you've introduced, go with `_`, but let's keep items that have an existing ID the same.

includes/admin/payments/orders.php
```diff
@@ -351,27 +351,27 @@ function edd_order_details_addresses( $order ) {
...
-				<input type="text" name="edd_order_address[address]" value="<?php echo esc_attr( $address->address ); ?>" />
+				<label for="edd_order_address[address]"><?php esc_html_e( 'Line 1:', 'easy-digital-downloads' ); ?></label>
+				<input type="text" name="edd_order_address[address]" id="edd_order_address[address]" value="<?php echo esc_attr( $address->address ); ?>" />
 			</div>
 
 			<div class="input-wrap">
-				<label><?php esc_html_e( 'Line 2:', 'easy-digital-downloads' ); ?></label>
-				<input type="text" name="edd_order_address[address2]" value="<?php echo esc_attr( $address->address2 ); ?>" />
+				<label for="edd_order_address[address2]"><?php esc_html_e( 'Line 2:', 'easy-digital-downloads' ); ?></label>
+				<input type="text" name="edd_order_address[address2]" id="edd_order_address[address2]" value="<?php echo esc_attr( $address->address2 ); ?>" />
```
ashleyfae, line 360: I'd like to avoid using `[` in IDs if we can. Keep it for the name because we do need it there, but change the ID to just `edd_order_addresses_address2` (and similar for all others affected).

PR #7756, opened 2020-04-20
Exclude taxes with SUM(total - tax) instead of SUM(subtotal) #7692

### LINKED ISSUE #7692: 3.0 - Reports - "Exclude Taxes" also incorrectly excludes discounts
### Expected behavior

When I check on "Exclude Taxes", I expect the amounts to reflect the total price with only taxes subtracted.

### Actual behavior

When I check on "Exclude Taxes", the amounts reflect the pre-tax **and pre-discount** price.

Notice how the earnings actually _increase_ when I exclude taxes:

Screenshot from 2020-03-30 13-49-13

Screenshot from 2020-03-30 13-49-24

When unchecked, the query does `SUM(total)` . When checked, the query does `SUM(subtotal)`. That would be okay if the subtotal was just the total without taxes. However, the subtotal does not include discounts, so when you do have a discount, your actual earnings are reduced and that isn't considered.

Unless we want to make some major changes around what `subtotal` is, we can correct this by doing `SUM(total - tax)`

1) Ensure you have at least one order with a discount, and at least one order with tax.
2) Go to Downloads > Reports and look at the Overview.
3) Toggle Exclude Taxes and notice the differences.

PHP Version:

WordPress Version:

Related: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7644, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7621 ("what is subtotal" discussion)

ashleyfae:
Fixes #7692

Proposed Changes:
1. Updates all reports to exclude taxes by querying `SUM(total - tax)` instead of `SUM(subtotal)`.

### DIFF
--- a/includes/reports/data/customers/class-top-five-customers-list-table.php
+++ b/includes/reports/data/customers/class-top-five-customers-list-table.php
@@ -69,7 +69,7 @@ class Top_Five_Customers_List_Table extends \EDD_Customer_Reports_Table {
 
 			// @todo DRY with Most_Valuable_Customers_List_Table
 
-			$column = Reports\get_taxes_excluded_filter() ? 'total = tax' : 'total';
+			$column = Reports\get_taxes_excluded_filter() ? 'total - tax' : 'total';
 
 			$sql = "SELECT customer_id, COUNT(id) AS order_count, SUM({$column}) AS total_spent
 					FROM {$wpdb->edd_orders}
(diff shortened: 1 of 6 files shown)


includes/reports/data/customers/class-top-five-customers-list-table.php
```diff
@@ -69,7 +69,7 @@ public function get_data() {
 
 			// @todo DRY with Most_Valuable_Customers_List_Table
 
-			$column = Reports\get_taxes_excluded_filter() ? 'subtotal' : 'total';
+			$column = Reports\get_taxes_excluded_filter() ? 'total = tax' : 'total';
```
spencerfinnell, line 72:
```suggestion
			$column = Reports\get_taxes_excluded_filter() ? 'total - tax' : 'total';
```

PR #7760, opened 2020-04-20
Stop saving payment_meta top order meta #7289

### LINKED ISSUE #7289: _edd_payment_meta backwards compatibility.
As we've moved to using order items, and order item meta, we no longer need to populate the `_edd_payment_meta` array, however we should still back-port data to it, for a period of backwards compatibility, including the order item options array we moved to meta in #7197 .

ashleyfae:
Fixes #7289

Proposed Changes:
1. No longer saves `payment_meta` (old key `_edd_payment_meta`). Meta is still dynamically retrieved, just not saved.

Do not merge until https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7555 has been merged. We'll want to do final testing after that PR has been merged in.


ashleyfae:
> Should this PR do anything with migrated data? Testing and migrated `_edd_payment_meta` remains in the `postmeta` table. If this is just for new orders, then disregard please.
> 
> Looking at a 3.0 site with orders placed before I switched to this branch, there are no `_edd_payment_meta` rows in the `postmeta` table, so is the PR removing code which no longer does anything?

Yes this is just to stop it getting added for new orders.

robincornett (approved):

PR #7761, opened 2020-04-20
Update discount amount to use user-defined currency symbol

### LINKED ISSUE #7757: 3.0 - Add/Edit discount flat rate hard-codes USD symbol despite currency settings
### Expected behavior

When I select a currency that is not USD, I expect all references to amounts to reflect my currency settings. I shouldn't see `$` to indicate currency. If I've selected GBP as  my currency then I should see `£` instead.

### Actual behavior

When I add a discount and want to select a flat rate, I see the `$` symbol to indicate a flat amount instead of `£`.

Screenshot from 2020-04-20 13-31-24

Screenshot from 2020-04-20 13-31-53

1) Go to Downloads > Settings > Currency and set your currency to anything that is not USD.
2) Go to Downloads > Discounts and add a discount.
3) For the `Amount` column, click the dropdown to change from `%` to a flat rate. You'll see a hard-coded `$` symbol.

PHP Version:

WordPress Version:

robincornett:
Fixes #7757 

Proposed Changes:
1. Changes hard coded `$` in discount type select to use the user-defined currency symbol
2. Changes `%` to no longer be a translatable string

### DIFF
--- a/includes/admin/discounts/add-discount.php
+++ b/includes/admin/discounts/add-discount.php
@@ -71,7 +71,7 @@ $minutes      = edd_get_minute_values();
 							<label for="edd-amount-type" class="screen-reader-text"><?php esc_html_e( 'Amount Type', 'easy-digital-downloads' ); ?></label>
 							<select name="amount_type" id="edd-amount-type">
 								<option value="percent">%</option>
-								<option value="flat"><?php echo esc_html( edd_currency_filter( '' ) ); ?></option>
+								<option value="flat"><?php echo esc_html( edd_currency_symbol() ); ?></option>
 							</select>
 						</span>
 						<p class="description"><?php esc_html_e( 'The amount as a percentage or flat rate. Cannot be left blank.', 'easy-digital-downloads' ); ?></p>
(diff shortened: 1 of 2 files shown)


includes/admin/discounts/add-discount.php
```diff
@@ -70,8 +70,8 @@
 							<input type="text" required="required" class="edd-price-field" id="edd-amount" name="amount" value="" placeholder="<?php esc_html_e( '10.00', 'easy-digital-downloads' ); ?>"/>
 							<label for="edd-amount-type" class="screen-reader-text"><?php esc_html_e( 'Amount Type', 'easy-digital-downloads' ); ?></label>
 							<select name="amount_type" id="edd-amount-type">
-								<option value="percent"><?php esc_html_e( '%', 'easy-digital-downloads' ); ?></option>
-								<option value="flat"><?php esc_html_e( '$', 'easy-digital-downloads' ); ?></option>
+								<option value="percent">%</option>
+								<option value="flat"><?php echo esc_html( edd_currency_filter( '' ) ); ?></option>
```
ashleyfae, line 74: What about using `edd_currency_symbol()` instead of `edd_currency_filter()` ?

I understand `edd_currency_filter()` does actually do exactly what we want here, but the original intention is for formatting a final amount with a price, even if it supports an empty string as a price.

spencerfinnell: `edd_currency_filter()` is used so much in the codebase I'm not even sure I knew `edd_currency_symbol()` existed. 😬

`edd_currency_symbol()` makes sense!

robincornett: I'll swap it--I looked in the `metabox.php` file to see how the currency symbol was added there and went off of that.

PR #7772, opened 2020-04-21
Allow 'complete' order items to be downloaded if payment is partially refunded #7758

### LINKED ISSUE #7758: 3.0 - Partially refunding an order removes ability to download all products from order
### Expected behavior

I have an order with two products: "WordPress Theme" and "WordPress Plugin". If I refund just "WordPress Theme" then I expect the customer is still able to download "WordPress Plugin" because this item was not refunded.

### Actual behavior

If one item in the order is refunded, then the download links for the **entire order** are removed.

Here are some screenshots:

**After purchase (before refund)**

Screenshot from 2020-04-20 13-49-01

**Processed partial refund**

Screenshot from 2020-04-20 13-49-58

**Customer's view after partial refund**

Screenshot from 2020-04-20 13-50-20

Screenshot from 2020-04-20 13-53-29

1) Complete an order that contains at least two different products.
2) Refund just one of the line items.
3) Refresh your customer account page with the `[download_history]` and/or `[purchase_history]` shortcodes. Download links are removed.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #7758

Proposed Changes:
1. Download `has_access` check now looks for a complete **order item** matching all the parameters (order ID + download ID + price ID) instead of a complete **order**.
2. Template conditionals looking at order statuses have been updated.
3. Refactored `history-downloads.php` to use new order functions. This allows us to better check order item status. All actions are backwards compatible with this change.
4. Introduced new `edd_order_grants_access_to_download_files()` function. This accepts an order ID, product ID, and optionally a price ID. With that information it determines whether or not access to the download files should be granted.
5. Introduced new unit tests on new `edd_order_grants_access_to_download_files()` function.

TK:

2. ~~Can an order item be `partially_refunded` ? I need to look into this more and see what happens if you buy 2 of the same item (quantities enabled) but 1 gets refunded. :shrug:~~ At this time it cannot be, so I'm not going to worry about it. Something to consider for the future.


cklosowski:
This seems like a pretty good area where we might want to get some unit tests in place, if at all possible. I know we cannot test the entire file download process but maybe this process download URL method can be tested if we ‘mock’ the data properly?

ashleyfae:
I'll see what I can do.

ashleyfae:
Introduced a new function `edd_order_grants_access_to_download_files()`. It accepts an order ID, product ID, and optionally a price ID. Given that information it determines whether or not access to the download files should be granted. I basically just moved the new logic I had put in `edd_process_signed_download_url()` into its own function so we could add unit tests just for that portion.

ashleyfae:
Thanks for testing @robincornett ! I've pushed up a change that displays `Refunded` next to the item if it was refunded.

PR #7775, opened 2020-04-21
Add special handling for the 'number' setting type #7774

### LINKED ISSUE #7774: 3.0 - Cannot set 0 as the Refund Window. Always defaults back to 30.
### Expected behavior
When trying to set an infinite refund window with the value of `0`, after saving the settings the value of `0` should remain.

### Actual behavior
When trying to set the refund window to `0`, it is always reset back to `30`, which is the default value.

1) Go to Downloads > Settings > Refunds
2) Set the value of `Refund Window` to `0`.
3) Click 'Save'

PHP Version: All

WordPress Version: All

cklosowski:
Fixes #7774

Proposed Changes:
1. Adds `number` case to the `switch` in `edd_settings_sanitize`.
2. Accounts for the `number` type in the `edd_get_option` call.
3. Adds a new function called  `edd_get_registered_setting_details` allowing getting the details of a registered setting, when knowing it's section and tab.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -24,9 +24,15 @@ defined( 'ABSPATH' ) || exit;
 function edd_get_option( $key = '', $default = false ) {
 	global $edd_options;
 
-	$value = ! empty( $edd_options[ $key ] )
-		? $edd_options[ $key ]
-		: $default;
+	$value = $default;
+
+	if ( isset( $edd_options[ $key ] ) ) {
+		if ( is_numeric( $edd_options[ $key ] ) ) {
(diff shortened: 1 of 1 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -1259,6 +1265,23 @@ function edd_settings_sanitize( $input = array() ) {
 						unset( $output[ $key ] );
 					}
 					break;
+				case 'number':
+					echo '<pre>';
+
```
spencerfinnell, line 1270:
```suggestion
```

cklosowski:
Fixed @spencerfinnell

PR #7776, opened 2020-04-22
3.0 - Orders - Allow Refund Overrides

spencerfinnell:
See https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7599#issuecomment-617444773

Proposed Changes:
1. Introduce `edd_is_order_refundable_by_override()` to allow Orders to be refunded in special circumstances.


cklosowski:
@spencerfinnell I think I like this but can we get one more 'override' filter that allows a specific filter for the order ID?

While this will work from WP Admin efforts, I'm wondering if a webhook call from a gateway would pose an issue as we don't have a 'capability' on a webhook call. So adding something like:
`$can_override = apply_filters( 'edd_is_order_id_' . $order_id . '_refundable_by_override', $can_override, $order_id );`

That may be overkill, but it would allow us to in run time, filter with something like:
`add_filter( 'edd_is_order_id_52_refundable_by_override', '__return_true' );`

That way a gateway running a webhook can just run a single filter, and then never have to 'unhook' that same filter inline.

spencerfinnell:
@cklosowski Alongside this one? https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7776/files#diff-d29d40f97d7b4cddeb7eb7727129a005R122-R134

I think WP core is trying to move away from dynamic actions, should we? It's a simpler one-liner to enable if it's order-specific, but it's also possible to create a filter that only affects the specific order with the global one, which wouldn't need to be unhooked.

cklosowski:
Ok, if that is the direction Core is going, we can go with it. Thanks @spencerfinnell

PR #7777, opened 2020-04-22
3.0 Form / Label Groups

### LINKED ISSUE #7739: 3.0 - Orders - Remove Right Aligned Labels and Standardize Form Controls

LisaCee:
Fixes #7739 

Proposed Changes:
1. Grouped all label / input fields into `edd-form-group` divs
2. Broke inputs into `edd-form-group__control` divs
3. Corrected some indentation according to phpcs

See discussion on formatting of groups in #7739 

I will be going through classes and CSS next.  I just wanted to start getting input on this phase.


assets/css/edd-admin.css
```diff
@@ -3706,3 +3685,78 @@ div:not(#edd-item-tab-wrapper) + #edd-item-card-wrapper .item-section { /** [1]
...
 }
+
+/**
+ * Forms
+ * ---------------------------------------------------------- */
+
+/**
+ * Form Group
+ *
+ * <div class="edd-form-group">
```
ashleyfae, line 3696: Huh. I think I really like having the HTML in here as a reference. I never thought to do that!

spencerfinnell:
This is looking good so far!

> I will be going through classes and CSS next.

When auditing any classes applied to labels, ensure nothing is applying `bold` to the `labels` anywhere. There are some inconsistent uses of bolded labels left over from the mis-matched markup and CSS.

Things like `.edd-order-details-label-mobile` should be able to be safely removed.

assets/css/edd-admin.css
LisaCee: I think that was @spencerfinnell s doing.

includes/admin/payments/orders.php
```diff
@@ -602,59 +668,82 @@ function edd_order_details_adjustments( $order ) {
...
+						);  // WPCS: XSS ok.
 						?>
+					</div>
 					</li>
 				</ul>
-
+			</div>
+			<div class="edd-form-group">
 				<ul>
+				<div class="edd-form-group">
```
robincornett, line 734: Is this duplication of line 732 intentional?

Also, I know the `ul/li` are inherited from previous code, but am not sure they are necessary. The `credit` class and inline style are likely necessary for the JS, but seems like the two text input fields could be logically grouped another way than a list with one item, maybe a `fieldset` along the lines of the potential address grouping?

includes/admin/payments/orders.php
```diff
@@ -602,59 +668,82 @@ function edd_order_details_adjustments( $order ) {
...
-						<input type="text" class="edd-add-order-credit-description" value="" placeholder="<?php echo esc_attr( 'Description', 'easy-digital-downloads' ); ?>" />
-						<input type="text" class="edd-add-order-credit-amount" value="" min="1" placeholder="<?php echo esc_attr( 'Amount', 'easy-digital-downloads' ); ?>" />
+						<label for="edd_add_order_credit_description" class="screen-reader-text"><?php echo esc_html_e( 'Description', 'easy-digital-downloads' ); ?></label>
+					<div class="edd-form-group__control">
+						<input type="text" class="edd-add-order-credit-description" value="" id="edd_add_order_credit_description" placeholder="<?php echo esc_attr( 'Description', 'easy-digital-downloads' ); ?>" />
+				</div>
+				<div class="edd-form-group">
+						<label for="edd_add_order_credit_amount" class="screen-reader-text"><?php echo esc_html_e( 'Amount', 'easy-digital-downloads' ); ?></label>
+					<div class="edd-form-group__control">
+						<input type="text" class="edd-add-order-credit-amount" value="" id="edd_add_order_credit_amount" min="1" placeholder="<?php echo esc_attr( 'Amount', 'easy-digital-downloads' ); ?>" />
```
robincornett, line 743: Should this be a number input field instead of text? Since there is text I am assuming/hoping there is numerical validation going on during sanitization, but this should be a number, shouldn't it? (I know this was inherited.)

includes/admin/payments/orders.php
```diff
@@ -761,18 +851,26 @@ function edd_order_details_extras( $order = false ) {
 					</div>
 				<?php endif; ?>
 
-				<div class="edd-order-payment-key edd-admin-box-inside">
-					<label for="edd_payment_key" class="label"><?php esc_html_e( 'Key', 'easy-digital-downloads' ); ?></label>
-					<input type="text" name="payment_key" id="edd_payment_key" <?php echo esc_attr( $readonly ); ?> value="<?php echo esc_attr( $order->payment_key ); ?>" />
+				<div class="edd-admin-box-inside">
+					<div class="edd-form-group">
+						<label for="edd_payment_key" class="edd-form-group__label label"><?php esc_html_e( 'Key', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 856: The `label` class can be removed because of the new `__label` class.

includes/admin/payments/orders.php
```diff
@@ -761,18 +851,26 @@ function edd_order_details_extras( $order = false ) {
...
 
 				<div class="edd-order-ip edd-admin-box-inside">
-					<label for="edd_ip" class="label"><?php esc_html_e( 'IP', 'easy-digital-downloads' ); ?></label>
-					<?php if ( edd_is_add_order_page() ) : ?>
-						<input type="text" name="ip" id="edd_ip" value="<?php echo esc_attr( edd_get_ip() ); ?>" />
-					<?php else : ?>
-						<span><?php echo edd_payment_get_ip_address_url( $order->id ); // WPCS: XSS ok. ?></span>
-					<?php endif; ?>
+					<div class="edd-form-group">
+						<label for="edd_ip" class="edd-form-group__label label"><?php esc_html_e( 'IP', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 865: `label` class can also be removed here (there are more but I'll stop repeating myself).

includes/admin/payments/orders.php
```diff
@@ -761,18 +851,26 @@ function edd_order_details_extras( $order = false ) {
...
+					<div class="edd-form-group__control">
+						<?php if ( edd_is_add_order_page() ) : ?>
+							<input type="text" name="ip" id="edd_ip" class="edd-form-group__input" value="<?php echo esc_attr( edd_get_ip() ); ?>" />
```
robincornett, line 868: If an order is being added, should this input maybe be `readonly`?

includes/admin/payments/orders.php
```diff
@@ -784,26 +882,39 @@ function edd_order_details_extras( $order = false ) {
...
-					<label for="edd_unlimited_downloads" class="label label--has-tip label--has-checkbox">
-						<input type="checkbox" name="edd-unlimited-downloads" id="edd_unlimited_downloads" value="1"<?php checked( true, $unlimited, true ); ?>/>
-						<?php esc_html_e( 'Unlimited Downloads', 'easy-digital-downloads' ); ?>
-						<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php _e( '<strong>Unlimited Downloads</strong>: checking this box will override all other file download limits for this purchase, granting the customer unliimited downloads of all files included on the purchase.', 'easy-digital-downloads' ); ?>"></span>
-					</label>
+					<div class="edd-form-group">
+						<div class="edd-form-group__control">
+							<input type="checkbox" name="edd-unlimited-downloads" id="edd_unlimited_downloads" class="edd-form-group__input" value="1"<?php checked( true, $unlimited, true ); ?>/>
+
+							<label for="edd_unlimited_downloads">
```
robincornett, line 899: Should this label and the one for the send receipt checkbox have the `__label` class added? I realize that class has the display set to `block`, but the CSS could be updated to something like this for consistency:

```css
input + label.edd-form-group__label {
    display: inline;
}
```

includes/admin/payments/orders.php
```diff
@@ -881,42 +996,55 @@ function edd_order_details_attributes( $order ) {
 					<label for="edd-payment-date" class="label">
 						<?php esc_html_e( 'Date', 'easy-digital-downloads' ); ?>
 					</label>
-					<input type="text" id="edd-payment-date" name="edd-payment-date" value="<?php echo esc_attr( $order_date->format( 'Y-m-d' ) ); ?>" class="medium-text edd_datepicker" placeholder="<?php echo esc_attr( edd_get_date_picker_format() ); ?>"/>
+					<input type="text" id="edd-payment-date" name="edd-payment-date" value="<?php echo esc_attr( $order_date->format( 'Y-m-d' ) ); ?>" class="edd_datepicker" placeholder="<?php echo esc_attr( edd_get_date_picker_format() ); ?>"/>
```
robincornett, line 999: This is not using the new classes (same for related label).

includes/admin/payments/orders.php
```diff
@@ -881,42 +996,55 @@ function edd_order_details_attributes( $order ) {
...
-						'show_option_all'  => false,
-					) ); // WPCS: XSS ok.
-					?>
+					<fieldset class="edd-form-group">
+						<legend class="edd-form-group__label">
+							<?php echo esc_html( __( 'Time', 'easy-digital-downloads' ) . ' (' . edd_get_timezone_abbr() . ')' ); ?>
+						</legend>
+
+						<div class="edd-form-group__control">
+							<label for="edd-payment-time-hour" class="screen-reader-text">
```
robincornett, line 1010: Because the `EDD()->html->select()` method converts IDs to use underscores instead of dashes, the label `for` should be `edd_payment_time_hour`, I think. Same for the minutes label. The ID for the hour doesn't _have_ to change but could, just so it's consistent with this label.

includes/admin/payments/orders.php
```diff
@@ -881,42 +996,55 @@ function edd_order_details_attributes( $order ) {
...
+							?>
+							:
+
+							<label for="edd-payment-time-min" class="screen-reader-text">
+								<?php esc_html_e( 'Minute', 'easy-digital-downloads' ); ?>
+							</label>
+							<?php
+							echo EDD()->html->select(
+								array(
+									'name'             => 'edd-payment-time-min',
```
robincornett, line 1036: An `id` parameter needs to be added here--the `select` method will not automatically populate it.

includes/admin/payments/orders.php
LisaCee: This makes sense as a number, since it's already got the min attribute.

includes/admin/payments/orders.php
```diff
@@ -184,17 +184,24 @@ function edd_order_details_customer( $order ) {
...
-				) ); // WPCS: XSS ok.
-				?>
+				<div class="edd-form-group">
+					<label for="customer_id" class="edd-form-group__label edd-order-details-label-mobile"><?php esc_html_e( 'Assign to an existing customer', 'easy-digital-downloads' ); ?></label>
+					<div class="edd-form-group__control">
+						<?php
+						echo EDD()->html->customer_dropdown(
+							array(
+								'class'         => 'edd-payment-change-customer-input edd-form-group__input',
+								'selected'      => $customer_id,
```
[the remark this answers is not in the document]
spencerfinnell: @robincornett It _should_ be sanitized further down in `customer_dropdown()`, and it wouldn't hurt to do it here as well, but I'd like to keep the changes in this PR specific to the new markup.

includes/admin/payments/orders.php
```diff
@@ -276,42 +298,60 @@ function edd_order_details_email( $order ) {
...
+
 					<label for="<?php echo rawurlencode( sanitize_email( $email ) ); ?>">
-						<input autocomplete="off" id="<?php echo rawurlencode( sanitize_email( $email ) ); ?>" class="edd-order-resend-receipt-email" name="edd-order-resend-receipt-address" type="radio" value="<?php echo rawurlencode( sanitize_email( $email ) ); ?>" <?php checked( true, ( 'primary' === $key ) ); ?> />
 						<?php echo esc_attr( $email ); ?>
 					</label>
-				</p>
+				</div>
 				<?php endforeach; ?>
+
+				<p class="edd-form-group__help">
```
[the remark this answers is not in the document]
spencerfinnell: @robincornett I think it's fine to add `.description`, but I'd like to keep the styles on `.edd-form-group__help` since core includes `p` or `span` specificity as well as alters it based on additional contexts: https://github.com/WordPress/WordPress/blob/fcff974c33f9bf2419d0d9f22ea7df8004e1122c/wp-admin/css/forms.css#L1495

includes/admin/payments/orders.php
spencerfinnell: @robincornett My thinking is that `.edd-form-group__label` applies _only_ to the item being used as the label for the control itself. `label`s inside of `edd-form-group__control` are not the label for the form group, they are just arbitrary markup inside of the control.

assets/css/edd-admin.css
```diff
@@ -3706,3 +3685,78 @@ div:not(#edd-item-tab-wrapper) + #edd-item-card-wrapper .item-section { /** [1]
...
+	margin-bottom: 12px;
+}
+
+.edd-form-group:last-of-type {
+	margin-bottom: 0;
+}
+
+.edd-form-group__label {
+	display: block;
+	margin-bottom: 4px;
```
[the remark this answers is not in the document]
spencerfinnell: Looking through some Gutenberg/block editor styles they have both `4px` and `8px` instances. Chatter around the "G2" UI updates for Gutenberg are to add more breathing room around a lot of things, so I'm fine moving to `8px` for now.

Since we can now easily update it in the future, why not experiment?! 😄

includes/admin/payments/orders.php
LisaCee: This one was intentional, though I'm open to suggestions.  I wasn't sure how to group these since it's technically only one `li` with two input fields and it seemed wrong to split up the `ul` into two `form-group`s.  

I'd love to hear @spencerfinnell s thoughts on changing this from a `ul/li` to ... something else.

spencerfinnell: Probably separate `.edd-form-group` inside each `li`, but since this will be replaced with the new Order Item management UI soon, we don't need to worry about it.

assets/css/edd-admin.css
```diff
@@ -1797,10 +1776,6 @@ input[class*="edd-price-field"] {
 .edd_repeatable_product_wrapper td {
 	overflow: visible;
 }
-.edd-order-details-label-mobile {
```
spencerfinnell, line 1800: There are a couple more instances of this class elsewhere  in this file that are no longer needed.

includes/admin/payments/orders.php
```diff
@@ -761,18 +857,26 @@ function edd_order_details_extras( $order = false ) {
...
-					<?php if ( edd_is_add_order_page() ) : ?>
-						<input type="text" name="ip" id="edd_ip" value="<?php echo esc_attr( edd_get_ip() ); ?>" />
-					<?php else : ?>
-						<span><?php echo edd_payment_get_ip_address_url( $order->id ); // WPCS: XSS ok. ?></span>
-					<?php endif; ?>
+					<div class="edd-form-group">
+						<label for="edd_ip" class="edd-form-group__label"><?php esc_html_e( 'IP', 'easy-digital-downloads' ); ?></label>
+					<div class="edd-form-group__control">
+						<?php if ( edd_is_add_order_page() ) : ?>
+							<input type="readonly" name="ip" id="edd_ip" class="edd-form-group__input" value="<?php echo esc_attr( edd_get_ip() ); ?>" />
```
spencerfinnell, line 874: `type` is still `text`. `readonly` is an attribute in the same way that `type` is, so they exist alongside each other.

https://www.w3schools.com/tags/att_input_readonly.asp

It is a boolean attribute, which means it doesn't need a value (where `text` is the value of the `type` attribte)

includes/admin/payments/orders.php
```diff
@@ -749,10 +841,14 @@ function edd_order_details_extras( $order = false ) {
...
-						<span class="label"><?php esc_html_e( 'Gateway', 'easy-digital-downloads' ); ?></span>
-						<span class="value"><?php echo edd_get_gateway_admin_label( $order->gateway ); ?></span>
-					</div>
+						<div class="edd-order-gateway edd-admin-box-inside edd-admin-box-inside--row">
+							<div class="edd-form-group">
+								<label for="edd_gateway" class="edd-form-group__label"><?php esc_html_e( 'Gateway', 'easy-digital-downloads' ); ?></label>
+								<div class="edd-form-group__control">
+									<input class="edd-form-group__input" id="edd_gateway"><?php echo edd_get_gateway_admin_label( $order->gateway ); ?>/>
+								</div>
+							</div>
```
spencerfinnell: The `edd-form-group` needs to be around the `<?php echo $gateway_select; ?>` and `label` below. The markup here is for viewing an existing Order, so the Gateway shouldn't be editable.

When viewing an Order let's make "Gateway" and "IP" appear the same as the rows in the "Order Amounts" box.

includes/admin/payments/orders.php
```diff
@@ -747,32 +839,43 @@ function edd_order_details_extras( $order = false ) {
...
+				<?php if ( ! edd_is_add_order_page() ) : ?>
 					<div class="edd-order-gateway edd-admin-box-inside edd-admin-box-inside--row">
-						<span class="label"><?php esc_html_e( 'Gateway', 'easy-digital-downloads' ); ?></span>
-						<span class="value"><?php echo edd_get_gateway_admin_label( $order->gateway ); ?></span>
+						<div class="edd-form-group">
+							<label for="edd_gateway" class="edd-form-group__label"><?php esc_html_e( 'Gateway', 'easy-digital-downloads' ); ?></label>
+							<div class="edd-form-group__control">
+								<input class="edd-form-group__input" id="edd_gateway"><?php echo edd_get_gateway_admin_label( $order->gateway ); ?>/>
+							</div>
+						</div>
```
spencerfinnell, line 850: The `edd-form-group` needs to be around the `<?php echo $gateway_select; ?>` and `label` below. The markup here is for viewing an existing Order, so the Gateway shouldn't be editable.

When viewing an Order let's make "Gateway" and "IP" appear the same as the rows in the "Order Amounts" box.

LisaCee: So, 845-850 should go back to being `span`s?

assets/css/edd-admin.css
LisaCee: I think I got them all!

includes/admin/payments/orders.php
```diff
@@ -276,42 +298,60 @@ function edd_order_details_email( $order ) {
...
-			'purchase_id' => $order->id,
-		) ) ); ?>" id="<?php if ( ! empty( $customer->emails ) && count( (array) $customer->emails ) > 1 ) {
-			echo esc_attr( 'edd-select-receipt-email' );
-		} else {
-			echo esc_attr( 'edd-resend-receipt' );
-		} ?>" class="button-secondary"><?php esc_html_e( 'Resend Receipt', 'easy-digital-downloads' ); ?></a>
+		<?php endif; ?>
+
+		<p>
+			<a href="<?php echo esc_url( add_query_arg( array(
```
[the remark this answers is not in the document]
spencerfinnell: @robincornett I think once #7728 is merged we can retroactively add enhancements to URLs/redirects such as this.

cklosowski:
Fixing up teh merge conflict on this quickly

assets/css/edd-admin.css
```diff
@@ -3981,3 +3939,78 @@ div:not(#edd-item-tab-wrapper) + #edd-item-card-wrapper .item-section { /** [1]
...
+ *
+ *   <div class="edd-form-group__control">
+ *   	<input type="radio" class="edd-form-group__input" />
+ *   	<label>Radio 2</label>
+ *   </div>
+ * </fieldset>
+ *
+ */
+.edd-form-group {
+	margin-bottom: 12px;
```
spencerfinnell, line 3975: With the `8px` under the labels, I think this needs to be increased to `16px` or `18px`. Thoughts @LisaCee @robincornett?

robincornett: I'm always a fan of white space. Could we also consider removing the `:last-of-type` modifier on it, at least to experiment? I'm seeing that applied more places than I expect.

includes/admin/payments/orders.php
```diff
@@ -742,32 +794,55 @@ function edd_order_details_extras( $order = false ) {
 			<div class="edd-admin-box">
 				<?php do_action( 'edd_view_order_details_payment_meta_before', $order->id ); ?>
 
-				<?php if ( ! edd_is_add_order_page() && ! empty( $order->id ) ) : ?>
-					<?php if ( ! empty( $order->gateway ) ) : ?>
+
+				<?php if ( ! edd_is_add_order_page() ) : ?>
```
robincornett, line 798: I'm comparing the behavior of the gateway and IP address fields on the existing orders and they're currently different, but I think they should be the same. However, I'm not sure of the best way to make them the same.
- the gateway field is no longer an input group, etc. at all, and the `<label>` has been removed
- the IP field still has a `<label>` element, but it's orphaned because an `<input>` is not added if it's an existing order

I think the gateway method makes more sense--partly because the IP address includes a link, so it can't be in a text input. Could we maybe define markup for a situation like this, and keep them in an `edd-form-group` container still, but then have inner `<span>` elements (or `<div>`, if we're adding CSS anyway) with `edd-` classes? Y'all will know better than I, but something like:

```html
<div class="edd-form-group">
	<span class="edd-form-group__label--readonly">Gateway</span>
	<span class="edd-form-group__input--readonly">Test Payment</span>
</div>
```
@spencerfinnell @LisaCee

assets/css/edd-admin.css
spencerfinnell: @robincornett Where is it causing issues?

includes/admin/payments/orders.php
spencerfinnell: @robincornett What about making a separate component that utilizes `dl` (and `dt` and `dd`) for these key + value read-only outputs? I'm not sure the `form` nomenclature is necessary for plain text output. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

includes/admin/payments/orders.php
LisaCee: Sorry for the delay.  I would have sworn that I did this one yesterday.

assets/css/edd-admin.css
LisaCee: With the way the `form-group`s are organized, I think most of them fall into the `:last-of-type` category.

LisaCee: Everything here is `:last-of-type`, with `margin-bottom:0` :
image

vs without `:last-of-type` and current `.edd-form-group {margin-bottom: 12px}`:

image

spencerfinnell: This is what I'm seeing as well -- removing `:last-of-type` will add a lot of extra space.

robincornett: Yes, thanks @LisaCee, that is primarily where I was feeling it.

Really, the only place right now where I could see it actually making a helpful difference to have the `:last-of-type` modifier is at the end of the address block field, but it actually does nothing there because the section content has this rule:

```css
.edd-sections-wrap .section-wrap .section-content > * {
    margin: 20px;
}
```
So removing the margin from the last-of-type wouldn't even affect that.

includes/admin/payments/orders.php
robincornett: Sure, we could try that--if something semantic exists which makes more potential sense than just generic `<span>` and `<div>` then I'm willing to try it.

@spencerfinnell are there order items, whether on the add or edit screen, which should not be editable? The IP address appears to be something which should never be edited, but would that have to stay in an input field on the add screen so that it would be saved to the order meta? On existing orders, the gateway and IP shouldn't be edited, but I'm thinking maybe the date also should not be changed (unless it was a manual payment in the first place, maybe)?

assets/css/edd-admin.css
spencerfinnell: @robincornett I think @LisaCee was showing the opposite. Currently it applies to all items in "Order Attributes" and "Order Extras", which is correct:

**With**
Screen Shot 2020-04-24 at 1 16 37 PM

**Without**
Screen Shot 2020-04-24 at 1 16 46 PM

includes/admin/payments/orders.php
spencerfinnell: @robincornett _What_ and _why_ things are editable is a lengthy discussion. The current items need to stay as-is. Unfortunately there isn't a discussion I can link to, but it has been decided previously to allow this.

assets/css/edd-admin.css
LisaCee: What about changing the css for `.edd-form-group__control:last-of-type`?  
Current
image

vs without
image

That gives it a bit more spacing at the bottom without making it really big.

spencerfinnell: @LisaCee If we want more space in those rows we should adjust the padding on `.edd-admin-box-inside` instead.

Let's do this:

- Change `.edd-form-group` to have `16px` bottom margin.
- Change `.edd-admin-box-inside` to have `12px` padding all around (increasing 2px on the top and bottom from previous)

includes/admin/payments/orders.php
spencerfinnell: @robincornett For the sake of keeping this PR moving forward and focused on `.edd-form-group` only, I think we should address creating a generic `edd-dl` at a later date.

@LisaCee For IP, when viewing an Order let's ensure we aren't using an actual `label` element for the label, and that the items align to the left and right -- so it's setup just like Gateway on both Add and Edit views.

LisaCee: Like this?
image

spencerfinnell: @LisaCee Exactly

robincornett: I don't know why I would ever need the key, but it bothers me a bit that it's not wide enough to show the entire thing?

And I think the Gateway/IP look great that way, thank you @LisaCee!

LisaCee: And you want it to look like that in add?  Right now I'm getting this: 
image

Do you want the gateway select to be right aligned?

spencerfinnell: @LisaCee It should still be a normal input when Adding an Order, so the IP can be set to something specific, other than the site admin's IP.

spencerfinnell:
@LisaCee 

We can remove

```css
.edd-admin-box-inside .edd_datepicker,
.edd-admin-box-inside input[type="text"] {
    max-width: 130px;
    width: 100%;
}
```

To allow things like `regular-text` classes applied to Key, Transaction ID, and any other fields to adjust the width accordingly.

includes/admin/payments/orders.php
robincornett: The gateway select should not be right aligned--it should look just like a normal select field. I mean, as much as it can. Would love to update our Chosen styles to look a little more WP-ish but that's a different issue.

spencerfinnell: Add

Gateway: editable `<select>` with label (standard `edd-form-control`)
IP: editable `<input>` with label (standard `edd-form-control`)

Edit/View

Gateway: right aligned static text
IP: right aligned static text

LisaCee: Alignment-wise for Add an Order?
image

assets/css/edd-admin.css
```diff
@@ -1825,7 +1817,7 @@ input[class*="edd-price-field"] {
 .edd-admin-box-inside {
 	border-bottom: 1px solid #f1f1f1;
 	clear: both;
-	padding: 10px 12px;
+	padding: 12px 12px;
```
robincornett, line 1820: Not a big deal, but since these are identical, you can just set this to `padding: 12px;`.

includes/admin/payments/orders.php
```diff
@@ -742,33 +794,60 @@ function edd_order_details_extras( $order = false ) {
...
 				<?php endif; ?>
 
-				<div class="edd-order-payment-key edd-admin-box-inside">
-					<label for="edd_payment_key" class="label"><?php esc_html_e( 'Key', 'easy-digital-downloads' ); ?></label>
-					<input type="text" name="payment_key" id="edd_payment_key" <?php echo esc_attr( $readonly ); ?> value="<?php echo esc_attr( $order->payment_key ); ?>" />
+				<div class="edd-admin-box-inside">
+					<div class="edd-form-group">
+						<label for="edd_payment_key" class="edd-form-group__label"><?php esc_html_e( 'Key', 'easy-digital-downloads' ); ?></label>
+						<div class="edd-form-group__control">
+							<input type="text" name="payment_key" id="edd_payment_key" class="edd-form-group__input" <?php echo esc_attr( $readonly ); ?> value="<?php echo esc_attr( $order->payment_key ); ?>" />
```
robincornett, line 831: If we could go ahead and add 'regular-text` to this as a class I would like that. Just so I can see the entire key.

includes/admin/payments/orders.php
```diff
@@ -742,33 +794,60 @@ function edd_order_details_extras( $order = false ) {
...
-					<?php else : ?>
-						<span><?php echo edd_payment_get_ip_address_url( $order->id ); // WPCS: XSS ok. ?></span>
-					<?php endif; ?>
-				</div>
+				<?php if ( edd_is_add_order_page() ) : ?>
+					<div class="edd-order-ip edd-admin-box-inside">
+						<div class="edd-form-group">
+							<label for="edd_ip" class="edd-form-group__label"><?php esc_html_e( 'IP', 'easy-digital-downloads' ); ?></label>
+							<div class="edd-form-group__control">
+								<input type="text" name="ip" id="edd_ip" class="edd-form-group__input" readonly value="<?php echo esc_attr( edd_get_ip() ); ?>" />
```
spencerfinnell, line 841: This shouldn't be `readonly`. It should be prefilled, but with the ability to edit it.

assets/css/edd-admin.css
LisaCee: Might as well make it more concise.

PR #7786, opened 2020-04-24
Fix Product Dropdown

### LINKED ISSUE #7551: 3.0 - Allow Product Dropdown to Show Variations + Standard Downloads

When using `EDD()->html->product_dropdown()` allow for downloads with variations to show just variations, but still show standard downloads alongside.

### Justification or use case

Currently when showing only variations (not `(All Price Options)`) standard Downloads are not added to the prepopulated list of Downloads or included in search results.

This limits the results to something like:

```
Download A: Price 1
Download A: Price 2
```

vs.

```
Download A: Price 1
Download A: Price 2
Download B
Download C
```

Previously/Related #6760 #7149

robincornett:
Fixes #7551

Proposed Changes:
1. Updates the product_dropdown list building to include non-variable products, regardless of variation options
2. Adds some inline documentation to clarify the `foreach` behavior
3. Adds variations to the discount screens ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7551#issuecomment-608110115))

All usages of the `product_dropdown` method were [researched](https://docs.google.com/spreadsheets/d/1fm2-KTs0bkiaP8-s5WUvezEdd3GGcn0v4Bi3ePsUztM/edit?usp=sharing); tests should be added for this.

### DIFF
--- a/includes/admin/discounts/edit-discount.php
+++ b/includes/admin/discounts/edit-discount.php
@@ -124,6 +124,7 @@ $minutes              = edd_get_minute_values();
 							'selected'    => $product_requirements,
 							'multiple'    => true,
 							'chosen'      => true,
+							'variations'  => true,
 							'placeholder' => sprintf( __( 'Select %s', 'easy-digital-downloads' ), edd_get_label_plural() )
 						) ); ?>
                         <div id="edd-discount-product-conditions"<?php echo $condition_display; ?>>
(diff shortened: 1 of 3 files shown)


includes/admin/discounts/edit-discount.php
```diff
@@ -160,7 +161,8 @@
                             'id'          => 'excluded_products',
                             'selected'    => $excluded_products,
                             'multiple'    => true,
-                            'chosen'      => true,
+							'chosen'      => true,
+							'variations'  => true,
```
ashleyfae, line 165: Let's fix these spacing inconsistencies.

PR #7793, opened 2020-04-27
Deactivate the manual purchases addon

### LINKED ISSUE #7719: 3.0 - Deactivate Manual Purchases addon
We should deactivate the Manual Purchases addon when 3.0 is active.

### Justification or use case
Since the manual purchases functionality has been merged into EDD core with 3.0, having the addon becomes unnecessary.

robincornett:
Fixes #7719


LisaCee:
Worked as expected.  I enabled Manual Purchase on release/3.0 and had an extra button to Create Payment on the Order Screen.  I switched over to issue/7719 and Manual Purchases deactivated itself and the extra button no longer appeared.

PR #7800, opened 2020-04-29
Use unserialized meta array when computing differences #7789

### LINKED ISSUE #7789: 3.0 - Migration - Custom payment meta keys not getting saved during order migration
### Expected behavior

I expect all payment meta from 2.9 to be copied over to 3.0 in some capacity, and I expect it to still be able to be retrieved by fetching "payment_meta".

### Actual behavior

I haven't dug into the cause yet, but I'm seeing some discrepancies on a clone of a live install. I think all the inconsistencies are custom meta or meta from integrations. As one example, I'm seeing `edd_send_prod_updates` in the 2.9 meta but not in 3.0. Can't find it anywhere.

Still need to do more work to reproduce this. Might do another migration just in case.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #7789

Proposed Changes:
1. Use the `$payment_meta` variable when computing key differences, as this has already been unserialized. `$meta['_edd_payment_meta']` is actually an array of unserialized data so this whole thing wasn't working.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -611,11 +611,13 @@ class Data_Migrator {
 		);
 
 		// Remove all the core payment meta from the array, and...
-		$remaining_payment_meta = array_diff_key( $payment_meta, array_flip( $core_meta_keys ) );
+		if ( is_array( $payment_meta ) ) {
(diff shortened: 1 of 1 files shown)


ashleyfae:
Confirmed this fixed my issue so this is ready for review.

PR #7813, opened 2020-04-30
Fix some CLI migrations not being marked as complete

### LINKED ISSUE #7805: 3.0 - CLI Migration - Some upgrades not marked as completed when run through CLI
### Expected behavior

When I run the 3.0 migration via CLI, I expect each stage to get marked as a completed upgrade via 
`edd_set_upgrade_complete()`.

### Actual behavior

The orders migration uses the wrong key.

CLI: `edd_set_upgrade_complete( 'migrate_payments' );`
UI: `migrate_orders`

These do not get set at all:

`migrate_customer_addresses`
`migrate_customer_email_addresses`

The method that handles those is `migrate_customer_data()` and it's actually checking and setting: `migrate_customer_data`, which is not a key used at all in the UI.

1) Run 3.0 migration via CLI.

PHP Version:

WordPress Version:

robincornett:
Fixes #7805

Proposed Changes:
1. Changes `migrate_payments` to `migrate_orders` to match the UI migration
2. Replaces `migrate_customer_data` with `migrate_customer_addresses` and `migrate_customer_email_addresses`, also to match the UI
3. Replaces `WP_CLI::error` with `WP_CLI::warning` since two processes are running

@ashleyfae setting this as a draft for the moment because the UI checks query the database and count before attempting either migration. Should the CLI process be updated to match this?


ashleyfae:
> the UI checks query the database and count before attempting either migration.

@robincornett Can you clarify for me where you're looking? Is it this: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/upgrade-functions.php#L1350-L1353 ?

robincornett:
Yes, that check, which is the result of this query?

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/7a4060dcca47a6d722aaa87d7dbf0a300a11faf8/includes/admin/upgrades/upgrade-functions.php#L1347

The CLI process doesn't do this check.

ashleyfae:
@robincornett I don't think we need to do that. It looks like the UI version only does that to determine whether or not they should show the upgrade option inside the UI. Because the CLI doesn't have a UI it's not necessary there.

PR #7815, opened 2020-05-01
Fix check for EDD_Customer migrator script

### LINKED ISSUE #7801: 3.0 - EDD_Customer out of namespace in migrator script
## Bug Report

I'm not sure what the actual consequences of this are because I don't think I've actually seen a real life error, but in the 3.0 `class-data-migrator.php` script, `EDD_Customer` is being referenced incorrectly. The file is namespaced under `EDD\Admin\Upgrades\v3` but the class `EDD_Customer` is in the global namespace. Therefor this would throw an error:

```php
// Maybe add email address to customer record
if ( ! empty( $customer ) && $customer instanceof EDD_Customer ) {
	$primary = ( $customer->email === $purchase_email );
	$customer->add_email( $purchase_email, $primary );
}
```

PHP Version:

WordPress Version:

robincornett:
Fixes #7801

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -568,7 +568,7 @@ class Data_Migrator {
 		edd_maybe_add_customer_address( $customer_id, $customer_address_data );
 
 		// Maybe add email address to customer record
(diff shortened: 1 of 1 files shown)


ashleyfae:
This looks good, but I'm also going to rerun the 3.0 migration and see if this actually changes any behaviour now that it works properly.

PR #7816, opened 2020-05-01
Remove `edd_product_type` for default download

### LINKED ISSUE #7811: 3.0 - Downloads - "0" Values Saving Meta for Download Type
### Expected behavior

When saving a Download with the "Default" product type, no `_edd_product_type` metadata should be saved.

### Actual behavior

`_edd_product_type` saves a `0` value.

1) Save a Download with the "Default" product type.

Introduced in https://github.com/easydigitaldownloads/easy-digital-downloads/commit/8da7e908d7e86920293ffc7c79e4329f3542b8f2

This is breaking things that rely on meta queries using `NOT EXISTS` comparators, such as the Products dropdown excluding Bundles.

LisaCee:
Fixes #7811 

Proposed Changes:
1. Search and delete `edd_product_type` field from metadata for "default" products
2. Update meta query to match AJAX search

To Test:
- Create a new default download and check database
- Before database would have `_edd_product_type` of `"0"`
- Now there is no entry for default

### DIFF
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -137,6 +137,7 @@ function edd_download_meta_box_save( $post_id, $post ) {
 
 			update_post_meta( $post_id, $field, $new_default_price_id );
 		} elseif ( '_edd_product_type' === $field && '0' === $_POST[ $field ] ) {
+			// No value stored when product type is "default" ("0") for backwards compatibility.
 			delete_post_meta( $post_id, '_edd_product_type' );
 		} else {
 			if ( isset( $_POST[ $field ] ) ) {
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -108,7 +108,6 @@ function edd_download_metabox_fields() {
  * @return void
  */
 function edd_download_meta_box_save( $post_id, $post ) {
-
 	if ( ! isset( $_POST['edd_download_meta_box_nonce'] ) || ! wp_verify_nonce( $_POST['edd_download_meta_box_nonce'], basename( __FILE__ ) ) ) {
 		return;
 	}
@@ -127,7 +126,6 @@ function edd_download_meta_box_save( $post_id, $post ) {
 
 	// The default fields that get saved
 	$fields = edd_download_metabox_fields();
-
 	foreach ( $fields as $field ) {
 		if ( '_edd_default_price_id' == $field && edd_has_variable_prices( $post_id ) ) {
 
@@ -138,8 +136,9 @@ function edd_download_meta_box_save( $post_id, $post ) {
 			}
 
 			update_post_meta( $post_id, $field, $new_default_price_id );
+		} elseif ( '_edd_product_type' === $field && '0' === $_POST[ $field ] ) {
+			delete_post_meta( $post_id, '_edd_product_type' );
 		} else {
-
 			if ( isset( $_POST[ $field ] ) ) {
 				$new = apply_filters( 'edd_metabox_save_' . $field, $_POST[ $field ] );
 				update_post_meta( $post_id, $field, $new );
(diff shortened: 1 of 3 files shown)


includes/admin/downloads/metabox.php
```diff
@@ -138,16 +136,16 @@ function edd_download_meta_box_save( $post_id, $post ) {
 			}
 
 			update_post_meta( $post_id, $field, $new_default_price_id );
+		} elseif ( '_edd_product_type' === $field && '0' === $_POST[ $field ] ) {
```
spencerfinnell, line 139: @LisaCee Let's add a comment here explaining that for backwards compatibility no value is stored when Product Type is set to "Default" (represented by a `0` value).

includes/class-edd-html-elements.php
```diff
@@ -86,7 +86,7 @@ public function product_dropdown( $args = array() ) {
 				array(
 					'key'     => '_edd_product_type',
 					'value'   => 'bundle',
-					'compare' => 'NOT EXISTS',
+					'compare' => '!=',
```
robincornett, line 89: The `meta_query` needs to be updated to account both for products which do have the `_edd_product_type` meta key, and for products which no longer have the meta key. This appears to work:

```php
			$product_args['meta_query'] = array(
				'relation' => 'OR',
				array(
					'key'     => '_edd_product_type',
					'value'   => 'bundle',
					'compare' => '!=',
				),
				array(
					'key'     => '_edd_product_type',
					'value'   => 'bundle',
					'compare' => 'NOT EXISTS',
				),
			);
```

spencerfinnell: @robincornett I believe since this was an unreleased bug and the meta can't exist on `2.9.x` we don't need to account for the times it was saved while testing `3.0`. 

@ashleyfae I know we have some 3.0 database migration routines that are required for those using pre-release 3.0 -- are we also trying to account for temporary data inaccuracies like this?

ashleyfae: @spencerfinnell Hm introduced in 2018 by the looks of it? That was quite a while ago...

I guess I'm okay with not accounting for it provided that re-saving the download would immediately fix the problem (which I'm guessing it would?).

My preference would be:

1. Don't account for it
2. Write an upgrade routine to fix it

Meta queries are already gross, so I'd rather not build one grosser than it has to be.

spencerfinnell: @ashleyfae That meta query has been around since then but there hasn't been a possibility to save a value other than `bundle` since `3.0`.  The change by @LisaCee is to provide parity with the AJAX product search that has existed for a while too: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7811#issuecomment-621924359

robincornett: @spencerfinnell the PR changes the meta query check to `!=` instead of `NOT EXISTS`, so shouldn't that be reverted if we are not checking for both?

spencerfinnell: @robincornett You're right -- we should update this instance and the other mentioned in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7811#issuecomment-621924359 to check if it is empty, or `!=` bundle. That will ensure Default (no meta) are returned, as well as potential non-`bundle` types (possibly added by extensions).

Sorry for the confusion ☕

cklosowski:
Do we need to do some further fixes based off [this comment](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7816/files#r419454184) concerning using `!=` vs `NOT EXISTS`?

LisaCee:
> Do we need to do some further fixes based off [this comment](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7816/files#r419454184) concerning using `!=` vs `NOT EXISTS`?

I think this is taken care of.  @robincornett , can you confirm ?

robincornett:
I believe the update you made should correct for it--the meta queries now check for any downloads which either do not have the product type meta key at all ( `NOT EXISTS`--preferred default scenario) `OR` do not have it set to `bundle` (`!=`). That second one also leaves room for other "types" of products to be added in the future, if such a thing were desired, so I think the meta query is correct.

cklosowski:
:+1: Thanks @robincornett and @LisaCee

PR #7821, opened 2020-05-04
Remove edd_payment_personal_details_list hook from add order screen

### LINKED ISSUE #7819: 3.0 - Add Order UI - "Call to a member function get_meta() on bool" fatal error
### Expected behavior

When I go to the Add New Order page, I expect to not see any fatal errors.

### Actual behavior

The following error is thrown:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_meta() on bool in /var/www/edd3/wp-content/plugins/easy-digital-downloads/includes/admin/payments/orders.php:278
```

Related to this change: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7788

We just missed the fact that this hook was also running on the Add Order UI, at which point the order (or `$payment`) doesn't yet exist.

Perhaps we should consider not triggering the `edd_payment_personal_details_list` hook at all if there's no order/payment. The hook didn't used to run on the manual payments Add Payment UI.

1) Go to Downloads > Orders and click to add a new one.
2) Error will immediately trigger.

PHP Version:

WordPress Version:

robincornett:
Fixes #7819

### DIFF
--- a/includes/admin/payments/orders.php
+++ b/includes/admin/payments/orders.php
@@ -275,7 +275,7 @@ function edd_order_details_customer( $order ) {
 	<?php
 
 	// The edd_payment_personal_details_list hook is left here for backwards compatibility
-	if ( ! edd_is_add_order_page() ) {
+	if ( ! edd_is_add_order_page() && $payment instanceof EDD_Payment ) {
 		do_action( 'edd_payment_personal_details_list', $payment->get_meta(), $user_info );
 	}
 	do_action( 'edd_payment_view_details',          $order->id );
--- a/includes/admin/payments/orders.php
+++ b/includes/admin/payments/orders.php
@@ -275,7 +275,9 @@ function edd_order_details_customer( $order ) {
 	<?php
 
 	// The edd_payment_personal_details_list hook is left here for backwards compatibility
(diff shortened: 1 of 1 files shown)


includes/admin/payments/orders.php
```diff
@@ -275,7 +275,9 @@ function edd_order_details_customer( $order ) {
 	<?php
 
 	// The edd_payment_personal_details_list hook is left here for backwards compatibility
-	do_action( 'edd_payment_personal_details_list', $payment->get_meta(), $user_info );
+	if ( ! edd_is_add_order_page() ) {
```
ashleyfae, line 278: Just for sanity, can we add an additional check if `$payment` is an object?

robincornett: @ashleyfae since `edd_get_payment` returns either a payment object or `false` is a `! $payment` check sufficient?

ashleyfae: @robincornett I personally really like to use `instanceof` in cases like this. So:

`&& $payment instanceof EDD_Payment`

That will confirm it's an object and of the expected class, which also means we'll know the `get_meta()` method exists too.

PR #7823, opened 2020-05-04
Do not add email address if it already exists for the customer

### LINKED ISSUE #7822: 3.0 - Migration - Secondary customer email addresses can be added twice
### Expected behavior

When I perform the migration, I expect each customer email address to only be added once. There should not be any duplicates in the customer_email_addresses table.

### Actual behavior

A few email addresses are being duplicated. All of the duplicates are `secondary` emails. Here's where the issue is:

First, **orders** are migrated. During this step, emails are added here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L570-L574 With my data set, this is where a secondary customer address is added for the first time. This is coming from the order information record.

Second, the **customer email addresses** are migrated. This runs this segment: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L76-L89 However, this block of code just blindly adds another record for a `secondary` email address. It doesn't first check to see if a record with that email already exists.

To fix this, we should check if the email already exists before adding it.

Don't have exact reproduction steps since this is coming from my live data set. From the code, it looks like this could happen when one customer has a set up like this:

- Primary email set to primary@test.com
- Has an `additional_email` saved in customer meta. Let's call this secondary@test.com.
- Has an order associated with secondary@test.com.

So they key is having an order record associated with that secondary email. That causes the email to get imported twice.

PHP Version:

WordPress Version:

robincornett:
Fixes #7822

This can be tested with this code:

```php
add_action( 'wp_head', function() {
	edd_add_customer_email_address( 
		array(
			'customer_id' => 142, // replace with a valid customer ID
			'email'       => 'Aurelio.Kogut@affwp.example', // replace with an email belonging to that customer
		) 
	);
} );
```
In `release/3.0` the above code will add a duplicate email address to the database; switch to this branch and that should no longer happen.

### DIFF
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -749,8 +749,8 @@ function edd_get_customer_address_counts( $args = array() ) {
  */
 function edd_add_customer_email_address( $data ) {
 
-	// A customer ID must be supplied for every address inserted.
-	if ( empty( $data['customer_id'] ) ) {
+	// A customer ID and email must be supplied for every address inserted.
+	if ( empty( $data['customer_id'] ) || empty( $data['email'] ) ) {
 		return false;
 	}
 
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -760,9 +760,8 @@ function edd_add_customer_email_address( $data ) {
 	// Check if the address already exists for this customer.
 	$existing_addresses = $customer_email_addresses->query(
 		array(
-			'customer_id'   => $data['customer_id'],
-			'no_found_rows' => true,
-			'email'         => $data['email'],
+			'customer_id' => $data['customer_id'],
+			'email'       => $data['email'],
 		)
 	);
 	if ( ! empty( $existing_addresses ) ) {
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -753,14 +753,23 @@ function edd_add_customer_email_address( $data ) {
 	if ( empty( $data['customer_id'] ) ) {
 		return false;
 	}
-	$exists = edd_get_customer_email_address_by( 'email', $data['email'] );
-	if ( $exists ) {
-		return false;
-	}
 
 	// Instantiate a query object.
 	$customer_email_addresses = new EDD\Database\Queries\Customer_Email_Address();
 
+	// Check if the address already exists for this customer.
+	$existing_addresses = $customer_email_addresses->query(
+		array(
+			'customer_id'   => $data['customer_id'],
+			'no_found_rows' => true,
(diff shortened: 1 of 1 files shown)


robincornett:
Actually, an easier check would be something like this:

```php
	$exists = edd_get_customer_email_address_by( 'email', $data['email'] );
	if ( $exists ) {
		return false;
	}
```
which is used elsewhere, but appears to check just if the email address exists anywhere in the database already, not whether it's assigned to a specific customer. Is it better to do it this way, @ashleyfae, so that there is no duplication anywhere in the database, or is it possibly okay (but maybe weird) to have multiple customers with a shared email address?

ashleyfae:
@robincornett I think this method is better: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7823/commits/425ef14fe493d68dd2e2c229581f00b1f3bd9716 However, you should be able to modify that query to check for `email` as well directly in the query.

I guess my reason is that people do weird stuff and might add their developer's email address to their account or something, in which case it's perfectly reasonable that the same developer might be associated with multiple clients. Let's focus on keeping them unique _within a single customer_.

includes/customer-functions.php
```diff
@@ -753,14 +753,23 @@ function edd_add_customer_email_address( $data ) {
...
-	}
 
 	// Instantiate a query object.
 	$customer_email_addresses = new EDD\Database\Queries\Customer_Email_Address();
 
+	// Check if the address already exists for this customer.
+	$existing_addresses = $customer_email_addresses->query(
+		array(
+			'customer_id'   => $data['customer_id'],
+			'no_found_rows' => true,
```
ashleyfae, line 764: (I know you know this already - just noting it down anyway.)

The `no_found_rows` argument can be removed because `true` is already the default.

ashleyfae:
```php
if ( empty( $data['customer_id'] ) ) {
	return false;
}
```

Let's replicate this for `$data['email']` too.

As it is, right now `$data['email']` isn't required, which means this could technically have an undefined index error if someone calls the function without `email` :

```php
// Check if the address already exists for this customer.
$existing_addresses = $customer_email_addresses->query(
	array(
		'customer_id' => $data['customer_id'],
		'email'       => $data['email'],
	)
);
```

But rather than only running that if `$data['email']` exists, I think we should just make it required just like `customer_id` is. I can't see any reason to add an email address record without an email address. :shrug: 

---

I have confirmed this PR does fix the problem as-is though.

PR #7846, opened 2020-05-08
Remap missing EDD_Payments_Query arguments #7845

### LINKED ISSUE #7845: 3.0 - EDD_Payments_Query missing some backwards compatibility in arg mapping
### Expected behavior

When I use `EDD_Payments_Query` I expect all parameters that worked in 2.9 to still work in 3.0 and get remapped accordingly.

### Actual behavior

Not all arguments get remapped.

#### `meta_key`

Previously you could pass in `meta_key` without a value. But now, `meta_key` only gets included if `meta_value` was also provided: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-payments-query.php#L571 This should not be necessary.

#### "Individual" date arguments

None of the "individual" date arguments get mapped. Such as: `year`, `month`, `day`. Also `hour` used to work, though it's not documented in the defaults.

#### `number` including support for `-1`

`number` used to be accepted. But the remapping is only looking for `posts_per_page`. Also we need to consider support for `-1` to fetch all (related: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7748)

There might be other issues as well. These are just the ones that were found in Software Licensing: https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1644

1) Have Software Licensing installed.
2) Visit the SL Upgrades Report.
3) The results on the graph will be inaccurate. This is due to the query not working as expected.

Alternatively:

1) Have Software Licensing installed.
2) Manually run this: `$renewals = edd_sl_get_renewals_by_date( 8, 5, 2020, 12 );`
3) Note how results are inaccurate.

The arguments being passed in to `EDD_Payments_Query` to run that are:

```
array (
  'number' => -1,
  'status' => 
  array (
    0 => 'revoked',
    1 => 'publish',
    2 => 'complete',
  ),
  'meta_key' => '_edd_sl_is_renewal',
  'year' => 2020,
  'month' => 5,
  'fields' => 'ids',
  'day' => 8,
  'hour' => 12,
)
```

(this is before merging with defaults)

But after the remapping, we only come out with these:

```
array (
  'orderby' => 'id',
  'type' => 'sale',
  'order' => 'DESC',
  'status__in' => 
  array (
    0 => 'revoked',
    2 => 'complete',
    3 => 'complete',
  ),
)
```

Most notably missing to actually make this query work are: meta_key and all the date parameters.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #7845

Proposed Changes:
1.`meta_key` can now be included without a corresponding `meta_value`.
2. Add support for "individual" date intervals: `array( 'year', 'month', 'week', 'day', 'hour', 'minute', 'second' ) `
3. Add support for `number`.
4. If `-1` is passed as `number` then `nopaging` is set to `true`.
5. If `nopaging` is set to `true` then this is meant to show all results. This was doing `unset( $arguments['number'] )` but the default in `edd_get_orders()` was then just being used instead, which was effectively setting the number to 30. So this has been adjusted to set the number to a really large integer instead.

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -568,13 +568,22 @@ class EDD_Payments_Query extends EDD_Stats {
 		}
 
 		// Meta key and value
-		if ( isset( $this->initial_args['meta_key'] ) && isset( $this->initial_args['meta_value'] ) ) {
-			$arguments['meta_query'] = array(
-				array(
-					'key'   => $this->initial_args['meta_key'],
-					'value' => $this->initial_args['meta_value'],
-				),
+		if ( isset( $this->initial_args['meta_key'] ) ) {
+			$meta_query = array(
+				'key' => $this->initial_args['meta_key']
 			);
+
+			if ( isset( $this->initial_args['meta_value'] ) ) {
+				$meta_query['value'] = $this->initial_args['meta_value'];
+			}
+
+			$arguments['meta_query'] = array( $meta_query );
+		}
+
+		foreach ( array( 'year', 'month', 'week', 'day', 'hour', 'minute', 'second' ) as $date_interval ) {
+			if ( isset( $this->initial_args[ $date_interval ] ) ) {
+				$arguments['date_created_query'][ $date_interval ] = $this->initial_args[ $date_interval ];
+			}
 		}
 
 		if ( $this->args['start_date'] ) {
@@ -607,6 +616,14 @@ class EDD_Payments_Query extends EDD_Stats {
 			);
 
 			$arguments['date_created_query']['inclusive'] = true;
+		}
+
+		if ( isset( $this->initial_args['number'] ) ) {
+			if ( -1 == $this->initial_args['number'] ) {
+				$this->args['nopaging'] = true;
+			} else {
+				$arguments['number'] = $this->initial_args['number'];
+			}
 		}
 
 		$arguments['number'] = isset( $this->args['posts_per_page'] )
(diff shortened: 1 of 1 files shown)


includes/payments/class-payments-query.php
```diff
@@ -609,12 +618,22 @@ private function remap_args() {
 			$arguments['date_created_query']['inclusive'] = true;
 		}
 
+		if ( isset( $this->initial_args['number'] ) ) {
+			if ( -1 == $this->initial_args['number'] ) {
```
spencerfinnell, line 622: Is it safe to do a strict check here?

ashleyfae: You could technically pass in the number as a string and if we do a strict check here then it wouldn't work the way we want (to fetch all results).

Both of these should work the same:

```php
$query = new EDD_Payments_Query( array(
	'number' => -1
) );
$query->get_payments();
```

```php
$query = new EDD_Payments_Query( array(
	'number' => '-1'
) );
$query->get_payments();
```

spencerfinnell: What about forcing `$this->initial_args['number']` to a string -- that seems to be the most common?

ashleyfae: If we're going to force it to something I think I'd rather do an integer. That's what all our default number values are already.

Is there something in particular you're concerned about here with not using a strict check though?

spencerfinnell: No big deal since we don't technically enforce it, but ideally it will eventually become a rule: https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/PHP/StrictComparisonsSniff.php

ashleyfae:
Updated to remap `post__not_in` as well. Found that issue while testing and fixing #7910

PR #7853, opened 2020-05-08
Move Send Receipt Option to Top of Order

### LINKED ISSUE #7738: 3.0 - Orders - Move "Send Receipt" to Top Bar

Move "Send Receipt" checkbox to the top bar next to "Create Order" and have it checked by default. 
### Justification or use case

When creating an order the option to "Send Receipt" is the very last item on the page. You can resend the receipt once the Order is created, but I think presenting the option on initial creation already opted-in better matches the similar WordPress pattern of creating a user:

Screen Shot 2020-04-14 at 9 53 54 AM

LisaCee:
Fixes #7738  

Proposed Changes:
1. Changed label for email receipt
2. Moved send receipt checkbox to top of order page
3. Removed `edd-admin-box-inside` class from receipt b/c it gave unnecessary styling

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -481,7 +481,17 @@ a.edd-delete:hover {
 	}
 }
 
-.download_page_edd-payment-history .edit-post-header #publishing-action {
+.download_page_edd-payment-history .edit-post-header .edit-post-header__toolbar {
+	order: 0;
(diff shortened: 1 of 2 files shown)


LisaCee:
I broke something.  I only changed 1 file, so I'm just going to close this and do it again.

PR #7854, opened 2020-05-09
Move Send Purchase Receipt to Top of Order

### LINKED ISSUE #7738: 3.0 - Orders - Move "Send Receipt" to Top Bar

LisaCee:
Fixes #7738 

Proposed Changes:
1. Changed wording on label for send receipt checkbox
2. Move send receipt checkbox to top of Order
3. Removed `edd-admin-box-inside` class to remove unnecessary styling

### DIFF
--- a/includes/admin/payments/orders.php
+++ b/includes/admin/payments/orders.php
@@ -907,7 +907,7 @@ function edd_order_details_extras( $order = false ) {
 
 							<label for="edd_unlimited_downloads">
 							<?php esc_html_e( 'Unlimited Downloads', 'easy-digital-downloads' ); ?></label>
-							<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php _e( '<strong>Unlimited Downloads</strong>: checking this box will override all other file download limits for this purchase, granting the customer unliimited downloads of all files included on the purchase.', 'easy-digital-downloads' ); ?>"></span>
+							<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php esc_html_e( 'Checking this box will override all other file download limits for this purchase, granting the customer unlimited downloads of all files included on the purchase.', 'easy-digital-downloads' ); ?>"></span>
 						</div>
 					</div>
 				</div>
--- a/includes/admin/payments/orders.php
+++ b/includes/admin/payments/orders.php
@@ -904,7 +904,7 @@ function edd_order_details_extras( $order = false ) {
 								<input type="checkbox" name="edd_order_send_receipt" id="edd-order-send-receipt" class="edd-form-group__input" value="1" />
 
 								<label for="edd-order-send-receipt">
-								<?php esc_html_e( 'Send Receipt', 'easy-digital-downloads' ); ?>
+								<?php esc_html_e( 'Send Purchase Receipt', 'easy-digital-downloads' ); ?>
 								</label>
 								<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php _e( '<strong>Send Receipt</strong>: checking this box will send the purchase receipt to the selected customer.', 'easy-digital-downloads' ); ?>"></span>
 							</div>
--- a/includes/admin/payments/orders.php
+++ b/includes/admin/payments/orders.php
@@ -33,6 +33,21 @@ function edd_order_details_publish( $order ) {
 		<div class="edit-post-header">
 
 			<div class="edit-post-header__settings">
+				<?php if ( edd_is_add_order_page() ) : ?>
+					<div class="edd-send-purchase-receipt">
+						<div class="edd-form-group">
+							<div class="edd-form-group__control">
+								<input type="checkbox" name="edd_order_send_receipt" id="edd-order-send-receipt" class="edd-form-group__input" value="1" />
+
+								<label for="edd-order-send-receipt">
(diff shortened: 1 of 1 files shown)


includes/admin/payments/orders.php
```diff
@@ -33,6 +33,21 @@ function edd_order_details_publish( $order ) {
 		<div class="edit-post-header">
 
 			<div class="edit-post-header__settings">
+				<?php if ( edd_is_add_order_page() ) : ?>
+					<div class="edd-send-purchase-receipt">
+						<div class="edd-form-group">
+							<div class="edd-form-group__control">
+								<input type="checkbox" name="edd_order_send_receipt" id="edd-order-send-receipt" class="edd-form-group__input" value="1" />
```
robincornett, line 40: If this is supposed to be checked on page load (I think that's how the issue suggests), it needs to include `checked` as an attribute (just `checked`, doesn't `=` anything ([ref](https://www.w3schools.com/tags/att_input_checked.asp))

includes/admin/payments/orders.php
```diff
@@ -33,6 +33,21 @@ function edd_order_details_publish( $order ) {
...
+
+								<label for="edd-order-send-receipt">
+								<?php esc_html_e( 'Send Purchase Receipt', 'easy-digital-downloads' ); ?>
+								</label>
+								<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php _e( '<strong>Send Receipt</strong>: checking this box will send the purchase receipt to the selected customer.', 'easy-digital-downloads' ); ?>"></span>
```
robincornett, line 45: While we're in here, could we go ahead and escape the help-tip text on output? Since it has a `strong` tag in it, it may need to be split into to separate translation strings, not sure about that, since `esc_html_e` would break that.

includes/admin/payments/orders.php
LisaCee: Good catch!

includes/admin/payments/orders.php
spencerfinnell: I think we can just remove the `<strong>Send Receipt</strong>:` prefix in the tooltip. The tooltip is given the context of the label, so I don't think it's necessary to repeat it.

I think the tooltip should change to `Checking this box will email` (replacing `send` with `email`).

LisaCee: For the sake of consistence, at least on this page, should I change the `edd-help-tip` for Unlimited Downloads, too?

spencerfinnell: @LisaCee Sure! This instance still needs @robincornett's original suggestion of `esc_html_e()` too.

includes/admin/payments/orders.php
```diff
@@ -33,6 +33,21 @@ function edd_order_details_publish( $order ) {
...
+				<?php if ( edd_is_add_order_page() ) : ?>
+					<div class="edd-send-purchase-receipt">
+						<div class="edd-form-group">
+							<div class="edd-form-group__control">
+								<input type="checkbox" name="edd_order_send_receipt" id="edd-order-send-receipt" class="edd-form-group__input" value="1" checked />
+
+								<label for="edd-order-send-receipt">
+								<?php esc_html_e( 'Send Purchase Receipt', 'easy-digital-downloads' ); ?>
+								</label>
+								<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php esc_html_e( 'Checking this box will email the purchase receipt to the selected customer.', 'easy-digital-downloads' ); ?>"></span>
```
ashleyfae, line 45:
```suggestion
								<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php esc_attr_e( 'Checking this box will email the purchase receipt to the selected customer.', 'easy-digital-downloads' ); ?>"></span>
```

`esc_attr_e()` is better here since we're inside an HTML attribute.

includes/admin/payments/orders.php
```diff
@@ -892,26 +907,11 @@ function edd_order_details_extras( $order = false ) {
 
 							<label for="edd_unlimited_downloads">
 							<?php esc_html_e( 'Unlimited Downloads', 'easy-digital-downloads' ); ?></label>
-							<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php _e( '<strong>Unlimited Downloads</strong>: checking this box will override all other file download limits for this purchase, granting the customer unliimited downloads of all files included on the purchase.', 'easy-digital-downloads' ); ?>"></span>
+							<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php esc_html_e( 'Checking this box will override all other file download limits for this purchase, granting the customer unlimited downloads of all files included on the purchase.', 'easy-digital-downloads' ); ?>"></span>
```
ashleyfae, line 910:
```suggestion
							<span alt="f223" class="edd-help-tip dashicons dashicons-editor-help" title="<?php esc_attr_e( 'Checking this box will override all other file download limits for this purchase, granting the customer unlimited downloads of all files included on the purchase.', 'easy-digital-downloads' ); ?>"></span>
```

Same as the other one.

PR #7863, opened 2020-05-11
Fix error when viewing notes associated with a deleted user

### LINKED ISSUE #7860: 3.0 - PHP Notice When Viewing an Order for a Non-Existent User
### Expected behavior
When I view order details, I expect no errors.

### Actual behavior
After migrating to 3.0, when I view an order for a customer, I sometimes, but not always, get this notice:

`PHP Notice:  Trying to get property 'display_name' of non-object in \includes\admin\notes\note-functions.php on line 79`

I'm looking at a customer with five orders and one of them consistently triggers this notice multiple times on the order screen. I can't readily see a difference between the order details for this order and the ones which are not throwing the error.

Honestly not sure how to reproduce this. Looking at the affected order, these are notes for Stripe, Mailchimp, and general store info for this order--on the live site for this order, the customer name is showing as the note "author", but if I view another order for this customer, the note "author" is EDD Bot (or Store Bot in 3.0). So I think there may be some kind of data error with this order, maybe (it's not one of the ones which migrated incorrectly, but it is the same customer).

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/cac1134dcb2276c543c3197f5d91a3a626d4d906/includes/admin/notes/note-functions.php#L78-L80

It does seem like we could separate the use of `get_userdata` from immediately retrieving the `display_name` in case `get_userdata` does not return an object.

robincornett:
Fixes #7860

### DIFF
--- a/includes/admin/notes/note-functions.php
+++ b/includes/admin/notes/note-functions.php
@@ -77,7 +77,7 @@ function edd_admin_get_note_html( $note_id = 0 ) {
 	$user_id = $note->user_id;
 	$author  = edd_get_bot_name();
 	if ( ! empty( $user_id ) ) {
-		$author      = sprintf( '%s #%s', __( 'User ID' ), $user_id );
+		$author      = sprintf( '%s #%s', __( 'User ID', 'easy-digital-downloads' ), $user_id );
 		$user_object = get_userdata( $user_id );
 		if ( $user_object ) {
 			$author = ! empty( $user_object->display_name ) ? $user_object->display_name : $user_object->user_login;
--- a/includes/admin/notes/note-functions.php
+++ b/includes/admin/notes/note-functions.php
@@ -77,7 +77,8 @@ function edd_admin_get_note_html( $note_id = 0 ) {
 	$user_id = $note->user_id;
 	$author  = edd_get_bot_name();
 	if ( ! empty( $user_id ) ) {
-		$author      = sprintf( '%s #%s', __( 'User ID', 'easy-digital-downloads' ), $user_id );
+		/* translators: user ID */
+		$author      = sprintf( __( 'User ID #%s', 'easy-digital-downloads' ), $user_id );
 		$user_object = get_userdata( $user_id );
 		if ( $user_object ) {
 			$author = ! empty( $user_object->display_name ) ? $user_object->display_name : $user_object->user_login;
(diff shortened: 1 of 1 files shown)


includes/admin/notes/note-functions.php
```diff
@@ -75,9 +75,14 @@ function edd_admin_get_note_html( $note_id = 0 ) {
...
 	$user_id = $note->user_id;
-	$author  = ! empty( $user_id )
-		? get_userdata( $user_id )->display_name
-		: edd_get_bot_name();
+	$author  = edd_get_bot_name();
+	if ( ! empty( $user_id ) ) {
+		$author      = sprintf( '%s #%s', __( 'User ID' ), $user_id );
+		$user_object = get_userdata( $user_id );
+		if ( $user_object ) {
+			$author = $user_object->display_name;
```
ashleyfae, line 83:
```suggestion
			$author = ! empty( $user_object->display_name ) ? $user_object->display_name : $user_object->user_login;
```

Just in case the display name is blank for some reason. :woman_shrugging:

includes/admin/notes/note-functions.php
```diff
@@ -75,9 +75,14 @@ function edd_admin_get_note_html( $note_id = 0 ) {
 
 	// User
 	$user_id = $note->user_id;
-	$author  = ! empty( $user_id )
-		? get_userdata( $user_id )->display_name
-		: edd_get_bot_name();
+	$author  = edd_get_bot_name();
+	if ( ! empty( $user_id ) ) {
+		$author      = sprintf( '%s #%s', __( 'User ID' ), $user_id );
```
ashleyfae, line 80:
```suggestion
		$author      = sprintf( '%s #%s', __( 'User ID', 'easy-digital-downloads' ), $user_id );
```

Missing the text domain.

includes/admin/notes/note-functions.php
```diff
@@ -75,9 +75,14 @@ function edd_admin_get_note_html( $note_id = 0 ) {
 
 	// User
 	$user_id = $note->user_id;
-	$author  = ! empty( $user_id )
-		? get_userdata( $user_id )->display_name
-		: edd_get_bot_name();
+	$author  = edd_get_bot_name();
+	if ( ! empty( $user_id ) ) {
+		$author      = sprintf( '%s #%s', __( 'User ID', 'easy-digital-downloads' ), $user_id );
```
spencerfinnell, line 80: This string could be given more context by not using a placeholder for `User ID`. @robincornett is there a specific reason to use a placeholder for that here?

robincornett: Sorry, with missing the text domain, I obviously was missing more than just that. Should be improved now.

PR #7867, opened 2020-05-11
When migrating, use order date as address creation date if possible

### LINKED ISSUE #7859: 3.0 - After Migration, Email and Physical Address Dates are Based on Migration Date
### Expected behavior
When I see a date in the admin, I expect it to be meaningful.

### Actual behavior
After migrating data from EDD 2.9 to 3.0, all migrated email and physical addresses show the migration date as the Date in the list tables, and as the "first used" date (for physical addresses) when viewing the customer record.

1) Migrate data from EDD 2.9 to 3.0
2) View the list tables for email and physical addresses

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

When viewing the individual customer screen, the "first used" date can be helpful information, but I'm not sure it's really needed on the list table. Wherever the date shows, I think it should be meaningful...I don't know if it is possible to pull the order date in when setting the address date during migration, but that seems like it would be a logical option.

robincornett:
Proposed changes:
1. When adding the customer address to the database, use the related order completion date as the address creation date
2. Change the way the email address is added to the database so the order completion date will be the email address creation date

If there is not an associated order, the creation date for the address record will be the migration date.

Fixes #7859

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -52,17 +52,20 @@ class Data_Migrator {
 		) );
 
 		if ( $customer ) {
-			edd_add_customer_address( array(
-				'customer_id' => $customer->id,
-				'type'        => 'primary',
-				'name'        => $customer->name,
-				'address'     => $address['line1'],
-				'address2'    => $address['line2'],
-				'city'        => $address['city'],
-				'region'      => $address['state'],
-				'postal_code' => $address['zip'],
-				'country'     => $address['country'],
-			) );
+			edd_add_customer_address(
+				array(
+					'customer_id'  => $customer->id,
+					'type'         => 'primary',
+					'name'         => $customer->name,
+					'address'      => $address['line1'],
+					'address2'     => $address['line2'],
+					'city'         => $address['city'],
+					'region'       => $address['state'],
+					'postal_code'  => $address['zip'],
+					'country'      => $address['country'],
+					'date_created' => $customer->date_created,
+				)
+			);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ashleyfae:
This looks good so far but I think we can do a bit more here.

We could add the same date for the order address: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb5820e258bffc25cd2d4fa9bcea782cce10c1c3/includes/admin/upgrades/v3/class-data-migrator.php#L546-L561

Following this conversation for the next two items: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7859#issuecomment-626759249

For the `customer_addresses()` migration, you can set `date_created` to `$customer->date_created`. https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb5820e258bffc25cd2d4fa9bcea782cce10c1c3/includes/admin/upgrades/v3/class-data-migrator.php#L55-L65

Same for `customer_email_addresses()` https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb5820e258bffc25cd2d4fa9bcea782cce10c1c3/includes/admin/upgrades/v3/class-data-migrator.php#L85-L88 (this will require getting the customer object first because it isn't yet being retrieved)

This will cover setting a more relevant date for addresses and emails that weren't already migrated from an order record.

robincornett:
@ashleyfae should there be additional work if an order is not completed? For example, if an order was abandoned, `date_completed` is `NULL`, so that order shows in the list table and the address has the migration date, and maybe it should be `date_created` as a fallback.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -556,15 +565,17 @@ public static function orders( $data = null ) {
 			'state'   => '',
 		) );
 
+		$address_created    = ! empty( $date_completed ) ? $date_completed : $date_created_gmt;
```
ashleyfae, line 568: Perhaps we should just use `$date_created_gmt` for this? The difference isn't huge (in most cases probably a matter of seconds/minutes), but technically the address was created as soon as the order was created -- not when the order was _completed_.

robincornett:
Note: the times for the migrated addresses do not show a time zone.

ashleyfae:
> Note: the times for the migrated addresses do not show a time zone.

Let's make a separate issue for that.

cklosowski:
@ashleyfae I think by default it does but I want to confirm, this would make the `modified` date be whatever the date the migration is run, as it is defaulted to `NOW()` upon insertion.

ashleyfae:
> @ashleyfae I think by default it does but I want to confirm, this would make the `modified` date be whatever the date the migration is run, as it is defaulted to `NOW()` upon insertion.

Yes that is correct. BerlinDB sets that automatically to the current time.

So `date_created` will be the order date.
And `date_modified` will be the date the migration was run.

PR #7871, opened 2020-05-12
Ensure UI & CLI migrations delete the same legacy data #7824

### LINKED ISSUE #7824: 3.0 - Migration - Deleting legacy data inconsistent between CLI & UI migrations
### Expected behavior

I expect migrating via the CLI to be exactly the same as migrating via the UI... just within the CLI. That means all the same data should be processed.

### Actual behavior

I haven't dug into this 100% yet, but at quick glance, it looks like there are inconsistencies with legacy data removal.

The UI runs `Remove_Legacy_Data`. This class removes data from two places:

1. Customer meta - deleting data `WHERE meta_key = 'additional_email'`
2. Deleting posts `WHERE post_type = 'edd_payment' OR post_type = 'edd_discount' OR post_type = 'edd_log'`

The CLI has the following deletion options:

1. Legacy discounts
2. Legacy logs
3. Legacy notes

So it appears that the UI is missing deleting legacy notes and the CLI is missing options for deleting customer legacy email records and legacy payment records.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #7824

Proposed Changes:
1. CLI - add support for deleting legacy customer emails
2. CLI - add support for deleting legacy customer addresses
3. CLI - add support for deleting legacy orders
4. UI - add support for deleting legacy customer addresses
5. UI - add support for deleting legacy notes

---

## To test:

1. Run a migration via the UI. Opt to delete legacy data at the end. Once complete, ensure all legacy data has been deleted. (see below)
2. Run a migration via WP-CLI. Opt to delete legacy data at the end. Once complete, ensure all legacy data has been deleted. (see below)

These queries can be run manually to check that all data has been deleted:

```mysql
SELECT COUNT(*) FROM wp_edd_customermeta WHERE meta_key = 'additional_email';
SELECT COUNT(*) FROM wp_usermeta WHERE meta_key = '_edd_user_address';
SELECT COUNT(*) FROM wp_posts WHERE post_type IN('edd_payment', 'edd_discount', 'edd_log');
SELECT COUNT(*) FROM wp_comments WHERE comment_type = 'edd_payment_note';
```

## Note:

In the UI, I've combined deleting post types and comments in the same process. This should work fine, but the percentage bar isn't guaranteed to be accurate. That's because we're deleting data as we go and don't keep track of exactly how many have been deleted in each step. Because although it's set up to delete a max of 25 post types and 25 comments each step, there may not necessarily be that full amount each time. You might get to a point where you're deleting 25 post types and 0 comments each step because the comments have all been fully deleted.


ashleyfae:
Anyone have any thoughts on deleting legacy notes via the UI?

The issue I'm looking at is that all the legacy note removal is in one class - `Remove_Legacy_Data`. The "stepping" in this class is already being used by the post type removal. I'm hesitant to do a one-liner removal like is being done for additional emails and addresses, because there could be a ton of notes. Batching them feels better.

The best solution I can think of involves a second class so we can step it properly, but that also involves either reworking the UI or doing some fancy JavaScript to run the second class after the first finishes. Just wondering if I'm missing a simpler/better option.

@spencerfinnell @cklosowski

robincornett (commented):
**Upgrade/remove legacy data via UI:** migration appears to have gone smoothly, and all legacy data has been removed per the instructions for testing. Noting, however, that reloading the migration screen shows active buttons for:
- Upgrade Orders
- Upgrade Notes
- Upgrade Customer Notes

**Upgrade/remove legacy data via CLI:** migration has gone smoothly, and the legacy data has all been removed per the instructions. Additionally, visiting the migration screen shows no active buttons, so I think somehow the `edd_set_upgrade_complete()` function is not being triggered for the above via the UI.

ashleyfae:
> Noting, however, that reloading the migration screen shows active buttons for:
> 
> * Upgrade Orders
> * Upgrade Notes
> * Upgrade Customer Notes

Can you replicate in release/3.0? This PR didn't touch that part so I don't think it would have been affected by what's going on here.

robincornett:
Yes, it's the same in `release/3.0` so I'll make a new issue.

robincornett (approved):

PR #7874, opened 2020-05-12
WIP: Customer address table changes #7254

### LINKED ISSUE #7254: 3.0 - EDD_Customer->get_address method not working
### Expected behavior
When I have a customer with an address in the Customer Addresses table, and I run this code:

```
$customer = new EDD_Customer( CUSTOMER_ID_HERE );
print_r( $customer->get_address() );
```

I expect to see the customer's active address. Note that the "Real world" scenario here is the auto pre-filling of the address upon checkout for logged in customers.

### Actual behavior
I get no output.

1) Purchase a product with taxes enabled, and fill out your address.
2) Attempt to run this code in debug bar console:

```
$customer = new EDD_Customer( CUSTOMER_ID_HERE );
print_r( $customer->get_address() );
```

3) You will get no output. 

### Reason
The address is being queried with the paramater "primary", but that isn't a queryable paramater:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-customer.php#L1074

EDD Version (or branch): 3.0

cklosowski:
Fixes #7254

Proposed Changes:
1.
2.
3.


cklosowski:
@robincornett If you could test this out a bit, I think this should account for the migration of empty values, but I don't have an amazing data set for this yet. So a little testing from other data sets like yours and @ashleyfae would be awesome.

includes/database/schemas/class-customer-addresses.php
```diff
@@ -52,6 +52,17 @@ class Customer_Addresses extends Schema {
...
 
+		// is_primary
+		array(
+			'name'       => 'is_primary',
+			'type'       => 'tinyint',
+			'length'     => '1',
+			'unsigned'   => false,
+			'default'    => '0',
+			'sortable'   => true,
+			'searchable' => true,
```
ashleyfae, line 63: I don't really think we need this to be `searchable`.

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -66,9 +66,10 @@ public function column_default( $item, $column_name ) {
 		switch ( $column_name ) {
 
 			case 'type' :
-				$value = ( 'primary' === $item['type'] )
-					? esc_html_e( 'Primary',   'easy-digital-downloads' )
-					: esc_html_e( 'Secondary', 'easy-digital-downloads' );
+				$value = ucfirst( $item['type'] );
```
ashleyfae, line 69: This won't be translatable now. Perhaps we translate the values we know exist (like `billing`) and as a fallback for anything else resort to `ucfirst( $item['type'] )`.

cklosowski:
Fixed up those issues.

cklosowski:
@spencerfinnell @ashleyfae this is ready for another review after making the requested changes.

includes/admin/customers/customers.php
```diff
@@ -737,7 +737,12 @@ function edd_customers_view( $customer = null ) {
 								: '&mdash;';
 							?>
 						</td>
-						<td><time datetime="<?php echo esc_attr( EDD()->utils->date( $address->date_created, null, true )->toDateTimeString() ); ?>"><?php echo edd_date_i18n( EDD()->utils->date( $address->date_created, null, true )->toDateTimeString(), 'M. d, Y' ) . '<br>' . edd_date_i18n( strtotime( $address->date_created ), 'H:i' ) . ' ' . edd_get_timezone_abbr(); ?></time></td>
+						<td>
+							<time datetime="<?php echo esc_attr( EDD()->utils->date( $address->date_created, null, true )->toDateTimeString() ); ?>"><?php echo edd_date_i18n( EDD()->utils->date( $address->date_created, null, true )->toDateTimeString(), 'M. d, Y' ) . ' ' . edd_date_i18n( strtotime( $address->date_created ), 'H:i' ) . ' ' . edd_get_timezone_abbr(); ?></time>
+							<?php if ( ! empty( $address->is_primary ) ) : ?>
+								<span class="edd-chip"><?php _e( 'Primary', 'easy-digital-downloads' ); ?></span>
```
spencerfinnell, line 743:
```suggestion
								<span class="edd-chip"><?php esc_html_e( 'Primary', 'easy-digital-downloads' ); ?></span>
```

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -66,9 +66,10 @@ public function column_default( $item, $column_name ) {
 		switch ( $column_name ) {
 
 			case 'type' :
-				$value = ( 'primary' === $item['type'] )
-					? esc_html_e( 'Primary',   'easy-digital-downloads' )
-					: esc_html_e( 'Secondary', 'easy-digital-downloads' );
+				$value = ucwords( edd_get_address_type_label( $item['type'] ) );
+				if ( ! empty( $item['is_primary'] ) ) {
+					$value .= ' <span class="edd-chip">git ' . __( 'Primary', 'easy-digital-downloads' ) . '</span>';
```
spencerfinnell, line 71:
```suggestion
					$value .= ' <span class="edd-chip">' . esc_html__( 'Primary', 'easy-digital-downloads' ) . '</span>';
```

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -581,8 +591,11 @@ public static function orders( $data = null ) {
 		unset( $customer_address_data['first_name'] );
 		unset( $customer_address_data['last_name'] );
 
-		// Maybe add address to customer record.
-		edd_maybe_add_customer_address( $customer_id, $customer_address_data );
+		// Maybe add address to customer record, if it is not empty.
+		$customer_address_data = array_filter( $customer_address_data );
```
spencerfinnell, line 595: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7906/files#diff-15bee25f82f2eaea59129e303e1bbf13R643 will take care of this a little later down the pipe, so I don't think this extra check will be required going forward.

Some discussion related to that is here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7895#issuecomment-629185205

robincornett:
When an order is created with no physical address, an address record is still being added to the customer record. This is happening during migration as well.

Looking through customer records, it looks like the addresses are generally not being duplicated, but sometimes they are. For one I can pinpoint specifically, it's because the `name` in the address array is different, but everything else is the same. My thought on that is that it's confusing, because the `name` is not displayed with the address in any of the customer tables, so the address appears to be duplicate information, even though technically it is different in the database.

Note: a PR which was attempting to deal with empty addresses has been merged into 3.0, but no longer works because of additional data (date) being added to the address data. There is a subsequent open PR which corrects that and prevents empty addresses being added to the database (#7906), which ignores the `name` parameter, at least for now, so if the address/city/region/country/postal code all match, the new address is not added.

Should this PR be updated with the latest changes in 3.0? I ran a migration from just this PR, and then one with 3.0 merged in, and there are differences.

includes/admin/customers/class-customer-addresses-table.php
spencerfinnell, line 71: It also appears `$item['is_primary']` will never not be empty because it's not added to the normalized data here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/4bccc655e12c874ff20e923305c9c2f9a15977ab/includes/admin/customers/class-customer-addresses-table.php#L365-L378

ashleyfae: Confirmed that `Primary` isn't showing up in the UI for a primary address.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -27,9 +27,10 @@ class Data_Migrator {
 	 *
 	 * @since 3.0
 	 *
-	 * @param object $data Data to migrate.
+	 * @param object $data       Data to migrate.
+	 * @param string $type       The type of address this is.
 	 */
-	public static function customer_addresses( $data = null ) {
+	public static function customer_addresses( $data = null, $type = 'billing' ) {
```
ashleyfae, line 33: Do we need the `$type` parameter given that it's always `billing` at this point?

cklosowski: @ashleyfae I'd like to include it so that when we write the upgrade routine for the Simple Shipping migration, we can hook in and use this same method to handle things.

robincornett:
@cklosowski should #7906 be finalized before this, or rolled into this?

cklosowski:
@robincornett lets go ahead and finalize #7906, and then we can merge this in.

cklosowski:
@robincornett Ok I've merged in #7906 into `release/3.0` and then updated this branch `issue/7254` from `release/3.0`.

Let me know if we're good to merge here.

tests/tests-formatting.php
```diff
@@ -110,4 +110,12 @@ public function test_decimal_filter_with_currency_passed_should_return_0() {
 
 		$this->assertSame( 0, edd_currency_decimal_filter( 2, 'JPY' ) );
 	}
+
+	public function test_address_type_label_billing() {
```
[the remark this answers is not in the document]
cklosowski: @robincornett Tests don't need docblocks if they aren't testing a class. When testing a class the docblock is needed in order for it to trigger the code coverage report, but tests that only test singular functions don't require it.

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -134,8 +135,8 @@ public function column_address( $item ) {
 			'view' => '<a href="' . esc_url( $customer_url ) . '">' . __( 'View', 'easy-digital-downloads' ) . '</a>'
 		);
 
-		// Non-primary email actions
-		if ( 'primary' !== $item_status ) {
+		// Non-primary address actions
+		if ( empty( $item['is_primary'] ) ) {
 			$actions['delete'] = '<a href="' . admin_url( 'edit.php?post_type=download&page=edd-customers&view=delete&id=' . $item['id'] ) . '">' . __( 'Delete', 'easy-digital-downloads' ) . '</a>';
```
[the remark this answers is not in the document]
cklosowski: Correct, we should fix that up in another issue, since we mentioned that we'll need to do quite a few UI updates to the linking.

cklosowski:
Ready for your next look over @robincornett

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -66,9 +66,10 @@ public function column_default( $item, $column_name ) {
 		switch ( $column_name ) {
 
 			case 'type' :
-				$value = ( 'primary' === $item['type'] )
-					? esc_html_e( 'Primary',   'easy-digital-downloads' )
-					: esc_html_e( 'Secondary', 'easy-digital-downloads' );
+				$value = ucwords( edd_get_address_type_label( $item['type'] ) );
```
robincornett, line 69: Just to clarify: as I read the code, if the address type is something other than `billing`, but the type has not been registered with the `edd_address_type_labels` filter, it will still show as "Billing" here. Is that the desired behavior?

Asking because in testing another issue, I added an address to the database with a different `type`, but it still displays as "Billing".

cklosowski: Fixed that up @robincornett. I've set a fallback to just use the label that was passed in if it it's registered. If it is, it'll use it.

robincornett: Yes, it used the one I hard coded in.

cklosowski:
Sweet thanks, once approved @robincornett I'll get this merged if you are good then.

spencerfinnell:
@cklosowski It looks like there are still a few places passing `primary` as `type`, such as:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/bee5562ad0205fec6b5ef333815f42e6bbdd55b3/includes/shortcodes.php#L905-L9087

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/bee5562ad0205fec6b5ef333815f42e6bbdd55b3/includes/compat/class-customer.php#L289-L292

Should we add any validation to what is allowed or is the plan to keep the column flexible for extensions to use any way they like?

robincornett:
@cklosowski outside of a migration, how does a customer address become designated as primary?

I just placed a manual order for a new customer, and the address was added to the database, but not marked as primary.

Note: manually created orders with addresses do not appear to populate the `name` column for the address, which is likely an oversight from earlier.

cklosowski:
@robincornett Right now we don't have a clear designation of that yet. The is primarily a future proofing preset that will help with address improvements later on, like the shipping integration.

cklosowski:
@spencerfinnell fixed those up.

includes/admin/customers/class-customer-addresses-table.php
robincornett, line 69: This has just come up on a support ticket--not all languages handle capitalization the same way (German is a specific example). We should not be forcing upper/lower case on translatable strings.

includes/formatting.php
```diff
@@ -387,3 +387,39 @@ function edd_get_allowed_tags() {
...
+	 *     Array of the address type labels, in key/value form. The key should match the database entry for the
+	 *         wp_edd_customer_addresses table in the 'type' column. The value of each array entry should be a translatable
+	 *         string for output in the UI.
+	 */
+	$address_type_labels = apply_filters( 'edd_address_type_labels', $address_type_labels );
+
+	// Fallback to just applying an upper case to any words not in the filter.
+	return array_key_exists( $address_type, $address_type_labels ) ?
+		$address_type_labels[ $address_type ] :
+		ucwords( $address_type );
```
robincornett, line 423: I don't think my translation comment applies here, as `$address_type` as the fallback would not be a translatable string, but if I'm wrong on that, then this is also forcing capitalization.

PR #7875, opened 2020-05-12
 edd-item-tab-wrapper rule changed to wildcard instead of div

### LINKED ISSUE #7866: 3.0 - Backwards Compatible CSS Issue for #edd-item-card-wrapper
### Expected behavior
When I visit an admin screen with legacy wrapper markup, I expect it to look "correct".

### Actual behavior
PR #7095 partially addressed this, but the CSS appears to no longer be quite valid, specifically to add padding to the `.item-section` div. Example from Recurring:

image

1) In EDD 3.0, visit a subscription screen (EDD Recurring) or add a new commission (EDD Commissions)
2) There is no padding around the `.item-section` containers.

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

I think the issue is in this CSS, as it's pretty specific:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/8acae18587802b10b685a73cf0b074792b4458b4/assets/css/edd-admin.css#L3808-L3813

I believe that CSS is targeting an `#edd-item-card-wrapper` immediately following a `div` which is not `#edd-item-tab-wrapper`, but in both Commissions and Recurring, the card wrapper follows an `h2` (which should actually be an `h1` in both cases, as they are the page title), so no padding is applied (other CSS is missing, but the padding is the most noticeable).

LisaCee:
Fixes #7866 

Proposed Changes:
1. Changed `#edd-item-tab-wrapper` to use `*` instead of `div` to add more consistent padding

I'm still working on testing each instance of this `ids` use, but you all can probably find any issues before me anyway, since I haven't yet used `commissions` or `usage-tracking` yet.

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -3813,7 +3813,7 @@ body.download_page_edd-reports {
 	box-sizing: border-box;
 }
 
-div:not(#edd-item-tab-wrapper) + #edd-item-card-wrapper .item-section { /** [1] */
+*:not(#edd-item-tab-wrapper) + #edd-item-card-wrapper .item-section { /** [1] */
 	margin: 25px 0;
 	padding: 20px;
 	border: 1px solid #e5e5e5;


spencerfinnell:
@robincornett Unfortunately like most markup used throughout the codebase there isn't any documentation for ensuring consistency. I think we spoke before about how having these "components" documented somewhere like the Wiki would be nice. Perhaps we can open a discussion in Basecamp about the best method to start documenting things like this?

PR #7895, opened 2020-05-13
Prevent empty addresses from being added to the database

### LINKED ISSUE #7814: 3.0 - Empty Physical Addresses Are Added to the Database
### Expected behavior
If my storefront is not collecting user address information, I expect that address records will not be created for them.

### Actual behavior
Generally, when a new customer is added to the database, an empty address record is created. Have checked this with three different methods of creating an order:
* shopping cart/front end: no address is collected, but an address is added to the customer record
* wp-cli: an empty address is added to the customer record/database
* admin add order: no address is collected and no address is added to the customer record (if a new customer is created and an address is provided, that is saved correctly)

1) disable taxes (and anything else which may require a physical address)
2) create some orders through the standard shopping cart, or using wp-cli
3) check the new customer records (or visit `/wp-admin/edit.php?post_type=download&page=edd-customers&page_type=physical`) and see empty address records

PHP Version: 7.3.5
WordPress Version: 5.4.1

potentially related: #7254, [comment on 7464](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7464#issuecomment-622017779)

robincornett:
Fixes #7814

### DIFF
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -623,6 +623,13 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
 		return false;
 	}
 
+	// Temporarily remove the name so we can see if the address is empty.
+	$name = $data['name'];
+	unset( $data['name'] );
+	if ( empty( $data ) ) {
+		return;
+	}
+	$data['name']        = $name;
 	$data['customer_id'] = $customer_id;
 
 	$c = edd_count_customer_addresses( $data );
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -627,7 +627,7 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
 	$name = $data['name'];
 	unset( $data['name'] );
 	if ( empty( $data ) ) {
-		return;
+		return false;
 	}
 	$data['name']        = $name;
 	$data['customer_id'] = $customer_id;
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -626,6 +626,7 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
 	// Temporarily remove the name so we can see if the address is empty.
 	$name = $data['name'];
 	unset( $data['name'] );
+	$data = array_filter( $data );
 	if ( empty( $data ) ) {
 		return false;
 	}


includes/customer-functions.php
```diff
@@ -623,6 +623,13 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
 		return false;
 	}
 
+	// Temporarily remove the name so we can see if the address is empty.
+	$name = $data['name'];
+	unset( $data['name'] );
+	if ( empty( $data ) ) {
+		return;
```
ashleyfae, line 630:
```suggestion
		return false;
```

The function expects:

```
@return int|false ID of the insert customer address. False otherwise.
```

includes/customer-functions.php
```diff
@@ -623,6 +623,13 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
 		return false;
 	}
 
+	// Temporarily remove the name so we can see if the address is empty.
+	$name = $data['name'];
+	unset( $data['name'] );
```
ashleyfae, line 628:
```suggestion
	unset( $data['name'] );
	$data = array_filter( $data );
```

This will remove any empty values that have been passed through. Without this, the address would still be submitted if empty values were provided like this:

```php
$address = edd_maybe_add_customer_address( 1, array(
	'address' => '',
	'address2' => '',
	'country' => '',
	'region' => ''
) );
```
`array_filter()` removes any empty keys. So if they're all empty strings like that, we'll end up with a truly empty array and then the address won't be saved.

ashleyfae:
I've been thinking about the name removal part of this PR and I'm wondering if we should move that logic to `edd_build_order()`.

So `edd_maybe_add_customer_address()` would keep this part:

```php
$data = array_filter( $data );
if ( empty( $data ) ) {
	return false;
}
```

And we'd move the temporary name stripping stuff to here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/functions/orders.php#L782 (remove first name, run `array_filter()`, if it's not empty add first name back in, and then run `edd_maybe_add_customer_address()`). To me that's just feeling like a better place for it. What do you think @robincornett ?

robincornett:
I think the issue with tucking it in there is that the `edd_maybe_add_customer_address` is invoked in more places than just building the order. It's also in the [data migrator class](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L585) and the [manual order completion](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/functions/actions.php#L205).

Also, the `name` is part of the address, so I think removing it elsewhere but trying to check an empty array here may not work, as the `name` will be set. You would have to move the empty logic with the `name` piece as well, wouldn't you?

PR #7899, opened 2020-05-14
Update Refund UX/UI

### LINKED ISSUE #7778: 3.0 - Orders - Refund UX/UI
As requested, moving this comment to a new issue based on some things I've encountered with #7599. Running into some JS issues, I think:

- if I click the "Issue Refund" button, a modal pops up. If I hit `esc` or click the X to close the modal, the order screen reloads, even though I haven't made any changes. If I have made a different change to the order, such as changing the linked customer, that change is lost.
- if I click the `disabled` "Issue Refund" button in the modal, without selecting any order items, the refund is processed and I get a PHP warning:
```
PHP Warning:  array_map(): Expected parameter 2 to be an array, null given in C:\Users\robin\Documents\Websites\plugins\easy-digital-downloads.30\includes\admin\payments\actions.php on line 675
```
- should a refund process before I click the "Save Order" button? I'm sure it would be more complicated, but it might be nice to select the items for the refund, and then somehow a notice could show somewhere that a refund is pending, so there is an additional confirmation step?
- when I click on the overlay, the modal does not close--the add media overlay does and I'm used to being able to close the modal with a sloppy click
- if the focus is somehow outside of the modal, the "esc" key no longer works to close it
- I'm not sure, but I think it would be preferable for the initial focus in the modal to be one of the inputs, not the X (but tabbing gets me straight to the checkboxes so maybe less of an issue)
- on the add order screen, I have some of the same general issues with the modals there, but the page does not reload when I close the modal
- I noticed that I could not click the "submit" button on a discount modal if I had not selected one, but that appeared to be an HTML validation catch, not JS. But preventing the modal closing if the button is disabled would be helpful
- the `edd-submit-refund-submit` button in the modal should be a `<button>`, not an `<a>` please (this may help with the disabled issue but I am not sure)
- minor detail, but if the empty `<td>` in the submit button row had a `colspan` of 4 instead of five, and then the button wrapper td had its at two, and CSS to `text-align:right;`, it would look correct even on smaller screens (or with the text alignment, the empty cell might not be necessary)

I know there is a Refund UI issue open, but I think the work is going on here, so if I should report this elsewhere, please let me know.

_Originally posted by @robincornett in https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7599#issuecomment-618557944_

robincornett:
Fixes #7778

Proposed Changes:
1. Change "Issue Refund" button to a button
2. Keep "Submit Refund" button disabled until there are items to refund
3. Allow free products to be refunded


includes/admin/payments/class-refund-items-table.php
```diff
@@ -459,13 +459,12 @@ public function display_rows() {
 			</td>
 		</tr>
 		<tr id="edd-refund-submit-total" class="edd-refund-submit-line-total">
-			<td colspan="<?php echo $this->get_column_count() - 1; ?>"></td>
 
-			<td id="edd-refund-submit-button-wrapper">
-				<a id="edd-submit-refund-submit" class="disabled button-primary"><?php _e( 'Submit Refund', 'easy-digital-downloads' ); ?></a>
+			<td id="edd-refund-submit-button-wrapper" colspan="<?php echo esc_attr( $this->get_column_count() ); ?>">
```
spencerfinnell, line 463: I think we should keep the two columns and then remove the `text-align: right;` CSS to better align the submit button. It _can_ align on larger screens but gets out of place on smaller ones with a single column:

Screen Shot 2020-05-15 at 8 49 05 AM

robincornett:
@spencerfinnell instead of hiding the tablenav, could we prevent it being output in the PHP ([top](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/payments/class-refund-items-table.php#L352), [bottom](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/payments/class-refund-items-table.php#L372))? I believe this class is just for the modal, not any other display.

spencerfinnell:
@robincornett Much better idea.

includes/admin/payments/class-refund-items-table.php
robincornett: The `text-align:right;` is actually for the rows above (Refund Subtotal, Total, and taxes) and doesn't affect the button. I think one issue with the button alignment is that the spinner is hidden with `visibility:hidden;` which means it still takes up physical space in the table cell. Also, with the empty cell added back in and the button wrapper taking up one column, I end up with this:

image

(Note: I do have my browser zoom set to 125%, and this is not my full screen, but it is over 1200px wide. It is worse at smaller screen sizes.)

Another option might be to keep the two cells, but set the submit button to `colspan="2"` and adjust accordingly, so the button has more potential space to occupy.

spencerfinnell: @robincornett What about keeping the bottom tablenav and moving the button there, out of the table?

robincornett: @spencerfinnell to do that, would I reintroduce the `display_tablenav` method and just override it within this class?

spencerfinnell: @robincornett Yes, override `display_tablenav` in the Refunds list table. 

If `$which` is `top`, show nothing. Otherwise show the submit button.

https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-list-table.php#L1229-L1255

robincornett: Sounds good. I've got it moved but doing that has broken the nonce validation, so I may need help figuring that out.

robincornett:
@spencerfinnell I am struggling a bit with the button labels:
- "Issue Refund" feels vague, and I'm not sure if I am refunding the entire order or what when I click that. In reality, I'm initiating a refund process, so this is really step one. Could we consider changing this to "Initiate Refund" or "Begin Refund" or something similar?
- "Submit Refund" isn't vague, but with the modal, I confess that I was surprised the first time I clicked it and realized that I had completed the refund process completely--somehow, I had expected that there would still be a final "update order" button to click which would finalize the refund. I think I mentioned this in my initial issue, but instead of going with my original thought of adding an extra stage to the process, could we maybe change this button label to something like "Finalize Refund" or make _this_ one say "Issue Refund", or possibly add a description before the button saying "clicking this button will issue the refund automatically", so that it's clear this is really the final step in returning someone's money?

spencerfinnell:
@robincornett I'm not sure "Begin Refund" feels _actionable_ enough. 

If we use the same terminology of creating an Order (`sale`, technically the same level of record as a `refund`), it would be:

Add Refund -> Create Refund

But I don't know if I love "Add Refund" in the context of the Order overview area. It makes it seem like it's not necessarily related to the current Order. What about:

Initialize Refund -> Create Refund

includes/admin/payments/class-refund-items-table.php
```diff
@@ -349,6 +349,7 @@ public function single_row( $item ) {
 	public function display() {
 		$singular = $this->_args['singular'];
 
+		wp_nonce_field( 'bulk-' . $this->_args['plural'] );
```
spencerfinnell, line 352: @robincornett Since we are adding this manually now lets create a nonce that is specific to the refund action.

spencerfinnell:
@robincornett It looks like there is some improper `table` nesting:

Screen Shot 2020-05-15 at 3 14 42 PM

The list table class is outputting a `<table>` but so is `edd_ajax_generate_refund_form`

robincornett:
Thanks, I had wondered how the nonce field was outputting a `table` at all. Turns out it wasn't.

robincornett:
Note: currently trying to sort out focus issues when the ajax function is complete.

robincornett:
@spencerfinnell I've noticed this extra space in other code locations as well, and it appears to be because the markup is created by switching between HTML and PHP, or maybe because the `echo` usages are in separate places. For example, the current subtotal output is this:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb020f7b062c2d68d756fc749ac078c54906b973/includes/admin/payments/class-refund-items-table.php#L420-L435

which does create the visual extra space, and the change when an item is selected. If I change the markup to something like this instead:

```php
                $currency_symbol_output = sprintf( '<span>%s</span>', $currency_symbol );
                $before                 = 'before' === $currency_position ? $currency_symbol_output : '';
                $after                  = 'after' === $currency_position ? $currency_symbol_output : '';
                $subtotal               = edd_format_amount( 0.00 );
                printf(
                    '%1$s<span id="edd-refund-submit-subtotal-amount" data-refund-subtotal="%2$s">%2$s</span>%3$s',
                    $before,
                    esc_attr( $subtotal ),
                    $after
                );
```
the resulting output has no space between the subtotal and the currency.

So changing this would be totally doable, and I think would fix the issue, but it's a departure from how I have generally seen the markup done.

Another option would be to use something like this, which is in the customer details screen for the orders table:

```php
echo edd_currency_filter( edd_format_amount( '0.00' ), $order->currency );
```

but that would remove the `<span>` tags around the currency/amounts here, and that may break the subtotal/etc dynamic update.

What would you recommend for approaching this, or am I misunderstanding where/how that space is being generated?

spencerfinnell:
@robincornett I like using the `print()` style for this case. 

This will eventually be rewritten to use the same techniques as the Order Overview which will make this a non-issue, so a quick fix here is fine.

robincornett:
@spencerfinnell while I am looking at this markup, I'm seeing that the PHP adds `data-refund-*` attributes to the spans, and it looks like the attributes are maybe supposed to be updated in the script, but the attributes don't match:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb020f7b062c2d68d756fc749ac078c54906b973/assets/js/admin/orders/order-overview/_refund.js#L87-L89

The attributes do not update when the amounts do, but evidently it hasn't impacted functionality. But should the JS be updated to match, and I think use `.attr()` instead of `.data()` there?

spencerfinnell:
@robincornett if they aren't used (I don't believe they are), let's remove them. `edd_ajax_process_refund_form()` only needs a list of Order Item IDs, not any totals.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/eb020f7b062c2d68d756fc749ac078c54906b973/assets/js/admin/orders/order-overview/_refund.js#L139-L147

robincornett:
> Out of curiosity, why allow free products to be refunded?

Great question! The scenario I can think of is when Software Licensing is involved--refunding a product disables the license automatically. The license can be disabled/revoked by hand, but I think keeping a record in the order as well is helpful.

Also, I believe if an entire order is refunded, the free products would be refunded as part of that, so if we allow them to be refunded as part of a larger group, I think we need to allow them to be refunded individually.

robincornett:
I just figured out how to close the dialog with a click on the overlay. Since I have to pull in from release/3.0 again anyway, @spencerfinnell, should I go ahead and add this in? It appears to just be this in the _refund.js file:

```js
$( document.body ).on( 'click', '.ui-widget-overlay', function ( e ) {
	$( '#edd-refund-order-dialog' ).dialog( 'close' );
} );
```

spencerfinnell:
@robincornett Yep, you can add it here.

robincornett:
Added and re-requesting reviews; thank you!

PR #7901, opened 2020-05-14
Fix deleting customer emails & addresses via admin tables #7825

### LINKED ISSUE #7825: 3.0 - Clicking the "delete" link on addresses tables returns an error
### Expected behavior
When I click a link to delete a record, I expect for it to be deleted.

### Actual behavior
In the customer list tables for both physical and email addresses, clicking the "Delete" link in the action links results in:

> Error: Invalid Customer ID Provided.

and the record is not deleted.

1) visit Downloads > Customers
2) switch to either Addresses or Email Addresses
3) click the "Delete" link for an address record

PHP Version: 7.3.5
WordPress Version: 5.4.1

Related: #7814, #7762

Removing an email address within an individual customer screen does work, so fixing that piece may be relatively straightforward. It is currently not possible to delete a physical address in either location.

ashleyfae:
Fixes #7825

Proposed Changes:
1. Correct URL used for deleting customer email addresses.
2. Introduce function for deleting customer addresses.
3. Correct URL used for deleting customer addresses to use new function.
4. Introduce new admin notices for successfully/unsuccessfully removing addresses.
5. Add a "Delete" link when hovering over an individual address on Customer Details page. This isn't quite consistent with the Customer Emails table, which has an "Actions" column, but given the number of columns already present on the physical addresses table I though this was better.


cklosowski (dismissed):
@ashleyfae this needs to _not_ allow immediately deleting what is considered the primary email address of a customer.

When viewing the customer details, the 'primary' email address cannot be deleted, only tertiary email addresses. The same should hold true for the list table display of the 'delete' action.

includes/admin/customers/class-customer-email-addresses-table.php
```diff
@@ -131,7 +131,14 @@ public function column_email( $item ) {
 
 		// Non-primary email actions
 		if ( 'primary' !== $item_status ) {
```
cklosowski, line 133: @ashleyfae I believe the issue I had with being able to delete a primary is that we're checking 'status', not 'type'.

ashleyfae:
@cklosowski Updated. Primary addresses now have no row action and the checkbox is disabled. I disabled the whole checkbox because our only bulk action is delete anyway.

Turns out bulk actions weren't working at all due to the nonce check failing. This has now been fixed.

Something to note related to that:

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7901/commits/18d117faed77b146e5a75d0927c16adb15456e75#diff-fc382348ba7ddea17a07f25d727c9abeL34-L35

These were being passed in via `__()` however I do not believe these arguments are intended for display/translation. They're used when creating the nonce action and in various HTML attributes (like CSS classes). Therefore we should always know exactly what they are (rather than being dynamic due to being translated). So those are now lowercase and hard-coded with that in mind.

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -31,8 +31,8 @@ class EDD_Customer_Addresses_Table extends List_Table {
 	 */
 	public function __construct() {
 		parent::__construct( array(
-			'singular' => __( 'Address',   'easy-digital-downloads' ),
-			'plural'   => __( 'Addresses', 'easy-digital-downloads' ),
+			'singular' => 'address',
```
robincornett, line 34: Are these intentionally no longer translatable strings (same for emails)? I thought these were label definitions.

ashleyfae:
> > Add a "Delete" link when hovering over an individual address on Customer Details page. This isn't quite consistent with the Customer Emails table, which has an "Actions" column, but given the number of columns already present on the physical addresses table I though this was better.
> 
> @ashleyfae I like the row actions instead of the columns. Should we update "Remove" action in "Customer Emails" to use the same `Delete` terminology? I think it helps convey the destruction more, especially with no confirmation.

I'm happy with that. Also if we'll be removing the `Actions` column we could use this opportunity to add a new column for `Date Added`.

spencerfinnell:
@ashleyfae `Date Added` sounds good too. Up to you if you want to change just the string for now and tackle the columns in the future, or if you think it makes sense to do it now.

cklosowski:
:+1: to adding `Date Added` if we're removing a column.

cklosowski:
@LisaCee do you think you can get a column added to this? @spencerfinnell, if needed can you point her in the right dirction?

ashleyfae:
1. Moved Customer Email actions to row actions.
2. Added new column for Date Added.
3. Fixed an undefined `customer_id` variable error when you leave the `Email Address` field blank but click `Add Email` anyway.

includes/admin/customers/customers.php
```diff
@@ -763,7 +769,7 @@ function edd_customers_view( $customer = null ) {
 			<thead>
 				<tr>
 					<th><?php esc_html_e( 'Email',   'easy-digital-downloads' ); ?></th>
-					<th><?php esc_html_e( 'Actions', 'easy-digital-downloads' ); ?></th>
+					<th><?php esc_html_e( 'Date Added', 'easy-digital-downloads' ); ?></th>
```
robincornett, line 772: The physical addresses table uses "First Used" for the column header. It feels like the data is a similar type and maybe we should use the same for each?

includes/admin/customers/customers.php
```diff
@@ -773,23 +779,25 @@ function edd_customers_view( $customer = null ) {
...
-							<span><?php echo esc_html( $email ); ?></span>
+							<span><?php echo esc_html( $email->email ); ?></span>
 
-							<?php if ( 'primary' === $key ) : ?>
+							<?php if ( 'primary' === $email->type ) : ?>
 								<span class="edd-chip"><?php esc_html_e( 'Primary', 'easy-digital-downloads' ); ?></span>
+							<?php else : ?>
+								<div class="row-actions">
+									<?php
+									$base_url    = admin_url( 'edit.php?post_type=download&page=edd-customers&view=overview&id=' . urlencode( $customer->id ) );
```
robincornett, line 789: Not a big deal either way, but could this use `add_query_arg` instead? I find it easier to parse mentally, or actually maybe just use the `edd_get_admin_url` function to build this base URL.

includes/admin/customers/class-customer-email-addresses-table.php
```diff
@@ -289,10 +302,22 @@ public function process_bulk_action() {
 			$ids = array( $ids );
 		}
 
+		/*
+		 * Only non-primary email addresses can be deleted, so we're building up a whitelist using the provided
```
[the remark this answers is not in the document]
ashleyfae: Good catch - thank you.

includes/admin/customers/customers.php
LisaCee: I like this suggestion.  I don't care which label we use, but consistency is nice.

cklosowski: I don't mind that they are different as the context is different. One is an object, the other is a property of an object than can have multiple values related to the context (email/physical address).

includes/admin/customers/customers.php
```diff
@@ -708,6 +711,9 @@ function edd_customers_view( $customer = null ) {
 								? esc_html( $address->address2 )
 								: '';
 							?>
+							<div class="row-actions">
```
robincornett, line 714: Adding the row actions does interesting things on mobile view because no cells are hidden:
image
(If the row/cell is not clicked/active, then there is just the gap.)

includes/admin/customers/customers.php
```diff
@@ -768,7 +774,7 @@ function edd_customers_view( $customer = null ) {
 			<thead>
 				<tr>
 					<th><?php esc_html_e( 'Email',   'easy-digital-downloads' ); ?></th>
-					<th><?php esc_html_e( 'Actions', 'easy-digital-downloads' ); ?></th>
+					<th><?php esc_html_e( 'Date Added', 'easy-digital-downloads' ); ?></th>
```
robincornett, line 777: The CSS for this table also needs to be updated for responsiveness, with the new column being added.

cklosowski:
@robincornett Do you mind taking this over so @ashleyfae can focus on the other tasks she has currently?

includes/admin/customers/class-customer-email-addresses-table.php
```diff
@@ -204,10 +211,16 @@ public function column_customer( $item ) {
 	 * @return string Displays a checkbox
 	 */
 	public function column_cb( $item ) {
+		$is_primary         = ! empty( $item['type'] ) && 'primary' === $item['type'];
+		$primary_attributes = $is_primary ? ' disabled' : '';
+		$title              = $is_primary ? __( 'Primary email addresses cannot be deleted.', 'easy-digital-downloads' ) : '';
+
 		return sprintf(
-			'<input type="checkbox" name="%1$s[]" value="%2$s" />',
+			'<input type="checkbox" name="%1$s[]" value="%2$s" title="%3$s"%4$s />',
```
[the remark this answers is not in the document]
robincornett: @cklosowski I think I've addressed the rest of the PR issues, but do you want me to go ahead and fix this one here, or wait for the other PR (which would need to be updated to account for this change either way)?

includes/admin/customers/customers.php
```diff
@@ -709,39 +712,42 @@ function edd_customers_view( $customer = null ) {
 								: '';
 							?>
 						</td>
-						<td>
+						<td data-colname="<?php esc_attr_e( 'City', 'easy-digital-downloads' ); ?>">
 							<?php
 							echo ! empty( $address->city )
-								? esc_html( $address->city )
+								? esc_attr( $address->city )
```
spencerfinnell, line 718: This should be able to stay `esc_html()`, I think? The other columns have changed as well except for Postal Code.

includes/admin/customers/customers.php
```diff
@@ -777,43 +783,58 @@ function edd_customers_view( $customer = null ) {
...
+											'page' => 'edd-customers',
+											'view' => 'overview',
+											'id'   => urlencode( $customer->id ),
+										)
+									);
+									$promote_url = wp_nonce_url( add_query_arg( array( 'email' => rawurlencode( $email->email ), 'edd_action' => 'customer-primary-email' ), $base_url ), 'edd-set-customer-primary-email' );
+									$remove_url  = wp_nonce_url( add_query_arg( array( 'email' => rawurlencode( $email->email ), 'edd_action' => 'customer-remove-email'  ), $base_url ), 'edd-remove-customer-email' );
+									?>
+									<a href="<?php echo esc_url( $promote_url ); ?>"><?php esc_html_e( 'Make Primary', 'easy-digital-downloads' ); ?></a>
+									&nbsp;|&nbsp;
```
spencerfinnell, line 805: Small thing -- the extra whitespace makes these a little inconsistent with core row actions:

Screen Shot 2020-07-10 at 1 33 16 PM

includes/admin/class-edd-notices.php
```diff
@@ -660,6 +660,19 @@ private function add_user_action_notices( $notice = '' ) {
 						'class'   => 'error'
 					) );
 					break;
+				case 'address-removed' :
+					$this->add_notice( array(
+						'id'      => 'edd-customer-address-removed',
+						'message' => __( 'Customer address removed', 'easy-digital-downloads')
```
spencerfinnell, line 666:
```suggestion
						'message' => __( 'Customer address deleted.', 'easy-digital-downloads')
```

`deleted` matches the "Delete" action link.

includes/admin/class-edd-notices.php
```diff
@@ -660,6 +660,19 @@ private function add_user_action_notices( $notice = '' ) {
...
+					) );
+					break;
+				case 'address-remove-failed' :
+					$this->add_notice( array(
+						'id'      => 'edd-customer-address-remove-failed',
+						'message' => __( 'Failed to remove customer address', 'easy-digital-downloads'),
```
spencerfinnell, line 672:
```suggestion
						'message' => __( 'Failed to delete customer address.', 'easy-digital-downloads'),
```

includes/admin/customers/customers.php
```diff
@@ -709,39 +712,42 @@ function edd_customers_view( $customer = null ) {
...
+						<td class="has-row-actions" data-colname="<?php esc_attr_e( 'First Used', 'easy-digital-downloads' ); ?>">
+							<time datetime="<?php echo esc_attr( EDD()->utils->date( $address->date_created, null, true )->toDateTimeString() ); ?>"><?php echo edd_date_i18n( EDD()->utils->date( $address->date_created, null, true )->toDateTimeString(), 'M. d, Y' ) . '<br>' . edd_date_i18n( strtotime( $address->date_created ), 'H:i' ) . ' ' . edd_get_timezone_abbr(); ?></time>
 							<?php if ( ! empty( $address->is_primary ) ) : ?>
 								<span class="edd-chip"><?php esc_html_e( 'Primary', 'easy-digital-downloads' ); ?></span>
 							<?php endif; ?>
+							<div class="row-actions">
+								<span class="delete"><a href="<?php echo esc_url( $delete_url ); ?>"><?php esc_html_e( 'Delete', 'easy-digital-downloads' ); ?></a></span>
+							</div>
```
spencerfinnell, line 750: I think this makes more sense in the "Address" column. This matches core row actions always being in the first column as well as where the row actions are when viewing the "Physical Addresses" tab.

robincornett: I agree with that, except for two concerns:
1. It looks/feels weird on small screens, because the row actions are after the street address and before everything else
2. Because the street address can be empty, having the row actions there presents potential accessibility issues. Generally, WP has the row actions in the first column because that's a title column, and pretty self explanatory. This is less so, because the first column isn't necessarily informative on its own, even if the street address is populated.

I ended up setting up the physical addresses table to perform somewhat like the tax rates table (or WP Core tables) for responsive behavior, but I'm not sure its data "behaves" the same way.

spencerfinnell: 👍 I think I'm more okay with this now. I was thinking maybe matching the global Physical Addresses table more by adding the City and Postal Code beneath the Street Address... but now I actually think doing a partially formatted version of the address like that might be more confusing than having things in separate columns (something to think about for the global table at a later date).

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -135,9 +135,14 @@ public function column_address( $item ) {
 			'view' => '<a href="' . esc_url( $customer_url ) . '">' . __( 'View', 'easy-digital-downloads' ) . '</a>'
```
spencerfinnell, line 135:
```suggestion
			'view' => '<a href="' . esc_url( $customer_url ) . '">' . esc_html__( 'View', 'easy-digital-downloads' ) . '</a>'
```

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -135,9 +135,14 @@ public function column_address( $item ) {
...
-		// Non-primary address actions
 		if ( empty( $item['is_primary'] ) ) {
-			$actions['delete'] = '<a href="' . admin_url( 'edit.php?post_type=download&page=edd-customers&view=delete&id=' . $item['id'] ) . '">' . __( 'Delete', 'easy-digital-downloads' ) . '</a>';
+			$delete_url = wp_nonce_url( edd_get_admin_url( array(
+				'page'       => 'edd-customers',
+				'view'       => 'overview',
+				'id'         => urlencode( $item['id'] ),
+				'edd_action' => 'customer-remove-address'
+			) ), 'edd-remove-customer-address' );
+			$actions['delete'] = '<a href="' . esc_url( $delete_url ) . '">' . __( 'Delete', 'easy-digital-downloads' ) . '</a>';
```
spencerfinnell, line 145:
```suggestion
			$actions['delete'] = '<a href="' . esc_url( $delete_url ) . '">' . esc_html__( 'Delete', 'easy-digital-downloads' ) . '</a>';
```

includes/admin/customers/class-customer-email-addresses-table.php
```diff
@@ -130,8 +130,15 @@ public function column_email( $item ) {
...
-			$actions['delete'] = '<a href="' . admin_url( 'edit.php?post_type=download&page=edd-customers&view=delete&id=' . $item['id'] ) . '">' . __( 'Delete', 'easy-digital-downloads' ) . '</a>';
+		if ( empty( $item['type'] ) || 'primary' !== $item['type'] ) {
+			$delete_url = wp_nonce_url( edd_get_admin_url( array(
+				'page'       => 'edd-customers',
+				'view'       => 'overview',
+				'id'         => urlencode( $customer_id ),
+				'email'      => rawurlencode( $email ),
+				'edd_action' => 'customer-remove-email'
+			) ), 'edd-remove-customer-email' );
+			$actions['delete'] = '<a href="' . esc_url( $delete_url ) . '">' . __( 'Delete', 'easy-digital-downloads' ) . '</a>';
```
spencerfinnell, line 141:
```suggestion
			$actions['delete'] = '<a href="' . esc_url( $delete_url ) . '">' . esc_html__( 'Delete', 'easy-digital-downloads' ) . '</a>';
```

robincornett (approved):

PR #7906, opened 2020-05-14
Prevent duplicate addresses being added to the database

### LINKED ISSUE #7873: 3.0 - Physical Addresses May Be Added Multiple Times During Migration
### Expected behavior
When I migrate data from 2.9 to 3.0, I expect physical addresses not to be duplicated.

### Actual behavior
When I migrate data, a new physical address record is created for each order, which can result in multiple duplicates for a long time customer.

1) Migrate a storefront with physical addresses to 3.0.
2) Check address records for customers with multiple orders.

PHP Version: 7.4.1
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

Related: #7822, #7814 (empty addresses are also added multiple times)

robincornett:
Fixes #7873
Should be merged after PR #7895 (this is branched off of that)

Proposed Changes:
Change the `edd_maybe_add_customer_address` function from checking if a customer has _any_ addresses to querying whether the specific address already exists for the customer.


robincornett:
Note: tested this with storefront data which has many addresses and this PR prevented a lot of duplication. Some seeming duplicates left, but they're due to variations in how the address was entered (space in postal code v. not, for example).

ashleyfae:
Brought up to date with `release/3.0` now that https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7895 has been merged.

robincornett:
Testing again now that it's been updated and have run into a couple of issues I'm trying to track down. One issue I know I can fix, but the other seems to be preventing accurate duplicate detection, so let me get back to you on this please.

robincornett:
@ashleyfae what I pushed up yesterday works, but because it broke once due to additional pieces being added to the new address, I'm left feeling like it's more fragile than it could be.

What would you think if, instead of using `unset` to remove elements from the array, and then having to add them back, I created a new array specifically for checking, which would include _only_ the data elements needing to be added as part of a new address (`address`, `address2`, `city`, `region`, `country`, `postal_code` I think), and using that for the two checks? That way, if additional elements are added to the address array for whatever reason, this won't be affected.

ashleyfae:
@robincornett Sounds good - go for it!

ashleyfae:
@robincornett This looks good. A few small things:

1. How about adding `'type' => 'billing'` to the whitelist as well? This will also require assigning `$data['type']        = 'billing';` higher up. We only use `billing` now but I imagine we might possibly want to allow someone to add the same address but with different types if/when we actually expand on this option.
2. Perhaps add a note [to the doc block](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/d333f5ac248931331aff43bafed9bf32da2b37b6/includes/customer-functions.php#L592) saying only unique addresses can be added.
3. Also maybe elaborate some more on this comment: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/d333f5ac248931331aff43bafed9bf32da2b37b6/includes/customer-functions.php#L626 to explain that we're checking if the address already exists. You've explained _what_ is going on, but not _why_. Might be worth noting.

spencerfinnell:
Is there a need or benefit to including `name` in the query to check for duplicates? I could see how it could be considered a unique address at that point, but you could also argue that a separate customer should be used instead?

robincornett:
At the point this function is implemented, the customer ID is already established, so even if the name is different, it will be assigned to the same customer.

I think one potential source of confusion here is that although the `name` is stored with the address, it's generally not displayed as part of it (thinking through the address list table, customer details screen, even creating an order)....so if "John Doe" has two orders with the same address, but does one order as "John Doe" and one as "John H Doe", then it would appear to be a duplicate, even though the `name` in the database is different, I think.

robincornett:
Just ran this again for a 2.9-3.0 migration on a storefront with a lot of addresses and it's working as expected--no errors during the migration process and spot checking currently showing no unexpected duplicates.

Created a new order for a new customer and did not include an address, and no address was created.

PR #7912, opened 2020-05-15
Update recount tools to use new tables & queries #7910

### LINKED ISSUE #7910: 3.0 - Stat recount tools for downloads and customers were not working.
### Expected behavior

When I use the `Recount Stats` tool to "Recount Earnings and Sales for a download", I expect all my sale and earning numbers to be corrected.

### Actual behavior

When I run the tool for a specific download, the stats all get set to 0, which is not accurate.

Here's how it looks prior to recounting:

Screenshot from 2020-05-15 13-04-56

Proof there are sales:

Screenshot from 2020-05-15 13-05-35

Recounting stats:

Screenshot from 2020-05-15 13-06-14

New numbers:

Screenshot from 2020-05-15 13-06-36

1) Have a download set up that you know has had at least one sale and some earnings.
2) Recount the earnings and sales for that download.
3) Refresh the table and you'll see the numbers are now all `0`.

I assume `Recount Earnings and Sales for All downloads` is equally broken. Haven't tested any others.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #7910

### `EDD_Tools_Recount_All_Stats` and `EDD_Tools_Recount_Download_Stats`

1. Classes have been rewritten to use a new `edd_get_order_items()` query.
2. `edd_recount_download_stats_args` and `edd_recount_download_stats_total_args` filters have been deprecated. I chose to deprecate them because like 99% of the arguments are no longer applicable at all, but we could also choose to keep them so that the new queries can still be filtered. :shrug: 

### `EDD_Tools_Recount_Customer_Stats` and `EDD_Tools_Recount_Single_Customer_Stats`

1. Recount processes largely weren't working due to `post__not_in` no longer working with payment queries. This has been fixed separately here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7846#issuecomment-629236644 but I took the opportunity to rewrite this to use a new query instead. We no longer need to do the two separate queries (attached and unattached) because those errors should have been corrected during the migration. Orders should now all be associated with the correct `customer_id`.
2. Note: still using `edd_get_payment()` in the recount processes due to filters being in place that take in an `EDD_Payment` object.
3. Now enqueuing the `edd-admin-tools-export` JavaScript on the Customer Details page. This is due to that screen utilizing the exporter script to recount a single customer.

### General

1. Updated some incorrect PHP docs.
2. Corrected a few items that were incorrectly passing in a second parameter to `get_stored_data()` although the function only takes one parameter. Example: `$items = $this->get_stored_data( 'edd_reset_tables_to_truncate', false );` => `$items = $this->get_stored_data( 'edd_reset_tables_to_truncate' );`

## Testing

All of the tools in Downloads > Tools > General should be tested. A good way to do this:

1. Already have a tab open that shows the current, accurate stats. Keep this open.
2. Switch to `release/3.0` branch and run the tool.
3. Open a new window and navigate to the same page from step 1. Check out the broken stats.
4. Switch to this branch, `issue/7910` and run the tool again.
5. Refresh your window from step 3 (or open yet another) and check that the data has been updated and matches the data you had originally in step/window 1. (Assuming your data was accurate to begin with. :P )

To test the individual customer tool you need to go to Downloads > Customers, click through to a customer, click "Tools", and click `Recount Stats`.


includes/admin/tools/class-edd-tools-recount-all-stats.php
```diff
@@ -72,73 +76,57 @@ public function get_data() {
...
-			unset( $payment_ids );
+		$order_items = edd_get_order_items( $new_args );
 
-			foreach ( $payments as $payment ) {
+		if ( $order_items ) {
+			foreach ( $order_items as $order_item ) {
 
 				// Prevent payments that have all ready been retrieved from a previous sales log from counting again.
-				if ( in_array( $payment->ID, $processed_payments ) ) {
+				if ( in_array( $order_item->id, $processed_payments ) ) {
```
robincornett, line 104: Could this and the next `in_array` check include `true` as the third parameter?

includes/admin/tools/class-edd-tools-recount-customer-stats.php
```diff
@@ -66,40 +66,22 @@ public function get_data() {
...
-				);
-
-				$unattached_payments = edd_get_payments( $unattached_args );
-
-				$payments = array_merge( $attached_payments, $unattached_payments );
+				$payment_ids = edd_get_orders( array(
+					'type'        => 'sale',
+					'customer_id' => $customer->id,
+					'status__in'  => $allowed_payment_status,
+					'number'      => 999,
```
robincornett, line 73: Is 999 a large enough number?

robincornett:
The monthly sales/earnings averages when I view the report for all downloads has still been reset to 0, but all had numbers before.

LisaCee:
Should running steps 4-5 correct the broken stats?  Or does this just fix the stats going forward and not retroactively?

includes/admin/tools/class-edd-tools-recount-all-stats.php
cklosowski: For this one, since it is coming from the database, there is the possibility that a database abstraction could possibly alter it from a string to an int or an int to a string. Since database dropins could modify the way serialization happens. I can't think of a certain situation where using `true` would benefit us further at this time, unless you have an example.

Leaving it without `true` for the other param means we'll be able to process either way.

Same goes for the accepted statuses. Since those can be filtered, I think it is safer to just leave them as 'loose' searches to account for possible data type changes. I can't imagine someone using numeric statuses, but it's not impossible that it could be an integer in the array above, but a string coming from the DB.

includes/admin/tools/class-edd-tools-recount-customer-stats.php
cklosowski: I think so. This would be orders for a single customer in the allowed status list, that are sales only. Based off those numbers the customer would be able to have the following to burst past 999:
2.7 years of daily purchases or
19.21 years of 1 purchase a week or
83 years of a customer making a single purchase per month.

Based off those numbers, I think it's pretty safe to assume that 999 is within reason.

cklosowski:
@LisaCee You should be able to just run it on steps 4-5. The only thing we're altering is some cached values, so there is no alteration of the underlying data set that is done.

PR #7916, opened 2020-05-15
Remove report view filters when getting dashboard widget data

### LINKED ISSUE #7914: 3.0 - Unhook "edd_report_views" filter outside of Reports page to prevent excessive deprecated error logging

Related conversation here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7904

In our dashboard widget, we initialize `EDD\Stats()` here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/dashboard-widgets.php#L62 This in turn initializes `Reports\Init`, which then loads _legacy reports_. This uses the `edd_report_views` filter, which throws this deprecation notice:

> PHP Notice: The edd_report_views hook is <strong>deprecated</strong> since Easy Digital Downloads version 3.0! Use the edd_reports_get_reports hook instead.

If you have any legacy reports hooked in, this excessively gets logged any time the heartbeat runs.

We don't actually need to load legacy reports for the dashboard stats, so we should unhook legacy reports when we're on the dashboard to prevent this excessive deprecation notice.

We could stop the notice from triggering by running this before initializing `EDD\Stats()` in the dashboard widget: `remove_all_filters( 'edd_report_views' );`

### Justification or use case

Prevent the deprecation notice from excessively clogging error logs. We only need to trigger it when actually on the Reports page.

robincornett:
Fixes #7914

### DIFF
--- a/includes/admin/dashboard-widgets.php
+++ b/includes/admin/dashboard-widgets.php
@@ -59,6 +59,7 @@ function edd_get_dashboard_sales_widget_data() {
 		if ( 'total' === $range ) {
 			unset( $args['range'] );
 		}
+		// Remove filters so that deprecation notices are not unnecessarily logged outside of reports.
 		remove_all_filters( 'edd_report_views' );
 		$stats          = new EDD\Stats( $args );
 		$data[ $range ] = array(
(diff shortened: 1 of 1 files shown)


includes/admin/dashboard-widgets.php
```diff
@@ -59,6 +59,7 @@ function edd_get_dashboard_sales_widget_data() {
 		if ( 'total' === $range ) {
 			unset( $args['range'] );
 		}
+		remove_all_filters( 'edd_report_views' );
```
spencerfinnell, line 62: @robincornett Can you add a comment here as to why these filters are removed?

PR #7918, opened 2020-05-15
Load Localization Scripts

### LINKED ISSUE #7847: 3.0 - Assets - Localizations Not Always Loaded
### Expected behavior

When using `edd_load_scripts()` directly the relevant scripts should be enqueued and localized.

### Actual behavior

Scripts are only enqueued, not localized. 

1) Call `edd_load_scripts()` directly.

Observe `edd_global_vars` is not available.

Related #7839

LisaCee:
Fixes #7847 

Proposed Changes:
1. Called `edd_localize_script` inside of `edd_load_scripts`
2. Unhooked (removed hook?) for `edd_localize_script` from `wp_enqueue_scripts`

### DIFF
--- a/includes/scripts.php
+++ b/includes/scripts.php
@@ -96,6 +96,7 @@ add_action( 'init', 'edd_register_styles' );
  */
 function edd_load_scripts() {
 	edd_enqueue_scripts();
+	edd_localize_scripts();
 }
 add_action( 'wp_enqueue_scripts', 'edd_load_scripts' );
 
@@ -204,7 +205,6 @@ function edd_localize_scripts() {
 		) ) );
 	}
 }
-add_action( 'wp_enqueue_scripts', 'edd_localize_scripts' );
 
 /**
  * Load head styles


cklosowski:
@spencerfinnell based off this example of the script localization:
https://developer.wordpress.org/reference/functions/wp_localize_script/#comment-1301

Shouldn't we call `edd_localize_scripts` first, then call `edd_enqueue_scripts`? So that the localizations are ready when the `enqueue` happens?

spencerfinnell:
@cklosowski Scripts need to be registered or enqueued first. From those docs:

> Works only if the script has already been added.

> IMPORTANT! wp_localize_script() MUST be called after the script has been registered using wp_register_script() or wp_enqueue_script().

Scripts are now _registered_ [on `init`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/scripts.php#L17-L37) so the exact order of enqueueing/localizing on the `wp_enqueue_scripts` hook shouldn't matter much, but I tend to always put the localization last.

cklosowski:
@spencerfinnell works for me.

PR #7924, opened 2020-05-16
Register jquery-ui-css 

### LINKED ISSUE #7884: 3.0 - Assets - jquery-ui-css No Longer Registered
### Expected behavior

When I set a style dependency of `jquery-ui-css` my custom styles should load.

### Actual behavior

The styles are not loaded because there is no `jquery-ui-css` style handle registered.

1) `wp_enqueue_style( 'my-custom-style', get_template_directory_uri() . '/style.css', array( 'jquery-ui-css' ) );`

We don't need to use/enqueue it, but the style handle should still be registered. `release/2.9.x` also registers a different CSS file based on the admin color scheme, which we don't need to keep in `release/3.0`.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/scripts.php#L274

LisaCee:
Fixes #7884

Proposed Changes:
1. Added `jquery-ui-css` to admin styles

* Let me know if there is a better place for this.  In 2.9.x it was registered in `edd_load_admin_scripts`, but since it's a CSS library that didn't seem right. *


includes/scripts.php
```diff
@@ -353,6 +353,7 @@ function edd_register_admin_styles() {
 
 	// Register styles
 	wp_register_style( 'jquery-chosen',         $css_dir . 'chosen'               . $css_suffix, array(), $version );
+	wp_register_style( 'jquery-ui-css',         $css_dir . 'jquery-ui-'           . $css_suffix, array(), $version );
```
spencerfinnell, line 356:
```suggestion
	wp_register_style( 'jquery-ui-css',         $css_dir . 'jquery-ui-fresh'      . $css_suffix, array(), $version );
```

LisaCee: Okay, I think I've got this resolved.

includes/scripts.php
```diff
@@ -353,6 +353,8 @@ function edd_register_admin_styles() {
...
```
spencerfinnell, line 356: @LisaCee This line (357) should be removed.

includes/scripts.php
```diff
@@ -353,6 +353,7 @@ function edd_register_admin_styles() {
 
 	// Register styles
 	wp_register_style( 'jquery-chosen',         $css_dir . 'chosen'               . $css_suffix, array(), $version );
+	wp_register_style( 'jquery-ui-fresh',       $css_dir . 'jquery-ui-fresh'      . $css_suffix, array(), $version );
```
spencerfinnell, line 356: 🤦 Sorry @LisaCee, these diffs are messing with my brain today. 

The script handle (first argument) should still be `jquery-ui-css`, and then `jquery-ui-fresh` is used to reference the file names: `jquery-ui-fresh.css` and `jquery-ui-fresh.min.css`.

Sorry for the back and forth.

PR #7934, opened 2020-05-18
Update report graph values for all reports

### LINKED ISSUE #7907: 3.0 - Reports - Graphs not showing complete data
### Expected behavior

I expect the graphs to show accurate data for the reported period.

### Actual behavior

It's possible all graphs are affected, but I'll focus on `Sales and Earnings`.

If I set the graph to `Last Year` then I get a month-by-month view. Each tick is supposed to show the total number of sales and total earnings in that period. However, the numbers don't add up. What's actually happening is that only the sale count/earnings for _the last day in that month that has data_ is being displayed.

This becomes particularly clear when you compare the values on the ticks with the tile further up that shows `Sales / Earnings`.

This is the problem: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports-callbacks.php#L86-L103

This is a `foreach` loop, but the values are being _overwritten_ each loop.

```php
$sales[ $timestamp ][1]    = $result->sales;
$earnings[ $timestamp ][1] = $result->earnings;
```

They should be getting added together like this:

```php
$sales[ $timestamp ][1]    += $result->sales;
$earnings[ $timestamp ][1] += $result->earnings;
```

1) Go to Downloads > Reports.
2) Look at the `Sales and Earnings` graph.
3) Count up the numbers for each tick and you'll see they're massively lower than the tile further up that reports the overall Sales / Earnings.

PHP Version:

WordPress Version:

robincornett:
Fixes downloads, gateways, file downloads, discounts, and customer reports.

Fixes #7907


ashleyfae:
> @robincornett For a Download's `Sales and Earnings` we also need to be limiting the Order Item's to a `complete` status. It's currently pulling items that have been Refunded or Trashed.
> 
> The same for `EDD\Stats::get_order_item_earnings()` and `EDD\Stats::get_order_item_count()`

I'm concerned about some inconsistency we have with this throughout reports in general. The `Overview` page intentionally includes refunded orders in its count. This was decided after a Slack discussion noted here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7637 The full list of statuses it includes is:

```php
$statuses = array( 'complete', 'publish', 'revoked', 'refunded', 'partially_refunded' );
```

This is true for the graph and the file counts.

Now a separate page is excluding `refunded` items and only showing `complete`.

To avoid confusion in reporting, I think we should be consistent across the pages.

Perhaps we should introduce a filter like the tax one for `Exclude Refunded Orders`.

spencerfinnell:
@ashleyfae Good catch. I like the idea of adding a new filter in a follow up as well.

cklosowski:
I'm in agreement as well, lets get another issue opened up to allow filtering statuses for reporting, this will allow extensions to determine how they affect reporting as well.

Apart from the filter, @ashleyfae @spencerfinnell are we good here?

spencerfinnell:
@cklosowski There is an carry over filter from pre 3.0 that we've used in other places: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-stats.php#L181

It was originally used here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/payments/class-payment-stats.php#L388

While the _name_ of the filter is a little more specific to Payments the original _description_ still seems to suite the current needs:

> Filters Order statuses that should be included when calculating stats.

Since we already have a fair amount of deprecations surrounding Reports my vote would be to continue using `edd_payment_stats_post_statuses`.

ashleyfae:
Confirmed the numbers all look right in the graph.

I think for now we should revert the status changes so we can handle that all separately in a new issue.

robincornett:
So revert the last two commits then?

ashleyfae:
> So revert the last two commits then?

Yes please. We'll still fix the statuses, but we'll handle it in a new issue (just created one here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7975). Just keep this PR strictly about the graph amounts and we'll handle all the status related logic in the new issue.

cklosowski:
@ashleyfae @spencerfinnell @robincornett Should we go ahead and use the #7985? This way we can start using that function to get the proper statuses for a `net` reporting?

spencerfinnell:
Since #7985 has landed we might as well start introducing it now to help minimize future changes when we do a full audit of status usage separately.

robincornett:
Yes, but my understanding from [Ashley's comment](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7934#issuecomment-638839389) is that the status fixes will not go into this PR, but will be part of #7975?

ashleyfae:
> Yes, but my understanding from [Ashley's comment](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7934#issuecomment-638839389) is that the status fixes will not go into this PR, but will be part of #7975?

Yes I still think that's what we should do. The scope of this issue was to address the numbers on the ticks. We should do status changes in the new issue.

PR #7936, opened 2020-05-19
Update admin tables' responsive CSS

### LINKED ISSUE #7922: 3.0 - Some tables, elements display oddly on smaller screens
### Expected behavior
I would like for admin screens, including tables, to display well on a variety of screen sizes.

### Actual behavior
When I view certain tables on smaller screen sizes, it's not handled gracefully. Example on the customer details screen:

image

Recent orders, also on the customer details screen:

image

Refund Modal:

image

These are just some samples I found in some quick checking.

1) In EDD 3.0, open your browser's dev tools.
2) Set the view to responsive/iPad size.
3) Visit a Customer's details screen--one with orders and multiple addresses.
4) Edit an order and initiate a refund.

PHP Version:
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

The licenses table on the customer details screen does not have this issue, but it also doesn't display well after a certain screen size--it is not responsive. It looks like these tables are inheriting some of the WordPress list tables behaviors, as those tables generally show only the "primary" column on small screens, and hide all others behind a toggle.

robincornett:
Fixes #7922

Proposed Changes:
1. Update the refund modal for smaller screens (iPads and lower). The WP Core primary column toggle doesn't work with the modal, so that has been removed, and the table CSS overridden to display necessary data on small screens. (It will look better in conjunction with PR #7899, but is not required.)
2. Fix the customer details physical address table to display only the primary column, although this feels awkward because just the street address doesn't feel helpful. Reworking to be more like the refund modal might make more sense.
3. Remove the primary column from the customer emails table (this was added in 3.0 but should not have been, as it hides the ability to remove emails on smaller screens).

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -1105,7 +1105,6 @@ textarea[name="edd-note"] {
 }
 
 .wp-list-table.addresses .column-type,
-.wp-list-table.emails .column-type,
 .wp-list-table.posts .column-download_category,
 .wp-list-table.posts .column-download_tag,
 .wp-list-table.posts .column-price {
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -3340,28 +3340,9 @@ body.download_page_edd-reports {
 #edd-item-tables-wrapper .no-items {
 	text-align: left;
 }
-#edd-item-tables-wrapper .column-actions {
-	width: 25%;
-}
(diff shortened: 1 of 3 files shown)


robincornett:
There is an open issue/PR to prevent empty physical addresses being added to the database (#7814), so hopefully we won't see that screen, but that logic has all been moved within the media query now, as I should have done.

PR #7938, opened 2020-05-19
headings on order page changed to h2, #7751

### LINKED ISSUE #7751: 3.0 - Review Heading Levels on Admin Screens
### Expected behavior
Heading levels should only increase/adjust by one (ref: [Deque](https://dequeuniversity.com/rules/axe/3.5/heading-order))

### Actual behavior
On an edit/add order admin screen, where the `postbox` containers are used, headings are currently set to `h3`. They follow the page title, which is an `h1`, and so the heading levels are not correct (`h2` is skipped).

1) on a site running 3.0, create or edit an order
2) inspect the heading levels for the postboxes, such as "Order Amounts" and "Order Items"

PHP Version: all
EDD Version (or branch): 3.0
WordPress Version: all

Updating the heading to `h2` does not change the visual appearance, as WordPress core applies the same styling to both heading levels (`#poststuff h3.hndle, #poststuff .stuffbox > h3, #poststuff h2`). I suspect it was flagged as an accessibility issue in Core at some point and fixed there, and the CSS was updated to maintain visual consistency for similar situations.

Currently, EDD adds this CSS to override the WordPress default behavior of changing the cursor to show that a postbox is movable:

```css
#edd-dashboard-widgets-wrap .postbox h3 {
    cursor: default;
}
```
EDD is having to do this because the `h3` element here includes the `hndle` class, but the postboxes on this screen are not movable. If we consider removing the `hndle` class from the heading, we can remove that CSS rule; otherwise, we need to update it when the heading level is updated.

I've scanned for other uses of `postbox` and titles in EDD and this one seems to be the main one that needs to be updated. Some are output on Tools screens, such as "Recount Stats", and since the Tools have `h2` output already, the heading level for "Recount Stats" is already correct.

LisaCee:
Fixes #7751

Proposed Changes:
1. Changed `<h3 class="hndle">` to `<h2>`

- I did not change the Billing Address to `<h2>` as that does change the appearance.  

- I have not modified any CSS, yet, until it is decided what to do with the `hndle` class on the screen.

There is also an `<h3>` on License Keys, from Software Licencing.  I can put in an issue/PR there too for consistency if you want.


includes/admin/payments/orders.php
```diff
@@ -768,9 +768,9 @@ function edd_order_details_sections( $order ) {
 ?>
 
 	<div id="edd-customer-details" class="postbox">
-		<h3 class="hndle">
+		<h2 class="hndle">
 			<span><?php esc_html_e( 'Order Details', 'easy-digital-downloads' ); ?></span>
```
robincornett, line 772: I think the `<span>` tag can be removed from each of these headings.

robincornett:
Should also note that any extensions which add information here, such as Software Licensing, should also review their headings (Software Licensing uses `<h3>` and should update to `<h2>` to match.)

spencerfinnell:
@robincornett I like the idea of adding a heading inside the panel content that is hidden when the tab label is visible, then toggled when the tab label is hidden.

robincornett:
@spencerfinnell does EDD have a utility "show on mobile" class that would work like screen reader text on desktop? Reports would benefit from something similar, especially as several of them use the same icons.

spencerfinnell:
@robincornett No. Can you open an issue outlining a reusable CSS pattern we can utilize in the future? The extra context labels aren't a blocker here so we can get it set up at a later date.

LisaCee:
It looks like all of these changes are in PR #8152.  Can we close this or should we keep it open for the label / show on mobile discussion?

ashleyfae:
> It looks like all of these changes are in PR #8152. Can we close this or should we keep it open for the label / show on mobile discussion?

I think we can move that to a new issue. :+1:

PR #7948, opened 2020-05-20
Align tiles 

### LINKED ISSUE #7928: 3.0 - Reports - Tiles Alignment Incorrect
### Expected behavior

When viewing Reports the Tiles should align with other endpoint displays.

### Actual behavior

They are inset too far:

Screen Shot 2020-05-18 at 10 05 19 AM

1) View "Reports > Customers"

I believe the CSS might have been lost in a merge conflict, as I don't believe they were misaligned before. 

Previously: #7669 

Relevant CSS: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/css/edd-admin.css#L2652-L2688

LisaCee:
Fixes #7928

Proposed Changes:
1. adjusted margins on edd-reports-tiles-wrap id

I'm not crazy about how hacky this solution looks, but is this what you were thinking visually @spencerfinnell?

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -2650,14 +2650,14 @@ body.download_page_edd-reports {
 }
 
 #edd-reports-tiles-wrap {
+	display: -ms-grid;
 	display: grid;
 	grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
-	grid-gap: 10px;
+	grid-gap: 20px;
 }
 
 .edd-reports-tile {
 	text-align: center;
-	margin: 0 10px 20px;
 	padding: 30px 10px;
 	display: flex;
 	flex-direction: column;
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -2652,8 +2652,7 @@ body.download_page_edd-reports {
 #edd-reports-tiles-wrap {
 	display: grid;
 	grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
-	padding: 20px 10px;
-	margin-bottom: -10px;
+	grid-gap: 10px;
 }
 
 .edd-reports-tile {
@@ -2661,7 +2660,6 @@ body.download_page_edd-reports {
 	margin: 0 10px 20px;
 	padding: 30px 10px;
 	display: flex;
-	flex: 0 1 calc(33% - 20px);
 	flex-direction: column;
 	justify-content: center;
 	border: 1px solid #e5e5e5;
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -2650,13 +2650,10 @@ body.download_page_edd-reports {
 }
 
 #edd-reports-tiles-wrap {
-	display: flex;
-	flex-wrap: wrap;
-	justify-content: center;
-	margin-left: 10px;
-	margin-right: 10px;
+	display: grid;
+	grid-template-columns: auto auto auto;
+	padding: 20px 10px;
 	margin-bottom: -10px;
-	padding: 10px;
 }
 
 .edd-reports-tile {
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -2668,26 +2668,6 @@ body.download_page_edd-reports {
 	box-sizing: border-box;
 }
 
-@media only screen and (max-width: 782px) {
-
-	.edd-reports-tile {
-		flex-basis: calc(50% - 20px);
-	}
-}
-
-@media screen and (max-width: 600px) {
-
-	.edd-reports-tile {
-		flex-basis: 100%;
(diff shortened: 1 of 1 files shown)


assets/css/edd-admin.css
```diff
@@ -2652,8 +2652,8 @@ body.download_page_edd-reports {
 #edd-reports-tiles-wrap {
 	display: flex;
 	flex-wrap: wrap;
-	margin-left: 10px;
-	margin-right: 10px;
+	margin-left: -13px;
+	margin-right: -10px;
```
spencerfinnell, line 2656: My first thought was to do negative margins too, but looking again it might be better to override the padding applied to `.edd-sections-wrap .section-wrap .section-content > *`

Perhaps we just override the padding to be `10px` which will be combined with the `10px` margin on the actual tiles. 

This also makes it easier because the padding override will only need to be applied on the breakpoint where the tiles go in to multiple columns (https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/css/edd-admin.css#L2725)

robincornett: Since the styling here is by ID, overriding the padding is pretty straightforward. I would also see about adding `justify-content: center;` because the columns don't completely fill the space even with the calculated margins, so technically there is a bigger gap on the right than the left.

LisaCee: I don't think the tiles are quite out to the edge of the other sections.  Personally I like how it looks, but I am happy to tweak as needed.

assets/css/edd-admin.css
```diff
@@ -2652,9 +2652,11 @@ body.download_page_edd-reports {
 #edd-reports-tiles-wrap {
 	display: flex;
 	flex-wrap: wrap;
+	justify-content: center;
 	margin-left: 10px;
 	margin-right: 10px;
```
robincornett, line 2657: What about removing all of the margins, so they would be zero, and adjusting the padding to 20px 10px, so that the room at the top looks less cramped?

Note: if the number of tiles are not divisible by 3, the last row tiles will be centered (because of my justify-content suggestion), which I'm okay with, but does bring them out of alignment with the grid.

LisaCee: @spencerfinnell ?  :point_up:

spencerfinnell: What about using actual CSS grid to make a true grid? 

https://www.w3.org/TR/css3-grid-layout/
https://caniuse.com/css-grid

It seems appropriate for a use case like this.

assets/css/edd-admin.css
```diff
@@ -2650,10 +2650,9 @@ body.download_page_edd-reports {
 }
 
 #edd-reports-tiles-wrap {
-	display: flex;
-	flex-wrap: wrap;
-	margin-left: 10px;
-	margin-right: 10px;
+	display: grid;
```
robincornett, line 2653: I haven't used CSS grid, but is there a way to force the columns to be the same width? They are varying widths for me with this in place, depending on the tile content.

LisaCee: I'm running into that, too.  I'm still playing with making them equal, but also not forcing there to be 3 columns on each row.

LisaCee: Hey @robincornett.  I just did another push.  Let me know what you think about the sizing.

robincornett: @LisaCee gorgeous! It's working great for me on multiple screen sizes.

Note: with the change to CSS Grid, I think we can remove the `flex: 0 1 calc(33% - 20px);` from `.edd-reports-tile` since it's no longer using flexbox (the tile itself is still using it so everything else should stay, but I think that's the only one applying to the tile as a child of a flexbox container). Line 2664, it looks like.

spencerfinnell: We can also now remove the `margin` in place of `grid-gap: 20px;` (and remove the `padding` override).

LisaCee: New push to review.  `grid-gap: 20px` looked too big so I did`10px`, but let me know what you guys think.

robincornett: What about the `grid-gap: 20px`, and removing the margins set on `.edd-reports-tile`, which I think is part of the reason the tiles are causing issues and requiring the negative margins in the first place? It looks like the `grid-gap` property on the parent container makes the appearance of margins but without the extra math on the user end?

spencerfinnell: @LisaCee Robin is correct we can remove the spacing on the individual tile. The `20px` previously was added on top of the `10px` left/right margin which would make it `40px` total. 

We also need to add `-ms-` prefixes for some of these rules. Some more info here: https://medium.com/@elad/supporting-css-grid-in-internet-explorer-b38669e75d66

PR #7954, opened 2020-05-22
Update chosen select widths

### LINKED ISSUE #7953: 3.0 - Chosen dropdown can be too narrow
### Expected behavior
When I click on a select field, I expect the values to be readable.

### Actual behavior
This is occurring specifically on the Payment Gateways report screen:

image

I think it's because the gateway select field is using the default "All", and when Chosen renders the field, it uses the first option to set the width of the entire element.

1) In EDD 3.0, go to Downloads > Reports > Payment Gateways.
2) Click to select a specific gateway.

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

We could fix the issue on this specific screen by changing the label here to "All Gateways" instead of using the default, but I think it may be better to fix the rendering behavior so that other instances will display well.

robincornett:
Resolves super narrow select issues when the first option has a short
label. Adds an override for the customer dropdown on the orders screens.

Fixes #7953

### DIFF
--- a/assets/css/edd-admin-chosen.css
+++ b/assets/css/edd-admin-chosen.css
@@ -246,8 +246,12 @@
 }
 
 .edd-select-chosen {
-	width: 100% !important;
+	width: 100%;
 	max-width: 300px;
+}
+
+.edd-select-chosen.edd-customer-select {
+	width: 100% !important;
 }
 
 .edd-select-chosen.edd-time {


assets/css/edd-admin-chosen.css
```diff
@@ -246,10 +246,14 @@
...
 
 .edd-select-chosen {
-	width: 100% !important;
+	width: 100%;
 	max-width: 300px;
 }
 
+.edd-select-chosen.edd-customer-select {
+	width: 100% !important;
+}
```
spencerfinnell, line 255: @robincornett Is this because Chosen is initially hidden inside of a tab section?

LisaCee:
I have no idea where this is coming from, but the two fields look different.  Border radius, dropdown arrow and outline/selected are different.  This is with them being deselected, but it is more noticeable when active.  Still trying to figure out a screenshot for that. 
image

robincornett:
Yes, the chosen script includes a lot of CSS to mimic a select field, but it's all custom CSS (EDD overrides some, but it's an external style). It used to look more like core WP inputs, but looks very different now. And when you have a `chosen` field right next to a regular select field, the difference is even more apparent.

> Would love to consider updating the chosen dropdowns to look more like WP select fields--the box shadow, border, and arrow are all super different. _(from the issue conversation)_

Issue #7583, opened 2020-02-12
Audit Use of "Chosen" Dropdowns


spencerfinnell:
### Explain your enhancement (please be detailed)

Remove use of [`Chosen`](https://harvesthq.github.io/chosen/) `<select>` tags when not needed.

### Justification or use case

There are many instances of using Chosen in the admin area where the list of available options is limited and predefined. These smaller lists that do not need the ability to quickly filter them or search for dynamic results should use standard `<select>` fields.

Items with just two options should either be a checkbox or list of radio options.

Limiting Chosen usage improves accessibility (currently cannot access a Chosen dropdown with a keyboard) and reduces possible UI bugs/discrepancies with core style changes.

spencerfinnell:
Once the number of items is down we should look at deprecating Chosen (which itself has been Deprecated https://github.com/harvesthq/chosen) and replacing it with a more maintained alternative, such as:

- https://github.com/jshjohnson/Choices
- https://github.com/woocommerce/selectwoo

Related https://github.com/easydigitaldownloads/easy-digital-downloads/issues/2758 https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4375

cklosowski:
We’ll have to be congnisant of the fact that some of these lists might have filters on them, allowing them to eventually have more than a few options, so we need to be careful not to convert to a massive radio list if multiple extensions have hooked into that filter.

Overall I’m a :+1: for this though for sure.

spencerfinnell:
Also, since Chosen doesn't transform `<select>` inputs/run at all on touch-enabled devices any of the fields that rely on AJAX search do not work on mobile devices.

cklosowski:
That is a fair point @spencerfinnell.

spencerfinnell:
Having a terrible time with my keyboard so I'm going to chunk this up so maybe we can get some in 3.0.

Removing some in settings via #7671

spencerfinnell:
@cklosowski I’d like to keep this open as a tracking issue for smaller PRs. There are some more places that can still be removed.

cklosowski:
No problem @spencerfinnell, thanks for the heads up.

robincornett:
Just noting here that select fields using `chosen` are not accessible at all for keyboard users or for screen readers.

ashleyfae:
[...]

PR #7956, opened 2020-05-22
3.0 - Remove More Chosen Instances

### LINKED ISSUE #7750: 3.0 - Update Linked Hour/Minute Fields
Where two separate inputs are used to set a time, update them to use number inputs. Additionally, update the markup to clearly and semantically indicate that these fields are related.

### Justification or use case
On the new/update order screen, the time is currently set using two separate `<select>` inputs, which also implement `chosen`. These two inputs are ideal for changing to an actual number input, because we can set the minimum/maximum values, and the values can be adjusted incrementally.

Similar hour/minute select fields are implemented on the discounts screens, but without implementing `chosen`.

> Also, thinking through our discussions about radio inputs and grouped controls, I did some more digging and I think the hour/minute inputs would benefit from being "grouped" using a `<fieldset>` as they are interdependent.
> - see especially the [SSN example](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA17) on the W3 site
> - [address field](https://www.w3.org/WAI/WCAG21/Techniques/html/H71) as an example as well, so maybe we should consider doing the same with our address field groups
> - consider using `aria-labelledby` to provide additional context for [fields in groups](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria)

_Originally posted by @robincornett in https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7744_

spencerfinnell:
Chosen keeps causing issues and the UI is out of date in inaccessible. This is another round of removing Chosen on `<select>` inputs that do not need to be filterable or dynamically searchable.

See #7583 
Fixes #7750

Proposed Changes:
1. Remove unneeded Chosen usage throughout interfaces.
2. Use `number` fields on Order screen Time inputs.


includes/admin/payments/orders.php
```diff
@@ -1026,39 +1021,13 @@ function edd_order_details_attributes( $order ) {
...
-									'options'          => edd_get_hour_values(),
-									'selected'         => $order_date->format( 'H' ),
-									'chosen'           => true,
-									'class'            => 'edd-time edd-form-group__input',
-									'show_option_none' => false,
-									'show_option_all'  => false,
-								)
-							); // WPCS: XSS ok.
-							?>
+							<input type="number" min="0" max="24" step="1" name="edd-payment-time-hour" id="edd-payment-time-hour" value="<?php echo esc_attr( $order_date->format( 'H' ) ); ?>" />
```
robincornett, line 1024: The related labels for the hour and minute inputs need to be updated to match the new ID (`EDD()->html->select()` converts `-` to `_` in the ID, which is no longer applicable here).

includes/admin/reporting/class-base-logs-list-table.php
```diff
@@ -250,7 +242,7 @@ public function downloads_filter() {
 		) );
 
 		if ( $downloads ) {
-			echo '<select name="download" id="edd-log-download-filter" class="edd-select-chosen ' . esc_attr( $rtl_class ) . '">';
+			echo '<select name="download" id="edd-log-download-filter">';
```
robincornett, line 245: Why remove `chosen` from this dropdown? Everywhere else we display a select for downloads/products, it uses `chosen`, although I see that this does not use the `product_dropdown` method, which is maybe another question. But presumably this could be rather a long list, so keeping `chosen` here seems like it would make sense.

spencerfinnell: @robincornett I removed it because it doesn't have the AJAX searching built in. It could be a long list but it is long with Chosen enabled as well. 

I wavered on this one but I think having a accessible long list is better than an inaccesible long list. It should eventually be updated to support AJAX searching.

robincornett: I did not realize it didn't include the search--so it really wasn't using the script at all in the first place.

spencerfinnell: @robincornett Chosen will show a "Search" field if there are more than X amount of items, but since the Downloads are listed alphabetically it's just as easy to select an option by typing the name and selecting with built in `<select>` behavior.

ashleyfae:
@spencerfinnell Can you also remove chosen from the Discount `Start date` and `End date` time dropdowns ? (Add/Edit discount UI)

includes/admin/discounts/edit-discount.php
```diff
@@ -177,31 +177,19 @@
...
-							'chosen'           => true,
-							'class'            => 'edd-time',
-							'show_option_none' => false,
-							'show_option_all'  => false
-						) );
-
-						echo esc_html( ' (' . edd_get_timezone_abbr() . ')' );
-						?>
+
+												<label class="screen-reader-text" for="start-date-hour">
```
ashleyfae, line 181: Funky spacing here.

includes/admin/discounts/edit-discount.php
```diff
@@ -214,31 +202,19 @@
...
-							'chosen'           => true,
-							'class'            => 'edd-time',
-							'show_option_none' => false,
-							'show_option_all'  => false
-						) );
-
-						echo esc_html( ' (' . edd_get_timezone_abbr() . ')' );
-						?>
+
+												<label class="screen-reader-text" for="end-date-hour">
```
ashleyfae, line 206: Spacing issue.

includes/admin/discounts/edit-discount.php
spencerfinnell: @ashleyfae Whole file seems funky: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/discounts/edit-discount.php

Should I format it all?

ashleyfae: @spencerfinnell Yeah, go for it.

includes/admin/discounts/edit-discount.php
```diff
@@ -127,186 +128,162 @@
...
+					<th scope="row" valign="top">
+						<label for="edd-start"><?php _e( 'Start date', 'easy-digital-downloads' ); ?></label>
+					</th>
+					<td>
+						<input name="start_date" id="edd-start" type="text" value="<?php echo esc_attr( false !== $discount->start_date ? $start_date : '' ); ?>" class="edd_datepicker" data-format="<?php echo esc_attr( edd_get_date_picker_format() ); ?>" placeholder="<?php echo esc_attr( edd_get_date_picker_format() ); ?>" />
+
+						<label class="screen-reader-text" for="start-date-hour">
+							<?php esc_html_e( 'Start Date Hour', 'easy-digital-downloads' ); ?>
+						</label>
+						<input type="number" min="0" max="24" step="1" name="start_date_hour" id="start-date-hour" value="<?php echo esc_attr( false !== $discount->start_date ? $start_hour : '0' ); ?>" />
```
ashleyfae, line 185:
```suggestion
						<input type="number" min="0" max="24" step="1" name="start_date_hour" id="start-date-hour" value="<?php echo esc_attr( false !== $discount->start_date ? $start_hour : '00' ); ?>" />
```

For clarity on date formatting and to be consistent with the default on Add Discount page.

Alternatively, it could default to blank. It is a bit weird to have the date field be blank but have the times filled out.

Screenshot from 2020-05-27 15-53-59

I'm fine with either though.

On a related note, should we add placeholders for clarity? like `placeholder="hh"` and `placeholder="mm"` ?

spencerfinnell: @ashleyfae I like the `00` because it helps make it clear it's 24 hour format. The `hh` and `mm` placeholders might be confusing for those who do not understand the formatting notation. We could do `00` placeholders as well.

I know it's not consistent with the current `yyyy-mm-dd` placeholder for the date, but I don't think that's great to have there either -- but a discussion for a different PR.

spencerfinnell: Or perhaps they should _actual_ blank values instead of `00` to better match the "leave blank" descriptions below.

ashleyfae: How about blank values with `00` placeholders?

Also we could adjust the description to read _"Pick the date **and time** .."_ (bold = new)

robincornett:
Note: super finicky, but the `vertical-align` on the EDD datepicker field now results in that input being slightly out of line with the number inputs:

image

The `.chosen-container` for the hour/minute fields using `chosen` had the same `vertical-align` applied so it appeared to match, even with the difference in box shadow/borders.

includes/admin/discounts/edit-discount.php
```diff
@@ -127,186 +126,162 @@
...
+					<th scope="row" valign="top">
+						<label for="edd-expiration"><?php _e( 'Expiration date', 'easy-digital-downloads' ); ?></label>
+					</th>
+					<td class="edd-discount-datetime">
+						<input name="end_date" id="edd-expiration" type="text" value="<?php echo esc_attr( false !== $discount->end_date ? $end_date : '' ); ?>"  class="edd_datepicker" data-format="<?php echo esc_attr( edd_get_date_picker_format() ); ?>" placeholder="<?php echo esc_attr( edd_get_date_picker_format() ); ?>" />
+
+						<label class="screen-reader-text" for="end-date-hour">
+							<?php esc_html_e( 'Expiration Date Hour', 'easy-digital-downloads' ); ?>
+						</label>
+						<input type="number" min="0" max="24" step="1" name="end_date_hour" id="end-date-hour" value="<?php echo esc_attr( false !== $discount->end_date ? $end_hour : '' ); ?>" placeholder="00" />
```
ashleyfae, line 208:
```suggestion
						<input type="number" min="0" max="24" step="1" name="end_date_hour" id="end-date-hour" value="<?php echo esc_attr( false !== $discount->end_date ? $end_hour : '' ); ?>" placeholder="23" />
```

To match the same placeholder value as on the `Add New Discount` page.

includes/admin/discounts/edit-discount.php
```diff
@@ -127,186 +126,162 @@
...
-                <tr>
-                    <th scope="row" valign="top">
-                        <label for="notes"><?php _e( 'Discount Notes', 'easy-digital-downloads' ); ?></label>
-                    </th>
-                    <td>
+
+						<label class="screen-reader-text" for="end-date-minute">
+							<?php esc_html_e( 'Expiration Date Minute', 'easy-digital-downloads' ); ?>
+						</label>
+						<input type="number" min="0" max="59" step="1" name="end_date_minute" id="end-date-minute" value="<?php echo esc_attr( false !== $discount->end_date ? $end_minute : '' ); ?>" placeholder="00" />
```
ashleyfae, line 214:
```suggestion
						<input type="number" min="0" max="59" step="1" name="end_date_minute" id="end-date-minute" value="<?php echo esc_attr( false !== $discount->end_date ? $end_minute : '' ); ?>" placeholder="59" />
```

To match the same placeholder value as on the `Add New Discount` page.

PR #7959, opened 2020-05-23
Add 4 new values to sales REST endpoint

fpcorso:
Building out a data pipeline for extracting and combining data from our store, support inbox, email list, other plugins/EDD extensions, and others to build out reports. However, we need some additional data points in the sales endpoint.

Proposed Changes:
1. To be able to ensure correct joins with data from other plugins, we need user ID.
2. To ensure we are only building reports on our live data, we need the mode.
3. To combine data with other EDD reports, it would be useful to have customer ID.
4. We will also be adding in other payments with different statuses through some of the filters, such as from the Recurring Payments extension, so we need to be able to see Complete vs Renewal statuses.

Here they are in the response after adding these fields on a local test site:
example-edd-sales

### DIFF
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -390,7 +390,7 @@ class EDD_API_V2 extends EDD_API_V1 {
 
 				$sales['sales'][ $i ]['ID']             = $payment->number;
 				$sales['sales'][ $i ]['mode']           = $payment->mode;
-				$sales['sales'][ $i ]['status']         = $payment->status_nicename;
+				$sales['sales'][ $i ]['status']         = $payment->status;
 				$sales['sales'][ $i ]['transaction_id'] = ( ! empty( $payment->transaction_id ) ) ? $payment->transaction_id : null;
 				$sales['sales'][ $i ]['key']            = $payment->key;
 				$sales['sales'][ $i ]['subtotal']       = $payment->subtotal;
--- a/includes/api/class-edd-api-v2.php
+++ b/includes/api/class-edd-api-v2.php
@@ -389,6 +389,8 @@ class EDD_API_V2 extends EDD_API_V1 {
 				$user_info    = $payment->user_info;
 
 				$sales['sales'][ $i ]['ID']             = $payment->number;
+				$sales['sales'][ $i ]['mode']           = $payment->mode;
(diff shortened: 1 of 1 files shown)


fpcorso:
All three of the failed tests in Travis CI are due to a composer install error.

includes/api/class-edd-api-v2.php
```diff
@@ -389,13 +389,17 @@ public function get_recent_sales() {
 				$user_info    = $payment->user_info;
 
 				$sales['sales'][ $i ]['ID']             = $payment->number;
+				$sales['sales'][ $i ]['mode']           = $payment->mode;
+				$sales['sales'][ $i ]['status']         = $payment->status_nicename;
```
ashleyfae, line 393: I think this would be better as `$payment->status`

fpcorso: @ashleyfae, I went back-and-forth on which to do here. If someone was using the REST API with another integration just to quickly generate a report for stakeholders, not sure if "publish" or "edd_subscription" would mean as much as "Complete" or "Renewal". 

But, on the other side, most people who are interacting with the endpoint are probably developers/analysts who could filter/modify the status, if needed.

Yeah, probably better to just go with "status" then.

fpcorso: Changed.

cklosowski:
Just updating the base branch for this to `release/2.9.24`, we just practice release branching so we can avoid possibly conflicts in `master` :+1:

fpcorso:
@cklosowski No worries. EDD's [contributing guidelines](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/CONTRIBUTING.md#making-changes) say to set the pull request to `master`. For in the future, should I still do that since I wouldn't know which is the current release branch or is there a better branch I should initialize the PR's on? Thanks!

cklosowski:
@fpcorso it's perfectly fine to just go against master, so long as your fork is reasonably up to date, it should have no issues directly merging in. If we need another base for your PR to be submitted we can let you know.

Appreciate the PR!

includes/api/class-edd-api-v2.php
ashleyfae: Thanks for the change!

`status_nicename` is really the "display" value, whereas "status" is the internal one. I think the internal one is better for the API and on the other side people you can convert it to your desired display.

PR #7969, opened 2020-05-27
Update customer calculations for lifetime counts/values

### LINKED ISSUE #7947: 3.0 - When a customer has refunded orders, the lifetime value is incorrect
### Expected behavior
If a customer has multiple orders, and one has been refunded, I expect that order to not count as part of the customer's lifetime value (or basically, that the order and refund would be added together to make $0).

### Actual behavior
When I view a customer with two orders and one refund, the refund is calculated as part of the lifetime value, but not the original order, so the lifetime value is/can be negative. Example:

image

1) In 3.0, create two orders for a customer.
2) Refund one order.
3) Visit the customer's details screen to see their lifetime value.

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

robincornett:
Fixes #7947

Proposed Changes:
1. Split the `recalculate_totals` method into two separate methods to get the count/value for the customer (these do not update the database)
2. Change the order status values for retrieving counts and values
3. Update the customer details screen and customer list table to use the new methods

`$customer->purchase_value` and `$customer->purchase_count` are used quite a lot of places throughout the codebase, and I'm not sure if they could all be updated to use a class method instead of a variable, if this is an appropriate direction to go. Also, this could result in discrepancies between what's in the database and what's displayed on a screen until a recalculation has been performed.

Note: the new methods use different statuses because including refunds in the lifetime value makes sense (per the original issue), but including them in the counts may be confusing. The different queries means that the `number` parameter for the purchase value isn't necessarily the same as the count of orders, so that's been changed.


robincornett:
The new `get_purchase_count` and `get_purchase_value` methods now evaluate `sale` type orders only, and only `complete` and `partially_refunded` statuses. I've named these methods based on the related variables, but not sure that's the most clear names I could have chosen.

At this point, customer count and purchase value are successfully updated/recalculated when:
- an order is completed
- an order is refunded (was not working in release/3.0)
- an order is moved to the trash (also not working in release/3.0)

However, at this point it feels like right now I'm just looking for all of the places where an order _might_ be changed or updated and making sure that the customer data is checked/recalculated at that point as well. It seems like the better option would be to make sure that the customer record is updated within `edd_update_order` (or in the database `$orders->update_item()` method), which should catch refunds, new orders, trash, but the customer ID is not always passed through to that order.

cklosowski:
@robincornett I see where you are going with this but I think this will create some performance impacts that we're not ready for.

Right now calls to the purchase value and purchase count will run against the orders DB, which is not the goal. We want to maintain getting that from the column for the customer. The only thing we want to change is how we keep this 'ledger'.

The current method we've used is just adding/subtracting from the column, and what I'm suggesting we do is when a modification to an order is made that will affect the stats of the customer, instead of doing a ledger calculation, we just fully recalculate the values for the order count and purchase value, and store that on the customer. Lookups of the customer's purchase value and purchase count, should be able to use the column in the customer table.

cklosowski:
Also, the likely reason the tests are failing is that we're not actually affecting the DB when we do the order creation in the tests. They are abstracts and avoid actually affecting the true database layer in some cases, leaving them in memory so that we can clear the data set easily and avoid any test pollution between classes.

robincornett:
Hey @cklosowski, thank you for looking over this for me. I'm confused by your comment here:

> Right now calls to the purchase value and purchase count will run against the orders DB...

I have not changed any implementation of `$customer->purchase_count` and `$customer->purchase_value` that I am aware of, although I did start out with that and realized that it would be problematic. What I thought I was getting to at this point is that I've removed the use of the increase and decrease methods in favor of the `recalculate_stats` method as instructed ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-customer.php#L762)). I did split that method into two separate ones, which is likely not necessary, but that method has always queried the orders table. So I thought that all I was doing was making sure the stats (count/value) are recalculated when an order status is changed (because I am not calling the new functions outside of using the recalculation, I thought).

I was struggling initially because in the `release/3.0` branch, the stats are not being updated (even by ledger) when an order is refunded or trashed. That's fixed at this point, but since I felt like I was just looking for potential places where an order _might_ be updated, I took a break because I assumed I would just be missing instances.

cklosowski:
@robincornett in `release/3.0` the `recalculate_stats` method does the queries for the data:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-customer.php#L766

And the methods `get_purchase_value` and `get_purchase_count` do not exist in `release/3.0`.

What I think we should move forward with is:
1) `recalculate_stats` does the work as linked above.
2) Remove the functions `get_purchase_value` and `get_purchase_count`
3) Update any areas in the `EDD_Customer` class where we increase or decrease value or purchase count, we just run the `recalculate_stats` method.

robincornett:
I've removed the new functions and moved their updated functionality back into the `recalculate_stats` method.

As for the increase/decrease methods, right now the only places I can find where they are being used is:
- in tests (I assume we need to update these)
- in [`compat/class-customer.php`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/compat/class-customer.php#L96-L124), where they do a comparison/return boolean, so I'm not sure how to handle those.

Note: the increase/decrease methods have been marked as deprecated in their respective doc blocks, but not in the code. Should that be addressed?

cklosowski:
@robincornett the addition of the deprecation notices has caused the unit tests to fail further. I'm going to dig into them and see what I can figure out.

robincornett:
Please let me know what I can help with, @cklosowski -- I am unable to run unit tests here so haven't tried updating them, although I expect they need to be updated due to the new methods.

cklosowski:
Honestly, I think we just need to kill all resemblance of using the `increse/decrease_stats` methods, and just move towards the new `recalculate_stats` method, as I mentioned on the issue, but we'll wait for some feedback from Ashley on it.

robincornett:
I think this should be held until PR #7986 has been merged so that the new status functions can be used.

robincornett:
@cklosowski, per your comment on the related issue, I'm removing the tests.

I believe the only place left where the increase/decrease functions are called are in [`compat/class-customer.php`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/compat/class-customer.php#L96-L124), which maybe is managing backwards compatibility, and returns a boolean, so I am not sure how to handle those.

Although I did just discover that the recount tool under Tools was basically ledgering without using the functions from the customer class, so I need to look for instances of that ([d92f4f8](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7969/commits/d92f4f864c726cf085f0dcf460f25d6bac6c5adf)).

Also, the deprecated functions all include action hooks in them. Do these also need to be deprecated somehow, or is it enough that the parent functions are deprecated already?

robincornett:
Current status:
- All customer stats related tests have been removed, so currently all tests are passing.
- The only place where the increase/decrease functions are called are in the `compat/class-customer.php` file, which I'm not sure I can do anything about.
- I _think_ I have dug out all of the places where ledgering was being done and have removed those.
- The customer recount tools ended up going on a significant diet, especially the single one, and I may have butchered it. The single recount progress bar doesn't seem to display anything now, which I suspect may be due the calculations being reduced? So I wonder if that is necessary (I have been running it on a customer with 100+ orders and it's just a flash).

cklosowski:
@robincornett that is quite the diet! nice. I would say for the individual recount took, could we just switch to a spinner, instead of a progress bar?

robincornett:
In theory, it's possible, but I don't know how practical it is. Looking at the JS involved, it's the export JS (I am not sure why, but all of the recount functionality appears to be stemmed off of the export functionality). The spinner is already present in the markup, but it's not used at all, because the script manages its own progress bar/notice-wrap output.

I expect it _can_ be done if we want to refactor the relevant JS to handle a "spinner only" option.

includes/admin/tools/class-edd-tools-recount-single-customer-stats.php
```diff
@@ -52,49 +52,8 @@ class EDD_Tools_Recount_Single_Customer_Stats extends EDD_Batch_Export {
...
-				}
-
-				$should_process_payment = 'complete' == $payment->status || 'revoked' == $payment->status ? true : false;
-				$should_process_payment = apply_filters( 'edd_customer_recount_should_process_payment', $should_process_payment, $payment );
-
-				if( true === $should_process_payment ) {
-
-					$found_payment_ids[] = $payment->ID;
-
-					if ( apply_filters( 'edd_customer_recount_sholud_increase_value', true, $payment ) ) {
```
JJJ, line 85: `sholud` 🙃

includes/admin/tools/class-edd-tools-recount-single-customer-stats.php
```diff
@@ -145,49 +107,16 @@ public function set_properties( $request ) {
...
-			// Find non-existing payments (deleted) and total up the purchase count
-			$purchase_count   = 0;
-			foreach( $payment_ids as $key => $payment_id ) {
-				if ( in_array( $payment_id, $removed_payments ) ) {
-					unset( $payment_ids[ $key ] );
-					continue;
-				}
-
-				$payment = edd_get_payment( $payment_id );
-				if ( apply_filters( 'edd_customer_recount_sholud_increase_count', true, $payment ) ) {
```
JJJ, line 172: `sholud` 🙃

includes/admin/tools/class-edd-tools-recount-single-customer-stats.php
```diff
@@ -145,49 +107,16 @@ public function set_properties( $request ) {
...
-			$this->delete_data( 'edd_stats_customer_pending_total' . $customer->id );
-			$this->delete_data( 'edd_recount_customer_stats_' . $customer->id );
-			$this->delete_data( 'edd_recount_customer_payments_' . $this->customer_id );
+		$customer = new EDD_Customer( $this->customer_id );
+		$customer->recalculate_stats();
 
-			$payment_ids    = implode( ',', $payment_ids );
-			$customer->update( array( 'payment_ids' => $payment_ids, 'purchase_count' => $purchase_count, 'purchase_value' => $pending_total ) );
+		$this->done    = true;
+		$this->message = __( 'Customer stats successfully recounted.', 'easy-digital-downloads' );
```
JJJ, line 117: This can be `esc_html__()` like the others that were changed to earlier.

includes/admin/tools/class-edd-tools-recount-single-customer-stats.php
```diff
@@ -145,49 +107,16 @@ public function set_properties( $request ) {
...
 		if ( ! $this->can_export() ) {
-			wp_die( __( 'You do not have permission to modify this data.', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
+			wp_die( esc_html__( 'You do not have permission to modify this data.', 'easy-digital-downloads' ), esc_html__( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
 		}
 
-		$had_data = $this->get_data();
-
-		if( $had_data ) {
-			$this->done = false;
-			return true;
```
JJJ, line 155: This method will no longer ever set `$this->done` to `false` like it does here.

It also always appears to always return `false`.

Is that OK?

includes/admin/tools/class-edd-tools-recount-single-customer-stats.php
```diff
@@ -207,92 +136,4 @@ public function export() {
...
-	 *
-	 * @since  2.5
-	 * @param  string $key The option_name to delete
-	 * @return void
-	 */
-	private function delete_data( $key ) {
-		global $wpdb;
-		$wpdb->delete( $wpdb->options, array( 'option_name' => $key ) );
-	}
-
```
JJJ, line 297: Lots of code deleted here! ⭐️⭐️⭐️⭐️⭐️

includes/class-edd-customer-query.php
```diff
@@ -45,7 +45,7 @@ class EDD_Customer_Query extends EDD\Database\Queries\Customer {
 	 *                                        or 'meta_value_num', `$meta_key` must also be provided.
 	 *                                        To sort by a specific `$meta_query` clause, use that
 	 *                                        clause's array key. Accepts 'id', 'user_id', 'name',
-	 *                                        'email', 'payment_ids', 'purchase_value', 'purchase_count',
+	 *                                        'email', 'purchase_value', 'purchase_count',
```
JJJ, line 48: Good catch deleting this!

includes/orders/functions/orders.php
```diff
@@ -127,6 +127,11 @@ function edd_trash_order( $order_id ) {
 			}
 		}
 
+		// Update the customer records when an order is trashed.
+		if ( ! empty( $order->customer_id ) ) {
+			$customer = new EDD_Customer( $order->customer_id );
+			$customer->recalculate_stats();
+		}
```
JJJ, line 134: Nice catch doing the recount here. I wonder if this was causing bugs in 2.x?

includes/orders/functions/orders.php
```diff
@@ -336,8 +341,14 @@ function edd_destroy_order( $order_id = 0 ) {
  */
 function edd_update_order( $order_id = 0, $data = array() ) {
 	$orders = new EDD\Database\Queries\Order();
+	$update = $orders->update_item( $order_id, $data );
+
+	if ( ! empty( $data['customer_id'] ) ) {
+		$customer = new EDD_Customer( $data['customer_id'] );
+		$customer->recalculate_stats();
+	}
```
JJJ, line 349: This is probably OK for now, but I have a small concern.

* Putting this here requires that `edd_update_order()` is always the function to call in order to ensure that a customer's stats get recounted when updating an order. If the `update_item()` method is called elsewhere, it won't get the same treatment.
* The customer's stats will _always_ get recounted even if data being changed in the order results in no change to the customer.

Could this maybe be hooked into a `_transition_` type action instead. Like maybe when `$data['status']` changes into or out of a status that would necessitate altering the customer stats.

Again... this good, but _perhaps_ not _completely_ optimized.

includes/orders/functions/orders.php
```diff
@@ -1282,3 +1293,62 @@ function edd_clone_order( $order_id = 0, $clone_relationships = false, $args = a
...
 }
+
+/**
+ * Get the order status array keys that can be used to run reporting related to gross reporting.
+ *
+ * @since 3.0
+ *
+ * @return array An array of order status array keys that can be related to gross reporting.
+ */
+function edd_gross_order_statuses() {
```
JJJ, line 1304: Should this be `edd_get_gross_order_statuses()`?

(I know this existed already in the 3.0 branch. Maybe bring up renaming this in a different PR?)

includes/orders/functions/orders.php
```diff
@@ -1282,3 +1293,62 @@ function edd_clone_order( $order_id = 0, $clone_relationships = false, $args = a
...
+}
+
+/**
+ * Get the order status array keys that can be used to run reporting related to net reporting.
+ *
+ * @since 3.0
+ *
+ * @return array An array of order status array keys that can be related to net reporting.
+ */
+function edd_net_order_statuses() {
```
JJJ, line 1334: Should this be `edd_get_net_order_statuses()`?

(I know this existed already in the 3.0 branch. Maybe bring up renaming this in a different PR?)

includes/payments/class-edd-payment.php
```diff
@@ -961,23 +961,6 @@ public function save() {
...
-					// Decrease the customer's purchase stats
-					$customer->decrease_value( $total_change );
-					edd_decrease_total_earnings( $total_change );
-				} elseif ( $total_change > 0 ) {
-					// Increase the customer's purchase stats
-					$customer->increase_value( $total_change );
-					edd_increase_total_earnings( $total_change );
-				}
-			}
-
```
JJJ, line 980: Good catch removing this bit 👍

includes/payments/class-edd-payment.php
```diff
@@ -2828,12 +2814,8 @@ private function maybe_alter_stats( $alter_store_earnings, $alter_customer_value
 		if ( ! empty( $this->customer_id ) ) {
 			$customer = new EDD_Customer( $this->customer_id );
 
-			if ( true === $alter_customer_value ) {
-				$customer->decrease_value( $this->total );
-			}
-
-			if ( true === $alter_customer_purchase_count ) {
-				$customer->decrease_purchase_count();
+			if ( $alter_customer_value || $alter_customer_purchase_count ) {
```
JJJ, line 2817: I like to use `! empty()` around variable checks like these.

Keep in mind also that this is changing from a strict `true ===` checks to a loose one.

includes/payments/functions.php
```diff
@@ -222,6 +215,10 @@ function edd_delete_purchase( $payment_id = 0, $update_customer = true, $delete_
 		}
 	}
 
+	if ( $customer && $customer->id && $update_customer ) {
+		$customer->recalculate_stats();
+	}
+
```
JJJ, line 221: Again, I wonder if this should have been down here the entire time, and was causing problems in 2.x.

Good catch.

includes/reports/reports-functions.php
```diff
@@ -1278,65 +1278,6 @@ function filter_items( $report = false ) {
...
-	 *
-	 * @since 3.0
-	 *
-	 * @param array $statuses {
-	 *     An array of order status array keys.
-	 *
-	 */
-	return apply_filters( 'edd_net_order_statuses', $statuses );
-}
-
```
JJJ, line 1339: Good idea to move these functions IMO 👍

tests/customers/tests-customers-db.php
```diff
@@ -217,42 +217,4 @@ public function test_legacy_remove_payment_should_return_false() {
...
-	public function test_legacy_decrement_stats_purchase_count_should_return_0() {
-		EDD()->customers->increment_stats( self::$customers[0], 10 );
-		EDD()->customers->decrement_stats( self::$customers[0], 10 );
-
-		/** @var $customer \EDD_Customer */
-		$customer = edd_get_customer( self::$customers[0] );
-
-		$this->assertSame( 0, $customer->purchase_count );
-	}
-}
```
JJJ, line 258: It's a little bit unusual to delete unit tests, even for formally deprecated functions & methods. It would be nice to know that these deprecated methods still work the way they used to in perpetuity.

tests/customers/tests-customers.php
```diff
@@ -115,32 +115,6 @@ public function test_remove_payment_should_return_true() {
...
-		$this->assertSame( 90.0, self::$customers[4]->purchase_value );
-	}
-
-	public function test_decrease_purchase_count_should_return_0() {
-		self::$customers[3]->increase_purchase_count();
-		self::$customers[3]->decrease_purchase_count();
-
-		$this->assertSame( 0, self::$customers[1]->purchase_count );
-	}
-
```
JJJ, line 143: See above. Would be nice to keep these tests unless the methods are completely going away.

tests/orders/tests-edd-payment.php
```diff
@@ -647,55 +647,6 @@ public function test_user_id_mismatch() {
...
-		// Store sales are based off 'publish' & 'revoked' status. So it reduces this count
-		$this->assertEquals( $store_sales - 1, edd_get_total_sales() );
-
-		remove_filter( 'edd_decrease_earnings_on_undo', '__return_false' );
-		remove_filter( 'edd_decrease_sales_on_undo', '__return_false' );
-		remove_filter( 'edd_decrease_customer_value_on_pending', '__return_false' );
-		remove_filter( 'edd_decrease_customer_purchase_count_on_pending', '__return_false' );
-		remove_filter( 'edd_decrease_store_earnings_on_pending', '__return_false ' );
-	}
-
```
JJJ, line 698: See above.

tests/orders/tests-edd-payment.php
```diff
@@ -731,51 +682,6 @@ public function test_pending_affecting_stats() {
...
-		// Store sales are based off 'publish' & 'revoked' status. So it reduces this count
-		$this->assertEquals( $store_sales - 1, edd_get_total_sales() );
-
-		remove_filter( 'edd_decrease_earnings_on_undo', '__return_false' );
-		remove_filter( 'edd_decrease_sales_on_undo', '__return_false' );
-		remove_filter( 'edd_decrease_customer_value_on_refund', '__return_false' );
-		remove_filter( 'edd_decrease_customer_purchase_count_on_refund', '__return_false' );
-		remove_filter( 'edd_decrease_store_earnings_on_refund', '__return_false ' );
-	}
-
```
JJJ, line 778: See above.

tests/reports/tests-reports-functions.php
```diff
@@ -792,7 +792,7 @@ public function test_clear_filter_should_default_to_last_30_days() {
 
 	public function test_gross_order_status() {
 		$expected = array(
-			'completed',
+			'complete',
```
JJJ, line 795: I know you asked about this in Slack, so I'm just noting here to make doubly sure `complete` is correct. 📝

tests/reports/tests-reports-functions.php
```diff
@@ -803,7 +803,7 @@ public function test_gross_order_status() {
 
 	public function test_net_order_status() {
 		$expected = array(
-			'completed',
+			'complete',
```
JJJ, line 806: See above.

includes/admin/tools/class-edd-tools-recount-customer-stats.php
```diff
@@ -61,57 +61,8 @@ public function get_data() {
...
-					foreach ( $payment_ids as $payment_id ) {
-
-						$payment = edd_get_payment( $payment_id );
-
-						$should_process_payment = 'complete' == $payment->status || 'revoked' == $payment->status ? true : false;
-						$should_process_payment = apply_filters( 'edd_customer_recount_should_process_payment', $should_process_payment, $payment );
-
-						if( true === $should_process_payment ) {
-
-							if ( apply_filters( 'edd_customer_recount_sholud_increase_value', true, $payment ) ) {
```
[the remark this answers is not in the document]
robincornett: I guess removing a filter completely is one way to fix a typo?

includes/orders/functions/orders.php
robincornett: Well, it's a brand new function and so far this PR is the only place it's used that I know of. It's moved in this PR because I was trying to use it and discovered it was in a namespaced file. I would be open to changing both of these to include `get` if @ashleyfae approves?

tests/customers/tests-customers-db.php
robincornett: The tests are deleted per [this comment](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7947#issuecomment-641486431).

robincornett: I expect new tests may need to be written but I do not know how.

includes/orders/functions/orders.php
robincornett: It's a great question, and I'm honestly not sure. I feel like there are plenty of places that `edd_update_order` is used, but not every usage includes the customer ID. As I was looking for places where the order status might change, and tucking the recalculation call in each, it seemed like a lot. Putting it here felt like the logical place to me because it was really the last gateway, but I think it's quite likely the recalculation could be running more than once in the process of changing an order because of it now being in multiple places. Definitely worth looking into more.

includes/admin/tools/class-edd-tools-recount-single-customer-stats.php
robincornett: Updated to be more consistent with how it was and the bulk customer recount tool. I am not sure it matters because the progress bar doesn't actually show anything either way, since the new recounting is not incremental.

tests/customers/tests-customers-db.php
JJJ: Unit tests are fuuuuuuuuuuuuunnnn! 🥳

It's completely OK to not know how, and it's also completely OK to not want to write them once you do know how. We can talk privately about them more and come up with a plan. 🌻

LisaCee:
This worked for the reproduction workflow described in #7947 

But... and this may be outside the scope of this issue/PR:
If I have a customer with multiple items in 1 order and I do a partial refund, the lifetime value doesn't update correctly.

image

image


Also, did we lose the tiles on Reports at some point?  This is not specific to this issue, but in testing it I noticed the change.  
image

robincornett:
> 
> 
> But... and this may be outside the scope of this issue/PR:
> If I have a customer with multiple items in 1 order and I do a partial refund, the lifetime value doesn't update correctly.

I think this is in the scope of this PR, yes. I feel like I'm tucking calls to `recalculate_stats` in multiple places, as I find them, but with the various ways an order can be modified (credits, discounts, refunds, trash, partial refunds), I am not feeling like I'm getting that done close enough to the database. For example, it's been added to `edd_trash_order`, but not `edd_add_order` ... on the other hand, we don't necessarily want to recount every time an order action happens, either, because in the case of processing a refund, two orders are being touched (the original and the refund is a new order being created). I'll look into this more.

robincornett:
Update: I think the issue is not necessarily in when `recalculate_stats` is being called--although that's still something to look into. The issue is in the `recalculate_stats` method itself. It's looking for `sale` only for the type of orders, and then statuses of `complete`, `partially_refunded`, and `revoked` (the net order statuses). So if an order is partially refunded, the recounter is looking at its value, which is original order value, and not including the refund at all.

It _may_ be correct to use only those statuses and type to get the customer's lifetime counts, but not to get their lifetime value, I think.

ashleyfae:
Failing unit test:

> 1) EDD\Orders\EDD_Payment_Tests::test_refund_affecting_stats
> Failed asserting that 0.0 matches expected -120.0.

I believe the actual way the _test_ is written is the issue.

What's happening here is:

- Customer starts out with 0 orders and 0 earnings. `$customer_sales` = `0`; `$customer_earnings` = `0`
- Customer makes a $120 purchase. NOTE: The `$customer_sales` variable is NOT updated at this point.
- That purchase is then refunded.
- The unit test then asserts:

```php
$this->assertEquals( $customer_earnings - $this->payment->total, $customer->purchase_value );
```

But `$customer_earnings` was never refreshed anywhere. So that's still `0`. So it's doing `0` - `120` and coming out with `-120`, which is never a value we want to end up with.  `$customer->purchase_value` is giving us `0`, which is correct, whereas the test is asserting `-120`, which is incorrect.

@robincornett 
> Question: if a refund is not counted as part of the customer's total counts, what about the original purchase, if it has been fully refunded? I feel like there is a difference between a customer with 1 completed sale and one with 1 completed sale but 10 refunds, for example, but if we count only complete and partially refunded sales, they will show the same lifetime purchase count.

Two examples, using the same $10 product:

- Customer A
    - 1 completed sale, 0 refunds
    - Purchase Count = 1; Purchase Value = $10
- Customer B
    - 1 completed sale, 5 refunded sales
    - Purchase Count = 1; Purchase Value = $10

I think those are the numbers we're expecting to have. (Someone correct me if I'm wrong.) I think the goal here is: _"How much money have I earned from this customer, and from how many orders?"_ So "Purchase Count" is looking at the number of successful orders the customer has made, where you ultimately didn't have to send back the money completely (which is why we exclude `refunded`). And "Purchase Value" is looking within the same parameters, but also being sure to subtract any partial refunds. It's representing the total money collected _and kept_ from that customer.

robincornett:
Thanks--I did some testing on 2.9 and the behavior you describe is what we do there. It makes sense, although personally I view Customer A and Customer B differently. But it makes sense and I think the current iteration is now handling this the same way.

cklosowski:
I agree with @ashleyfae on the final point here:

> Two examples, using the same $10 product:

    Customer A
        1 completed sale, 0 refunds
        Purchase Count = 1; Purchase Value = $10
    Customer B
        1 completed sale, 5 refunded sales
        Purchase Count = 1; Purchase Value = $10

> I think those are the numbers we're expecting to have. (Someone correct me if I'm wrong.) I think the goal here is: "How much money have I earned from this customer, and from how many orders?" So "Purchase Count" is looking at the number of successful orders the customer has made, where you ultimately didn't have to send back the money completely (which is why we exclude refunded). And "Purchase Value" is looking within the same parameters, but also being sure to subtract any partial refunds. It's representing the total money collected and kept from that customer.

includes/orders/functions/refunds.php
```diff
@@ -373,9 +373,13 @@ function edd_refund_order( $order_id = 0, $order_items = array() ) {
...
 		: 'partially_refunded';
 
-	edd_update_order( $order_id, array(
-		'status' => $order_status,
-	) );
+	edd_update_order(
+		$order_id,
+		array(
+			'status'      => $order_status,
+			'customer_id' => $order->customer_id,
```
ashleyfae, line 380: Why is this here now?

robincornett: If the `customer_id` is not passed through to the `edd_update_order` function, the customer's stats are not updated (commit [cc4f12c](https://github.com/easydigitaldownloads/easy-digital-downloads/commit/cc4f12ce372fc4127599ae228e41cc3218665e49))

ashleyfae: Ah I see. Makes sense. Thanks!

ashleyfae (approved):
This has fixed a ton of bad numbers in my database! Previously the lifetime values were including failed orders and things like that. Not anymore!

Well done @robincornett .

ashleyfae:
I did re-add the unit tests covering increasing and decreasing stats just because the methods do still exist and actually make the changes. Therefore the unit tests do still work, so I think we can keep them a while longer.

PR #7970, opened 2020-05-30
implode(): Passing glue string after array is deprecated.

sultann:
Fixes #
Since PHP 7.4 implode(): Passing glue string after the array is deprecated. So, I fixed with swapping the arguments.


ashleyfae:
Thank you @sultann !

This has already been fixed in version 3.0, but given that we don't have a release date on that yet, I see no harm in getting this in a 2.9.x release as well.

Issue #7629, opened 2020-03-05
3.0 - Auto save tax rate changes


nosegraze:
### Explain your enhancement (please be detailed)

It would be nice if tax rate changes were automatically saved and didn't require clicking the `Save Changes` button. Example:

- When adding a new tax rate, it should be saved in the database as soon as it appears in the table.
- When deactivating or activating a tax rate, clicking the link should make that change in the database.

### Justification or use case

I love the new tax rate UI. It's super slick and super user friendly. However, it feels so "react-y" that it makes me think my changes have been saved when they haven't. I add a tax rate, it appears in the table, and my mind fells me that it's been "saved". But in reality it hasn't. I deactivate a tax rate, it disappears, and my mind tells me that it's been saved. But it hasn't.

Every time I've used this feature, I've made the mistake of thinking my data has been saved. I added a tax rate (didn't save), added something to my cart, and got confused when taxes weren't being added. I'd forgotten to save.

I deactivated a tax rate, I added something to my cart, and got confused when taxes _were_ being added. I'd forgotten to save.

I think it doesn't help that the `Save Changes` button feels far away from the actual table, and thus doesn't feel related. This is due to the "indenting" of the tax rate table.

![Screenshot from 2020-03-05 16-48-56](https://user-images.githubusercontent.com/6324272/76004369-4bd3d000-5f01-11ea-9e88-041996a84318.png)

spencerfinnell:
@nosegraze There should be an "Are you sure" when you leave the tab if changes are unsaved, but you of course need to leave first to realize it hasn't saved.

Previously https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6902#issuecomment-418135312

Autosaving was a future enhancement when the new UI was added -- maybe something good for an earlish 3.x?

nosegraze:
> @nosegraze There should be an "Are you sure" when you leave the tab if changes are unsaved, but you of course need to leave first to realize it hasn't saved.

Yeah in my cases I had two tabs open. In one I made the tax rate changes and didn't try to navigate away, so I never got that notice. I just switched to the other tab for my checkout process.

spencerfinnell:
I think this is slightly less likely to happen when not testing a bunch of things in wp-admin but it happened to @ashleyfae again so this is certainly something we should keep an eye on; especially if it starts affecting support.

spencerfinnell:
[...]

PR #7973, opened 2020-06-02
3.0 - Orders - Update Tax Rate When Selecting Existing Address

### LINKED ISSUE #7958: 3.0 - Selecting an existing address for a manual order does not retrieve the region
### Expected behavior
When I manually create an order and select a customer's preexisting address, if it is within a region which collects taxes, I expect the taxes to be applied.

### Actual behavior
When I select the address, the `getTaxRate` JS function retrieves the country, but not the region/state, so the tax rate is not applied.

1) In an EDD store with taxes enabled for a region (not country wide), create a new order.
2) The order should be for an existing customer with an existing address which meets the criteria for being taxable (so I picked a customer with a TN address).
3) Add a product to the order. It will show a 0.00% tax rate.
4) Steps 2 and 3 can be done in reverse order; it doesn't seem to change it.

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

I thought we had this resolved at one point ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7781#discussion_r414800661)), but maybe that applied to addresses being typed in, not retrieving an existing address.

spencerfinnell:
Fixes #7958 

Proposed Changes:
1. Ensure tax rate is updated correctly when selecting an existing address.
2. Allow address to be cleared by selecting blank existing address.


includes/ajax-functions.php
```diff
@@ -1227,8 +1233,8 @@ function edd_ajax_customer_addresses() {
 
 			// Fetch the select
 			if ( ! empty( $options ) ) {
-				$html  = '<select data-nonce="' . wp_create_nonce( 'edd-country-field-nonce' ) . '" data-placeholder="Select a previously used address" class="edd-select-chosen ' . esc_attr( $rtl_class ) . ' add-order-customer-address-select">';
-				$html .= '<option></option>';
+				$html  = '<select id="edd_customer_existing_addresses" data-nonce="' . wp_create_nonce( 'edd-country-field-nonce' ) . '" data-placeholder="Select a previously used address" class="edd-select-chosen ' . esc_attr( $rtl_class ) . ' add-order-customer-address-select">';
```
robincornett, line 1236: I don't think the `edd-select-chosen` class is necessary here--that script is not implemented or applied.

robincornett:
@spencerfinnell I just realized that on smaller screens, the `#edd_customer_existing_addresses` select overflows the bounds of the container. It looks like the core `select` CSS sets a max-width, but not a width, which may be the issue here.

spencerfinnell:
@robincornett I added a `max-width` rule for `select.edd-form-group__input` in https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7973/commits/65eeb93b0b02c49721e6b22c25676f4bcf73a18f

robincornett:
Yes, that fixes it, thanks!

includes/ajax-functions.php
```diff
@@ -538,9 +538,15 @@ function edd_ajax_get_states_field() {
 	// Maybe setup the new listbox.
 	if ( ! empty( $states ) ) {
 		$field_name = sanitize_text_field( $_POST['field_name'] );
+		$field_id   = sanitize_text_field( $_POST['field_id'] );
```
ashleyfae, line 541: This causes the following notice when adding a tax rate in settings

```
PHP Notice:  Undefined index: field_id in /var/www/edd3/wp-content/plugins/easy-digital-downloads/includes/ajax-functions.php on line 541
```

1. Go to Downloads > Settings > Taxes.
2. Select a country via the dropdown.
3. Notice triggers when states are loaded for that country.

spencerfinnell: @ashleyfae Updated. I also added a check and fallback for `field_name` which should never be empty but could be with incorrect implementation.

ashleyfae:
Someone help me... what am I doing wrong.. lol

## Tax rate:

Screenshot from 2020-06-03 14-37-24

## Order address

Screenshot from 2020-06-03 14-38-01

## Order tax (incorrect)

Screenshot from 2020-06-03 14-38-28

Should be 20% tax.

spencerfinnell:
@ashleyfae Were you stung by https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7629 -- the rate should say "Deactivate" -- it stays on "Removed" until settings are saved.

ashleyfae:
> @ashleyfae Were you stung by #7629 -- the rate should say "Deactivate" -- it stays on "Removed" until settings are saved.

sdfsiokdsjfoswpfijoifwj

You're absolutely right. FML.

After _actually saving_ this works perfectly.

PR #7974, opened 2020-06-02
Report Tiles Incorrect #7921

### LINKED ISSUE #7921: 3.0 - Customer report screen may be incorrect
### Expected behavior
I expect I should understand what the numbers mean for a report when I'm looking at it.

### Actual behavior
I don't really understand the tiles for the Customer reports, and I don't think some of the numbers are accurate. Example:

image

The average number of orders is more than are in my test storefront. I am not sure what the average age number is supposed to be, so I can't tell if that's correct or weird. I will assume the other two numbers are correct, though.

1) In EDD 3.0 with quite a few orders/customers, visit Downloads > Reports > Customers.

PHP Version: any
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

LisaCee:
Fixes #7921 

Proposed Changes:
1. I removed the `Average Age` tile and endpoint based on [this comment](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7921#issuecomment-630328620)
2. Changed name of the `$accepted_functions` argument to match the name of the AVG function name

**Steps to reproduce the behavior**
Visit Downloads > Reports > Customers
   - Average customer age should be gone
   - Average number of orders should be correct

I don't have a ton of customer data to pull from, but I _think_ the average number of orders is correct.  Please check in your own stores.  If it is not correct, I can just pull out the tile until I / we can figure out where that number is going sideways.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -2312,6 +2312,8 @@ function edd_register_customer_report( $reports ) {
 				'tiles'  => array(
 					'lifetime_value_of_customer',
 					'average_customer_value',
+					// this line removes tile
+					'average_number_of_orders_per_customer',
 					'customer_average_age',
 					'most_valuable_customer',
 				),
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -2312,7 +2312,6 @@ function edd_register_customer_report( $reports ) {
 				'tiles'  => array(
 					'lifetime_value_of_customer',
 					'average_customer_value',
-					'average_number_of_orders_per_customer',
 					'customer_average_age',
 					'most_valuable_customer',
 				),
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -2311,9 +2311,8 @@ function edd_register_customer_report( $reports ) {
 			'endpoints' => array(
 				'tiles'  => array(
 					'lifetime_value_of_customer',
-					'average_customer_value',
 					'average_number_of_orders_per_customer',
-					'most_valuable_customer',
+					'average_customer_value',
 				),
 				'tables' => array(
 					'top_five_customers',
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -2312,9 +2312,7 @@ function edd_register_customer_report( $reports ) {
 				'tiles'  => array(
 					'lifetime_value_of_customer',
 					'average_customer_value',
-					// this line removes tile
 					'average_number_of_orders_per_customer',
-					'customer_average_age',
 					'most_valuable_customer',
 				),
 				'tables' => array(
@@ -2373,20 +2371,6 @@ function edd_register_customer_report( $reports ) {
 						return apply_filters( 'edd_reports_customers_average_order_count', $stats->get_customer_order_count( array(
 							'function' => 'AVG',
 						) ) );
-					},
-				),
-			),
-		) );
-
-		$reports->register_endpoint( 'customer_average_age', array(
-			'label' => __( 'Average Age', 'easy-digital-downloads' ),
-			'views' => array(
-				'tile' => array(
-					'data_callback' => function () {
-						global $wpdb;
-						$average_value = (int) $wpdb->get_var( "SELECT AVG(DATEDIFF(NOW(), date_created)) AS average FROM {$wpdb->edd_customers}" );
-
-						return apply_filters( 'edd_reports_customers_average_age', $average_value . ' ' . __( 'days', 'easy-digital-downloads' ) );
 					},
 				),
 			),
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -2086,7 +2086,6 @@ class Stats {
 	 * @return int Number of orders made by a customer.
 	 */
 	public function get_customer_order_count( $query = array() ) {
-
 		// Add table and column name to query_vars to assist with date query generation.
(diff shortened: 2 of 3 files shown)


includes/class-stats.php
```diff
@@ -2096,7 +2096,7 @@ public function get_customer_order_count( $query = array() ) {
 		$this->pre_query( $query );
 
 		// Only `COUNT` and `AVG` are accepted by this method.
-		$accepted_functions = array( 'COUNT', 'AVG' );
+		$accepted_functions = array( 'COUNT', 'AVG(id)' );
 
 		$function = isset( $this->query_vars['function'] ) && in_array( strtoupper( $this->query_vars['function'] ), $accepted_functions, true )
 			? $this->query_vars['function'] . "({$this->query_vars['column']})"
```
spencerfinnell, line 2102: @LisaCee

Now, with your change the [existing stats usage](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports.php#L2373) will still not use the correct query:

```php
$stats = new EDD\Stats();
$stats->get_customer_order_count( array(
  'function' => 'AVG',
) );
```

This does not pass the `in_array()` check above now (`AVG` is not in `COUNT` or `AVG(id)`).

Instead I think we could keep the array of accepted functions the same and if the passed function string passes validation set that as the value, falling back to an empty string. If the passed function is not considered valid/accepted and we end up with an empty string we can just return early with a `0` value because we don't have a way of determining which SQL query to run.

The original implementation of this stats method half written to be flexible (referencing dynamic column values in `$this->query_vars['column']` and dynamic function values in `$this->query_vars['function']`) with the other half uses very static comparisons and SQL statements. 

Because we know the SQL calculations for determining the average or total counts can really only happen one way I think it makes sense to remove some of the "dynamicness" if this method and hard code the calculation portions of the SQL queries (which is already happening for `AVG`) and simplify the checks for the type of count we want.

includes/class-stats.php
```diff
@@ -2123,7 +2123,7 @@ public function get_customer_order_count( $query = array() ) {
 					) o
 					WHERE 1=1 {$this->query_vars['status_sql']} {$user} {$customer} {$email} {$this->query_vars['where_sql']} {$this->query_vars['date_query_sql']}";
 		} else {
-			$sql = "SELECT {$function}
+			$sql = "SELECT COUNT(id)
```
robincornett, line 2126: This is throwing a database error for me, but from what I can tell, it's actually due to the use of undefined variables in lines 2119 and 2122 above (`$this->query` should be `$this->query_vars`).

The error reads:
`WordPress database error You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') o
					WHERE 1=1 AND wp_edd_orders.status IN ('complete', 'revoked')' at line 6 for query SELECT COUNT(id) / total_customers AS average
					FROM wp_edd_orders
					CROSS JOIN (
						SELECT COUNT(DISTINCT customer_id) AS total_customers
						FROM 
					) o
					WHERE 1=1 AND wp_edd_orders.status IN ('complete', 'revoked')      made by require_once('wp-admin/admin.php'), do_action('download_page_edd-reports'), WP_Hook->do_action, WP_Hook->apply_filters, edd_reports_page, edd_reports_sections, EDD\Admin\Reports_Sections->display, EDD\Admin\Sections->get_all_section_contents, call_user_func_array, edd_output_report_callback, EDD\Reports\Data\Report->display, call_user_func, EDD\Reports\default_display_report, EDD\Reports\Data\Report->display_endpoint_group, call_user_func, EDD\Reports\default_display_tiles_group, EDD\Reports\Data\Tile_Endpoint->display, EDD\Reports\Data\Endpoint->display, EDD\Reports\Data\Endpoint->get_data, call_user_func, {closure}, EDD\Stats->get_customer_order_count`

and then a PHP notice (same for the two lines I mentioned):

`PHP Notice:  Undefined property: EDD\Stats::$query in \plugins\easy-digital-downloads.30\includes\class-stats.php on line 2119`

I adjust those, your updated code shows an average number of orders which makes sense to me.

LisaCee: Great catch!  I'm still getting in the habit of checking logs.

includes/admin/reporting/reports.php
```diff
@@ -2313,7 +2313,6 @@ function edd_register_customer_report( $reports ) {
 					'lifetime_value_of_customer',
 					'average_customer_value',
 					'average_number_of_orders_per_customer',
-					'customer_average_age',
 					'most_valuable_customer',
```
spencerfinnell, line 2316: While we are here `most_valuable_customer` is not a valid tile and this can be removed as well.

cklosowski:
Is it correct that we are also removing the `most_valuable_customer` tile? Just curious off hand.

LisaCee:
@spencerfinnell said it's not a valid tile, so I removed it.  https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7974#discussion_r437696577

spencerfinnell:
@cklosowski as far as I can tell there was never an endpoint added for `most_valuable_customer`. `most_valuable_customers` (plural) is a registered endpoint for the table of Most Valuable Customers but no tile for `most_valuable_customer`.

https://github.com/easydigitaldownloads/easy-digital-downloads/blame/bee5562ad0205fec6b5ef333815f42e6bbdd55b3/includes/admin/reporting/reports.php#L2317

cklosowski:
:+1: Ok just making sure it was an intentional change.

PR #7983, opened 2020-06-05
Update download labels for reports

### LINKED ISSUE #7972: 3.0 - Downloads report is sometimes unable to get the name for a single variation
### Expected behavior
When I view a report for a specific download/variation, I expect to see a complete report with no errors.

### Actual behavior
For some of my products with variations, but not all, when viewing the sales report for individual variations, I get PHP notices:

```
Notice | Undefined offset: 0 /plugins/easy-digital-downloads.30/includes/admin/reporting/reports.php:496
Notice | Undefined offset: 0 /plugins/easy-digital-downloads.30/includes/admin/reporting/reports.php:1570
Notice | Undefined offset: 0 /plugins/easy-digital-downloads.30/includes/admin/reporting/reports.php:1711
```

Not 100% sure, but my two affected products are from the EDD sample XML file, so:
1) in 3.0, import the EDD `sample-products-import.xml` file
2) make sales
3) visit Downloads > Reports > Downloads and try filtering the report by a variation for a product.
4) if you are able to trigger the notice, you will still see a graph, but the variation name will not be appended to the product name:

image

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

This does not happen with every product with variations--I have one which I created manually, and it's fine, and there is a sample music album from the import which also does not trigger a notice.

robincornett:
Fixes #7972

Proposed Changes:
1. Use `edd_get_price_name` to get price option names, if viewing a price ID.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -489,8 +489,11 @@ function edd_register_downloads_report( $reports ) {
 		if ( $download_data ) {
 			$download = edd_get_download( $download_data['download_id'] );
 			if ( $download_data['price_id'] ) {
-				$args                  = array( 'price_id' => $download_data['price_id'] );
-				$download->post_title .= ': ' . edd_get_price_name( $download->ID, $args );
+				$args       = array( 'price_id' => $download_data['price_id'] );
+				$price_name = edd_get_price_name( $download->ID, $args );
+				if ( $price_name ) {
+					$download->post_title .= ': ' . $price_name;
+				}
 			}
 			$download_label = esc_html( ' (' . $download->post_title . ')' );
 		}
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/reports.php
```diff
@@ -485,18 +485,14 @@ function edd_register_downloads_report( $reports ) {
...
-			$prices   = $download->get_prices();
-
 			if ( $download_data['price_id'] ) {
-				$prices = array_values( wp_filter_object_list( $download->get_prices(), array( 'index' => absint( $download_data['price_id'] ) ) ) );
-
-				$download_label = esc_html( ' (' . $download->post_title . ': ' . $prices[0]['name'] . ')' );
-			} else {
-				$download_label = esc_html( ' (' . $download->post_title . ')' );
+				$args                  = array( 'price_id' => $download_data['price_id'] );
+				$download->post_title .= ': ' . edd_get_price_name( $download->ID, $args );
```
ashleyfae, line 493: `edd_get_price_name()` could return `false` so we should probably be checking if the return value isn't empty before appending it.

ashleyfae (approved):

PR #7984, opened 2020-06-05
Deactivate any country wide tax rates when importing a country wide tax rate that is for the same country #7602

### LINKED ISSUE #7602: 3.0 - Taxes - Multiple Country Wide Taxes can exist on migration
### Expected behavior
When migrating multiple tax rates at a country level, the most recent should be set as active, and the remaining country wide tax rates for a specific country should be marked as ‘deactivated’.

### Actual behavior
Found in #6037 
When migrating from EDD 2.9.x, where it was possible to have multiple country wide taxes (bug), we still migrate both active tax rates as active tax rates.

1) On EDD 2.9.x, setup two tax rates that are for the same country and set to be country wide.
2) Migrate to EDD 3.0
3) You’ll have multiple country wide tax rates for the same country

PHP Version: All

WordPress Version: All

cklosowski:
Fixes #7602 

Proposed Changes:
1. In the `Data_Migrator::tax_rates` method, check for any other country wide tax rates for the same country and deactivate them.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -1085,6 +1085,27 @@ class Data_Migrator {
 			? sanitize_text_field( $data['state'] )
 			: '';
 
+		// If the scope is 'country', look for other active rates that are country wide and set them as 'inactive'.
+		if ( 'country' === $scope ) {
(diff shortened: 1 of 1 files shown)


cklosowski:
@robincornett that is best handled in #7629. This issues is specifically about the import process, so we can keep the changes isolated to that part of the process.

PR #7986, opened 2020-06-05
Add gross/net order status functions to aid in consistent reporting #7985

### LINKED ISSUE #7985: 3.0 - Add gross/net status functions with filters.
We need to add two functions (at a minimum) that just return an array of order statuses. The two functions should be related to returning:

- Statuses that would affect `net` revenue
- Statuses that would affect `gross` revenue

These functions should return a filterable array of status keys.

### Justification or use case
These two functions will help us generate more consistent repoting.

cklosowski:
Fixes #7985 

Proposed Changes:
1. Introduces the `edd_gross/net_order_statuses()` functions into the reporting functions.
2. Introduces two filters to allow filtering these statuses (with doc blocks)
3. Adds unit tests for the two functions.

### DIFF
--- a/includes/reports/reports-functions.php
+++ b/includes/reports/reports-functions.php
@@ -887,7 +887,7 @@ function default_display_tiles_group( $report ) {
 
 	<div id="edd-reports-tiles-wrap" class="edd-report-wrap">
 		<?php
-		foreach ( $tiles as $endpoint_id => $tile ) : 
+		foreach ( $tiles as $endpoint_id => $tile ) :
 			$tile->display();
 		endforeach;
 		?>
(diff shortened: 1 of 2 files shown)


includes/reports/reports-functions.php
```diff
@@ -1278,6 +1278,65 @@ function filter_items( $report = false ) {
...
+ *
+ * @since 3.0
+ *
+ * @return array An array of order status array keys that can be related to net reporting.
+ */
+function edd_net_order_statuses() {
+	$statuses = array(
+		'completed',
+		'partially_refunded',
+		'revoked',
```
robincornett, line 1322: Should revoked really be included in the net statuses? I would have thought just `completed` and `partially_refunded`.

ashleyfae: > Revoked: access to purchased items is disabled, perhaps due to policy violation or fraud.

I don't know for sure, but it sounds like maybe this is a step you might use before refunded? You revoke while investigating, then later you might actually refund (or not). So at this point, while it's `revoked`, you haven't given the money back, so because you still have it, it would make sense to be in net?

robincornett: Thanks, that makes sense--I didn't realize the financial side of that status.

cklosowski: Yep, the point of these are statuses as related to where the funds exist. You can use a Revoked if someone is violating your TOS, and is not entitled to a refund, so the money stays within your account as a store owner.

PR #7988, opened 2020-06-06
Payment key and helper func #7529

### LINKED ISSUE #7529: 3.0 - Orders - Key Not Generated for Manually Created Order
### Expected behavior

When manually creating an Order and leaving the "Key" field blank, a key should be generated.

### Actual behavior

No key is generated.

1) Switch to `release/3.0`
2) Manually create a new order.

When a Refund is created a key is automatically generated and attached to that record.

LisaCee:
Fixes #7529

Proposed Changes:
1. Created helper function modeled after [this](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/process-purchase.php#L125) for `$payment_key`
2. Added `$payment_key` value to `$order_id` array
3. Adjusted the array spacing per phpcs


cklosowski:
@spencerfinnell If I'm not mistaken, we wanted to get a function to generate this key right? So we can reuse the methods of creating a payment key without duplicating the code twice for manual and user-generated orders?

If we do make this a function, we have the opportunity to also add in a filter to it.

includes/orders/functions/actions.php
```diff
@@ -151,22 +151,24 @@ function edd_add_manual_order( $args = array() ) {
...
+		array(
+			'status'       => 'pending', // Always insert as pending initially.
+			'user_id'      => $user_id,
+			'customer_id'  => $customer_id,
+			'email'        => $email,
+			'ip'           => sanitize_text_field( $data['ip'] ),
+			'gateway'      => sanitize_text_field( $data['gateway'] ),
+			'mode'         => $mode,
+			'currency'     => edd_get_currency(),
+			'payment_key'  => edd_generate_order_payment_key( $email ),
```
spencerfinnell, line 164: @LisaCee The function is working great to generate a key. However when adding an Order in the admin we have the `Key` field available to enter a custom one:

Screen Shot 2020-06-11 at 2 18 06 PM
 
We'll need to check if `$data['payment_key']` is empty and only use a generated key if it is -- otherwise use the input value stored in `$data['payment_key']` that needs to be sanitized via `sanitize_text_field()`

Issue #7946, opened 2020-05-20
3.0 - Create a reusable "show on mobile" class to enhance labels


robincornett:
### Explain your enhancement (please be detailed)
Per the [discussion regarding order screen headings](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7938#issuecomment-631508326), create an EDD utility class which would:
- display on mobile
- be hidden visually on larger screens, but still accessible to assistive technology

### Justification or use case
Specifically, the orders screen shows a customer section with vertical tabs. For visual users, the navigation labels are visible on larger screens, but not on smaller screens. The associated tab content does not include a heading to label the content, so adding a heading which would be visible on smaller screens would be very helpful.

This is especially applicable to the reports screens, as the reports do not all have unique icons, and on smaller screens, only the tiles and graphs give an idea of what context the screen/report being displayed is.

spencerfinnell:
We currently have a _ton_ of seemingly arbitrary media queries:

```css
@media only screen and ( max-width: 782px ) {
@media all and ( min-width: 1080px ) {
@media all and ( min-width: 1240px ) {
@media (min-width: 280px) {
@media handheld, only screen and (max-width: 640px) {
@media screen and ( max-width: 1284px ) {
@media screen and ( max-width: 1024px ) {
@media screen and ( max-width: 782px ) {
@media screen and ( max-width: 600px ) {
@media handheld, only screen and (max-width: 1000px) {
@media screen and ( max-width: 1100px ) {
@media screen and ( max-width: 810px ) and ( min-width: 656px ) {
@media screen and ( max-width: 781px ) {
@media screen and ( max-width: 656px ) {
@media screen and ( max-width: 480px ) {
@media only screen and ( min-width: 1200px ) {
@media only screen and ( max-width: 782px ) {
```

We'll want to find out the current set core is using and create the new classes around those standards -- and eventually clean up our CSS.

robincornett:
A CSS reorganization might be helpful as well, to try to get CSS for specific screens grouped as much as possible.

Switching to Sass could allow for logical grouping in the readable CSS file, but efficient grouping in the compiled (using a media query packer type processor).

robincornett:
[...]

PR #7990, opened 2020-06-08
Update tax rates table markup and CSS for improved responsiveness

### LINKED ISSUE #7957: 3.0 - Tax rate tables display should be updated for responsiveness
The CSS (and markup, at least of the `tfoot` section) should be updated to display better on smaller screens.

### Justification or use case
Once a user's screen size is narrower than 1100px, the fixed/percentage widths in the CSS cause display issues, especially for the Rate/Actions columns. On iPad/smaller screens, the default WP behavior to set table cells to display as blocks causes the cells to line up vertically, although the table header does not:

image

Markup: there is an empty `<th>` in the `<tfoot>` which could be removed, and the country form given a little more space--I think it would help visually, especially on the smaller screens, without impacting functionality.

Not markup related, but worth noting, I think: once a tax rate is added, it can be deactivated, but it appears that there is no way to delete it.

robincornett:
Fixes #7957

Proposed Changes:
1. Update table markup for consistency with Core tables.
2. Add `data-colname` attributes to rows for managing small screen output.

The input labels for the "add new" row are currently hidden with `screen-reader-text`, which I'm currently overriding on mobile screens so they can be visible for all users. This is a good use case for the reusable "show on mobile" CSS class requested in #7946.

### DIFF
--- a/includes/admin/views/tmpl-tax-rates-table-row.php
+++ b/includes/admin/views/tmpl-tax-rates-table-row.php
@@ -15,12 +15,12 @@
 	<input type="checkbox" <# if ( data.selected ) { #>checked<# } #> />
 </th>
 
-<td class="edd-tax-rates-table-country" data-colname="<?php esc_html_e( 'Country', 'easy-digital-downloads' ); ?>">
+<td class="edd-tax-rates-table-country" data-colname="<?php esc_attr_e( 'Country', 'easy-digital-downloads' ); ?>">
 	{{ data.country }}
 	<input type="hidden" name="tax_rates[{{ data.id }}][country]" value="{{ data.country }}" />
 </td>
 
-<td data-colname="<?php esc_html_e( 'Region', 'easy-digital-downloads' ); ?>">
+<td data-colname="<?php esc_attr_e( 'Region', 'easy-digital-downloads' ); ?>">
 	<# if ( data.global ) { #>
 	&mdash;
 	<input type="hidden" name="tax_rates[{{ data.id }}][global]" value="{{ data.global }}" />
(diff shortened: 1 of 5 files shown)


robincornett:
@LisaCee nice catch on the borders, not sure how I missed that! I've made changes to the bulk actions as well, but it involved updating markup--see what you think?

includes/admin/views/tmpl-tax-rates-table-row.php
```diff
@@ -11,16 +11,16 @@
...
 ?>
 
-<td class="edd-tax-rates-table-checkbox">
+<th class="edd-tax-rates-table-checkbox check-column">
 	<input type="checkbox" <# if ( data.selected ) { #>checked<# } #> />
-</td>
+</th>
 
-<td>
+<td class="edd-tax-rates-table-country" data-colname="<?php esc_html_e( 'Country', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 18:
```suggestion
<td class="edd-tax-rates-table-country" data-colname="<?php esc_attr_e( 'Country', 'easy-digital-downloads' ); ?>">
```

includes/admin/views/tmpl-tax-rates-table-row.php
```diff
@@ -11,16 +11,16 @@
...
 	{{ data.country }}
 	<input type="hidden" name="tax_rates[{{ data.id }}][country]" value="{{ data.country }}" />
 </td>
 
-<td>
+<td data-colname="<?php esc_html_e( 'Region', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 23:
```suggestion
<td data-colname="<?php esc_attr_e( 'Region', 'easy-digital-downloads' ); ?>">
```

includes/admin/views/tmpl-tax-rates-table-row.php
```diff
@@ -30,12 +30,12 @@
 	<# } #>
 </td>
 
-<td>
+<td data-colname="<?php esc_html_e( 'Rate', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 33:
```suggestion
<td data-colname="<?php esc_attr_e( 'Rate', 'easy-digital-downloads' ); ?>">
```

includes/admin/views/tmpl-tax-rates-table-row.php
```diff
@@ -30,12 +30,12 @@
...
 	{{ data.formattedAmount }}
 	<input type="hidden" name="tax_rates[{{ data.id }}][rate]" value="{{ data.amount }}" />
 </td>
 
-<td class="edd-tax-rates-table-actions">
+<td class="edd-tax-rates-table-actions" data-colname="<?php esc_html_e( 'Actions', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 38:
```suggestion
<td class="edd-tax-rates-table-actions" data-colname="<?php esc_attr_e( 'Actions', 'easy-digital-downloads' ); ?>">
```

PR #7993, opened 2020-06-09
3.0 - Orders - Split Flat Rate Discounts

### LINKED ISSUE #7955: 3.0 - Manual orders apply flat rate discounts for each product instead of per order
### Expected behavior
When I add a discount to a manually created order, I expect it to be applied one time.

### Actual behavior
For flat rate discounts, the discount is applied to each item in the order instead of the entire order. This is different from the front end behavior. Percentage discounts appear to be fine.

1) Create a flat rate discount.
2) Create a new order and add a product to it.
3) Add the discount to the order.
4) Add another product to the order. The discount is doubled.

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

Tested two orders--one front end and one admin/manual--with identical products and a $10 discount. Discount applied was $20 for the manual order, but $10 for the front end order.

spencerfinnell:
Fixes #7955 

Proposed Changes:
1. Ensure "Flat Rate" discounts are spread across Order Items when manually creating an Order.
2. Introduce `edd_get_item_discount_amount()` that can be used for calculating discount amounts in different contexts.


spencerfinnell:
@robincornett Is that on the frontend Checkout or admin Order screen? Can you share a screenshot of the items in the Order?

robincornett:
This is in the admin, and happens with varying products:

image

(This discount is for $27--it's the one I made to see if the $10 amount was the issue.)

robincornett:
Applying one discount seems to be working okay, but applying two does not:

image

In this screenshot, `sample` is a flat $10 discount; `NOTTEN` is $27.

If I remove `sample`, the cart recalculates and is correct. If I add it back, I get this:

image

And if I add a second item to the cart:

image

The line item discounts appear to add up to the $30.77, but `sample` should be $10.

The discounts have nothing set other than the name, code, amount, and type. No limitations, exclusions, or dates.

robincornett:
@spencerfinnell I have been looking at the `edd_order_adjustments` table and it's raising a question here. If an order has multiple items in it, should the discount be applied to each order item, or to the entire order? I did a test transaction with a tax rate and the tax rate was applied to the order item, but on a transaction with a discount, the database is showing the discount applied to the order, even though there are multiple items in the order.

spencerfinnell:
@robincornett The current goal is to match what happens on the frontend Checkout. The frontend and backend need to be in sync at a current point then we can make adjustments to both if we figure out something in both needs to change based on #8012

ashleyfae:
Started working through a bunch of tests. So far I've done:

- [x] Percentage, with download requirements.
    - [x] Manual order
    - [x] Front end
    - [x] with taxes
- [x] Flat, no requirements or restrictions
    - [x] Manual order
    - [x] Front end
    - [x] with taxes

My next one was "flat with exclusions" and I think I found an issue there. To be fair, I've not tested this on `master` branch so I don't know if this works differently in 2.9.

My discount settings:

Screenshot from 2020-06-16 12-18-24

So it's a flat $10 discount, but "WordPress Theme" is excluded.

Then I created a new order containing two items:

1. A download priced at $1
2. The "WordPress Theme" product I excluded.

Result:

Screenshot from 2020-06-16 12-19-02

and front-end:

Screenshot from 2020-06-16 12-19-55

Perhaps worth noting: in the admin view when I hover over "$1 item" it says "Discount: $10.00". When I hover over "WordPress Theme" it has no discount noted.

My expectation here is that overall the discount would only be $1, because the only item eligible for a discount costs $1. WordPress Theme isn't eligible, so it shouldn't even be factored in.

spencerfinnell:
Thanks @ashleyfae. The Manual Purchases extension (required for 2.9) doesn't support Discounts at all so we just need to match frontend Checkout behavior in 3.0 where possible. I'll take a look at this scenario.

ashleyfae:
> Oh did you mean on the frontend? I'm seeing different behavior on `issue/7955` and `release/3.0` with the exclusions so something is messing up in the newly abstracted function.

I get the exact same behaviour on the admin and front-end. Sorry if my screenshots weren't clear. Maybe it's because the front-end one has taxes and the admin one doesn't. But in both cases the discount is applied as the full $10.

robincornett:
I created an order with two downloads and applied two flat rate discounts. The latest code somehow feels snappier to me.

I initiated a refund and the discounts showed as prorated on the refund modal. However, the refund order appears to not have the correct information (should have been $37.50):

image

Original order screenshot:

image

ashleyfae:
Confirmed Robin's findings. 

I re-ran all the tests I ran before and everything is looking good order/checkout-wise. Just refunds that are off now, I think.

spencerfinnell:
Thanks for testing. I confirmed as well. I'll try to get this fixed up/completed early next week.

includes/ajax-functions.php
```diff
@@ -1437,39 +1437,47 @@ function edd_admin_order_get_item_amounts() {
...
+		$edd_flat_discount_total = 0;
+
 		$d = edd_get_discount( $discount_id );
 
-		if ( empty( $d ) ) {
-			continue;
+		// Retrieve total flat rate amount.
+		if ( 'flat' === $d->get_type() ) {
+			foreach ( $products as $product ) {
+				edd_get_item_discount_amount( $product, $products, array( $d ) );
```
robincornett, line 1450: Should this be assigned to a variable, or be added to `$edd_flat_discount_total`? I'm not sure what's going on here.

includes/orders/functions/orders.php
```diff
@@ -1049,16 +1049,23 @@ function edd_build_order( $order_data = array() ) {
...
+
+				if ( is_array( $items ) && ! empty( $items ) ) {
+					foreach ( $items as $key => $item ) {
+						$discount_amount += edd_get_item_discount_amount( $item, $items, array( $discount ) );
+					}
+				}
 
 				edd_add_order_adjustment( array(
 					'object_id'   => $order_id,
 					'object_type' => 'order',
```
robincornett, line 1063: Since you have worked out how to calculate the discount per item, what about storing the order adjustment per `order_item` directly instead of per `order`? This seems like it could make retrieving the discount amounts easier when looking at refunds, and would be consistent with how we store `tax_rate` on orders.

includes/ajax-functions.php
spencerfinnell: @robincornett The `global $edd_flat_discount_total;` inside `edd_get_item_discount_amount()` gets updated. I'll add some comments to clear this up.

includes/orders/functions/orders.php
spencerfinnell: @robincornett I know you've done some more research regarding Adjustments since this comment -- do you still feel the same way?

I think it could be done separately as this implementation is meant to mimic the current frontend checkout.

robincornett: per @cklosowski, Discounts should [stay assigned](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8014#issuecomment-655160388) to `order`s, so I think this can stay as is:
> As far as discounts, this gets a little difficult. The discount code is applied to the order as a whole. This is ideal as it allows for accurate reporting of how often a discount is used. As far as it affecting each item in the cart, that's likely because discounts can be cart wide or specific to a single item (based off the discount code settings). As EDD has always done, a discount code being attached to an order simply adjusts the order total, by adjusting each order item, so that items that can be affected are and others that are not affected by the discount code can be skipped. This also allows for accurate reporting of earnings of a product as well as discount code usages from a reporting standpoint.

In the comment after that one, I added a link to a spreadsheet I created with a record of every use of `edd_add_order_adjustment`, trying to evaluate purely at a data level, along with some notes/questions. If you have some time to look over the GH comments and weigh in, I would appreciate it, @spencerfinnell.

includes/admin/views/tmpl-order-totals.php
```diff
@@ -10,6 +10,18 @@
  */
 ?>
 
+<# if ( false !== data.state.isRefund ) { #>
+	<tr class="is-expanded">
+		<td></td>
+		<td colspan="{{ data.config.colspan }}" class="column-primary">
+			<?php esc_html_e( 'Discount', 'easy-digital-downloads' ); ?>
+		</td>
+		<td class="column-right" data-colname="<?php esc_html_e( 'Discount', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 19:
```suggestion
		<td class="column-right" data-colname="<?php esc_attr_e( 'Discount', 'easy-digital-downloads' ); ?>">
```

ashleyfae (changes requested):

ashleyfae:
I was retesting the issue I noted here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7993#issuecomment-644702810 (so same discount / product settings). I found an issue when manually adjusting prices during manual order creation.

1. Add a download that's excluded from your discount ("WordPress Theme" in my screenshots).
2. Add a download that's included in your discount. Mine is a $1 product. Select "Automatically calculate amounts" and set the price to $5 instead.
3. Add a discount code that's a flat rate and only applies to your $1 product. Mine is a flat rate for $10.
4. Total discount is only $1 -- the original price of the product. I'd expect it to be $5 -- the amount I changed the product to.

Screenshot from 2020-09-14 12-58-13

spencerfinnell:
@ashleyfae It limits it to the Download price to prevent the issue you originally reported. We can add a bunch of extra checks and workarounds but the logic surrounding this is already pretty complex. For the initial implementation of supporting discounts in manual Orders what about relying on adding manual adjustments to account for discrepancies after manually adjusting amounts?

ashleyfae:
> For the initial implementation of supporting discounts in manual Orders what about relying on adding manual adjustments to account for discrepancies after manually adjusting amounts?

@spencerfinnell That works for me. :+1:

spencerfinnell:
Rested the last Refund issue reported and that's working for me after https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7993/commits/2c9dc1b4fd064bd559307ca8c9e0519d8ce19d5e

ashleyfae (approved):

PR #7996, opened 2020-06-09
Adding notice to tax report if default tax rate is detected #6840

### LINKED ISSUE #6840: 3.0 - Reports: Tax report does not display any data related to 'global' tax rates, just tax regions.

cklosowski:
Fixes #6840

Proposed Changes:
1. Changes the `object_id` of reports class to `public` so that it can be used in the hooks in the reports output. Everything was previously `private` so nothing could be used or filtered.
2. Adds a notice at the bottom of the tax reports if a default tax rate is detected.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -2972,7 +2972,7 @@ add_action( 'admin_footer', 'edd_add_screen_options_nonces' );
  * that we cannot report on the default tax rate.
  *
  * @since 3.0
- * @param $report
+ * @param \EDD\Reports\Data\Report|\WP_Error $report The current report object, or WP_Error if invalid.
  */
 function edd_tax_report_notice( $report ) {
 	if ( 'taxes' === $report->object_id && false !== edd_get_option( 'tax_rate' ) ) {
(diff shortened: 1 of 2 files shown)


includes/admin/reporting/reports.php
```diff
@@ -2966,3 +2966,22 @@ function edd_add_screen_options_nonces() {
 	wp_nonce_field( 'meta-box-order',  'meta-box-order-nonce', false );
 }
 add_action( 'admin_footer', 'edd_add_screen_options_nonces' );
+
+/**
+ * This function adds a notice to the bottom of the Tax reports screen if a default tax rate is detected, stating
+ * that we cannot report on the default tax rate.
+ *
+ * @since 3.0
+ * @param $report
```
spencerfinnell, line 2975:
```suggestion

 * @param \EDD\Reports\Data\Report|\WP_Error $report The current report object,
 *                                                   or WP_Error if invalid.
```

Sorry meant to include this in the previous suggestion.

cklosowski:
Fixed @spencerfinnell

PR #8001, opened 2020-06-10
3.0 - Migration - Don't Automatically Mark Upgrades Complete via UI

### LINKED ISSUE #7840: 3.0 - Confusing message about upgrade having already run when it hasn't
### Expected behavior

If I haven't actually run a migration, I shouldn't get a message saying I have.

### Actual behavior

If you have no data to migrate for one particular routine -- say tax rates -- then you can get a message saying you have when you run the migration via the CLI.

1) Have a 2.9 install that has no tax rates configured.
2) Switch to release/3.0.
3) Visit the 3.0 upgrade **UI** but don't actually run anything. However, just having visited this UI will mark the tax rate migration as having been completed.
4) Then run the 3.0 migration via the CLI. You'll get this:

Screenshot from 2020-05-07 13-55-16

I never ran the tax rate migration before so this is confusing. However, it was automatically marked as completed just by visiting the UI once. I wonder if people will be confused by this and think something has gone wrong.

**Perhaps we shouldn't be marking upgrades as complete simply by viewing the migration UI.**

PHP Version:

WordPress Version:

spencerfinnell:
Fixes #7840 

Proposed Changes:
1. Avoid marking migrations complete automatically when no migration needs to happen.

### DIFF
--- a/includes/admin/upgrades/upgrade-functions.php
+++ b/includes/admin/upgrades/upgrade-functions.php
@@ -1343,7 +1343,6 @@ function edd_upgrade_render_v30_migration() {
 	$orders_complete = edd_has_upgrade_completed( 'migrate_orders' );
 
 	if ( empty( $orders ) ) {
-		edd_set_upgrade_complete( 'migrate_orders' );
 		$orders_complete = true;
 	}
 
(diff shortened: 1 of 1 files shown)


spencerfinnell:
@cklosowski I thought it was handled during actual processing here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-base.php#L131

Running the UI updates the `edd_completed_upgrades` option.

cklosowski:
But those never get run, because the markup for the step is never output to the screen, since no tax rates were detected. therefore the migrator never get's triggered by the UI, never setting it to complete.

cklosowski:
Going to merge this into the branch I'm working on for migration UI improvements so I can work through getting it included there. Thanks @spencerfinnell

Issue #6880, opened 2018-08-27
3.0: Upgrade Screen is Confusing


spencerfinnell:
I've tested 3.0 on a couple of sites and using the upgrade screen has been a bit confusing to me during all experiences.

- When I click "Upgrade Database" on the first step it triggers all of the steps automatically. If they happen all at once why not make one button initially?
- The "Upgrades" page title is pretty generic. I think "Easy Digital Downloads 3.0 Upgrades" would give more context.
- Not a huge fan of the mdash: https://cloudup.com/cNqOeq_GbuQ Either have it on a new line without the dash or remove it completely. It just duplicates what the success alert below displays.
- Same thing with this state: https://cloudup.com/cO3ydsEbGRb -- I think it would be better to use a warning similar to the alert in the previous note about how migration is no longer possible.

Functionality wise:

- When I refresh I can still upgrade orders, order notes, and customer notes again. There are no errors so I'm not sure if it's failing or not. I think it's just not auto removing the legacy data and customer addresses are? 

---

I think only surfacing a single step for the upgrades will be a lot easier for customers. In the end they really don't care what is going on as long as things don't break. 

If something fails they shouldn't be using their site and should contact support. If everything goes according to plan they should just continue their day as normal; which means they probably also shouldn't have to think about legacy data. 

Anyone who cares about legacy data after a successful upgrade knows what they are doing and has made appropriate plans for it already. What about setting a future (like 6 months) to delete that data automatically? Provides a pretty long safety net for catching edge cases and keeps the user as a user not a database admin.

---

Proposal:

Simplify the upgrade screen: https://cloudup.com/cMWAvdHbeAH

Perform all the upgrades at once and schedule a cron to remove legacy data in the very far future (and create a tool that users can trigger manually). If something goes wrong display an error and a very clear way to create a support ticket with relevant information.

---

As someone who experienced the Ninja Forms 3.0 update on many sites (and dealt with many customers using the plugin) I've realized how "unimportant" these kind of upgrades really are and how behind the scenes they should be kept.  Obviously the user needs to trigger it so backups (lol) can be made but other than that it should be irrelevant to them.

cklosowski:
Actually the first step is required to go first, it's the initial migration, if it doesn't migrate all items it should not auto start the next steps.

We used this same method for software licensing so people knew what was happening instead of blindly failing at some point. Failing on a specific step helps us debug better.

spencerfinnell:
[...]

PR #8008, opened 2020-06-11
Improve the upgrade UI to be less confusing.

### LINKED ISSUE #6810: 3.0 - Upgrade Screen: Interaction is allowed on upgrade routines out of order
### Expected behavior
When viewing the upgrade screen, actions should be required to be completed in any necessary orders. If moving legacy order data over is required for the next steps to be one, it must be done first.

### Actual behavior
All upgrade routines appear to be able to be accessed immediately, including remove legacy data.

1) View the 3.0 upgrade screen

PHP Version: All

WordPress Version: All

This was in Firefox. Keep in mind we did this same type of screen with SL 3.6, but we locked out any upgrades that required the initial license upgrade be done first.:

cklosowski:
Fixes #6810 #7840 

Proposed Changes:
1. Does not allow migration steps to be done out of order.
2. Stops migration routines from being marked as completed, just by visiting the UI.
3. Exposes all the steps, regardless of if the steps are necessary (for transparency)


includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1550,13 +1513,13 @@ function edd_upgrade_render_v30_migration() {
 						<?php wp_nonce_field( 'edd_ajax_export', 'edd_ajax_export' ); ?>
 
 						<?php if ( ! $customer_addresses_complete ) : ?>
-							<span class="edd-migration allowed" style="<?php echo ! $migration_complete ? '' : 'display: none'; ?>">
+							<span class="edd-migration allowed" style="<?php echo $migration_complete ? '' : 'display: none'; ?>">
```
robincornett, line 1516: Could we update the markup and script so that the inputs are not duplicated? This is causing validation errors because of duplicate IDs. Instead of using show/hide on the allowed/unavailable elements, we could enable/disable the inputs and change classes.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1550,13 +1513,13 @@ function edd_upgrade_render_v30_migration() {
...
 								<input type="submit" id="migrate-customer-addresses-submit" value="<?php esc_attr_e( 'Upgrade Database', 'easy-digital-downloads' ); ?>" class="button-primary"/>
 							</span>
 
-							<span class="edd-migration unavailable" style="<?php echo $migration_complete ? '' : 'display: none'; ?>">
+							<span class="edd-migration unavailable" style="<?php echo ! $migration_complete ? '' : 'display: none'; ?>">
 								<input type="submit" disabled id="migrate-customer-addresses-submit" value="<?php esc_attr_e( 'Upgrade Database', 'easy-digital-downloads' ); ?>" class="button-secondary"/>
```
robincornett, line 1521: Since `<input type="submit">` are more commonly for submitting forms, and buttons for triggering actions, could we consider converting these to buttons?

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1353,19 +1352,9 @@ function edd_upgrade_render_v30_migration() {
 	$addresses = $wpdb->get_results( "SELECT umeta_id FROM {$wpdb->usermeta} WHERE meta_key = '_edd_user_address' LIMIT 1" );
```
spencerfinnell, line 1352: Since we no longer need to check if these are empty we can avoid a lot of extra queries by removing the initial value checks.

includes/admin/upgrades/upgrade-functions.php
cklosowski: We do both but the messaging changes depending on what happens or what step gets hung up on

includes/admin/upgrades/upgrade-functions.php
cklosowski: No this is not possible with the current batch processor class.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1550,13 +1513,13 @@ function edd_upgrade_render_v30_migration() {
...
-							<span class="edd-migration allowed" style="<?php echo ! $migration_complete ? '' : 'display: none'; ?>">
+							<span class="edd-migration allowed" style="<?php echo $migration_complete ? '' : 'display: none'; ?>">
 								<input type="submit" id="migrate-customer-addresses-submit" value="<?php esc_attr_e( 'Upgrade Database', 'easy-digital-downloads' ); ?>" class="button-primary"/>
 							</span>
 
-							<span class="edd-migration unavailable" style="<?php echo $migration_complete ? '' : 'display: none'; ?>">
+							<span class="edd-migration unavailable" style="<?php echo ! $migration_complete ? '' : 'display: none'; ?>">
 								<input type="submit" disabled id="migrate-customer-addresses-submit" value="<?php esc_attr_e( 'Upgrade Database', 'easy-digital-downloads' ); ?>" class="button-secondary"/>
-								&mdash; <?php esc_html_e( 'Your customer addresses been upgraded.', 'easy-digital-downloads' ); ?>
+								&mdash; <?php esc_html_e( 'Please complete the previous steps before updating customer addresses.', 'easy-digital-downloads' ); ?>
```
robincornett, line 1522: I ran the migration via the UI and once it was done, all of these "Please complete the previous steps..." messages were displaying. If the migration step has completed successfully, this message should disappear or be replaced with something like "Migration step complete!".

So maybe the UI needs to have three possible statuses:
- unavailable (not ready)
- allowed
- complete

With messages shown/hidden accordingly.

spencerfinnell: @robincornett I slightly addressed that in https://github.com/easydigitaldownloads/easy-digital-downloads/compare/issue/6810...issue/6810-2 by not showing the "unavailable" message and instead disabling the input inside of "allowed".

I think a fair amount will still change if we have time in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6880 so I'm not sure making too many changes in this PR beyond the initial interaction blocking is necessary.

ashleyfae:
Confirmed Spencer's comment here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8008#pullrequestreview-429767175

Otherwise I think it's looking good.

includes/admin/upgrades/upgrade-functions.php
cklosowski: @spencerfinnell I've merged in `issue/6810-2` as well as removed queries for data, have a look over that please.

cklosowski:
Regarding wording, phrasing, messaging. I'd like to look at those post-alpha. As they are mostly usability, so long as the data is migrated properly, and in the right order (with the help of #8012).

cklosowski:
Ok fixed this up, took some work, but looked like we ahd a merge conflict somewhere.

PR #8009, opened 2020-06-12
Add name to manually created order addresses

### LINKED ISSUE #8007: 3.0 - Manual orders with new addresses for existing customers are missing the name
### Expected behavior
When I create a manual order, I expect the database record for the address to be complete.

### Actual behavior
The `name` parameter is not added to the address record for manually created orders ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/functions/actions.php#L181-L188)).

1. Using the admin interface, create a new order. Make sure that there is an address entered for the order.
2. Check the database entry for that order in both the `customer_addresses` and the `order_addresses` tables.

PHP Version: 7.4.1
WordPress Version: 5.4.2

robincornett:
Fixes #8007

### DIFF
--- a/includes/orders/functions/actions.php
+++ b/includes/orders/functions/actions.php
@@ -80,7 +80,7 @@ function edd_add_manual_order( $args = array() ) {
 			: '';
 
 		// Combine
-		$name = $first_name . ' ' . $last_name;
+		$name = trim( $first_name . ' ' . $last_name );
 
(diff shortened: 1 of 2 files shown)


robincornett:
Thanks, @spencerfinnell. I added that change, and also added it to the same place in the `edd_build_order` function.

Issue #8012, opened 2020-06-12
3.0 - Issues with edd_order_adjustments table


robincornett:
### Expected behavior
I expect the order adjustments table data to be consistent and complete.

### Actual behavior
I have migration related issues:
- No discounts are migrated to the new adjustments table.
Checking an order with a discount shows the correct amount in the customer details screen, but when viewing the order details screen, the discount does not display and the total is incorrect.
- The adjustments which did migrate all use the migration date for the created/ modified date, instead of the order date (related: #7859).
- In a storefront which collects sales tax, as I read the code, orders placed in the region associated with my tax rate should have a row for the tax rate; otherwise, they should not. Spot checking suggests that free orders have a tax rate associated with them, regardless of the order address. This doesn't seem to happen with orders placed within 3.0, so I'm listing as a migration issue.
- Migrated orders with sales tax applied do correctly show the tax.

Possible table issues (all tax rate items):
- The `type_id` is `0`, instead of the adjustment `id`.
- The `description` is empty, but should probably inherit the adjustment `description`, which for my store is the region (TN). For a storefront with multiple tax rates, I think it would be important to populate this.
- The `subtotal` and `tax` data information are all `0.0000000`, but this could be intentional.
- The `total` is the tax rate.

Manually adding a new order with tax applied does not add a row to this table. Completing an order with tax applied on the front end does add a row.

Interestingly, the new order row shows the tax applied to the `order_item`, not the `order`, which is the case for the migrated orders. This may be a difference in accounting practice, but mentioning in case it is not.

### Steps to reproduce the behavior
1) Migrate a 2.9 storefront with tax data to 3.0.
2) Check the `edd_order_adjustments` table. Mine shows only `tax_rate`, no `discount`, even though there should be.
3) Add a manual order with tax applied. The row for it will not be added to the database.

### Information (if a specific version is affected):
PHP Version: 7.4.1
EDD Version (or branch): release/3.0
WordPress Version: 5.4.1

Any other relevant information:
A note for myself so I don't forget, but in the `edd_order_adjustments` table, the `object_id` refers to the specific ID of whatever the `object_type` is, which could be an `order`, an `order_item`, or maybe something else (so far this is what I have seen). Thanks to @JJJ for touring me through the adjustments tables so I could begin to wrap my head around them.

robincornett:
Some notes as I'm looking into this:

[...]

PR #8014, opened 2020-06-15
Fix order adjustments table

robincornett:
#8012

Proposed Changes:
1. Use original order dates for adjustment dates.
2. Reorder migration steps so that taxes and discounts are migrated before orders.
This solves the issue for migrating discounts in orders, but would require updating the Upgrade Screen behavior.
3. Create a new function to retrieve the tax rate object, so that full tax rate data can be added to the order adjustments table for both manual and front end orders.

This should be tested to evaluate migrated data and new data both.

**Migrations:**
1. Start with a 2.9 storefront with both discounts and tax rates. Some discounts and taxes need to have been used on orders.
2. If you upgrade to 3.0 and run a migration with the current `release/3.0` branch, no discounts on individual orders will be migrated. You can confirm this by looking at the `edd_order_adjustments` table.
3. Note the dates on the tax rate rows: they are the migration date, not the order dates.
4. Reset the database, switch to this branch, and run the migration again. Discounts on orders will now show properly in the table. Discounts and tax rate rows will use the order dates instead of the migration date.

Note: this PR reorders the migration process so that tax rates and discounts are migrated before orders. As the current release branch stands, it doesn't seem to matter, but this will require #8008 to be reworked, I expect. In this PR, CLI migrations work fine, and the UI migration does go through all steps when clicking on the first button.

**New data**
1. In a 3.0 store using `release/3.0`, generate a manual order with taxes applied. Check the `edd_order_adjustments` table. No row is created.
2. Generate an order from the front end with taxes applied. Check the table. The row is created, but is incomplete--no `type_id` and no `description` are added.
3. Switch to this branch and create taxable orders, both in the admin and front end. Check the database rows which have been added: `type_id` and `description` should both be present.


robincornett:
At this point, new orders and migrating discounts appear to be fixed.

Migrated tax rate data still feels very incomplete, but I think trying to complete the tax rate rows for 2.9 orders would be more educated guesswork, and not something to rely on if tax rates or situations changed from early use.

I am also still confused by the `subtotal`, `tax_rate`, and `total` columns as they display in this table.

Also, for new orders, tax rates are applied to each order item, but discounts are applied to the order. With the work going into #7993, it seems like discounts should be applied to order items.

includes/tax-functions.php
```diff
@@ -442,3 +410,63 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+/**
+ * Gets the tax rate object from the database.
+ * Used in `edd_get_tax_rate`, `edd_build_order`, `edd_add_manual_order`.
+ *
+ * @param array $args
+ * @return object|boolean
+ *
+ * @since 3.0
+ */
+function edd_get_tax_rate_object( $args ) {
```
ashleyfae, line 423: I wonder if we should make this more specific. It's really _"get tax rate object for given country / region"_, yes? So although it's wordy, maybe:

`edd_get_tax_rate_object_for_country_region()`
or
`edd_get_tax_rate_object_for_location()`
or
`edd_get_tax_rate_object_for_address()` (might be misleading since we don't need the whole address)

includes/tax-functions.php
```diff
@@ -442,3 +410,63 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
 
 	return (Bool) apply_filters( 'edd_download_is_tax_exclusive', $ret, $download_id );
 }
+
+/**
+ * Gets the tax rate object from the database.
```
ashleyfae, line 415:
```suggestion
 * Gets the tax rate object from the database for a given country / region.
```

includes/tax-functions.php
```diff
@@ -442,3 +410,63 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+ * Used in `edd_get_tax_rate`, `edd_build_order`, `edd_add_manual_order`.
+ *
+ * @param array $args
```
ashleyfae, line 418: Can you clarify what's expected in this array? Example:

```suggestion
  * @param array $args {
 *     Country and, optionally, region to get the tax rate for.
 *
 *     @type string $country Required - country to check.
 *     @type string $region  Optional - check a specific region within the country.
 * }
```

includes/tax-functions.php
```diff
@@ -442,3 +410,63 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+	// The region is not passed in deliberately in order to check for country-wide tax rates.
+	$tax_rates = edd_get_tax_rates(
+		array(
+			'name'   => $args['country'],
+			'status' => 'active',
+		),
+		OBJECT
+	);
+
+	// Save processing if only one tax rate is returned.
```
ashleyfae, line 440: Wondering if this section is really needed. Feels like unnecessary code duplication. If there's only one rate then the `foreach` loop would still handle it just fine.

includes/tax-functions.php
robincornett: I think location makes sense, while still being somewhat generic. I had originally used `object` in the name because `edd_get_tax_rate` is already in use...another option might be `edd_get_location_tax_rate`, which would remove the `for`, and I don't think the code base usually includes `object` or similar in function names?

includes/tax-functions.php
robincornett: I wasn't sure about editing too much of the code in this as it was extracted from the original tax rate function, but I agree. Will remove.

includes/tax-functions.php
```diff
@@ -442,3 +410,63 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+					continue;
+				}
+
+				$regional_rate = $tax_rate->amount;
+
+				if ( ( 0 !== $regional_rate || ! empty( $regional_rate ) ) && '' !== $regional_rate ) {
+					$rate = $tax_rate;
+				}
+			}
+		}
```
robincornett, line 468: @ashleyfae as it stands, if there are multiple tax rates retrieved, the last one is what will be chosen--even if `$rate` is defined in a prior loop, it can be overwritten by a later one. Is that the correct behavior, or should we tuck something in to go ahead and return the first matched object?

includes/tax-functions.php
ashleyfae: I think I prefer `edd_get_tax_rate_for_location()`. I know the `for` feels silly, but I like how the item I'm retrieving is closer to the start of the name. To me it just reads better/faster while skimming. I quickly know "I'm getting a tax rate".

includes/tax-functions.php
ashleyfae: Two things to consider I think:

1. You shouldn't be able to have multiple active country-wide tax rates for the same country. (Or if we haven't fixed that yet then we will.. I recall an issue on it at one point.) So hopefully we shouldn't run into that issue.
2. If you have a country-wide tax rate **and** a tax rate for the same country that's narrowed down to a region, then I think the region one should always win. I don't think that's currently happening. However, maybe that's a big/separate enough change that we should make a new issue for it.

robincornett: The multiple country-wide tax rates are still an issue, yes.

So I'm reviewing the logic in this foreach loop and it's currently a bit convoluted, and I think could stand to be simplified ( `0`, `''`, both evaluate the same as `empty` in the regional check). Going to work on that and see if I can finesse it.

robincornett:
Note: looking at `edd_order_adjustments` table, I have some discounts showing positive amounts for subtotal/total, and some showing negative amounts. I've been testing multiple branches which may have affected this but wondering which it should be.

includes/tax-functions.php
```diff
@@ -415,12 +415,17 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+ *
+ *     @type string $country Required - country to check.
+ *     @type string $region  Optional - check a specific region within the country.
+ * }
  * @return object|boolean
  *
  * @since 3.0
  */
-function edd_get_tax_rate_object( $args ) {
+function edd_get_tax_rate_for_location( $args ) {
```
robincornett, line 428: I guess another question here is should the parameters be updated to be more consistent with the `edd_get_tax_rate` function, which has these as parameters:

- `$country = ''`
- `$region = '`
- `$fallback = true` (would not need)

So pass each as its own item instead of array keys? Or does it matter?

includes/tax-functions.php
```diff
@@ -442,3 +410,60 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+			if ( ! empty( $regional_rate ) ) {
+				return $tax_rate;
+			}
+		} elseif ( 'country' === $tax_rate->scope ) {
+			// Countrywide tax rate.
+			$rate = $tax_rate;
+		}
+	}
+
+	return $rate;
```
JJJ, line 468: The original `edd_get_tax_rate()` function used `number_format()` internally to round to 4 decimal places, but I see this new function does not do that. Just noting in case this was unintentional.

includes/tax-functions.php
JJJ: 1. Suggest to keep this as an `$args` array so keys can be added later without changing the function signature.
1. Maybe should set `$args = array()` in the function parameters, and then use `$r = wp_parse_args( $args, array( whatever ) )` internally. That way the default keys and values are already taken care of, in the event something weird is passed in or the defaults need to change later on?
1. `_for_location()` is better than the original `_object()`, but it kinda hints to me that it's for a literal location, like an address. (I guess if it were for an address, it would be `_by_address()`, but that function doesn't exist either; it's just part of `edd_get_tax_rate()`.) I can't really think of anything better though.

includes/tax-functions.php
```diff
@@ -442,3 +410,60 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
 
 	return (Bool) apply_filters( 'edd_download_is_tax_exclusive', $ret, $download_id );
 }
+
+/**
+ * Gets the tax rate object from the database for a given country / region.
+ * Used in `edd_get_tax_rate`, `edd_build_order`, `edd_add_manual_order`.
+ *
```
JJJ, line 417: Suggest adding a blurb about the internal fallback logic so people using this function know what to expect.

I.E. Does a Region override a Country? (which is being discussed elsewhere)

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -970,26 +979,34 @@ public static function orders( $data = null ) {
...
+				edd_add_order_adjustment(
+					array(
+						'object_id'     => $order_id,
+						'object_type'   => 'order',
+						'type_id'       => $discount->id,
+						'type'          => 'discount',
+						'description'   => $discount,
+						'subtotal'      => $subtotal - $discount->get_discounted_amount( $subtotal ),
+						'total'         => $subtotal - $discount->get_discounted_amount( $subtotal ),
+						'date_created'  => $date_created_gmt,
```
JJJ, line 991: Should this be `$data->post_modified_gmt`? (If not, does it deserve an inline comment because it's different?)

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -920,13 +927,15 @@ public static function orders( $data = null ) {
...
-					'tax'         => $tax,
-					'total'       => floatval( $fee['amount'] ) + $tax,
+					'object_id'     => $order_id,
+					'object_type'   => 'order',
+					'type'          => 'fee',
+					'description'   => $fee['label'],
+					'subtotal'      => floatval( $fee['amount'] ),
+					'tax'           => $tax,
+					'total'         => floatval( $fee['amount'] ) + $tax,
+					'date_created'  => $date_created_gmt,
```
JJJ, line 937: Should this be `$data->post_modified_gmt`? If not, does it deserve an inline comment because it's different?

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -886,25 +886,32 @@ public static function orders( $data = null ) {
...
-				'total'       => $tax_rate,
-			) );
+			edd_add_order_adjustment(
+				array(
+					'object_id'     => $order_id,
+					'object_type'   => 'order',
+					'type_id'       => 0,
+					'type'          => 'tax_rate',
+					'total'         => $tax_rate,
+					'date_created'  => $date_created_gmt,
```
JJJ, line 898: Should this be `$data->post_modified_gmt`? If not, does it deserve an inline comment because it's different?

includes/tax-functions.php
ashleyfae: > Maybe should set $args = array() in the function parameters

I like everything you said, @JJJ , however, I'd personally keep the parameter as `$args` just because having a parameter is _required_. I'd only do `$args = array()` if the args were optional.

includes/admin/upgrades/v3/class-data-migrator.php
ashleyfae: What I gathered is that the _order adjustments_ used `date_created` and the _refund adjustments_ used `post_modified`. (Though Robin can confirm if that was the goal.)

includes/tax-functions.php
robincornett: I think the tax rate in 3.0 requires a location now--if the storefront has a default/global rate set, there are warnings to remove it. So an address (that is, country at minimum) is required to get the tax rate object.

includes/tax-functions.php
robincornett: This returns the entire object, not just the rate, but I should double check where the `amount` is used in the order creation process.

includes/tax-functions.php
robincornett: Update: have reworked the logic so that if a regional rate is set, it will automatically be returned and the `foreach` stopped. If a country rate is set, it will be stored, but can be overridden in the next iteration of the loop, I think.

@JJJ per your question above, a regional rate should take precedence over a country rate.

includes/admin/upgrades/v3/class-data-migrator.php
robincornett: I think my goal was to match it to the order dates, which are set here:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/ecb6b176ff67914eeebbd6fdd48a69401db7f1d1/includes/admin/upgrades/v3/class-data-migrator.php#L495-L496

This is different than what we ended up doing for customer/order addresses, which use the customer/order date for the creation date, but the migration date for the modified date ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7867#issuecomment-628248098)).

And yes, the post_modified is what's used for the refund adjustments to be consistent with refund dates, which are a best guess ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L538-L540)).

Mainly, the migrated order adjustments are ideally tracking with the migrated order data so that all are consistent.

includes/orders/functions/actions.php
```diff
@@ -324,7 +324,7 @@ function edd_add_manual_order( $args = array() ) {
 							'object_id'   => $order_item_id,
 							'object_type' => 'order_item',
 							'type'        => 'tax_rate',
-							'total'       => $tax_rate->amount,
+							'total'       => number_format( $tax_rate->amount, 4 ),
```
JJJ, line 327: I think this may be incorrect, but I might also be mistaken.

I believe EDD always wants to store the exact amount to be as precise as possible, without truncating or rounding it. Truncation or rounding happens on display only, just for simplicity sake.

In the event the tax rate is `0.52759` in this case `number_format()` here would be saving an inaccurate value.

includes/orders/functions/orders.php
```diff
@@ -982,7 +982,7 @@ function edd_build_order( $order_data = array() ) {
 						'object_id'   => $order_item_id,
 						'object_type' => 'order_item',
 						'type'        => 'tax_rate',
-						'total'       => $tax_rate->amount,
+						'total'       => number_format( $tax_rate->amount, 4 ),
```
JJJ, line 985: See above.

includes/orders/functions/actions.php
robincornett: That makes sense. I had added this back in because it was removed from the same code in the `edd_get_tax_rate` function, but based on what you're saying here, it should not have been added there, either.

robincornett:
I've just reverted this PR to a draft because I am seeing some confusing things with the data in the order adjustments table, even with these fixes, and would like to get some clarity. Note: at this point, the orders appear to _work_, in that things are totaled up correctly, and refunds happen correctly (except where they are being dealt with elsewhere), but as I look at the database table, I'm still not understanding what I'm seeing.

## Discounts
Discounts are applied to the `order`, even if a discount is intended for one product only. The discount amount is entered in both the `subtotal` and the `total` column.

_Question: if discounts need to be pro-rated across order items (thinking of refunds in particular), wouldn't it make sense to apply discounts to the `order_item` instead?_

## Taxes
Taxes are applied to each `order_item` in an order which is taxable. If two items are in the order, one taxable and one marked nontaxable, a database row is created for each `order_item`, but all other data in the rows are the same, so there is no way to tell visually if an `order_item` is being taxed or not.

Only the `total` has a value (the tax rate); the `subtotal` and `tax` columns are set to the default (`0.000000000`).

## Fees
During the manual order creation process, fees can be added, either to an individual download in the order, or to the entire order. The fee is added to the database for both the `subtotal` and the `total` amounts.

If I select an `order_item`, two rows are added to the database: one for the `order_item` and one for the `order`. This doesn't seem correct.

Additionally, in the data migration class, it appears that fees may have taxes applied, and so the `subtotal` would be the fee, the `tax` would be the tax rate, and the `total` is the fee plus the tax, but I've not been able to replicate that on a new order.

A fee registered using `EDD()->add_fee()` which is automatically added to a front end order is not automatically added to an admin/manual order.

## Credits
A credit is applied to the `order`, with the credit amount being both the `subtotal` and the `total`.

## General
Looking through all instances of `edd_add_order_adjustment`, it looks like the `tax` is used only for fees, and then only for fees added using `EDD()->add_fee()`.

cklosowski:
There have been many conversations over time about the concept of taxes and adjustments. For Taxes, you are correct the `total` is the tax rate, the other columns are not needed. We previously had a taxes table but that didn't work out as expected considering that taxes are adjustments as well and resulted in some even worse inconsistencies. I'm not sure we're at a point we can change that totally, or if we need to do that in a future 3.1 or 3.2. 3.0 might be the time but that's going to set us back pretty far I think.

As far as discounts, this gets a little difficult. The discount code _is_ applied to the order as a whole. This is ideal as it allows for accurate reporting of how often a discount is used. As far as it affecting each item in the cart, that's likely because discounts can be cart wide or specific to a single item (based off the discount code settings). As EDD has always done, a discount code being attached to an order simply adjusts the order total, by adjusting each order item, so that items that can be affected are and others that are not affected by the discount code can be skipped. This also allows for accurate reporting of earnings of a product as well as discount code usages from a reporting standpoint.

Fees (and by association, credits which are just negative fees), are roughly the same as discounts, and can sometimes be associated with items as well, like with Discounts Pro and also Signup Fees on Recurring.

I'd love to get some feedback from @jjj and @ashleyfae about their view on where we take these things. I think if we decide to completely adjust our usage of adjustments, that the one we should focus on is 'taxes' as it's the only one that is slightly different than discounts, fees, and credits.

If we choose to go this route, I think it's important that we get a strict write up, and know that this _will_ delay our movement forward quite a bit, so I ask just that we assess, is this something we can do in the future or is it completely necessary to completely re-write this now, or does enough data exist in the current schema that we can further improve this system in the next major release?

robincornett:
Thanks, Chris, appreciate it. If it helps the discussion, I've gone through the code (based on this issue branch) and [documented](https://docs.google.com/spreadsheets/d/1dMh-ezUI088XHdEvO2nYlocSim638tnpIeHmZ8Da87k/edit?usp=sharing) everywhere `edd_add_order_adjustment` is used, with what key parameters.

includes/orders/functions/orders.php
```diff
@@ -949,13 +949,17 @@ function edd_build_order( $order_data = array() ) {
 			// Store order item fees as adjustments.
 			if ( isset( $item['fees'] ) && ! empty( $item['fees'] ) ) {
 				foreach ( $item['fees'] as $fee_id => $fee ) {
+					$tax = ( isset( $fee['no_tax'] ) && false === $fee['no_tax'] && ! empty( $tax_rate->amount ) ) || ( $fee['amount'] < 0 && ! empty( $tax_rate->amount ) )
+						? floatval( floatval( $fee['amount'] ) - ( floatval( $fee['amount'] ) / ( 1 + $tax_rate->amount ) ) )
+						: 0.00;
```
JJJ, line 954: I am glad you were ultimately able to find what the heck was going on with this.

If I had to guess, it looks like this approach was nicked from the migrator. (It was unruly then, and me thinks it's still quite a handful to work out what's happening.)

----

Each comparison on the side of the `||` checks for:

```php
! empty( $tax_rate->amount )
```
...so that could happen first with an `&&` and then get removed from both sides.

Statements like this...
```php
isset( $fee['no_tax'] ) && false === $fee['no_tax']
```
...could be simplified down to:

```php
empty( $fee['no_tax'] );
```

(I suspect that strict type checking against `'no_tax'` was a Stylistic choice by Sunny, but I'm not convinced it's worth the additional code complexity on these lines specifically, especially considering that the `'no_tax'` logic-gate already introduces a double-negative.)

Perhaps this is an opportunity to introduce a public static method in the `EDD_Fees` class to centralize these repeated calls, along with some global function akin to `edd_calculate_tax()` but that knowingly does not calculate tax for the `floatval( floatval( ... ) )` business.

----

👍 again for figuring this out.

robincornett: I can't claim to have figured anything out--just when I was documenting all uses of `edd_add_order_adjustment`, I discovered that this particular bit of code was missing the `tax` for fees, although all other fees have the potential tax added. I did pull the tax calculation logic from the data migrator, since it was already in place for that, but I agree that it's a bit convoluted. Would be happy to update it here and the other places where it's done this way.

Making this consistent was just something I felt I could go ahead and resolve while we are still thinking about the other elements of this data table.

robincornett:
Per a conversation with @cklosowski today, marking this one as ready for review in the hopes that we can get it merged and answer some of the related questions in additional issues/PRs. So at this point, this PR:
1. Reorders the migration process so that settings (tax rates and discounts) are migrated _before_ orders.
2. Creates a new function to get the tax rate by location.
3. Uses original order dates for order adjustments on migration.
4. Makes one inconsistent `fee` creation consistent with the others.

#8056 has been opened to address deeper specifically with fees. I would suggest opening a separate issue to address tax rates. Discounts may not need to be addressed.

includes/tax-functions.php
spencerfinnell: I think using `_by_` is more consistent with the existing `edd_get_x_by_y()` patterns we use elsewhere. `edd_get_tax_rate_by_location()` makes sense to me.

includes/orders/functions/actions.php
```diff
@@ -297,6 +315,25 @@ function edd_add_manual_order( $args = array() ) {
 					}
 				}
 
+				// Maybe store order tax.
+				if ( $tax_rate ) {
+					$description = $tax_rate->name;
+					if ( ! empty( $tax_rate->description ) ) {
+						$description = $tax_rate->description;
+					}
+					// Always store tax rate, even if empty.
```
spencerfinnell, line 324: @robincornett I'm not sure what would be empty here/what is being forced here?

robincornett: @spencerfinnell  The `$tax_rate->name` is always set to the country code (eg `US`). The description may be blank. If the tax rate is regional, the `$tax_rate->description` will be the region code (eg `TN`). So this is setting the adjustment description to the name initially, and then overriding it with the more specific value if it exists.

spencerfinnell: @robincornett That makes sense. Perhaps a comment above the `! empty()` check explaining that would be clearer. I thought the comment might have been referring to the result of `edd_get_tax_rate_by_location()`.

includes/tax-functions.php
```diff
@@ -442,3 +410,62 @@ function edd_download_is_tax_exclusive( $download_id = 0 ) {
...
+ * If a regional tax rate is found, it will be returned immediately,
+ * so rates with a scope of `country` may be overridden by a more specific rate.
+ *
+ * @param array $args {
+ *     Country and, optionally, region to get the tax rate for.
+ *
+ *     @type string $country Required - country to check.
+ *     @type string $region  Optional - check a specific region within the country.
+ * }
+ * @return object|boolean
```
ashleyfae, line 426: One minor change:

Instead of `@return object` I'd like to actually specify the exact object. That led me to a discovery that I'm not sure if there's a reason for...

The `edd_get_tax_rates()` function specifies that if `$output = OBJECT` then it will return `an array of EDD\Adjustments\Adjustment objects`. But that's not actually correct. In `EDD\Database\Queries\Adjustment`, the `$item_shape` is set to `\\EDD\\Database\\Rows\\Adjustment`.

I guess for now, for the scope of this PR, we can specify what it actually currently is, but maybe it's worth making a new issue to look into this inconsistency? Whether that means changing the `$item_shape` to be `\\EDD\\Adjustments\\Adjustment` or changing the comment in `edd_get_tax_rates()`.

But for now...

```suggestion
 * @return \EDD\Database\Rows\Adjustment|false
```

PR #8018, opened 2020-06-16
Fix label update behavior for unlinked labels

### LINKED ISSUE #7836: Update Input Labels When Adding Repeatable Rows
When a row is added (for example, in the download metabox: Price ID, Download Files, or Software Licensing Upgrade Paths), labels associated with input fields should be updated to reference the new row input fields.

### Justification or use case
Currently, new row input field `name` and `id` are updated (if set), but the label `for` attribute is not. Many inputs do not have associated labels (#7701), but for those which do, updating the label attributes will improve both semantic markup and accessibility.

robincornett:
Fixes #7836

Proposed Changes:
1. Prevents a JS error if a repeatable row has a label which is missing a `for` attribute.


ashleyfae:
Code looks good, but how do I actually trigger the error? Just so I can test before and after.

Also can you please fix the merge conflict? Thanks!

robincornett:
Sorry, I mentioned it in the issue, but should have shared here as well for testing. If you try to add a new row to the Download Prices metabox, to add a variable price, it was breaking, because the "Default" radio input has a `label` element with no `for` attribute. That should be fixed, but there are already one or two issues related to inputs and accessibility in the downloads metaboxes.

ashleyfae:
For some reason I can't reproduce that error. :woman_shrugging: Works fine for me in both `release/3.0` and this branch.

robincornett:
When I click "Add New Price" in `release/3.0`, the new price row isn't added and I get the following error because the `for` is undefined:

`index.js:96 Uncaught TypeError: Cannot read property 'replace' of undefined
    at HTMLLabelElement.<anonymous> (index.js:96)`

PR #8020, opened 2020-06-17
Return false for discount ID if discount object is not retrieved

### LINKED ISSUE #8002: 3.0 - Retrieving a non-existent discount code causes a PHP notice
### Expected behavior
When I enter a checkout screen, I expect not to encounter a PHP error/notice.

### Actual behavior
I encountered the following error after a completely unexpected sequence of events:

`Trying to get property 'id' of non-object /plugins/easy-digital-downloads.30/includes/discount-functions.php:912`

My unexpected sequence of events was to drop all tables on an existing local site and to reinstall WordPress from scratch. Apparently a cart was still in cache, so when I added new products to the cart, an old discount code which had not been recreated was applied to the cart, which threw an error.

(Normally, adding a discount which does not exist is not possible, and does not cause an error, but I think this happened because it was already in the cart.)

I don't know if it is possible to reproduce the error, but the error is thrown in the `edd_get_discount_id_by_code`, which just calls `edd_get_discount_by_code`. This function can return an object, or it can return `false`, so any function which uses it should expect that an object will not always be returned.

PHP Version: 7.3.5
WordPress Version: 5.4.1

robincornett:
Fixes #8002

To test:

Use `edd_get_discount_id_by_code()` (`var_dump` or `error_log` is enough) to retrieve a nonexistent discount code. Example:

```php
var_dump( edd_get_discount_id_by_code( 'adsrfd' ) );
```

In `release/3.0` this will throw a PHP notice, because a discount object is not returned. This adds a check for that situation.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -910,7 +910,7 @@ function edd_is_discount_valid( $code = '', $user = '', $set_error = true ) {
 function edd_get_discount_id_by_code( $code = '' ) {
 	$discount = edd_get_discount_by_code( $code );
 
-	return is_object( $discount ) ? $discount->id : false;
+	return ( $discount instanceof EDD_Discount ) ? $discount->id : false;
 }
 
 /**
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -905,11 +905,12 @@ function edd_is_discount_valid( $code = '', $user = '', $set_error = true ) {
  * @since 3.0 Updated to call edd_get_discount_by_code()
  *
  * @param string $code Discount code.
- * @return int Discount ID.
+ * @return int|bool Discount ID, or false if discount does not exist.
  */
 function edd_get_discount_id_by_code( $code = '' ) {
 	$discount = edd_get_discount_by_code( $code );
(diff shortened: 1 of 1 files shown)


includes/discount-functions.php
```diff
@@ -905,11 +905,12 @@ function edd_is_discount_valid( $code = '', $user = '', $set_error = true ) {
...
  *
  * @param string $code Discount code.
- * @return int Discount ID.
+ * @return int|bool Discount ID, or false if discount does not exist.
  */
 function edd_get_discount_id_by_code( $code = '' ) {
 	$discount = edd_get_discount_by_code( $code );
-	return $discount->id;
+
+	return is_object( $discount ) ? $discount->id : false;
```
spencerfinnell, line 913: @robincornett I think checking for a `false` value or an `instanceof EDD_Discount` seems more appropriate since those are the two expected return values of `edd_get_discount_by_code()`

robincornett: Thanks, the the `instanceof` check is one I've only used once before and I couldn't remember it. Updated.

PR #8021, opened 2020-06-17
Prevent empty addresses being added to the edd_order_addresses table

### LINKED ISSUE #8010: 3.0 - Empty addresses are added to the edd_order_addresses table
### Expected behavior
If no address data is entered for an order, I do not expect a database row to be created.

### Actual behavior
When I create an order with no address, a row is added to the `edd_order_addresses` table.

1) Create an order with no physical address information.
2) In your database, look in the `edd_order_addresses` table for the row with that `order_id`.

PHP Version: 7.4.1
WordPress Version: 5.4.1

There are related issues/work regarding the same issue with empty addresses being added to the customer records, updating the `edd_maybe_add_customer_address` function to check this.

robincornett:
Fixes #8010

To test:

Generate a number of orders using CLI. These orders do not have physical addresses associated with them, but in `release/3.0`, will have rows created in the `edd_order_addresses` table anyway. Switching to this branch should prevent those rows being created.

Example CLI code (will generate 10 orders for unique customers over the past 10 days):

```
wp edd payments create --number=10 --date=random --range=10 --generate_users=1
```


robincornett:
This is currently not working with a unit test which inserts a country using the `edd_payment_meta` filter. I don't know if that means the test should be updated since we don't use payment meta anymore, or if the code needs to be reevaluated to allow for payment meta to be inserted.

ashleyfae:
We'll need to look into that more because that'll basically be testing backwards compatibility. We won't want to edit the test itself.

robincornett:
I just "tested" locally by adding the same filter the test uses and generating two orders via CLI. Both properly have the address set in the `edd_order_addresses` table, and I'm able to retrieve it the way the test does as well.

ashleyfae:
It looks like the issue is here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/c8777b4cb930775d72439ff428f9f211a2b98472/includes/payments/class-edd-payment.php#L2269

Changing that `if` statement to:

```php
if ( null !== $this->order && $this->order->get_address()->id ) {
```

fixes the issue (that's not necessarily pretty code and maybe we should check it better, but just proving a point).

When it runs, `$this->order->get_address()` is returning the fallback array with an empty `id` value because there is no address at that point.

PR #7987, opened 2020-06-05
Update manual gateway label

robincornett:
Fixes #7067
Fixes #7848

PR #8034, opened 2020-06-29
Update migration for orders created with Manual Purchases extension

### LINKED ISSUE #8033: 3.0 - Orders from the Manual Purchases plugin are not migrated accurately
### Expected behavior
I expect the customer data and orders in my 3.0 storefront to match what's in my 2.9 storefront.

### Actual behavior
When I migrate orders added using the Manual Purchases addon, the order totals are not correct. For example, an order in a 2.9 storefront which was a "free" purchase shows in 2.9 as a $0 purchase for a $30 download item (because the manual purchases extension allows me to override the total), but in 3.0, that purchase shows a $30 total.

Additionally, there is no gateway label for the `manual_purchases` gateway and that should be mapped as well.

1) Have a 2.9 storefront with an order generated using the Manual Purchases extension. This should include an item which is not free, and the total (entered for Amount in Manual Purchases) should be 0.00, not automatically calculated.
2) Migrate the storefront to 3.0.
3) Check the customer value (this may need to be recalculated using the tool). The customer value will not match.
4) Check the manual purchase order. The total is no longer `0.00`.

PHP Version: 7.4.1
EDD Version (or branch): 3.0
WordPress Version: 5.4.2

From the Manual Purchases doc:
>A field is provided for you to enter in a total purchase amount. If you do not enter anything the total will be calculated from the total of all the products. If you do enter something, that total will be used, regardless of whether it tallies the total of the products or not.

I believe the migration is automatically calculating the total, but the extension explicitly allows users to override that, so the manually entered total should be kept.

robincornett:
Fixes #8033

Proposed Changes:
1. Since the Manual Purchases extension allows users to override the purchase total, use that if it is set.
2. Add a fallback label for orders with the `manual_purchases` gateway. This is currently set to "Manual Payment", which is what the extension uses, but could be changed.

**To Test:**
1. In a 2.9 storefront, use the Manual Purchases extension to create some orders. Make sure to override the purchase total on some.
2. Check out the `issue/8033` branch and migrate the payments. The order totals should reflect the manually set total, instead of the calculated total. (In `release/3.0`, the orders will use a calculated total, and will override the `_edd_payment_total` amount.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -485,6 +485,10 @@ class Data_Migrator {
 			$parent = $wpdb->get_var( $wpdb->prepare( "SELECT edd_order_id FROM {$wpdb->edd_ordermeta} WHERE meta_key = %s AND meta_value = %d", esc_sql( 'legacy_order_id' ), $data->ID ) );
 		}
 
+		if ( 'manual_purchases' === $gateway && isset( $meta['_edd_payment_total'][0] ) ) {
+			$total = $meta['_edd_payment_total'][0];
+		}
+
 		// Build the order data before inserting.
 		$order_data = array(
 			'id'             => $data->ID,
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -486,7 +486,8 @@ class Data_Migrator {
 		}
 
 		if ( 'manual_purchases' === $gateway && isset( $meta['_edd_payment_total'][0] ) ) {
-			$total = $meta['_edd_payment_total'][0];
+			$gateway = 'manual';
+			$total   = $meta['_edd_payment_total'][0];
 		}
 
 		// Build the order data before inserting.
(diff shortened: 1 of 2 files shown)


ashleyfae:
> Add a fallback label for orders with the manual_purchases gateway. This is currently set to "Manual Payment", which is what the extension uses, but could be changed.

Maybe we should convert those to `manual` on migration, which will then use the "Store Gateway" label.

robincornett:
That works for me--updated the migrator and labels accordingly.

LisaCee:
I'm not sure, but when I migrate a manual purchase from `release/2.9.23` to `issue/8033`, what gateway should I expect to see in Orders?  I expect `manual` or `manual_purchases` but I'm getting `test payment`.  

When I migrate from `release/2.9.23` to `release/3.0` I get a gateway of `manual_purchases`.

Other than the gateway question, the dollar amounts and customer reports look correct for your PR.

ashleyfae:
@LisaCee Right now you should see `Test Payment` because that's what our `manual` payment gateway is currently called. However, we have a separate PR that just changes the label to `Store Gateway`. https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7987 So for now seeing `Test Payment` is normal, but once the other PR is merged, that will change to `Store Gateway`.

PR #8049, opened 2020-07-09
Add "doing it wrong" when -1 is used in a query which no longer supports it (#7748)

### LINKED ISSUE #7748: 3.0 - Trigger "doing it wrong" when "-1" is passed to query function that no longer supports it

Split from here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7690#issuecomment-608116847

> We’ve had a few discussions about if we should still support `-1` to mean `all the things`. In part, it can cause some serious scalability issues which was the goal to remove them and stop supporting them within the addition of BerlinDB.
> 
> I think we should maybe look at places where we stop supporting `-1` and throw a doing it wrong, but for this issue, we should just fix the usage to the correct function.

This is relevant to functions that have been reworked with the introduction of BerlinDB in 3.0.

### Justification or use case

Conveying to developers that `-1` should no longer be used as a way to "get all results".

robincornett:
Fixes #7748

Proposed Changes:
1. Replace `-1` with `9999999` when using new database queries
2. Add `_doing_it_wrong` if `-1` is used in `edd_get_customers` or `EDD_Payments`

What this doesn't currently cover is if someone were to bypass a function such as `edd_get_customers` in favor of calling the class directly using something like (if `$args['number'] => -1`):

```php
$customers = new EDD\Database\Queries\Customer();

return $customers->query( $args );
```

### DIFF
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -219,6 +219,7 @@ function edd_get_customers( $args = array() ) {
 
 	if ( -1 == $args['number'] ) {
 		_doing_it_wrong( __FUNCTION__, esc_html__( 'Do not use -1 to retrieve all results.', 'easy-digital-downloads' ), '3.0' );
+		$args['number'] = 9999999;
 	}
 
 	// Instantiate a query object
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -217,9 +217,9 @@ function edd_get_customers( $args = array() ) {
 		'number' => 30
 	) );
 
-	if ( -1 == $args['number'] ) {
+	if ( -1 == $r['number'] ) {
 		_doing_it_wrong( __FUNCTION__, esc_html__( 'Do not use -1 to retrieve all results.', 'easy-digital-downloads' ), '3.0' );
-		$args['number'] = 9999999;
+		$r['number'] = 9999999;
 	}
 
 	// Instantiate a query object
--- a/includes/customer-functions.php
+++ b/includes/customer-functions.php
@@ -217,6 +217,10 @@ function edd_get_customers( $args = array() ) {
 		'number' => 30
 	) );
 
+	if ( -1 == $args['number'] ) {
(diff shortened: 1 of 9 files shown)


includes/payments/class-payments-query.php
```diff
@@ -633,7 +634,6 @@ private function remap_args() {
 
 		if ( isset( $this->args['nopaging'] ) && true === $this->args['nopaging'] ) {
 			// Setting to a really large number because we don't actually have a way to get all results.
-			// @todo Maybe trigger doing_it_wrong.
 			$arguments['number'] = 99999;
```
spencerfinnell, line 637:
```suggestion
			$arguments['number'] = 9999999;
```

Just to match the number used elsewhere.

includes/customer-functions.php
```diff
@@ -217,6 +217,11 @@ function edd_get_customers( $args = array() ) {
 		'number' => 30
 	) );
 
+	if ( -1 == $args['number'] ) {
```
ashleyfae, line 220:
```suggestion
	if ( -1 == $r['number'] ) {
```

This needs to be `$r` because `$args` is only looking at the value passed in by the user, which may not contain `number` at all. `$r` is after we've merged the supplied arguments with our defaults.

Fixing this will fix the unit tests complaining about :

```
Undefined index: number
/home/travis/build/easydigitaldownloads/easy-digital-downloads/includes/customer-functions.php:220
```

includes/customer-functions.php
```diff
@@ -217,6 +217,11 @@ function edd_get_customers( $args = array() ) {
...
+		_doing_it_wrong( __FUNCTION__, esc_html__( 'Do not use -1 to retrieve all results.', 'easy-digital-downloads' ), '3.0' );
+		$args['number'] = 9999999;
```
ashleyfae, line 222:
```suggestion
		$r['number'] = 9999999;
```

To match with the above.

includes/customer-functions.php
robincornett: Thanks, I just caught that--fixed in latest.

PR #8052, opened 2020-07-09
Update product dropdown placeholders for exports (#8046)

### LINKED ISSUE #8046: Export options should an All option in all dropdown boxes
### Explain your enhancement 
Export options should always include an All option in the dropdown and should default to that. 

For example, in this screenshot Export Orders has an All Statuses option, selected by default. But Export Sales and Export Taxed Orders don't have All Downloads, All Customers or All Countries options.
Screen Shot 2020-07-08 at 12 02 02 PM

### Justification or use case
We should be consistent for all export options. Also, it's not obvious to all users or a good user experience that no selection would export all.

robincornett:
Fixes #8046

Proposed Changes:
1. Change the placeholder (default/`0` value) for the product dropdowns on Downloads > Reports > Export to say "All Downloads" (use `edd_get_label_plural` for storefronts which have modified the labels) instead of "Choose a Download".
2. Remove explanatory text which is no longer needed.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -2602,7 +2602,8 @@ function display_export_report() {
 									'name'        => 'download_id',
 									'id'          => 'edd_orders_export_download',
 									'chosen'      => true,
-									'placeholder' => __( 'All Downloads', 'easy-digital-downloads' ),
+									/* translators: the plural post type label */
+									'placeholder' => sprintf( __( 'All %s', 'easy-digital-downloads' ), edd_get_label_plural() ),
 								) );
 
 				                echo EDD()->html->customer_dropdown( array(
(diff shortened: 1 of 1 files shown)


robincornett:
Thanks, @dgoldak, have updated those as well.

robincornett:
@dgoldak can you confirm that you have pulled the latest changes, please? This is what those fields look like for me:

image

(I removed the "All.." options from countries/regions because it felt confusing.)

spencerfinnell:
@robincornett If the processing code doesn't need to change could `All Statuses` be moved to a `placeholder` vs `show_option_all`? If that affects what gets exported we can leave it.

Screen Shot 2020-07-22 at 11 09 45 AM

robincornett:
@spencerfinnell in this dropdown, the `placeholder` parameter is not honored--changing `show_option_all` to `placeholder` just results in `All` as the visible label. But looking at this, I wonder:
- should `show_option_none` be set to `false` because I don't know what that would even do with statuses?
- is `chosen` necessary for the status dropdowns?

spencerfinnell:
> is chosen necessary for the status dropdowns?

No, let's remove it from Status since the list isn't populated by a dynamic search.

> should show_option_none be set to false

I believe so. All Orders should have a status.

spencerfinnell:
@robincornett Let's do the same for "All Taxonomies" under "Export Customers" as well.

Issue #7648, opened 2020-03-17
Dynamic discount codes


tubiz:
### Explain your enhancement (please be detailed)

In AffiliateWP, we need to be able to validate/generate a dynamic coupon when a coupon is being applied on the EDD checkout page.

The dynamic coupon won't exist in EDD but it will be based on an existing coupon created in EDD.

### Justification or use case

In AffiliateWP, we have a new feature planned to auto-create tracked coupons for affiliates on registration https://github.com/AffiliateWP/AffiliateWP/issues/426

The currently proposed implementation is to select a discount code already created in EDD which will be the base when a dynamic coupon is being applied on the checkout page.

With this approach, if the site owner wants to change the details of the discount code assigned to affiliates only the base discount code will be updated in EDD.

cklosowski:
> The dynamic coupon won't exist in EDD but it will be based on an existing coupon created in EDD.

This is going to be interesting considering in EDD 3.0, we’re requiring all discounts to be applied as cart adjustments, and thus require a tie back to some sort of object ID, so that reports can be run on them.

Is the proposal a ‘ghost’ discount of sorts?

ashleyfae:
Related / RCP: https://github.com/restrictcontentpro/restrict-content-pro/issues/2754

@cklosowski Also a bit more info from Drew here: https://github.com/restrictcontentpro/restrict-content-pro/issues/2754#issuecomment-599632641

tubiz:
@cklosowski I have also been thinking about this with regards to reporting as well.

> Is the proposal a ‘ghost’ discount of sorts?

Yes, a discount that is valid but doesn't exist.

In the first iteration, only WooCommerce integration will be supported pending when the required changes in made in EDD/RCP for this feature.

cklosowski:
@tubiz Could you detail a little bit or link to some code in how you're doing this with Woocommerce so we can see what is going on there.

tubiz:
@cklosowski here is the current implementation in the WooCommerce integration https://github.com/AffiliateWP/AffiliateWP/blob/ed052d9604f5604765e20851d4bcc0ffa7fb218d/includes/integrations/class-woocommerce.php#L1358-L1417

PR #8057, opened 2020-07-14
Update fees added to the order adjustments table

### LINKED ISSUE #8056: 3.0 - Fees are not consistent with 2.9
### Expected behavior
I expect fees to be displayed consistently with EDD 2.9.

### Actual behavior
I have two sites with taxes enabled, one running 2.9 and one running 3.0. I have used `EDD()->fees->add_fee()` to add the same fee to each site, with `no_tax` set to `false`, so that taxes are applied to the fees as well as the items in the cart.

Although the order total is calculated correctly for each storefront, the fee is displayed differently on the purchase receipts. In 2.9, the fee is displayed without tax:

image

In 3.0, the fee is displayed with tax:

image

Add a fee to a site which collects taxes (based on [docs](https://docs.easydigitaldownloads.com/article/543-edd-fees)):

```php
add_action( 'init', 'pw_add_handling_fee' );
function pw_add_handling_fee() {
	if ( ! function_exists( 'EDD' ) ) {
		return;
	}
	// Add a $10 handling fee no matter what
	EDD()->fees->add_fee(
		array(
			'amount'      => 10,
			'label'       => 'Handling Fee',
			'id'          => 1235444444,
			'type'        => 'fee',
			'no_tax'      => false,
		)
	);
}
```

Place an order which collects sales tax in 2.9/3.0 and see how the fee shows differently on the purchase receipt.

One other inconsistency is that the fee ID is not added to the database in 3.0.

PHP Version: 7.4.1
WordPress Version: 5.4.2

There may be other inconsistencies/issues with fees as well, but digging into #8012 is my first time looking into the fees class. This issue is related to that one, but I felt that the fees maybe merited their own issue/research.

robincornett:
Fixes #8056

Branched from #8012 due to needing tax rates; should be held until after that's resolved.

Proposed Changes:
1. Add fee ID to the order adjustments table `type_id` column.
2. Create a new `EDD()->fees->get_calculated_tax()` method to get the tax amount to be applied to a fee and use it where possible (not possible in `EDD_Payment` meta functions that I can tell)
3. Update one out of sync implementation of the fee creation to be consistent with others (add tax/update total).

Not yet addressed: if a fee registered with `EDD()->fees->add_fee()` is assigned to a specific product, the fee is currently stored against both the `order_item` and the `order` and it should probably only be stored against the `order_item`.


robincornett:
@cklosowski or @ashleyfae: per the [EDD Fees docs](https://docs.easydigitaldownloads.com/article/543-edd-fees), a fee ID can be an alphanumeric value (the docs use `handling_fee`). However, the `type_id` schema is set to `type_id bigint(20) unsigned NOT NULL default '0'` when the order adjustments table is created, so any fees registered with a non-numeric ID will not be saved to the database.

Can we update the schema for this column?

robincornett:
Another question on this: as I remove fees from `order` to get rid of the duplication, I am discovering that doing so means that order calculations will also have to be updated, as they're done from the `order`, not the `order_item`. So it appears that getting rid of the duplication has potential for extended consequences and searching for all possible affected actions (orders, refunds, just for starters). So my question for @ashleyfae or @cklosowski is which is preferable: duplicate data or digging through affected calculations to manage the ramifications of removing duplication?

robincornett:
Further research shows that fees registered using `EDD->fees()->add_fee()` are not applied to manually created orders in 3.0, and are not available for use there either. However, this appears to be on par with the Manual Purchases extension behavior in 2.9--orders placed with that plugin also do not have registered fees added, for at least the few I've tested.

Extensions which use the EDD fees class:
- Recurring
- Discounts Pro
- Gateway Fees
- Content Restriction
- Commissions
- Simple Shipping
- Volume Discounts
- Wallet
- RCP Member Discounts
- Ninja Forms

Related, because of manual purchases, but when a custom fee is added to a manual order, there is no `type_id` added to the database for the fee, which makes sense, but should this kind of fee (and presumably the same for credit) have something other than `0` for the `type_id`?

ashleyfae:
> @cklosowski or @ashleyfae: per the [EDD Fees docs](https://docs.easydigitaldownloads.com/article/543-edd-fees), a fee ID can be an alphanumeric value (the docs use `handling_fee`). However, the `type_id` schema is set to `type_id bigint(20) unsigned NOT NULL default '0'` when the order adjustments table is created, so any fees registered with a non-numeric ID will not be saved to the database.
> 
> Can we update the schema for this column?

Loosely related: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7648#issuecomment-600414135 Just because it talks about order adjustments without an associated ID.

I'm not sure I love the `varchar` just because then it's no longer an _ID_ as the column suggests. (object_type + object_id = an object) If there's no object in the database, then the object ID (`type_id`) should be empty.

In 2.9, what is that fee ID used for once it's been saved to the database? Just trying to determine 1) how badly we need to save that; and 2) if there's somewhere else we could put it instead.


> Another question on this: as I remove fees from `order` to get rid of the duplication, I am discovering that doing so means that order calculations will also have to be updated, as they're done from the `order`, not the `order_item`. So it appears that getting rid of the duplication has potential for extended consequences and searching for all possible affected actions (orders, refunds, just for starters). So my question for @ashleyfae or @cklosowski is which is preferable: duplicate data or digging through affected calculations to manage the ramifications of removing duplication?

I'm not necessarily opposed to some duplication for now. We can always refactor that at a later date.



> Related, because of manual purchases, but when a custom fee is added to a manual order, there is no `type_id` added to the database for the fee, which makes sense, but should this kind of fee (and presumably the same for credit) have something other than `0` for the `type_id`?

If it's a custom fee created on the fly, then `0` makes sense. (Frankly I prefer `null` as a default, but that's a different discussion. :woman_shrugging: ) The `type_id` is meant to link to an object in the database. If the object doesn't exist, then there should be no ID (which is represented by `0` here).

robincornett:
Noted for fixing: when refunding an order, if a fee has been added with `EDD()->fees->add_fee()`, the fee is not refunded.

robincornett:
Looking through 2.9 code and tables, the fee ID is stored as the fee's array key in `_edd_payment_meta`. When a payment is retrieved with ` $payment = new EDD_Payment( $payment_id )`, the fees in the meta are added to the `$payment->fees` as an array. The key is not kept as a key (not sure why), but is copied over to an `id` key within each fee array retrieved. Here's an example of what's retrieved for a payment which includes a manually added fee, a gateway fee (from the extension), and a fee added using the Pro Discounts extension:

```
[29-Jul-2020 12:47:10 UTC] Array
(
    [0] => Array
        (
            [amount] => 20.00
            [label] => Fee Added
            [no_tax] => 
            [type] => fee
            [download_id] => 0
            [price_id] => 
            [id] => order_item_fee
        )

    [1] => Array
        (
            [amount] => 4.23
            [label] => Test Payment Gateway Fee
            [type] => fee
            [no_tax] => 
            [download_id] => 0
            [price_id] => 
            [id] => gateway_fee
        )

    [2] => Array
        (
            [amount] => -10.80
            [label] => Pro Discount Test - #1553 One Year
            [no_tax] => 1
            [type] => fee
            [download_id] => 1425
            [price_id] => 
            [id] => dp_1425
        )

)
```

The [order details screen](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/admin/payments/view-order-details.php#L160-L170) is an example of how that's output.

@ashleyfae, I think I understand what you are saying about not wanting to change the `type_id`, but it seems important to keep the fee ID as part of the data, and I'm thinking it could become something more useful in 3.0 (I'm thinking of things like reporting, for example, beyond just the day to day of viewing order details).

ashleyfae:
Here's what I'm leaning towards:

`type_id` - defaults to `null`. In the case of a custom fee, this would be `null`. This remains a `bigint` column.

We add a new, nullable column for `type_key` or something like that. In EDD core this will probably always be `null`, but it's where we can put the slug-style IDs.

This helps 1) clarify that the fee does not link to an "official" fee ID in the database; and 2) it still retains the information.

robincornett:
If that will allow a query for all manually added fees with a specific ID, eg every order with a `gateway_fee` added, then I think that sounds reasonable. Do you want to update the PR or do you want me to take a stab at it?

ashleyfae:
Before we proceed, @JJJ do you have any alternative thoughts?

JJJ:
Hi @ashleyfae and @robincornett! I think I'm all caught up now. Phew, quite a wild ride to get here.

## Vision

@sunnyratilal and I's ultimate vision for Fees as a feature for EDD3 and beyond, was to get them out of being registered in PHP just-in-time and on-the-fly (like a custom post type) and instead to move them into something that (like most other EDD3 things) was immutable-by-choice and persistently stored in the database.

In this way, a "Fee" would be similar to a "Discount" in that it has a primary record that lives in `edd_adjustments` and that can be pointed to as needed by `edd_order_adjustments` – or even `edd_order_items` if (for some crazy reason) we wanted a fee to feel more like an item than an adjustment.

(The `'fee_id'` would then be derived from the adjustment `'code'` rather than simply being derived from the description using `sanitize_key()`.)

In this way, a family of functions like `edd_get_fee()` would wrap around the adjustments API the way that Discounts does, and the `EDD_Fees` class would become the session/cart handler only.

## Adjustments

The reason we planned, conceptually, for Fees, Discounts, Taxes, and the theoretical "Credits" idea (which hasn't been brought up in a while) to all use the shared Adjustments API was so they could all be queried for together at once given a known set of IDs & codes from any set of Order(s)/Item(s)/Adjustment(s). If instead they all had their own more-strictly defined database schemas, each thing would always _need_ their own specific query.

Once we laid out what data all of those things were comprised of, they were all nearly identical (loosely based off of discounts originally) and so a meta-data table was introduced to help with any gaps.

The goal is for them all was to work and feel like how `wp_get_object_terms()` works in WordPress.

Given any number of Order IDs, get all of the relative Adjustments from the database and prime them in the cache. That way, with 1 database query, you can get everything, and subsequent calls to `edd_get_discount()` or `edd_get_tax_rate()` will already be hot in the cache and never hit the database a second time. This ends up being specifically useful in lists of Orders (in list-tables or Customer/Account pages) where we query for Orders first, have their IDs ready, go out and prime everything once, and never need to worry about function calls causing performance problems.

## type_id thoughts

I think... changing the `type_id` column to `varchar` might not be a good idea. 😕 That will result in that column being used to store _sometimes_ integers and _sometimes_ strings (depending on what kind of adjustment that it is)  which feels like a compromise that _will_ confuse people who go looking at the data directly.

Specifically, it will confuse ourselves later and forever forward - code maintainers and the people responsible for supporting EDD and the related add-ons to it who need to dig this deep will need to learn this sharp edge exists to avoid being cut by it a second time.

Performance wise, queries on that column will now perform worse, though that won't show up until the tables have enough data in them for it to matter. The way it was previously, unindexed integer queries would've performed better than they will now that it's a unicode lookup rather than a numeric one.

Looking at two unit tests specifically...

* `test_get_order_adjustments_with_orderby_type_id_and_order_asc_should_return_true()`
* `test_get_order_adjustments_with_orderby_type_id_and_order_desc_should_return_true()`

...that order by the `type_id` column. It being an unindexed `varchar` (that is mostly comprised of integers) means it is going to perform worse than when it was simply an integer column, even when some rows have `0` in there. 📉

Related to that, it seems undesirable (perhaps not intentional) to duplicate the storage of `'fee_id'` in order-adjustment meta-data as well as in the `type_id` column. If it's already in meta data (and queried by it via `meta_query` in a few places) why does it _also_ need to be in the `type_id` column, particularly when `'type_id'` is not being queried by for fees, and `->type_id` is not being used for fee IDs anywhere? 🤔

I'm afraid that database column change makes unwinding the rest of this PR somewhat of a challenge. Either that work is incomplete and the `fee_id` meta-data storage duplication & `meta_query` should be removed, or the `type_id` column work should be reverted and potentially moved into its own PR.

## Wrap-up

In general, everything but the `type_id` column changes make sense to me, which usually means I'm overlooking something or doing something wrong. I've read through #8012, #7859, #6810, and some related issues, but I still don't think I understand why it's necessary to recommend any alternatives. 🥸

Please accept my sincere apologies for the massive comment. Only wanting to be helpful and encouraging. 💙

robincornett:
Latest commits update the `edd_order_adjustments` table to add the new `type_key` column and I've hopefully found all of the places it's used.

Local unit tests are exploding spectacularly, all objecting to an unknown `type_key` column, so I've evidently done something incorrectly with getting that added to the database, although initial [manual] tests are getting fees added via the Fees class set up properly. So pushing here for more eyes.

The only thing I found about the database in the tests was the `tests\phpunit\factories\class-edd-factory-for-order-adjustments.php` class, and I tried adding the `type_key` there, but it didn't help with the tests' breaking.

includes/database/tables/class-order-adjustments.php
```diff
@@ -64,7 +65,8 @@ protected function set_schema() {
 		$this->schema = "id bigint(20) unsigned NOT NULL auto_increment,
 		object_id bigint(20) unsigned NOT NULL default '0',
 		object_type varchar(20) DEFAULT NULL,
-		type_id bigint(20) unsigned NOT NULL default '0',
+		type_id bigint(20) unsigned DEFAULT NULL,
+		type_key varchar(32) DEFAULT '',
```
ashleyfae, line 69:
```suggestion
		type_key varchar(32) DEFAULT NULL,
```

Prefer a null default.

includes/database/tables/class-order-adjustments.php
```diff
@@ -145,4 +147,29 @@ protected function __202002141() {
...
+		// Update `type_id`.
+		$result = $this->get_db()->query( "
+			ALTER TABLE {$this->table_name} MODIFY COLUMN `type_id` bigint(20) default NULL;
+		" );
+
+		// Add `type_key`.
+		$column_exists = $this->column_exists( 'type_key' );
+		if ( false === $column_exists ) {
+			$result = $this->get_db()->query( "
+				ALTER TABLE {$this->table_name} ADD COLUMN `type_key` varchar(32) default '' AFTER `type_id`;
```
ashleyfae, line 169:
```suggestion
				ALTER TABLE {$this->table_name} ADD COLUMN `type_key` varchar(32) default NULL AFTER `type_id`;
```

To match the above.

includes/database/tables/class-order-adjustments.php
ashleyfae: Also I'm wondering if we should make it longer than `32` given that the old-style ID field is kind of a free for all.

includes/database/schemas/class-order-adjustments.php
```diff
@@ -69,9 +69,18 @@ class Order_Adjustments extends Schema {
 			'type'       => 'bigint',
 			'length'     => '20',
 			'unsigned'   => true,
-			'default'    => '0',
+			'default'    => null,
 			'searchable' => true,
-			'sortable'   => true
+			'sortable'   => true,
```
ashleyfae, line 74:
```suggestion
			'sortable'   => true,
			'allow_null' => true,
```

Required to allow a `null` value.

includes/database/tables/class-order-adjustments.php
```diff
@@ -145,4 +147,29 @@ protected function __202002141() {
...
+			" );
+		}
```
ashleyfae, line 171:
```suggestion
		}
		
		// Change `type_id` with `0` value to `null` to support new default.
		$this->get_db()->query( "UPDATE {$this->table_name} SET type_id = null WHERE type_id = 0;" )
```

To convert old `0` values to `null`.

includes/orders/functions/actions.php
```diff
@@ -308,6 +308,7 @@ function edd_add_manual_order( $args = array() ) {
 							'object_id'   => $order_item_id,
 							'object_type' => sanitize_text_field( $order_item_adjustment['object_type'] ),
 							'type'        => sanitize_text_field( $order_item_adjustment['type'] ),
+							'type_key'    => sanitize_key( $order_item_adjustment['description'] ),
```
ashleyfae, line 311: Not sure about this. I'd rather remove it.

includes/orders/functions/actions.php
```diff
@@ -357,6 +358,7 @@ function edd_add_manual_order( $args = array() ) {
 				'object_id'   => $order_id,
 				'object_type' => 'order',
 				'type'        => sanitize_text_field( $adjustment['type'] ),
+				'type_key'    => sanitize_key( $adjustment['description'] ),
```
ashleyfae, line 361: Same as above.

includes/orders/functions/orders.php
```diff
@@ -1036,7 +1039,7 @@ function edd_build_order( $order_data = array() ) {
 			// Add the adjustment.
 			$adjustment_id = edd_add_order_adjustment( $args );
 
-			edd_add_order_adjustment_meta( $adjustment_id, 'fee_id', $key );
+			edd_add_order_adjustment_meta( $adjustment_id, 'fee_id', $fee_id );
```
ashleyfae, line 1042: Given that this is now stored in the main table, we should be able to remove this and all other such references (there are quite a few). For example:

`includes/orders/classes/class-order.php`, line 451:

```php
if ( 'fee' === $adjustment->type ) {
	$fee_id = edd_get_order_adjustment_meta( $adjustment->id, 'fee_id', true );

	$fees[ $fee_id ] = $adjustment;
}
```

We should be able to remove all instances of getting/setting this and use the new `type_key` instead.

ashleyfae (changes requested):

includes/payments/actions.php
```diff
@@ -70,12 +70,12 @@ function edd_complete_purchase( $order_id, $new_status, $old_status ) {
 					foreach ( $item->get_fees() as $key => $item_fee ) {
 						/** @var EDD\Orders\Order_Adjustment $item_fee */
 
-						$fee_id      = edd_get_order_adjustment_meta( $item_fee->id, 'fee_id', true );
 						$download_id = edd_get_order_adjustment_meta( $item_fee->id, 'download_id', true );
 						$price_id    = edd_get_order_adjustment_meta( $item_fee->id, 'price_id', true );
 						$no_tax      = edd_get_order_adjustment_meta( $item_fee->id, 'price_id', true );
+						// todo: should $no_tax be changed to (bool) 0.00 === $item_fee->tax;?
```
robincornett, line 76: Noted a potential issue on this `$no_tax` variable, also in `setup_cart_details` ... in `setup_fees`, the `$no_tax` variable is defined using:

```php
$no_tax   = (bool) 0.00 === $order_fee->tax;
```
and it seems these two should use the same? The `price_id` meta seems like a copy/paste error maybe?

includes/database/tables/class-order-adjustments.php
robincornett: The `meta_key` in the `postmeta` table is set to 255.

includes/orders/classes/class-order.php
```diff
@@ -446,12 +447,7 @@ public function get_fees() {
 		// Fetch the fees that applied to the entire order.
 		foreach ( $this->adjustments as $adjustment ) {
 			/** @var Order_Adjustment $adjustment */
-
-			if ( 'fee' === $adjustment->type ) {
-				$fee_id = edd_get_order_adjustment_meta( $adjustment->id, 'fee_id', true );
-
-				$fees[ $fee_id ] = $adjustment;
-			}
+			$fees[ $adjustment->type_key ] = $adjustment;
```
ashleyfae, line 450: Sorry I should have caught this earlier.. But I think we should add in some "just in case" checks to see if `$fees[ $adjustment->type_key ]` is empty or if it already exists in the `$fees` array before setting this. `type_key` can technically be null (and would be for manually created orders with manually added fees) and we want to make sure we're not overwriting any data in the array.

If `type_key` is empty, we could fall back to using `$adjustment->id` instead.

This would need to be updated any time we're looping over fees and adding to a `$fees` array.

includes/payments/actions.php
ashleyfae: What you've described seems fine to me. I think you're right this looks like a copy paste error.

ashleyfae (changes requested):

includes/orders/classes/class-order.php
robincornett: Thanks, I'm glad you caught that--I was actually wondering about that a little since we removed the `type_key` from the manual orders. Will update!

tests/orders/tests-edd-payment.php
```diff
@@ -316,14 +316,17 @@ public function test_payment_remove_fee() {
 		$this->assertEquals( 'Test Fee 2', $this->payment->fees[1]['label'] );
 
 		// Test that it saves to the DB
-		$payment_meta = edd_get_payment_meta( $this->payment->ID, '_edd_payment_meta', true );
-
-		$this->assertArrayHasKey( 'fees', $payment_meta );
-
-		$fees = $payment_meta['fees'];
+		$fees = edd_get_order_adjustments(
```
robincornett, line 319: I cannot replicate it by hand, but this is failing:

```php
$this->assertEquals( 3, count( $this->payment->fees ) );
```
The fee added with `$i = 0` is not being counted during the tests. However, copying the logic over and running it by hand, it is (I initially thought the issue was with doing an `empty` check, but the above code gives me `3` fees, just not with the tests).

Not sure how to resolve this one and would love any insight.

ashleyfae: @robincornett The issue is with this change (in various places):

```php
$id          = empty( $adjustment->type_key ) ? $adjustment->id : $adjustment->type_key;
$fees[ $id ] = $adjustment;
```

The reason is that, for `Test Fee 0`, the `type_key` is `0`, so it's hitting that `empty()` check, then falling back to using the ID instead, which is `1`, but then there's another fee that does have `type_key` `1`. So it's being overwritten.

There are two changes we should make here:

1. Instead of `empty( $adjustment->type_key )` we should use `is_null( $adjustment->type_key )`. This will allow `0` to be used, but not `null`.
2. After setting `$id` (but before assigning it) we should check if that already exists in the array, and if so, change it to something else. We could just append `_2` or something since the value isn't ultimately particularly important.

Full example:

```php
$id = is_null( $adjustment->type_key ) ? $adjustment->id : $adjustment->type_key;
if ( array_key_exists( $id, $fees ) ) {
	$id = $id . '_2';
}

$fees[ $id ] = $adjustment;
```

includes/database/schemas/class-order-adjustments.php
```diff
@@ -69,9 +69,19 @@ class Order_Adjustments extends Schema {
...
+			'allow_null' => true,
+		),
+
+		array(
+			'name'       => 'type_key',
+			'type'       => 'varchar',
+			'length'     => '255',
+			'default'    => '',
```
ashleyfae, line 82:
```suggestion
			'default'    => null,
			'allow_null' => true,
```

ashleyfae (changes requested):
One minor change to get the DB schema array lined up with the reality. But this is looking really good, thanks @robincornett !

includes/database/tables/class-order-adjustments.php
```diff
@@ -64,7 +65,8 @@ protected function set_schema() {
 		$this->schema = "id bigint(20) unsigned NOT NULL auto_increment,
 		object_id bigint(20) unsigned NOT NULL default '0',
 		object_type varchar(20) DEFAULT NULL,
-		type_id bigint(20) unsigned NOT NULL default '0',
+		type_id bigint(20) unsigned DEFAULT NULL,
+		type_key varchar(255) DEFAULT NULL,
```
ashleyfae, line 69: This is really unimportant, but can you move `type_key` to after `type` ? This will need to be done:

1. Here
2. In the migration further down.
3. In the schema file.

I just like having `type_id` and `type` next to each other. :P

robincornett:
Note: even though there wasn't a merge conflict, I merged `release/3.0` back into this branch because the date tests were failing for me. With that merged back in, all local unit tests passed for me.

ashleyfae:
It looks like `type_id` is still getting saved as `0` instead of `null`. It's due to this validation: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8056/includes/database/engine/class-query.php#L1999-L2010

Right before that, the value is `null`, but then the `$callback` is `intval` due to the column being `bigint`, and it's running `intval( null )` which is turning it into `0` instead.

We can patch this here but I'll also make a separate issue in BerlinDB.

If we land on this if statement: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8056/includes/database/engine/class-query.php#L1988-L1989 we don't really need to run through validation because we are explicitly setting the value. So perhaps the way to solve this is to set a variable here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8056/includes/database/engine/class-query.php#L1986 for `$should_validate = true` and if we land inside that null if statement, change it to `false`. Then we can adjust this if statement: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8056/includes/database/engine/class-query.php#L2000 to also check `&& $should_validate`

ashleyfae:
@robincornett Actually do you mind if I fix this particular issue? There's a better way which involves some BerlinDB stuff.

robincornett:
@ashleyfae please help yourself!

includes/orders/functions/orders.php
```diff
@@ -967,20 +967,21 @@ function edd_build_order( $order_data = array() ) {
 						? floatval( floatval( $fee['amount'] ) - ( floatval( $fee['amount'] ) / ( 1 + $tax_rate->amount ) ) )
```
ashleyfae, line 967: There's a tax variable up here that can be removed.

ashleyfae (changes requested):

ashleyfae (approved):
Pushed up some changes that tweak the null handling in BerlinDB, as per some upstream changes. Also added some unit tests to check the null values get saved and returned properly.

Unit tests are passing for me locally and I'm happy with this. :+1: Thanks @robincornett !

PR #8058, opened 2020-07-15
Fix missing/hidden headings on admin screens

### LINKED ISSUE #8040: 3.0 - Some admin screens are missing h1/title elements
### Expected behavior
All pages should have titles (`h1` elements) ([ref](https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one)).

### Actual behavior
Some EDD admin screens are missing `h1` elements altogether, while others do have the markup, but the `h1` has been hidden in an inaccessible manner.

Screens with no `h1`:
- Downloads > Categories
- Downloads > Tags

Screens with `h1` hidden using `display: none;`, which hides an element visually and for non-visual users both, and removes the element from the DOM:
- Downloads > Downloads
- Downloads > Orders
- Downloads > Refunds
- Downloads > Customers
- Downloads > Email Addresses
- Downloads > Physical Addresses
- Downloads > Discounts (hidden `h1` is "Adjustments")

Screens with correct/visible `h1`:
- Downloads > Orders > New/View Order (also View Refund)
- Downloads > Customers > Customer Details
- Downloads > Discounts > Add/Edit Discount
- Downloads > Reports
- Downloads > Tools
- Downloads > Settings
- Downloads > Extensions

1) Check out `release/3.0`.
2) Visit the admin screens listed and inspect the `h1` element.
3) Run an accessibility scan using an accessibility tool such as [axe](https://www.deque.com/axe/).

PHP Version: 7.4.1
WordPress Version: 5.4.2

Although at least one heading is hidden this way in 2.9, it looks like the bulk of the work was done as part of #2880, to update/refresh the EDD admin screens. For the screens where this visual behavior is desired/preferred, one option would be to remove the CSS rule and instead adding a `screen-reader-text` class to the `h1` elements to be hidden, which will hide them visually, but still allow them to be accessed by assistive technology. Another option would be to update the CSS rule so that it replicates the WordPress Core `screen-reader-text` rule.

robincornett:
Fixes #8040

Proposed Changes:
1. Remove CSS which inaccessibly hides `h1` elements on admin screens.
2. Relocate navigation tabs on download post type taxonomy screens.
3. Remove duplicate "Add New" button from download post type/taxonomy screens to maintain consistency with WP Core screens, and remove confusion about new "what" on taxonomy screens.


ashleyfae:
@robincornett There are some merge conflicts now - can you look into those?

ashleyfae (approved):

PR #8059, opened 2020-07-15
Update admin screens tabbed navigation

### LINKED ISSUE #8055: 3.0 - Update tabbed settings navigation
We should update our settings pages' tabbed navigation to be consistent with [WordPress Core](https://core.trac.wordpress.org/ticket/43398#comment:17).

### Justification or use case
Summary of what Core changed:

> Accessibility: Improve the semantics of "tabs" that are not real ARIA tabs.
>
> Some "visual tabs" in the WordPress admin used links wrapped in a `h2` heading. While these tabs look like ARIA tabs, they're just navigation menus and should be marked-up as such to be correctly exposed to assistive technologies.
>
> -  changes the wrapping `<h2>` to a `<nav>` element: worth reminding `<nav>` elements also define ARIA landmarks
> -  adds an `aria-label` to the `<nav>` elements so they can be distinguished from other `<nav>` elements in the page
> -  adjusts the headings level in the Credits page

It's worth noting that updating this improves markup and accessibility and should be backwards compatible with 4.x versions of WordPress, as the linked changeset doesn't modify Core CSS, except for the "About" page. Ran into this myself when trying to work on #8040, as that is related.

robincornett:
Fixes #8055

Proposed Changes:
1. Change admin screens' tabbed navigation to use `nav` instead of `h2`
2. Adds `aria-label` to tabbed navigation
3. Removes unneeded CSS

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -39,34 +39,41 @@ function edd_tools_page() {
 		<h1><?php esc_html_e( 'Tools', 'easy-digital-downloads' ); ?></h1>
 		<hr class="wp-header-end">
 
-		<nav class="nav-tab-wrapper edd-nav-tab-wrapper" aria-label="<?php esc_attr_e( 'Secondary menu', 'easy-digital-downloads' ); ?>"><?php
+		<nav class="nav-tab-wrapper edd-nav-tab-wrapper" aria-label="<?php esc_attr_e( 'Secondary menu', 'easy-digital-downloads' ); ?>">
+		<?php
 
-			foreach ( $tabs as $tab_id => $tab_name ) {
+		foreach ( $tabs as $tab_id => $tab_name ) {
 
-				$tab_url = edd_get_admin_url( array(
+			$tab_url = edd_get_admin_url(
+				array(
 					'page' => 'edd-tools',
-					'tab'  => $tab_id
-				) );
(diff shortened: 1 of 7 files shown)


includes/admin/tools.php
```diff
@@ -35,11 +35,11 @@ function edd_tools_page() {
...
 ?>
 
-    <div class="wrap">
-        <h1><?php _e( 'Tools', 'easy-digital-downloads' ); ?></h1>
+	<div class="wrap">
+		<h1><?php esc_html_e( 'Tools', 'easy-digital-downloads' ); ?></h1>
 		<hr class="wp-header-end">
 
-        <h2 class="nav-tab-wrapper edd-nav-tab-wrapper"><?php
+		<nav class="nav-tab-wrapper edd-nav-tab-wrapper" aria-label="<?php esc_attr_e( 'Secondary menu', 'easy-digital-downloads' ); ?>"><?php
```
ashleyfae, line 42: Inconsequential, but:

Can we move the `<?php` opening tag to a new line.
Then at the bottom where we have `?></nav>` can you split those onto separate lines and make sure the indents line up with the opening tags?

(I know it's not related to your changes, but since we're in the area making some markup tweaks, might as well make these line up a bit nicer.)

PR #8067, opened 2020-07-23
Update Chosen script/styles (#7722)

### LINKED ISSUE #7722: Replace "chosen" script with a more accessible option
## Enhancement Request

> ...we should look at deprecating Chosen (which itself has been Deprecated https://github.com/harvesthq/chosen) and replacing it with a more maintained alternative, such as:
> 
> - https://github.com/jshjohnson/Choices
> - https://github.com/woocommerce/selectwoo
> 
> Related https://github.com/easydigitaldownloads/easy-digital-downloads/issues/2758 https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4375
> 
> _Originally posted by @spencerfinnell in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7583#issuecomment-587152955_

### Justification or use case
Any select option using Chosen is currently not accessible to keyboard or screen reader users.

robincornett:
Fixes #7722

Proposed Changes:
1. Update to use @JJJ's Chosen fork
2. Update EDD specific styles to more closely match WordPress Core (5.2+)


spencerfinnell:
@robincornett The current `jquery.chosen.min.js` is moved from the Chosen NPM package via Webpack: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/7cd96a6752c9f8ff0e4126a7a6c4740572b349c7/webpack.config.js#L80-L83

If we are moving forward with @JJJ's fork we we can remove the handling from Webpack and the [`chosen-js` dependency](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/7cd96a6752c9f8ff0e4126a7a6c4740572b349c7/package.json#L27) until the fork is distributed via NPM.

JJJ:
🤩 awesome to see Chosen 2.0.0 making its way in here!

Unless my EDD assets are broken, I think I still prefer the styling that is in Sugar Calendar already.

## Sunrise Color Scheme

Here is the outline styling, matching the color scheme in WordPress admin:
| This PR | SC |
|-|-|
| jjj-sc-2020-07-24 at 14 26 10@2x | jjj-sc-2020-07-24 at 14 25 50@2x |

## Dropdown Expanded

Here is what a dropdown looks like expanded:
| This PR |  Sugar Calendar |
|-|-|
| jjj-sc-2020-07-24 at 14 20 50@2x | jjj-sc-2020-07-24 at 14 21 11@2x |

Thoughts?

assets/css/edd-admin-chosen.css
```diff
@@ -203,10 +193,13 @@
 /* WordPress Color Schemes */
 
 body.admin-color-fresh .chosen-container-active .chosen-single,
-body.admin-color-fresh .chosen-container-active .chosen-choices {
+body.admin-color-fresh .chosen-container-active .chosen-choices,
+body.admin-color-fresh .chosen-with-drop.chosen-dropup .chosen-drop {
 	border-color: #007cba;
 	box-shadow: 0 0 0 1px #007cba;
 }
+body.admin-color-fresh ..chosen-with-drop.chosen-dropup .chosen-single,
```
JJJ, line 201: Double period is causing styling issues with Fresh color scheme:

jjj-sc-2020-07-27 at 12 21 24@2x

JJJ:
The only extremely minor quirk left I see, is drop-ups have a 1px jiggle.

| Closed | Open |
|-|-|
| jjj-sc-2020-07-28 at 11 53 19@2x | jjj-sc-2020-07-28 at 11 53 25@2x |

spencerfinnell:
@robincornett/@JJJ the core `select` and `button` styles use a `3px` border radius. `input`s use `4px`. I think we should match the core `select` radius of `3px`. On that note, I don't think we should add an inset shadow either. Thoughts?

robincornett:
Updated the border-radius and removed the box-shadow. I removed a negative margin-top from the drop--had added that because I was battling a ghost of the admin color box shadows, but not seeing it today, so maybe the inset was contributing?

spencerfinnell:
@robincornett Nice! Last (I think) little nitpick: can the dropdown and highlights be updated to `3px` radius as well? 

I see @JJJ's drop-up issue as well, but I'm not sure what is causing that.

JJJ:
> the core `select` and `button` styles use a `3px` border radius. `input`s use `4px`.

LOL. Sore subject!
https://wordpress.slack.com/archives/C02S78ZAL/p1571857052457600

Selects _start as_ `4px`...
https://github.com/WordPress/wordpress-develop/blob/master/src/wp-admin/css/forms.css#L52

...but then get set to `3px`:
https://github.com/WordPress/wordpress-develop/blob/master/src/wp-admin/css/forms.css#L310

Anyone who brings the issue up in Core Design or Trac gets shut down:
https://core.trac.wordpress.org/ticket/48537

It appears that the intent is to match whatever Gutenberg is doing, with is all over the place between `2px` and `4px` depending on the thing.

> I think we should match the core `select` radius of `3px`. Thoughts?

I prefer matching `4px`, because `select` elements require user input vs. buttons that invoke actions. If we (or WordPress) want to visually group like-elements together, a select is closer to an input than a button.

## Elements near each other
jjj-sc-2020-07-28 at 13 21 00@2x

jjj-sc-2020-07-28 at 13 21 15@2x

spencerfinnell:
Thanks for the background @JJJ 

> because select elements require user input vs. buttons that invoke actions

_Technically_ these are collapsable listboxes and a button is toggling it, not taking input: https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html -- using an actual `button` element for the toggle vs an `a` might help relieve some additional CSS that is being added currently. That being said it looks like the radius could be changing pretty frequently since it seems to be all over the place so it's not a huge deal... I'm fine with either. Whatever makes it easiest to maintain moving forward.

JJJ:
> I'm fine with either. Whatever makes it easiest to maintain moving forward.

Same! 🥳

robincornett:
Who knew that border-radius would be so contentious? I went with 4px on everything Chosen related, but could be talked out of it again.

LisaCee:
@robincornett, where should I be able to see the new modern theme in the Chosen selects?  I'm seeing the theme change, but I haven't seen any difference in the selects.  With all of the themes, I'm still just seeing the Chosen selects as "default" blue.

image

robincornett:
@LisaCee the highlight color should be applied to the `.chosen-drop`, `.chosen-single`, and `.chosen-choices` elements (it has to be applied to two different ones to get the effect around the whole thing and is difficult to target). Could you maybe have another plugin active which might have more specific Chosen styling, at least for the highlights? I may need to make this more specific, if so--right now the styles in here would theoretically affect any Chosen rendered fields, not just EDD ones. It should look like this:

image

JJJ:
Remember these are to support WordPress color schemes, and the latest one is named “Modern.” To see it, you need to edit your User Profile and picked that new color scheme, then find a Chosen select to verify that it’s purpley instead of blue.

robincornett:
@ashleyfae got it, sorry--we were already forcing that input to be full width on single select elements, but missed it on multi.

ashleyfae (approved):
Awesome - looks good now. Great work!

Issue #8070, opened 2020-07-26
Use on() instead of change()


akshuvo:
### Explain your enhancement (please be detailed)
Filename: `/assest/js/edd-ajax.js`
Line No: `346`

The code is saying this:

```
	// Load the fields for the selected payment method
	$('select#edd-gateway, input.edd-gateway').change( function (e) {

		var payment_mode = $('#edd-gateway option:selected, input.edd-gateway:checked').val();

		if( payment_mode == '0' ) {
			return false;
		}

		edd_load_gateway( payment_mode );

		return false;
	});
```

I am developing a theme and the problem is when I want to use a custom trigger when payment gateway changed using this code: `jQuery('input.edd-gateway').trigger('change');` it is not working.

But, if I change your `edd-ajax.js` file with the below code I can use the trigger.

```
	$(document).on('change', 'select#edd-gateway, input.edd-gateway', function (e) {

		var payment_mode = $('#edd-gateway option:selected, input.edd-gateway:checked').val();

		console.log('hh');

		if( payment_mode == '0' ) {
			return false;
		}

		edd_load_gateway( payment_mode );

		return false;
	});
```

### Justification or use case
I would really happy if you change the code so I can use the custom trigger. I am going to pull a request with my changes. Thanks

akshuvo:
Another problem was when you load the checkout page or `[download_checkout]` using ajax then this section`<div id="edd_purchase_form_wrap"></div>` doesn't load any data for not triggering the change of `select#edd-gateway, input.edd-gateway`

![image](https://user-images.githubusercontent.com/16398283/88487005-8ff46980-cfa3-11ea-956a-54a53e4e7b68.png)

Using `$(document).on('change', 'select#edd-gateway, input.edd-gateway', function (e) {` will solve the trigger issue when checkout page/shortcode loaded by ajax.

![image](https://user-images.githubusercontent.com/16398283/88487099-1315bf80-cfa4-11ea-9368-07a75e8c5aae.png)

cklosowski:
I'm going to initially set this for 2.9.24.

spencerfinnell:
Here's a very rough/quick snippet to test loading the `[downloads_checkout]` shortcode via AJAX. Replace `[downloads_checkout]` with `[checkout_ajax]` on the checkout page and add the following code:

```php
add_shortcode( 'checkout_ajax', function() {
	wp_enqueue_script( 'jquery' );

	ob_start();
?>

[...]

PR #8071, opened 2020-07-26
fixed: Use on() instead of change() #8070

akshuvo:
Fixes #

I've fixed the code as described on #8070. Thanks


cklosowski:
@spencerfinnell Does this have any over-reaching ramifications that I'm not aware of? Could this be a simple change to slate into the 2.9.24 release?

spencerfinnell:
@cklosowski Not that I can think of, and it aligns it with the other event handlers such as https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/edd-ajax.js#L312

PR #8073, opened 2020-07-28
Make Refund Details consistent with Order Details (#8069)

### LINKED ISSUE #8069: 3.0 - Discrepancy in refund screens
### Expected behavior
When I issue a refund, I expect the refund amounts to be consistent when I'm viewing the refund in the admin.

### Actual behavior
When a discount was used on an order, and I issue a refund, I am able to view a version of the refund which does not show the discount (or the correct total). It looks like there are two different links/screens which can be used to view a refund:
- `https://3dd.local/wp-admin/edit.php?post_type=download&page=edd-payment-history&view=view-refund-details&id=269` (the discount does not show)
- `https://3dd.local/wp-admin/edit.php?post_type=download&page=edd-payment-history&view=view-order-details&id=269` (the discount does show)

Looking in the database, the row for the order information appears to be correct, so this could maybe just be a display issue, but I do wonder why there are two different links for viewing a refund.

1) Create/run an order which uses a discount code.
2) Refund the order.
3) On the original order details screen, click the link to view the refund. The discount does not show as part of this refund.
4) Go back to the original order and click "View Customer Record". The refund amount on the Customer Details screen is correct.
5) Click the link to view the refund details. This will take you to the order screen which does show the discount as part of the refund.
6) Visit the orders list table and view the refunds. The title/Edit links for the refund are for the `view-refund-details` screen (the incorrect one).

PHP Version: 7.4.1
WordPress Version: 5.4.2

I'm not sure if the `view-refund-details` screen needs to be updated, or removed (in favor of `view-order-details`), but I think the links to view the refunds should be consistent, so if the screen is updated to show the correct information, the link on the Customer Details screen should probably be changed to use `view-refund-details`, to be consistent with the order details and list table.

robincornett:
Fixes #8069

Proposed Changes:
1. Update Refund Details screen to retrieve and include order adjustments.
2. Update the refund link in the "Recent Refunds" table on the Customer Details screen.


robincornett:
@spencer, what about fees? I have one that is stored as a negative amount on this refund ID but it's showing positive on the refund screen:

image

(This could be part of what's going on in #8057, not sure. But the database amount definitely shows negative and it's displaying positive here.)

spencerfinnell:
@robincornett To me that display makes sense. It steps you through how $71.01 was totaled to become the refund amount. The item amount, fee, and tax are all positive values because those are what created the original $71.01 charge.

If the fee is negative then the price, amount, and tax would need to be displayed negative as well.

Some previous discussion here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7528#issuecomment-581515866

robincornett:
Okay, sorry, that makes sense! Then it is displaying correctly for me. :)

ashleyfae:
It's still possible to manually adjust a refund URL to end up on the order version.

1. View a refund record. The URL will be `view-refund-details`. Example: `http://develop/edd3/wp-admin/edit.php?post_type=download&page=edd-payment-history&view=view-refund-details&id=124`
2. Manually change `view-refund-details` to `view-order-details`. Example: `http://develop/edd3/wp-admin/edit.php?post_type=download&page=edd-payment-history&view=view-order-details&id=124`

tl;dr: I'm thinking if we're on `view-order-details` but the "order" is a refund, maybe we should force a redirect to the `view-refund-details` version.

robincornett:
One way I can do this is to update the [`switch` statement](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8069/includes/admin/payments/payments-history.php#L118) in `edd_payment_history_page` to something like this:

```php
	$view     = edd_get_payment_view();
	$order_id = absint( $_GET['id'] );
	$order    = edd_get_order( $order_id );
	if ( 'refund' === $order->type ) {
		$view = 'view-refund-details';
	}
```

and then use the `$view` variable as the switch. The URL still has `order` in it instead of `refund`, but it's using the correct view file.

But that's not performing an actual redirect.

spencerfinnell:
@robincornett Another option would be to check the type here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8069/includes/admin/payments/view-order-details.php#L30

spencerfinnell (changes requested):
@robincornett Since we are preventing the view when it's a Refund we can remove https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8069/includes/admin/payments/view-order-details.php#L52-L64

robincornett:
I like that option better, @spencerfinnell...so I could add this condition to the view-order-details:

```php
if ( 'refund' === $order->type ) {
	require_once EDD_PLUGIN_DIR . 'includes/admin/payments/view-refund.php';
	edd_view_refund_page_content();
	return;
}
```

That's working nicely here, as long as the URL is not an issue.

includes/admin/payments/view-order-details.php
```diff
@@ -31,6 +31,12 @@
 	wp_die( __( 'The specified ID does not belong to an order. Please try again', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ) );
 }
 
+if ( 'refund' === $order->type ) {
+	require_once EDD_PLUGIN_DIR . 'includes/admin/payments/view-refund.php';
+	edd_view_refund_page_content();
+	return;
+}
```
spencerfinnell, line 38: @robincornett I wonder if we should keep loading of the views to just the `edd_payment_history_page()` function? 

When I suggested checking the type I meant checking in the `wp_die()` condition above. I think since you can only get to the wrong page by manually changing the URL it's safe to put developer experience over user experience and present users with an "ugly" error message and keep our view loading logic contained.

cc @ashleyfae

spencerfinnell: We are doing a similar thing for Refunds here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/payments/view-refund.php#L29-L33

ashleyfae: I'd rather either do a redirect or a `wp_die()`.

includes/admin/payments/view-order-details.php
```diff
@@ -32,9 +32,14 @@
...
 
 if ( 'refund' === $order->type ) {
-	require_once EDD_PLUGIN_DIR . 'includes/admin/payments/view-refund.php';
-	edd_view_refund_page_content();
-	return;
+	$refund_link = edd_get_admin_url(
+		array(
+			'page' => 'edd-payment-history',
+			'view' => 'view-refund-details',
+			'id'   => $order->id,
```
ashleyfae, line 39:
```suggestion
			'id'   => urlencode( $order->id ),
```

ashleyfae (changes requested):

ashleyfae:
I'm seeing two discount rows:

Screenshot from 2020-11-10 11-04-15

robincornett:
I think the redundancy was introduced when we [added the `$_adjustments` to the localized order/refund data](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8069/includes/admin/payments/refunds.php#L128). That has more information than the row which shows in the order summary section, so I just removed that.

ashleyfae:
Not sure if we want to create a new issue for this, but I'm still seeing some inconsistent stuff with fees.

I manually created an order and applied a $50 fee to the whole order.

Screenshot from 2020-11-11 13-40-55

The refund modal does not show this fee at all.

Screenshot from 2020-11-11 13-41-07

Nor is it present on the refund UI:

Screenshot from 2020-11-11 13-41-35

robincornett:
I _think_ that might be technically a separate issue--with the modal behavior rather than the refund screen. $.02

The refund screen is _showing_ the fee, but it's just not being refunded.

I would wonder if fees should be automatically selected for refund (seems like yes if it's item level) or manually selected (order level)?

ashleyfae:
@robincornett Made a new issue for that: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8203

ashleyfae (approved):

spencerfinnell (approved):

PR #8085, opened 2020-08-17
Return 401 response on failed login and use wp_signon #6647

### LINKED ISSUE #6647: Add Limit Login Attempt support to checkout user login form
## Enhancement Request
A customer ticket came in recently where an EDD website was being brute forced using the checkout login form, causing high AJAX requests/resource usage on the server (which was flagged by Kinsta).

The login form in Restrict Content Pro already includes support for the "Limit Login Attempts" plugin to prevent against these types of attacks, I believe EDD should also.

All forms, be it checkout or short-code (login and registrations) should have protection against this using the same method as RCP already has.

### Justification or use case

To prevent against brute force attacks at checkout.

cklosowski:
Fixes #6647

Proposed Changes:
1. Adds a 401 status header when a login is failed.
2. Remove validation done by EDD so that core actions can run
3. Modify `edd_log_user_in` to use `wp_signon` instead of just setting cookies and doing actions, so that core actions can run.

### DIFF
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -9,8 +9,28 @@
  * @since       1.0
  */
 
-// Exit if accessed directly
-if ( ! defined( 'ABSPATH' ) ) exit;
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * While loading the template, see if an error was set for a filed login attempt and set the proper
+ * HTTP status code if there was a failed login attempt.
+ *
+ * @since 2.9.24
+ *
+ * @return void
+ */
+function edd_login_error_check() {
+	$errors = edd_get_errors();
+	if ( ! empty( $errors ) ) {
+		if ( array_key_exists( 'edd_invalid_login', $errors ) ) {
+			status_header( 401 );
+		}
+	}
+}
+add_action( 'template_redirect', 'edd_login_error_check', 10 );
 
 /**
  * Login Form
@@ -30,13 +50,6 @@ function edd_login_form( $redirect = '' ) {
 	$edd_login_redirect = $redirect;
 
 	ob_start();
-
-	$errors = edd_get_errors();
-	if ( ! empty( $errors ) ) {
-		if ( array_key_exists( 'edd_invalid_login', $errors ) ) {
-				status_header( 401 );
-		}
-	}
 
 	edd_get_template_part( 'shortcode', 'login' );
 
@@ -84,10 +97,6 @@ function edd_process_login_form( $data ) {
 
 			$user = edd_log_user_in( 0, $data['edd_user_login'], $data['edd_user_pass'], $rememberme );
 
-			if ( ! $user instanceof WP_User ) {
-					edd_set_error( 'edd_invalid_login', __( 'Invalid username or password.', 'easy-digital-downloads' ) );
-			}
-
 			// Check for errors and redirect if none present
 			$errors = edd_get_errors();
 			if ( ! $errors ) {
@@ -121,6 +130,18 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 	);
 
 	$user = wp_signon( $credentials );
+
+	if ( ! $user instanceof WP_User ) {
+		edd_set_error(
+			'edd_invalid_login',
+			sprintf(
+				__( 'Invalid username or password. %sReset Password%s', 'easy-digital-downloads' ),
+				'<a href="' . wp_lostpassword_url( edd_get_checkout_uri() ) . '">',
+				'</a>'
+			)
+		);
+	}
+
 	do_action( 'edd_log_user_in', $user_id, $user_login, $user_pass );
 
 	return $user;
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -135,7 +135,8 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 		edd_set_error(
 			'edd_invalid_login',
 			sprintf(
-				__( 'Invalid username or password. %sReset Password%s', 'easy-digital-downloads' ),
+				/* translators: %1$s Opening anchor tag, do not translate. %2$s Closing anchor tag, do not translate. */
+				__( 'Invalid username or password. %1$sReset Password%2$s', 'easy-digital-downloads' ),
 				'<a href="' . wp_lostpassword_url( edd_get_checkout_uri() ) . '">',
 				'</a>'
 			)
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -92,11 +92,19 @@ function edd_register_form( $redirect = '' ) {
 function edd_process_login_form( $data ) {
 
 	if ( ! empty( $data['edd_login_nonce'] ) && wp_verify_nonce( $data['edd_login_nonce'], 'edd-login-nonce' ) ) {
+		$login      = isset( $data['edd_user_login'] ) ? $data['edd_user_login'] : '';
+		$pass       = isset( $data['edd_user_pass'] ) ? $data['edd_user_pass'] : '';
 		$rememberme = isset( $data['rememberme'] );
 
-		$user = edd_log_user_in( 0, $data['edd_user_login'], $data['edd_user_pass'], $rememberme );
+		$user = edd_log_user_in( 0, $login, $pass, $rememberme );
 
-		// Check for errors and redirect if none present
+		// Wipe these variables so they aren't anywhere in the submitted format any longer.
+		$login = null;
+		$pass  = null;
+		$data['edd_user_login'] = null;
+		$data['edd_user_pass']  = null;
+
+		// Check for errors and redirect if none present.
 		$errors = edd_get_errors();
 		if ( ! $errors ) {
 			$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -91,19 +91,18 @@ function edd_register_form( $redirect = '' ) {
 */
 function edd_process_login_form( $data ) {
 
-	if ( wp_verify_nonce( $data['edd_login_nonce'], 'edd-login-nonce' ) ) {
+	if ( ! empty( $data['edd_login_nonce'] ) && wp_verify_nonce( $data['edd_login_nonce'], 'edd-login-nonce' ) ) {
+		$rememberme = isset( $data['rememberme'] ) ? (bool) $data['rememberme'] : false;
 
-			$rememberme = isset( $data['rememberme'] ) ? (bool) $data['rememberme'] : false;
+		$user = edd_log_user_in( 0, $data['edd_user_login'], $data['edd_user_pass'], $rememberme );
 
-			$user = edd_log_user_in( 0, $data['edd_user_login'], $data['edd_user_pass'], $rememberme );
-
-			// Check for errors and redirect if none present
-			$errors = edd_get_errors();
-			if ( ! $errors ) {
-					$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
-					wp_redirect( $redirect );
-					edd_die();
-			}
+		// Check for errors and redirect if none present
+		$errors = edd_get_errors();
+		if ( ! $errors ) {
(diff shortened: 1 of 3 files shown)


cklosowski:
It's working on the RCP and AffWP sites as we speak.

includes/login-register.php
```diff
@@ -31,6 +31,12 @@ function edd_login_form( $redirect = '' ) {
 
 	ob_start();
 
+	if ( ! empty( edd_get_errors() ) ) {
+		if ( array_key_exists( 'edd_invalid_login', edd_get_errors() ) ) {
+				status_header( 401 );
```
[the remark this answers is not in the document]
spencerfinnell: @cklosowski I see it now using Postman but Chrome doesn't show it as a `401` unless it's done on an earlier action.

cklosowski:
Chrome and FF both show it for me. Will confirm with you shortly.

includes/login-register.php
```diff
@@ -64,63 +70,58 @@ function edd_register_form( $redirect = '' ) {
...
+			}
 
-				edd_log_user_in( $user_data->ID, $data['edd_user_login'], $data['edd_user_pass'], $data['rememberme'] );
-			} else {
-				edd_set_error( 'password_incorrect', __( 'The password you entered is incorrect', 'easy-digital-downloads' ) );
+			// Check for errors and redirect if none present
+			$errors = edd_get_errors();
+			if ( ! $errors ) {
+					$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], $user_ID );
+					wp_redirect( $redirect );
```
robincornett, line 92: Should this be `wp_safe_redirect` instead?

includes/login-register.php
```diff
@@ -64,63 +70,58 @@ function edd_register_form( $redirect = '' ) {
...
+					edd_set_error( 'edd_invalid_login', __( 'Invalid username or password.', 'easy-digital-downloads' ) );
+			}
 
-				edd_log_user_in( $user_data->ID, $data['edd_user_login'], $data['edd_user_pass'], $data['rememberme'] );
-			} else {
-				edd_set_error( 'password_incorrect', __( 'The password you entered is incorrect', 'easy-digital-downloads' ) );
+			// Check for errors and redirect if none present
+			$errors = edd_get_errors();
+			if ( ! $errors ) {
+					$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], $user_ID );
```
robincornett, line 91: `$user_ID` is not defined.

includes/login-register.php
```diff
@@ -31,6 +31,12 @@ function edd_login_form( $redirect = '' ) {
 
 	ob_start();
 
+	if ( ! empty( edd_get_errors() ) ) {
```
robincornett, line 34: PHPunit error (Travis) for this line: `PHP Fatal error:  Can't use function return value in write context in /home/travis/build/easydigitaldownloads/easy-digital-downloads/includes/login-register.php on line 34`

cklosowski:
@spencerfinnell Here is my profile editor showing a 401:
image

includes/login-register.php
cklosowski: I think it _should_ be but we'll have to do that in a major release. I'll open an issue in 2.10 for it. We can't do that in a minor release as it might break customizations that are using EDD as an SSO or multisite situation.

cklosowski: Noted @robincornett https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8087

includes/login-register.php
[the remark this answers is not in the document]
robincornett: I'm not seeing a `401` in Chrome or Firefox using the `[edd_login]` shortcode form.

cklosowski:
@robincornett @spencerfinnell This was tested against the 'Profile Editor' from a logged out state. Maybe try that? Most of the bruteforce attempts actually come when people visit the 'account' pages of sites, not the checkout form.

includes/login-register.php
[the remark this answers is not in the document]
spencerfinnell: @cklosowski Testing on `[edd_login]` and logged out in `[edd_profile_editor]` (with Limit Login Attempts Reloaded both active and inactive) and I'm not able to get it to get past the `headers_sent()` check unless it's moved to an earlier hook. Is there anything in your environment/Pagely environment that would affect that?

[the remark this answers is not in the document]
cklosowski: Not that I can think of. I'm running a standard nginx install locally.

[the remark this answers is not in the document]
cklosowski: @spencerfinnell I'll see if I can find another area to hook this.

cklosowski:
@robincornett @spencerfinnell Give that a shot...

I've standardized the error to match that of the checkout login form, to have a forgot password. The checkout form is now supported, the only negative is that we cannot produce a 401 on the checkout form due to it breaking the AJAX response, however the limit login plugin still is supported as it's hooked into the failed login action.

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
-		$user_data = get_user_by( 'login', $data['edd_user_login'] );
-		if ( ! $user_data ) {
-			$user_data = get_user_by( 'email', $data['edd_user_login'] );
-		}
-		if ( $user_data ) {
-			$user_ID = $user_data->ID;
-			$user_email = $user_data->user_email;
 
-			if ( wp_check_password( $data['edd_user_pass'], $user_data->user_pass, $user_data->ID ) ) {
+			$rememberme = isset( $data['rememberme'] ) ? (bool) $data['rememberme'] : false;
```
spencerfinnell, line 96:
```suggestion
			$rememberme = isset( $data['rememberme'] );
```

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
+			'remember'      => $remember,
+	);
+
+	$user = wp_signon( $credentials );
+
+	if ( ! $user instanceof WP_User ) {
+		edd_set_error(
+			'edd_invalid_login',
+			sprintf(
+				__( 'Invalid username or password. %sReset Password%s', 'easy-digital-downloads' ),
```
spencerfinnell, line 138:
```suggestion
				/* translators: %1$s Opening anchor tag, do not translate. %2$s Closing anchor tag, do not translate. */
				__( 'Invalid username or password. %1$sReset Password%2$s', 'easy-digital-downloads' ),
```

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
-	if ( $user_id < 1 )
-		return;
 
-	wp_set_auth_cookie( $user_id, $remember );
-	wp_set_current_user( $user_id, $user_login );
-	do_action( 'wp_login', $user_login, get_userdata( $user_id ) );
+	$credentials = array(
+			'user_login'    => $user_login,
+			'user_password' => $user_pass,
+			'remember'      => $remember,
```
spencerfinnell, line 129: Looks like there is some extra indention here.

includes/login-register.php
cklosowski: This will provide `true` if `$data['rememberme']` is set to `false` since it will be set.

spencerfinnell: @cklosowski `rememberme` won't be set at all if the checkbox isn't checked. This is the same as the expanded logic that existed previously as well:

```php
if ( isset( $data['rememberme'] ) ) {
	$data['rememberme'] = true;
} else {
	$data['rememberme'] = false;
}
```

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
+
+	$user = wp_signon( $credentials );
+
+	if ( ! $user instanceof WP_User ) {
+		edd_set_error(
+			'edd_invalid_login',
+			sprintf(
+				/* translators: %1$s Opening anchor tag, do not translate. %2$s Closing anchor tag, do not translate. */
+				__( 'Invalid username or password. %1$sReset Password%2$s', 'easy-digital-downloads' ),
+				'<a href="' . wp_lostpassword_url( edd_get_checkout_uri() ) . '">',
```
spencerfinnell, line 139:
```suggestion
				'<a href="' . esc_url( wp_lostpassword_url( edd_get_checkout_uri() ) ) . '">',
```

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
+
+		$user = edd_log_user_in( 0, $data['edd_user_login'], $data['edd_user_pass'], $rememberme );
+
 		// Check for errors and redirect if none present
 		$errors = edd_get_errors();
 		if ( ! $errors ) {
-			$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], $user_ID );
-			wp_redirect( $redirect );
-			edd_die();
+				$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
```
robincornett, line 102: There is an extra indent on each of these lines.

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
+				wp_redirect( $redirect );
+				edd_die();
```
robincornett, line 104: phpcs says, "It is also important to remember to call exit() after a redirect so that no other unwanted code is executed." ... is calling `wp/edd_die` equivalent?

includes/login-register.php
```diff
@@ -64,63 +84,68 @@ function edd_register_form( $redirect = '' ) {
...
+		$user = edd_log_user_in( 0, $data['edd_user_login'], $data['edd_user_pass'], $rememberme );
+
 		// Check for errors and redirect if none present
 		$errors = edd_get_errors();
 		if ( ! $errors ) {
-			$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], $user_ID );
-			wp_redirect( $redirect );
-			edd_die();
+				$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
+				wp_redirect( $redirect );
```
robincornett, line 103: Should `esc_url` be run on the `$redirect` before using it in the function?

includes/login-register.php
```diff
@@ -9,8 +9,28 @@
...
+ * While loading the template, see if an error was set for a filed login attempt and set the proper
+ * HTTP status code if there was a failed login attempt.
+ *
+ * @since 2.9.24
+ *
+ * @return void
+ */
+function edd_login_error_check() {
+	$errors = edd_get_errors();
+	if ( ! empty( $errors ) ) {
```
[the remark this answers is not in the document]
cklosowski: @robincornett I choose this for future thinking. It's less likely we'll miss removing this in the future, if we end up with further error message handling in the future. It was a deliberate choice on my part here.

includes/login-register.php
cklosowski: @robincornett Yes
```
/**
 * Wrapper function for wp_die(). This function adds filters for wp_die() which
 * kills execution of the script using wp_die(). This allows us to then to work
 * with functions using edd_die() in the unit tests.
 *
 * @author Sunny Ratilal
 * @since 1.6
 * @return void
 */
function edd_die( $message = '', $title = '', $status = 400 ) {
	add_filter( 'wp_die_ajax_handler', '_edd_die_handler', 10, 3 );
	add_filter( 'wp_die_handler'     , '_edd_die_handler', 10, 3 );
	add_filter( 'wp_die_json_handler', '_edd_die_handler', 10, 3 );

	wp_die( $message, $title, array( 'response' => $status ));
}

```

includes/login-register.php
cklosowski: @robincornett `wp_redirect` calls `wp_sanitize_redirect` which helps prevent malicious intent.
https://core.trac.wordpress.org/browser/tags/5.5/src/wp-includes/pluggable.php#L1277

includes/login-register.php
cklosowski: Fixed

includes/login-register.php
robincornett: Thanks--I was not sure if `die` and `exit` were equivalent, but I see in the PHP manual that they are.

cklosowski:
Yeah Ill fix those up.

PR #8089, opened 2020-08-18
Issue 7611 Reverse args in implode

### LINKED ISSUE #7611: Deprecated: implode() error when viewing download in PHP 7.4
### Expected behavior
When adding and then viewing a download with variable pricing, I do not expect to get any PHP errors.

### Actual behavior
After adding a download with variable pricing, I view the download and get the PHP error: 
```
Deprecated: implode(): Passing glue string after array is deprecated. 
Swap the parameters in ../plugins/easy-digital-downloads/includes/template-functions.php on line 282
```

1) Use PHP 7.4
2) Add new download with variable pricing enabled
3) Add the ```[purchase_link]``` shortcode to a page 
4) Visit the page and check debug.log

PHP Version: 7.4

EDD Version (or branch): 

WordPress Version:

Error doesn't stop the users ability to complete the purchase, nothing breaks.

Solution: pass the glue string as a first argument in the implode function rather than the last.

LisaCee:
Fixes #7611 

Proposed Changes:
1. Reverse args in implode to be (separator, array)
2. Tidy formatting on array passed into implode


LisaCee:
@cklosowski, I think this can be closed since it looks like it's already been handled.

ashleyfae:
Fixed in above commit via https://github.com/easydigitaldownloads/easy-digital-downloads/pull/7970

PR #8091, opened 2020-08-18
Issue/8065 Data range styles 

### LINKED ISSUE #8065: 3.0 - Update Date Range Styles to Match Core

Update input date range styles to match WordPress 5.2+ styles.

### Justification or use case

Current CSS styles cause the inputs to look broken:

Screen Shot 2020-07-22 at 12 13 36 PM

LisaCee:
Fixes #8065 

Proposed Changes:
1. Updated border-radius, border color, padding to more closely match core 
2. Removed display: block from `#edd-filters span` so that fields stay together when resized
3. Added media query to add margin under edd-filters on small screens

These changes work well in Chrome, but the heights of the the inputs and selects vary in Firefox.  Firefox loads correctly for a second and then the selects change height.  Maybe a Chosen side effect?

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -3486,7 +3486,6 @@ body.download_page_edd-reports {
 	}
 
 	#edd-filters span {
-		display: block;
 		margin: 2px 0;
 	}
 
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -232,6 +232,11 @@ a.edd-delete:hover {
 	z-index: 9999 !important;
 }
 
+@media screen and (max-width: 782px) {
+	#edd-filters .filter-items > span:not(.edd-date-range-options) {
+		margin-bottom: 6px 6px 6px 0;
+	}
+}
 /* =Payment Icon Styling
 -------------------------------------------------------------- */
 
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -60,7 +60,6 @@ a.edd-delete:hover {
 .edd-from-to-wrapper input[name="start"],
 .edd-from-to-wrapper input[name="start-date"],
 .edd-from-to-wrapper input[name="filter_from"] {
-	border-right-color: #eee;
 	border-radius: 4px 0 0 4px;
 }
 
@@ -236,7 +235,7 @@ a.edd-delete:hover {
 }
 
 @media screen and (max-width: 782px) {
-	#edd-filters .filter-items > span:not(.edd-date-range-options) {
+	#edd-filters .filter-items {
 		margin: 6px 6px 6px 0;
 	}
 }
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -80,8 +80,8 @@ a.edd-delete:hover {
 .edd-from-to-wrapper input[name="start"]:focus,
 .edd-from-to-wrapper input[name="start-date"]:focus,
 .edd-from-to-wrapper input[name="filter_from"]:focus {
-	border-top-left-radius: 0;
-	border-bottom-left-radius: 0;
+	border-top-right-radius: 0;
+	border-bottom-right-radius: 0;
 }
 
 .edd-from-to-wrapper input[name="end"]:focus,
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -62,19 +62,21 @@ a.edd-delete:hover {
 	outline: auto 5px -webkit-focus-ring-color;
 	border-color: #eee;
 	box-shadow: none;
+	position: relative;
 }
 
 .edd-from-to-wrapper input[name="start"],
 .edd-from-to-wrapper input[name="start-date"],
 .edd-from-to-wrapper input[name="filter_from"] {
 	border-right-color: #eee;
+	border-radius: 4px 0 0 4px;
 }
 
 .edd-from-to-wrapper input[name="end"],
 .edd-from-to-wrapper input[name="end-date"],
 .edd-from-to-wrapper input[name="filter_to"] {
-	border-left-color: #eee;
 	margin-left: -1px;
+	border-radius: 0 4px 4px 0;
 }
 
 .edd-settings-sub-nav {
@@ -1250,7 +1252,6 @@ textarea[name="edd-note"] {
 #edd-filters input[type="text"].edd_datepicker,
 #edd-filters input[type="number"] {
 	max-width: 100px;
-	padding: 5px;
 }
 
 #edd-filters input[type="number"],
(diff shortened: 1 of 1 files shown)


assets/css/edd-admin.css
```diff
@@ -62,19 +62,21 @@ a.edd-delete:hover {
...
 	border-right-color: #eee;
+	border-radius: 4px 0 0 4px;
 }
 
 .edd-from-to-wrapper input[name="end"],
 .edd-from-to-wrapper input[name="end-date"],
 .edd-from-to-wrapper input[name="filter_to"] {
-	border-left-color: #eee;
 	margin-left: -1px;
+	border-radius: 0 4px 4px 0;
```
robincornett, line 79: `:focus` styles need to be added for these as well--when the inputs have focus, the border-radius disappears. Actually, I think it's not that the border-radius is changed or removed, but the outline added to `.edd-from-to-wrapper input:focus` (line 60, I think), overrides it.

assets/css/edd-admin.css
```diff
@@ -62,19 +62,21 @@ a.edd-delete:hover {
 	outline: auto 5px -webkit-focus-ring-color;
 	border-color: #eee;
 	box-shadow: none;
+	position: relative;
 }
 
 .edd-from-to-wrapper input[name="start"],
 .edd-from-to-wrapper input[name="start-date"],
 .edd-from-to-wrapper input[name="filter_from"] {
 	border-right-color: #eee;
```
robincornett, line 71: I believe the `#eee` can be removed here, maybe?

assets/css/edd-admin.css
```diff
@@ -230,6 +232,11 @@ a.edd-delete:hover {
 	z-index: 9999 !important;
 }
 
+@media screen and (max-width: 782px) {
```
robincornett, line 235: I wonder if the custom margin added to `#edd-filters .filter-items .edd-date-range-options` (line 2461) could be removed so the spans here could all be styled together, so you are not having to override them? Also see that there are some `inline-block` styles applied to these elements which may not be necessary since the parent container is using flexbox.

assets/css/edd-admin.css
LisaCee: I think I've got the focus fixed up, but what color should it be?  The `select`s focus in black while the search field is in the admin color scheme.  Again, I'm guessing it is a Chosen side effect.  What did you opt for in #7956?

assets/css/edd-admin.css
LisaCee: I think I made the specificity in this media query too specific, since the `.edd-date-range-options` isn't actually used here.  

It looks like it is specific to reports and as far as I can tell, `#edd-filters .filter-items .edd-date-range-options` isn't used anywhere.  I have removed the whole rule, but please double check me as I don't have all of the plugins/extensions installed.

assets/css/edd-admin.css
robincornett: For me, it's picking up the admin color scheme focus now and I think everything looks good:

image

assets/css/edd-admin.css
robincornett: I think what you've added resolves the primary issue. Although I do think there is some unnecessary, maybe conflicting and/or redundant CSS affecting the filter items, which could be fixed, that could be tabled for a future sassification. This is a big improvement overall!

assets/css/edd-admin.css
```diff
@@ -57,24 +57,28 @@ a.edd-delete:hover {
...
-	outline: auto 5px -webkit-focus-ring-color;
-	border-color: #eee;
-	box-shadow: none;
-}
-
 .edd-from-to-wrapper input[name="start"],
 .edd-from-to-wrapper input[name="start-date"],
 .edd-from-to-wrapper input[name="filter_from"] {
-	border-right-color: #eee;
+	border-radius: 4px 0 0 4px;
```
spencerfinnell, line 63: @LisaCee For the border radius changes let's only _remove_ the round corners where needed and leave the other radi as-is. That way if core changes to a `3px` or `5px` radius we don't need to update.

assets/css/edd-admin.css
spencerfinnell: There are definitely still issues with some of the filter layouts (and additional overly complicated CSS) but that can be followed up separately as I'm sure we'll want to do another sweep after the Chosen changes.

assets/css/edd-admin.css
```diff
@@ -57,24 +57,39 @@ a.edd-delete:hover {
...
+	border-top-left-radius: 0;
+	border-bottom-left-radius: 0;
+}
+
+.edd-from-to-wrapper input:focus {
+	z-index: 2;
+	position: relative;
+}
+
+.edd-from-to-wrapper input[name="start"]:focus,
```
robincornett, line 80: I'm not sure that focus are needed beyond the initial block at line 75-77--removing them seems to keep everything properly rounded and "focused" for me.

LisaCee:
> We need to increase the width slightly on the date input fields, at least on mobile--they're getting cut off when a date is selected. I didn't actually add a fix for that (TBH easiest would be to make those inputs 105px wide instead of 100px wide), because I got derailed trying to make it possible (widths are set multiple places). I ended up removing all of this CSS just to get to the point where I could adjust the date/text inputs without things going sideways:
> 
> ...
> 
> Personally, I would bump the `.edd-from-to-wrapper input` to 105 at all screen sizes, but if you want a more "fitted" look then we need to add a media query.

I made **some** of your changes and I'd love to hear your opinion.  I've only commented out the CSS so far so that you can play with it more easily and because I'm sure I've broken something else.  

I didn't use **all** of your suggestions because they caused the fields to stack on all screen sizes, but they could easily fit side-by-side on larger screens.  (Styling for stacked on small is on my todo list)

assets/css/edd-admin.css
```diff
@@ -1251,10 +1251,10 @@ textarea[name="edd-note"] {
 	margin-top: -30px;
 	margin-left: 2px;
 }
-
+/* Removed by Robin -- first selector only*/
 #edd-filters input[type="text"].edd_datepicker,
 #edd-filters input[type="number"] {
-	max-width: 100px;
+	max-width: 105px;
```
robincornett, line 1257: This won't take effect until the `width: 100px;` rule is removed (lines 2465-2467), as that's overriding this.

LisaCee: My screen is showing the fields as `105px`.  Where are you seeing it being overridden in the UI?

LisaCee: It looks like `max-width` overrides `width` https://developer.mozilla.org/en-US/docs/Web/CSS/max-width

robincornett: Sorry, the filter bar inputs are showing at 100px--the ones in the export form do not appear to be affected.

robincornett: Also, totally not sure why the filter bar is showing on the Exports screen--does that seem weird? The filters don't affect anything there as far as I can tell.

PR #8093, opened 2020-08-19
check that inputs have lengths before checking them on remove download #8092

### LINKED ISSUE #8092: Removing downloads from payments does not work in WordPress 5.5+
### Expected behavior
When viewing order details in admin, and clicking Remove download, no errors should present themselves, and the download should be removed.

### Actual behavior
When clicking remove download, a JS error is thrown:
`SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data`

The page is refreshed, and the download is not removed.

This is caused by a change in jQuery that no longer supports the conditional checks of inputs without the `length` attribute being checked.

1) View an order in the admin that has no fees and/or taxes. And has multiple downloads.
2) Click 'Remove' on one of the products.
3) Watch the page refresh without removing, and if you have an extension to view JS errors, the error above (or view it in console, with Persist Logs enabled)
etc, etc

PHP Version: All

EDD Version (or branch): `2.9.23`

WordPress Version: 5.5+

cklosowski:
Fixes #8092

Proposed Changes:
1. Adds `.length` attributes to the checks on the presense of the tax and fees inputs when removing a download from a payment in the admin.

### DIFF
--- a/assets/js/admin-scripts.js
+++ b/assets/js/admin-scripts.js
@@ -453,11 +453,11 @@ jQuery(document).ready(function ($) {
 					var quantity    = $('input[name="edd-payment-details-downloads['+key+'][quantity]"]').val();
 					var amount      = $('input[name="edd-payment-details-downloads['+key+'][amount]"]').val();
 
-					if ( $('input[name="edd-payment-details-downloads['+key+'][tax]"]') ) {
+					if ( $('input[name="edd-payment-details-downloads['+key+'][tax]"]').length ) {
 						var fees = $('input[name="edd-payment-details-downloads['+key+'][tax]"]').val();
 					}
 
-					if ( $('input[name="edd-payment-details-downloads['+key+'][fees]"]') ) {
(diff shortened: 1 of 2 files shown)


spencerfinnell:
> Works well for me. I could easily replicate the error in `2.9.23` and saw no such error here. I wasn't sure how to test for the `cart_wrapper` but it sounds like that was just added in an effort to avoid future errors there.

@LisaCee that only occurs when using the "Downloads Cart" widget:

Screen Shot 2020-08-19 at 8 14 40 PM

It depends on the active theme (this is Twenty Eleven) where it will appear:

Screen Shot 2020-08-19 at 8 15 47 PM

There shouldn't be any errors when removing the items via that widget and removing the last item should show an "Your cart is empty" item.

LisaCee:
> There shouldn't be any errors when removing the items via that widget and removing the last item should show an "Your cart is empty" item.

Thanks @spencerfinnell.  I tested this with Vendd and Themedd and it behaved as expected with no errors.

PR #8099, opened 2020-09-02
Fixing checkout registration form to set the current user #8098

### LINKED ISSUE #8098: Registering on checkout causes "Error: You must be logged in to purchase"
### Expected behavior
Creating an account on checkout should not cause `Error: You must be logged in to purchase`.

### Actual behavior
Properly filling in create account forms causes `Error: You must be logged in to purchase a` on checkout. Page refreshes and I'm logged in, so account was created.

1) Disable guest checkout
2) Register on checkout
3) Complete a purchase via PayPal or Stripe. Test Gateway does not appear to be affected.

PHP Version: 5.6, 7.2, 7.4

EDD Version (or branch): 2.9.24

WordPress Version: 5.5 and 5.5.1

This doesn't happen with EDD 2.9.23 so a change in .24 seems to be triggering the error.

cklosowski:
Fixes #8098

Proposed Changes:
1. Boat-load of PHPCS fixes
2. In the `edd_register_and_login_new_user` function, we're now setting the WordPress user if logged in successfully.

### DIFF
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -166,64 +166,66 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 */
 function edd_process_register_form( $data ) {
 
-	if( is_user_logged_in() ) {
+	if ( is_user_logged_in() ) {
 		return;
 	}
 
-	if( empty( $_POST['edd_register_submit'] ) ) {
+	if ( empty( $_POST['edd_register_submit'] ) ) {
 		return;
 	}
 
 	do_action( 'edd_pre_process_register_form' );
 
-	if( empty( $data['edd_user_login'] ) ) {
+	if ( empty( $data['edd_user_login'] ) ) {
 		edd_set_error( 'empty_username', __( 'Invalid username', 'easy-digital-downloads' ) );
 	}
 
-	if( username_exists( $data['edd_user_login'] ) ) {
+	if ( username_exists( $data['edd_user_login'] ) ) {
 		edd_set_error( 'username_unavailable', __( 'Username already taken', 'easy-digital-downloads' ) );
 	}
 
-	if( ! validate_username( $data['edd_user_login'] ) ) {
+	if ( ! validate_username( $data['edd_user_login'] ) ) {
 		edd_set_error( 'username_invalid', __( 'Invalid username', 'easy-digital-downloads' ) );
 	}
 
-	if( email_exists( $data['edd_user_email'] ) ) {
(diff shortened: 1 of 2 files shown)


includes/process-purchase.php
```diff
@@ -746,81 +753,99 @@ function edd_purchase_form_validate_guest_user() {
...
+	// Allow themes and plugins to hook.
 	do_action( 'edd_insert_user', $user_id, $user_data );
 
-	// Login new user
+	// Login new user.
 	edd_log_user_in( $user_id, $user_data['user_login'], $user_data['user_pass'] );
 
-	// Return user id
+	// Since wp_signon doesn't set the current user, we need to do this.
+	wp_set_current_user( $user_id );
```
robincornett, line 806: Should we set line 803 to be a variable assignment, since `edd_log_user_in` will return either a `WP_User` or `WP_Error` object (what `wp_signon` returns), and then only set the current user if that is a `WP_User` object? It may not be necessary, though, as it looks like `wp_set_current_user` will perform the check itself, if the `$user_id` here is not a valid user.

spencerfinnell: Related to `edd_log_user_in()` it might make more sense to always call that in that function? `wp_signon()` calls `wp_set_auth_cookie()` so always calling `wp_set_current_user()` inside of `edd_log_user_in()` would be closer to the previous behavior: https://github.com/easydigitaldownloads/easy-digital-downloads/commit/b7efaf07d5ce2b0ba23ca07fb5bec22bc878f173#diff-bf871927add1d88bdfed2583a5816e54L121

includes/login-register.php
```diff
@@ -148,9 +148,12 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 				'</a>'
```
robincornett, line 148: Note: this was done back in the previous PR, but the `wp_lostpassword_url` shouldn't necessarily presume that the redirect should go back to the checkout screen. I'm testing the profile edit screen and redirecting to checkout doesn't make sense (it's really line 147 but GH won't let me comment there).

spencerfinnell: We could probably derive context by checking for `$_POST['action']` being `edd_process_checkout_login` -- the problem is we don't know how to find the login or profile editor pages (there is no setting). We could just go back to the homepage if we aren't on the Checkout login form.

cklosowski: let's log that for a 2.10 release @robincornett

Issue #7903, opened 2020-05-14
Active spinner can be hidden behind progress bar


robincornett:
### Expected behavior
When I begin a process which shows a progress bar with an active spinner, I expect to be able to see the spinner.

### Actual behavior
When I begin this kind of process, the active spinner is hidden behind the progress bar. Example:

![image](https://user-images.githubusercontent.com/4853066/81960690-55ec0c00-95df-11ea-8fe3-f57ed1bd23e5.png)

Affected items:
- export actions
- migration/upgrade screen
- possibly others?

### Steps to reproduce the behavior
1) In EDD 3.0, visit a screen which will have buttons/actions which result in a progress bar, such as Downloads > Reports > Export.
2) Click a button to generate a CSV.
3) The spinner is hidden behind the progress bar.
etc, etc

### Information (if a specific version is affected):
PHP Version: any
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

Any other relevant information:
In 2.9.x, the progress bar/spinner looks like this:

![image](https://user-images.githubusercontent.com/4853066/81961074-ddd21600-95df-11ea-8195-54eac339cec5.png)

robincornett:
Update: this is actually an issue in 2.9 as well, depending on the screen size. It has to do with the absolute positioning of the progress bar combined with its width: on smaller screens, the spinner may take up enough space that the progress bar's 95% width will overlap it. It can be seen on screens less than 1600px wide.

I think it is possible to update the related CSS to not use absolute positioning or floats and get a more reliable visual output (flexbox).

PR #8106, opened 2020-09-10
Fields Visible on Reports - Exports

### LINKED ISSUE #8094: Some Export Select Fields Hidden
### Expected behavior
When I resize the Downloads -> Reports -> Export screen I expect all select fields to be visible

### Actual behavior
On screens sized 481px - 542px there are fields hidden / half-hidden and inaccessible to chose.

Export Taxed Orders is the most noticeable example since the `Country` and `Region` fields are not visible.
image

Also hidden or half-hidden:
- Export Sales and Earnings
- Export Orders
- Export Customers

1) Go to Downloads -> Reports -> Export
2) Resize screen to be between 481px - 542px and see hidden / half-hidden fields
3) Resize screen outside of 481px - 542px range and see all of the fields

PHP Version:
`7.3`
EDD Version (or branch):
`release/3.0`
WordPress Version:
`5.5`

~~- This issue is not fixed in #7722~~
- Sort of related to #8054
- I tested in both Chrome and Firefox

LisaCee:
Fixes #8094

Proposed Changes:
1. Change Reports -> Exports to use CSS Grid
2. Ensure all fields are visible at all sizes
3. Make inputs / selects more uniform


LisaCee:
Ready for initial review.  I am still digging for unintended CSS consequences in other places and looking for CSS rules that can be pruned.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
 	margin-bottom: 20px;
 }
 
+div#post-body-content {
```
spencerfinnell, line 2423: @LisaCee This will affect all screens EDD is loaded on, and used by core WordPress markup so you'll get things like:

Screen Shot 2020-09-16 at 2 25 16 PM

Instead you can add a class to the wrapping `<div>` here:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports.php#L2555

Maybe something like `.edd-reports-export` and then you can create the grid just on that class.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+	display: grid;
+	display: -ms-grid;
+	grid-template-columns: 1fr;
+	grid-gap: 20px;
+	max-width: 1000px;
+}
+
+.postbox [class^="edd-export"] {
```
spencerfinnell, line 2431: Smart!

`.postbox` could be updated to `.edd-reports-export` if you make the above change as well. It shouldn't be breaking anything as is, but a more specific selector will be less likely to break things in the future.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+#edd-order-export-start,
+#edd-order-export-end,
+#edd-orders-export-start,
+#edd-orders-export-end,
+#edd-taxed-orders-export-start,
+#edd-taxed-orders-export-end,
+#edd-api-requests-export-start,
+#edd-api-requests-export-end,
+#edd-file-download-export-start,
+#edd-file-download-export-end {
```
spencerfinnell, line 2464: Could the markup be updated to use `class="edd-export-start"` and `class="edd-export-end"` and then CSS applied to those class names vs. needing to list each ID here?

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+	display: flex;
+	flex-direction: column;
+	flex-wrap: wrap;
+}
+
+div.to-and-from-container {
+	height: 50px;
+}
+
+#to-label {
```
spencerfinnell, line 2479: This should be prefixed with `edd-` and perhaps made a bit more specific to Exports? Also using a class instead of an ID will prevent duplicate ID warnings Chrome sometimes shows in the console.

spencerfinnell (changes requested):

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+	margin-right: 0px;
+}
+
+.edd-import-export-form {
+	display: flex;
+	flex-direction: column;
+	flex-wrap: wrap;
+}
+
+div.to-and-from-container {
```
robincornett, line 2475: I'm not a fan of setting specific heights on elements because I feel like it's spackling over things we should fix instead. If you set this element to use CSS Grid, something like:

```css
div.to-and-from-container {
    display: grid;
    grid-template-columns: 1fr 1fr;
    grid-gap: 8px;
    margin-bottom: 16px;
}
```

(This may need to be more specific for the exports context actually.) Once you have that, though, you can remove the negative top margin on the `#to-label`, and get more appropriate padding or margins set.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+}
+
+span.edd-label.edd-export-to-label {
+	line-height: normal;
+	align-self: center;
+	margin-top: -7px;
+}
+
+.edd-reports-wrapper .postbox .edd-select {
+	max-width: inherit;
```
robincornett, line 2486: This might be better off at 100%. I haven't replicated it but I did get at least one screen width where the select was oozing out of the container box.

includes/admin/reporting/reports.php
```diff
@@ -2560,11 +2560,17 @@ function display_export_report() {
                         <div class="inside">
                             <p><?php esc_html_e( 'Download a CSV giving a detailed look into earnings over time.', 'easy-digital-downloads' ); ?></p>
                             <form id="edd-export-earnings" class="edd-export-form edd-import-export-form" method="post">
-								<?php echo EDD()->html->month_dropdown( 'start_month' ); ?>
-								<?php echo EDD()->html->year_dropdown( 'start_year' ); ?>
-								<span class="edd-label"><?php echo _x( 'to', 'Date one to date two', 'easy-digital-downloads' ); ?></span>
-								<?php echo EDD()->html->month_dropdown( 'end_month' ); ?>
-								<?php echo EDD()->html->year_dropdown( 'end_year' ); ?>
+								<div class="to-and-from-container">
```
robincornett, line 2563: I think wrapping the grouped selects together here is a great idea.

Looking at the HTML, I'm suddenly wondering about 1) should this use the `.edd-form-group` markup, although that may be out of context of this PR; and 2) I'm not sure that any of these inputs have labels (also probably out of context, although would be great to fix while you're fiddling with the HTML anyway?).

assets/css/edd-admin.css
robincornett: Finishing reading the code and I am guessing the 49% on the selects was the best option with the original markup, but I think your markup change was a great idea and will allow more flexible CSS.

robincornett (changes requested):
Overall, it looks great! Just some nitpicky changes based on a first pass but this looks SO GOOD.

assets/css/edd-admin.css
LisaCee: I just learned about that wildcard selector today!

ashleyfae:
This is looking so excellent @LisaCee !

What do you guys think about utilizing more space on the screen on larger screens? (so more than 2 columns)

Screenshot from 2020-09-17 15-41-12

I can't really decide. On the one hand, sometimes a smaller width is easier to manage. That would be in favour of keeping it as it is. On the other hand, the reports do extend to fill up all available space, so that would be an argument for more columns. :woman_shrugging:

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+	margin-right: 0;
+}
+
+@media screen and ( min-width: 768px ) {
+	div.edd-reports-export {
+		display: grid;
+		display: -ms-grid;
+		grid-template-columns: 1fr 1fr;
```
robincornett, line 2494: I believe that only the `grid-template-columns` needs to be set here, as the others are set on smaller screens and unchanged, so don't need to be overridden.

assets/css/edd-admin.css
LisaCee: Nice solution.  I wasn't happy using ALL of the ids here.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2420,92 @@ body.download_page_edd-reports {
...
+
+.edd-reports-wrapper .postbox {
+	margin-bottom: 0;
+}
+#edd-export-earnings select.edd-select {
+	width: 49%;
+}
+
+span.edd-from-to-wrapper {
+	margin-bottom: 15px;
```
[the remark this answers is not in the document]
LisaCee: Found another way around this with @spencerfinnell 's suggestion for a new class.

assets/css/edd-admin.css
LisaCee: @robincornett Take a look at this and let me know what you think.

assets/css/edd-admin.css
LisaCee: You are my hero!  Every time I changed breakpoints I would see that weird overflow until I reloaded the page.  I think this fixed it.

assets/css/edd-admin.css
```diff
@@ -2350,18 +2350,6 @@ body.download_page_edd-reports {
...
-.edd-reports-wrapper form .edd-from-to-wrapper,
-.edd-reports-wrapper form .edd-select,
-.edd-reports-wrapper form .edd-label,
-.edd-reports-wrapper form input.button {
-	vertical-align: middle;
-	float: left;
-}
 
 .edd-reports-wrapper form .edd-label {
 	height: 29px;
```
robincornett, line 2355: I think all of this can be removed, and replaced with a `margin-bottom: 16px;` to get it to match the spacing above.

includes/admin/reporting/reports.php
```diff
@@ -2552,19 +2552,25 @@ function display_export_report() {
...
-								<?php echo EDD()->html->year_dropdown( 'start_year' ); ?>
-								<span class="edd-label"><?php echo _x( 'to', 'Date one to date two', 'easy-digital-downloads' ); ?></span>
-								<?php echo EDD()->html->month_dropdown( 'end_month' ); ?>
-								<?php echo EDD()->html->year_dropdown( 'end_year' ); ?>
+								<div class="edd-to-and-from-container">
+									<?php echo EDD()->html->month_dropdown( 'start_month' ); ?>
+									<?php echo EDD()->html->year_dropdown( 'start_year' ); ?>
+								</div>
+
+								<span class="edd-label edd-export-to-label"><?php echo _x( '-- to --', 'Date one to date two', 'easy-digital-downloads' ); ?></span>
```
robincornett, line 2568: I know it's just a class name, but it kind of bothers me to have `label` in the class name when this isn't a label--it's more of a spacer or descriptive (okay not descriptive because you can read it) element. What would you think about changing the class(es) to something else entirely and updating the CSS accordingly?

robincornett (commented):
Just another little tweak to the `edd-to-and-from-container` and I'm liking that a lot more. I like Ashley's thought about allowing more than two columns on larger screens. Wondering about setting the columns to repeat with a minmax of 300px-1fr? Currently, for me, on my larger monitor, the `.postbox` elements are getting big, like 490px wide, and ending up with empty space on the right. Changing the columns keeps them narrower and I end up with three across, which I'm okay with.

assets/css/edd-admin.css
robincornett: Sorry, I mean all of the rules on the `.edd-reports-wrapper form .edd-label` element (which I suggest elsewhere we maybe rename as well).

assets/css/edd-admin.css
```diff
@@ -2347,33 +2347,20 @@ body.download_page_edd-reports {
...
-	vertical-align: middle;
-	float: left;
-}
 
 .edd-reports-wrapper form .edd-label {
-	height: 29px;
-	line-height: 29px;
-	vertical-align: middle;
-	margin-right: 4px;
+	margin-bottom: 16px;
```
robincornett, line 2356: This margin-bottom could be moved to the rule beginning at line 2460 so that the element is styled by just one rule.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2418,55 @@ body.download_page_edd-reports {
...
+
+.edd-export-sales-earnings {
+	grid-template-columns: 1fr;
+}
+
+.edd-from-to-wrapper {
+	margin-right: 0;
+}
+
+input.edd-export-start,
```
robincornett, line 2440: Setting the widths on these inputs works, but I'm wondering about removing this entire rule and using flexbox on the container instead, something like:

```css
.edd-from-to-wrapper {
    margin-right: 0;
    display: flex;
    width: 100%;
    margin-bottom: 16px;
}

#edd-start-wrap,
#edd-end-wrap {
    flex-grow: 1;
}
```

Really it would be better to set a class on the `edd-start-wrap` and `edd-end-wrap` elements and use that. Also pretty sure the `edd-from-to-wrapper` would be better as a `div` instead of a `span` (in the markup), at least with this new styling--maybe it needed to be a span when everything was on one line.

robincornett: Looking at the rest of the codebase, you could change these to `div`, I think, but the CSS would need to be more specific to manage just the exports, as this from-to input group is replicated elsewhere and styling just `.edd-from-to-wrapper` would do bad things to that!

LisaCee: I couldn't get this to work with `flex-grow: 1`.  My solution feels kinda hacky, so please let me know if there is a better way to do this with flexbox.

includes/admin/reporting/reports.php
LisaCee: Issue [#8112](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8112)

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,57 @@ body.download_page_edd-reports {
...
+
+.edd-reports-wrapper .postbox {
+	margin-bottom: 0;
+}
+
+.edd-export-sales-earnings {
+	grid-template-columns: 1fr;
+}
+
+[class^="edd-export"] .edd-from-to-wrapper {
```
robincornett, line 2432: Should this use the `.edd-reports-export` class instead of `[class^="edd-export"]`?

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,57 @@ body.download_page_edd-reports {
...
+	display: flex;
+	margin-right: 0;
+	margin-bottom: 16px;
+	width: 100%;
+}
+
+.download_page_edd-reports .inside input[type="text"] {
```
robincornett, line 2439: You could replace this with `.edd-reports-export .edd-from-to-wrapper input`, which might feel better. Then add the `flex-grow` to the input parent `span` elements:

```css
.edd-reports-export .edd-from-to-wrapper span {
    flex-grow: 1;
}
```
I think both the width on the inputs and the flex-grow are needed to get that whole thing full width.

robincornett: (Instead of `span` you could target the IDs of the `span` elements....I was hoping there was a way to add a class to those but it does not appear to be possible.)

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,57 @@ body.download_page_edd-reports {
...
+.edd-import-export-form {
+	display: flex;
+	flex-direction: column;
+	flex-wrap: wrap;
+}
+
+div.edd-to-and-from-container {
+	display: grid;
+	display: -ms-grid;
+    grid-template-columns: 1fr 1fr;
```
robincornett, line 2452: Little detail, but this line is indented with spaces instead of a tab.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,57 @@ body.download_page_edd-reports {
 	margin-bottom: 20px;
 }
 
+div.edd-reports-export {
+	display: grid;
+	display: -ms-grid;
+	grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
```
robincornett, line 2420: I just realized, with my browser zoom set at 125%, this nicely goes to three boxes across on my screen, but at 100%, it's five across. So although the 300px works nicely for me, it may not be the best option for people with better eyesight? 👓

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,57 @@ body.download_page_edd-reports {
...
+	grid-gap: 20px;
+}
+
+.edd-reports-wrapper .postbox {
+	margin-bottom: 0;
+}
+
+.edd-export-sales-earnings {
+	grid-template-columns: 1fr;
```
robincornett, line 2429: I am not sure this rule is necessary--it does not seem to be doing anything on my end?

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,57 @@ body.download_page_edd-reports {
...
+}
+
+span.edd-label.edd-export-to-label {
+	line-height: normal;
+	align-self: center;
+	margin-bottom: 16px;
+
+}
+
+.edd-reports-wrapper .postbox .edd-select {
```
robincornett, line 2463: I think this is intended to override the existing ruleset defined beginning at line 2355, which applies to all `select` elements on reports, so this should not use the `.edd-reports-wrapper` alone--needs `.edd-reports-export` instead or as well. (Also likely that the margin-bottom in the above ruleset should be moved to here, although I'm not sure it's negatively impacting other reports screens that I can see so far.)

includes/admin/reporting/reports.php
```diff
@@ -2780,8 +2792,10 @@ function display_export_report() {
 
 				                wp_nonce_field( 'edd_ajax_export', 'edd_ajax_export' );
 				                ?>
-                                <input type="hidden" name="edd-export-class" value="EDD_Batch_Customers_Export"/>
-                                <input type="submit" value="<?php esc_html_e( 'Generate CSV', 'easy-digital-downloads' ); ?>" class="button-secondary"/>
+								<input type="hidden" name="edd-export-class" value="EDD_Batch_Customers_Export"/>
+								<span>
```
robincornett, line 2796: I was going to ask why you added the `span` tags here and I still wonder that a little bit, not due to your changes but due to the export forms' behavior. I do see that the rest of the export forms have them, as well as the `<span class="spinner"></span>`, which the customer export does not have. However, I think that the spinner on these forms is never actually implemented, because the [JS appends this entire `notice-wrap` element](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/js/admin/tools/export/index.js#L27) to show a progress bar instead, which I am reasonably certain includes its own spinner.

#7903 is related, and currently milestoned for 3.0.5 (@spencerfinnell what do you think, since you assigned it?), but we may want to consider bumping that up as it impacts all of these boxes, since the `notice-wrap` is no longer the width of the screen.

But all that is to say that maybe we should consider removing the spinners from the export form PHP, as I do not think they're ever exposed. At that point, the `span` around each button may possibly also be dispensed with, but I'm not sure.

robincornett (commented):
Sorry, multiple comments--mostly finicky things. This continues to look great to me! And one "phone a friend" question at the end.

includes/admin/reporting/reports.php
LisaCee: I added the `<span>` to make it consistent with the other spinner fields and make the styling the same.  With that said, though, I'm not sure that the spinners are being used either.  I thought it was just me, but I couldn't trigger them either.

assets/css/edd-admin.css
LisaCee: Thoughts @spencerfinnell and @ashleyfae?

My screen has 4 columns across and I think it is very readable.  I tend to err towards covering the person with iffy sight.  We could go a little smaller as these do scale well with browser zooming.

includes/admin/reporting/reports.php
LisaCee: It looks like this will be resolved in #7903

includes/admin/reporting/reports.php
cklosowski: We'll have to be careful here and make sure we don't cross-contaminate the front end since `edd-label` is used on the card fields (where they are `label` tags).

LisaCee: Could I just make up a class and leave the `edd-label` as is?  Maybe `edd-export-to-spacing`?  Something more eloquent / creative?

assets/css/edd-admin.css
```diff
@@ -4060,3 +4093,4 @@ select.edd-form-group__input {
 	line-height: initial;
 	margin: 8px 0 0;
 }
+
```
spencerfinnell, line 4096: @LisaCee Very strangely this extra new line at the end of the file is stopping the file from being loaded in my environment. Not really sure why but when I remove it everything is fine. Otherwise I get:

Screen Shot 2020-09-25 at 1 42 21 PM

cklosowski: Could it be a difference in line ending format (Windows vs Linux/Unix?)

assets/css/edd-admin.css
spencerfinnell: I really like how it's behaving currently.

assets/css/edd-admin.css
LisaCee: I _thought_ most files were supposed to end in a new line, but I can remove it if it's causing issues.  I can't duplicate it here in Chrome or FF.

includes/admin/reporting/reports.php
robincornett: This `.edd-to-and-from-container` isn't anywhere else in the codebase, so maybe a variation on that? `.edd-to-and-from--separator` or something?

assets/css/edd-admin.css
robincornett: I agree--it's a disappointment to switch to another branch and visit this screen now.

assets/css/edd-admin.css
spencerfinnell: Could be, but it works fine on my other machine which is using generally the same local environment. Lots of random suggestions here: https://stackoverflow.com/questions/58215104/whats-the-neterr-http2-protocol-error-about

I've seen it on/off on random things locally the past few weeks but have never been able to nail it down to anything until the empty line here.

includes/admin/reporting/reports.php
LisaCee: I like that.  I can make that work with no `label`.

assets/css/edd-admin.css
robincornett: For some reason I have two blank lines at the end of my copy of the file (4096 and 4097) but it's not causing issues here.

includes/admin/reporting/reports.php
spencerfinnell, line 2568:
```suggestion
								<span class="edd-label edd-export-to-label"><?php echo _x( '&mdash; to &mdash;', 'Date one to date two', 'easy-digital-downloads' ); ?></span>
```

@LisaCee Using an `—` (mdash) looks a little cleaner than `--`

spencerfinnell (changes requested):

assets/css/edd-admin.css
LisaCee: So, do we think it should stay or should it go?

spencerfinnell (approved):
This is awesome @LisaCee. So much easier to use. 

The file with the exports has a lot of weird spacing issues but I think that should be handled separately since they weren't introduced here and will just make these changes hard to compare.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,56 @@ body.download_page_edd-reports {
...
+}
+
+div.edd-to-and-from-container {
+	display: grid;
+	display: -ms-grid;
+	grid-template-columns: 1fr 1fr;
+	grid-gap: 8px;
+}
+
+span.edd-to-and-from--seperator {
```
robincornett, line 2456: I like having the different class, but there is `a rat` in separate--spelling should be updated here and in the PHP.

LisaCee: Good catch.  Thank you.

robincornett (approved):

ashleyfae:
This is a side effect in Downloads > Tools > Import/Export, with Software Licensing activated:

## release/3.0

Screenshot from 2020-09-29 14-30-47

## this branch

Screenshot from 2020-09-29 14-29-20

This is due to this CSS:

```css
.edd-import-export-form {
    display: flex;
    flex-direction: column;
    flex-wrap: wrap;
}
```

Perhaps we should prefix that with something? Like: `.edd-reports-export .edd-import-export-form` ?

robincornett:
What if, instead of (or in addition to) using the `.edd-reports-export` class for styling this, we created a little more multipurpose class that was something like `.edd-something-something--has-grid` and made the grid specific styling based off of that? This might give a head start on #8124, by allowing the same rules to be used as a baseline starting point for styling the separate screens, and if we wanted to grid out other screens later, possibly get the same head start with just adding the class?

LisaCee:
@ashleyfae - Good catch.  I used your solution of adding a more specific class to fix this for now.

@robincornett - I love the idea of a reusable class for grid containers, and I made note of it in issue #8124

robincornett:
@LisaCee could the new class be done as part of this PR please? That would allow me to use it for #8119 as well, as it's based on this, which covers tools as well as Reports > Export. I know you flagged it for the Tools issue, but if your PR there is based on mine (if it is), which is based on this, then we could all use it from here.

LisaCee:
@robincornett, I'll give it a try.  What do you think of the class `edd-admin--has-grid`?

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,56 @@ body.download_page_edd-reports {
...
 }
 
+.edd-admin--has-grid {
+	display: grid;
+	display: -ms-grid;
+	grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
+	grid-gap: 20px;
+}
+
+.edd-reports-wrapper .postbox {
```
robincornett, line 2424: This could be updated to use the new `.edd-admin--has-grid` class.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,56 @@ body.download_page_edd-reports {
...
+
+.edd-admin--has-grid .edd-from-to-wrapper input {
+	width: 100%;
+}
+
+.edd-admin--has-grid .edd-from-to-wrapper span {
+    flex-grow: 1;
+}
+
+.edd-admin--has-grid .edd-import-export-form {
```
robincornett, line 2443: With the new class, could this be changed to just `.edd-admin--has-grid form` so it will apply to any form in a grid? That may be too generic, but I'm almost thinking not.

assets/css/edd-admin.css
```diff
@@ -2431,6 +2414,56 @@ body.download_page_edd-reports {
...
+	grid-gap: 8px;
+}
+
+span.edd-to-and-from--separator {
+	line-height: normal;
+	align-self: center;
+	margin-bottom: 16px;
+}
+
+.edd-reports-wrapper .edd-admin--has-grid .postbox .edd-select {
```
robincornett, line 2462: I think this could drop the `.edd-reports-wrapper` class.

robincornett (commented):
The new class name sounds like a good option to me--thank you for adding it!

ashleyfae (approved):
Excellent work @LisaCee ! :raised_hands:

PR #8109, opened 2020-09-16
Add fallback parameter to edd_calculate_tax() function #8107

### LINKED ISSUE #8107: 3.0 - Manual order - taxes being added when I haven't set a customer address
### Expected behavior

When I manually create a new order and don't have a customer address selected, I expect the tax rate to be 0. It should only be greater than 0 if I select a customer address in a country that has a tax rate.

### Actual behavior

When I create an order and add a download before selecting an address, tax is being applied anyway.

1) Enable taxes. These are my tax settings:
Screenshot from 2020-09-14 12-48-56
2) Go to Downloads > Orders > Add New.
3) Add a download to the order. You'll see this notice:
Screenshot from 2020-09-14 12-46-02
The expectation here is that no tax will be set because you haven't set an address.
4) But after clicking `Add Download` you can see that tax has been applied:
Screenshot from 2020-09-14 12-46-21

PHP Version: 7.4

WordPress Version:

ashleyfae:
Fixes #8107

Proposed Changes:
1. Adds a `$fallback` parameter to the `edd_calculate_tax()` function. Set this to `false` when calling this function from inside `edd_admin_order_get_item_amounts()`.
2. Removes the `edd_get_tax_rate()` call (returned variable wasn't being used).

To test:

1. Make sure you have a `Business Country` set in Downloads > Settings > General.
2. Enable taxes and add a tax rate for your base country.
3. Manually create a new order via Downloads > Orders > Add New.
4. Add an item to your order, but do not set a customer/address, and ensure no tax is applied.

Also ensure that if you _do_ set a country that has a tax rate, tax is still applied.

### DIFF
--- a/includes/tax-functions.php
+++ b/includes/tax-functions.php
@@ -151,8 +151,9 @@ function edd_active_tax_rates_query_clauses( $clauses ) {
  *
  * @param string  $country Country.
  * @param string  $region  Region.
- * @param boolean $fallback Fall back to the current Customer's address information
- *                          or server $_POST data. Default true.
+ * @param boolean $fallback Fall back to (in order): server $_POST data, the current Customer's
+ *                          address information, then your store's Business Country setting.
+ *                          Default true.
  *
  * @return mixed|void
  */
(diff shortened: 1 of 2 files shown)


includes/tax-functions.php
```diff
@@ -259,15 +259,18 @@ function edd_get_formatted_tax_rate( $country = false, $state = false ) {
  *
  * @since 1.3.3
  * @since 3.0 Renamed $state parameter to $region.
+ *            Added $fallback parameter.
  *
  * @param float  $amount  Amount.
  * @param string $country Country. Default base country.
  * @param string $region  Region. Default base region.
+ * @param boolean $fallback Fall back to the current Customer's address information
+ *                          or server $_POST data. Default true.
```
spencerfinnell, line 268: @ashleyfae Can we update this (and the `@param` on `edd_get_tax_rate()`) to mention the final fallback being the store's address settings?

spencerfinnell (changes requested):

includes/tax-functions.php
ashleyfae: Good idea - updated!

spencerfinnell (approved):

PR #8116, opened 2020-09-21
Ensure user has install_plugins capability to install Jilt #8115

### LINKED ISSUE #8115: Allow Jilt installation to super admins only when in MU
This was reported by a customer what he considers a security issue. He is using EDD in a MU setup and admins can install Jilt by just clicking the button available on the EDD Emails page.
He does not use Jilt and does not want it to be available to admins.

### Justification or use case
Customer does not use Jilt and does not want it to be available to admins on his network sites.

ashleyfae:
Fixes #8115

Proposed Changes:
1. Add additional check for `install_plugins` capability.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -2254,7 +2254,7 @@ function edd_jilt_callback( $args ) {
 
 		<?php endif; ?>
 
-	<?php else : ?>
+	<?php elseif( current_user_can( 'install_plugins' ) ) : ?>
 
 		<p>
 			<button id="edd-jilt-connect" class="button button-primary">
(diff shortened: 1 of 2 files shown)


robincornett (commented):
This works to hide the Jilt button, but it feels a little off to show the setting/description if there isn't anything a user can do (other than click the links).

ashleyfae:
> This works to hide the Jilt button, but it feels a little off to show the setting/description if there isn't anything a user can do (other than click the links).

I'm kind of in two minds about that. On the one hand, it's nice to say, _"this option exists"_ because then the user can ask the network admin about it. On the other hand I do agree it's a little weird. We could have some other kind of notice like, _"Ask the network admin to install this for you"_ but I can't think of good wording right now.

robincornett:
Yes, I agree--everything I can think of wording wise feels a little smarmy. It's a little complicated by the description being separated from the buttons.

The PR does address and resolve the immediate issue. I'm happy to approve and if we run into confusion later then we could look into updating the language. It may not be necessary.

robincornett (approved):

cklosowski:
I restarted the failed builds on Travis. All is good now.

PR #8117, opened 2020-09-21
Update redirect when directing to lost password URL

### LINKED ISSUE #8100: Failed login reset password link should not always go to the checkout page
### Expected behavior
When I click a password reset link and request a password reset, I expect the final redirect to take me back to where I was, or the front page of the site if that is not possible.

### Actual behavior
Since changes in 2.9.24, using an EDD login form and failing to log in (due to an incorrect username or password), the password reset link includes a redirect to the [EDD checkout screen](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/login-register.php#L147). This is the case even if the login shortcode is on the profile page, or the purchase history page.

1) Create a page with the EDD login shortcode.
2) Log in from that page with an invalid username or password.
3) Fill out the form. You are redirected to the checkout screen.

PHP Version: 7.4.1
EDD Version (or branch): master, 2.9.24
WordPress Version: 5.5.1

Introduced in PR #8085.

robincornett:
Fixes #8100

Proposed Changes:
1. Because the EDD login process isn't always part of the checkout process, the current referer is set as the redirect instead of hard coding in the checkout page.

Base set to 2.10.

### DIFF
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -171,7 +171,7 @@ add_filter( 'wp_login_errors', 'edd_login_register_error_message', 10, 2 );
 function edd_login_register_error_message( $errors, $redirect ) {
 	$action_is_confirm   = ! empty( $_GET['checkemail'] ) && 'confirm' === sanitize_text_field( $_GET['checkemail'] );
 	$edd_action_is_reset = ! empty( $_GET['edd_reset_password'] ) && 'confirm' === sanitize_text_field( $_GET['checkemail'] );
-	$redirect_url        = ! empty( $_GET['edd_redirect'] ) ? esc_url( $_GET['edd_redirect'] ) : false;
+	$redirect_url        = ! empty( $_GET['edd_redirect'] ) ? $_GET['edd_redirect'] : false;
 	if ( $action_is_confirm && $edd_action_is_reset && $redirect_url ) {
 		$errors->remove( 'confirm' );
 		$errors->add(
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -181,14 +181,17 @@ add_filter( 'wp_login_errors', 'edd_login_register_error_message', 10, 2 );
  * @return void
  */
 function edd_login_register_error_message( $errors, $redirect ) {
-	if ( 'confirm' === filter_input( INPUT_GET, 'edd_reset_password', FILTER_SANITIZE_STRING ) ) {
+	$action_is_confirm   = 'confirm' === filter_input( INPUT_GET, 'checkemail', FILTER_SANITIZE_STRING );
+	$edd_action_is_reset = 'confirm' === filter_input( INPUT_GET, 'edd_reset_password', FILTER_SANITIZE_STRING );
+	$redirect_url        = filter_input( INPUT_GET, 'edd_redirect', FILTER_SANITIZE_URL );
+	if ( $action_is_confirm && $edd_action_is_reset ) {
 		$errors->remove( 'confirm' );
 		$errors->add(
 			'confirm',
 			sprintf(
 				/* translators: %s: Link to the referring page. */
 				__( 'Check your email for the confirmation link, then <a href="%s">return to what you were doing</a>.', 'easy-digital-downloads' ),
-				filter_input( INPUT_GET, 'edd_redirect', FILTER_SANITIZE_URL )
+				$redirect_url
 			),
 			'message'
 		);
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -136,11 +136,19 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 		'remember'      => $remember,
 	);
 
-	$user     = wp_signon( $credentials );
-	$redirect = wp_get_referer();
-	if ( empty( $redirect ) ) {
-		$redirect = edd_get_checkout_uri();
+	$user = wp_signon( $credentials );
+	$url  = wp_get_referer();
+	if ( empty( $url ) ) {
+		$url = edd_get_checkout_uri();
 	}
+	$redirect = add_query_arg(
+		array(
+			'checkemail'         => 'confirm',
+			'edd_reset_password' => 'confirm',
+			'edd_redirect'       => $url,
+		),
+		'wp-login.php'
+	);
 
 	if ( ! $user instanceof WP_User ) {
 		edd_set_error(
@@ -163,6 +171,31 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 
 }
 
+add_filter( 'wp_login_errors', 'edd_login_register_error_message', 10, 2 );
+/**
+ * Changes the WordPress login confirmation message when using EDD's reset password link.
+ *
+ * @since 2.10
+ * @param object \WP_Error $errors
+ * @param string $redirect
+ * @return void
+ */
+function edd_login_register_error_message( $errors, $redirect ) {
+	if ( 'confirm' === filter_input( INPUT_GET, 'edd_reset_password', FILTER_SANITIZE_STRING ) ) {
+		$errors->remove( 'confirm' );
+		$errors->add(
+			'confirm',
+			sprintf(
+				/* translators: %s: Link to the referring page. */
+				__( 'Check your email for the confirmation link, then <a href="%s">return to what you were doing</a>.', 'easy-digital-downloads' ),
+				filter_input( INPUT_GET, 'edd_redirect', FILTER_SANITIZE_URL )
+			),
+			'message'
+		);
+	}
+
+	return $errors;
+}
 
 /**
  * Process Register Form
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -147,7 +147,7 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 			'edd_reset_password' => 'confirm',
 			'edd_redirect'       => $url,
 		),
-		'wp-login.php'
+		wp_login_url()
 	);
 
 	if ( ! $user instanceof WP_User ) {
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -137,18 +137,6 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 	);
 
 	$user = wp_signon( $credentials );
-	$url  = wp_get_raw_referer();
-	if ( empty( $url ) ) {
-		$url = edd_get_checkout_uri();
-	}
-	$redirect = add_query_arg(
-		array(
-			'checkemail'         => 'confirm',
-			'edd_reset_password' => 'confirm',
-			'edd_redirect'       => esc_url_raw( $url ),
-		),
-		wp_login_url()
-	);
 
 	if ( ! $user instanceof WP_User ) {
 		edd_set_error(
@@ -156,7 +144,7 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 			sprintf(
 				/* translators: %1$s Opening anchor tag, do not translate. %2$s Closing anchor tag, do not translate. */
 				__( 'Invalid username or password. %1$sReset Password%2$s', 'easy-digital-downloads' ),
-				'<a href="' . esc_url( wp_lostpassword_url( $redirect ) ) . '">',
+				'<a href="' . esc_url( edd_get_lostpassword_url() ) . '">',
 				'</a>'
 			)
 		);
@@ -198,6 +186,30 @@ function edd_login_register_error_message( $errors, $redirect ) {
 	}
 
 	return $errors;
+}
+
+/**
+ * Gets the lost password URL, customized for EDD. Using this allows the password
+ * reset form to redirect to the login screen with the EDD custom confirmation message.
+ *
+ * @since 2.10
+ * @return string
+ */
+function edd_get_lostpassword_url() {
+	$url = wp_get_raw_referer();
+	if ( empty( $url ) ) {
+		$url = edd_get_checkout_uri();
+	}
+	$redirect = add_query_arg(
+		array(
+			'checkemail'         => 'confirm',
+			'edd_reset_password' => 'confirm',
+			'edd_redirect'       => esc_url_raw( $url ),
+		),
+		wp_login_url()
+	);
+
+	return wp_lostpassword_url( $redirect );
 }
 
 /**
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -200,10 +200,21 @@ function edd_login_register_error_message( $errors, $redirect ) {
  * @return string
  */
 function edd_get_lostpassword_url() {
-	$url = wp_get_raw_referer();
+	$url = false;
+	// `wp_get_raw_referer` was added in WP 4.5; this check is required as we still support 4.4.
+	if ( function_exists( 'wp_get_raw_referer' ) ) {
+		$url = wp_get_raw_referer();
+	} else {
+		if ( ! empty( $_REQUEST['_wp_http_referer'] ) ) {
+			$url = wp_unslash( $_REQUEST['_wp_http_referer'] );
+		} elseif ( ! empty( $_SERVER['HTTP_REFERER'] ) ) {
+			$url = wp_unslash( $_SERVER['HTTP_REFERER'] );
+		}
+	}
 	if ( empty( $url ) ) {
 		$url = edd_get_checkout_uri();
 	}
+	$url      = wp_validate_redirect( $url );
 	$redirect = add_query_arg(
 		array(
 			'checkemail'         => 'confirm',
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -200,8 +200,7 @@ function edd_login_register_error_message( $errors, $redirect ) {
  * @return string
  */
 function edd_get_lostpassword_url() {
-	$url      = edd_get_current_page_url();
-	$url      = wp_validate_redirect( $url, edd_get_checkout_uri() );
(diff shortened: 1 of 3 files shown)


includes/login-register.php
```diff
@@ -136,15 +136,16 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
 		'remember'      => $remember,
 	);
 
-	$user = wp_signon( $credentials );
+	$user     = wp_signon( $credentials );
+	$redirect = $_SERVER['HTTP_REFERER'];
```
ashleyfae, line 140: Let's use `wp_get_referer()` instead. Also, add in a check if it's empty (it sometimes can be) and if so, maybe then we should fall back to `edd_get-checkout_uri()`.

ashleyfae (changes requested):

ashleyfae:
This is working well. :+1: However, it is a little weird there isn't a success message. It makes me wonder if we should even be supplying a redirect at all. :thinking: 

If we don't supply a redirect, WordPress core shows a nice message:

> Check your email for the confirmation link, then visit the login page.

We could add a query arg to our redirect URL, then show a success message where possible (for instance, we could show one above the login form if you happen to be going back there)... but because we're taking them back to the referer, we can't guarantee the success message will always show up.

LisaCee:
> This is working well.  However, it is a little weird there isn't a success message. It makes me wonder if we should even be supplying a redirect at all. 
> 
> If we don't supply a redirect, WordPress core shows a nice message:
> 
> > Check your email for the confirmation link, then visit the login page.

I tend to think maybe we should be using the WP core message, too.  It was a little jarring to be dropped right back into the site without having reset my password.  I know the screen where you put in your username/email says to check your email, but that can easily be overlooked.

Also, a few other things to consider here:
- Should the `[edd_profile_editor]` login and the `Lost Password` link redirect to the same screen as the `[edd_login]`?
- Reminder that the [docs](https://docs.easydigitaldownloads.com/article/222-eddlogin) will need to be updated

robincornett:
@ashleyfae, @LisaCee: see what you think now. Current changes:

1. Primary redirect link is now `wp-login.php`, consistent with WP.
2. This allows us to filter `wp_login_errors` and change the WP Core message.
3. Adding custom query args for the redirect link lets us also pass through the referer URL, so the "lost password" confirmation page links direct to that.

The wording of the message could be more inspired but wanted feedback before finessing.

This behavior should apply to any EDD form which uses the `edd_log_user_in` function, but should be checked with any screens which include a login shortcode/form.

includes/login-register.php
```diff
@@ -136,11 +136,19 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+	if ( empty( $url ) ) {
+		$url = edd_get_checkout_uri();
 	}
+	$redirect = add_query_arg(
+		array(
+			'checkemail'         => 'confirm',
+			'edd_reset_password' => 'confirm',
+			'edd_redirect'       => $url,
+		),
+		'wp-login.php'
```
ashleyfae, line 150: We shouldn't hard-code this in case someone has changed it (common with some security plugins).

```suggestion
		wp_login_url()
```

ashleyfae (changes requested):

robincornett:
Note: `wp_get_referer` is consistently returning `false` for me, so the redirect is still always set to the checkout screen. I think it is [due to this information in the doc block](https://core.trac.wordpress.org/browser/tags/5.5.1/src/wp-includes/functions.php#L1891) for that function:

> If it's the same as the current request URL, will return false.

Wondering if we should use `wp_get_raw_referer()` instead, which does return the current URL.

LisaCee:
With `wp_get_referer` replaced with `wp_get_raw_referer`, this works well for me.  

_Aside: can anyone explain why `referer` is [intentionally misspelled](https://developer.wordpress.org/reference/functions/wp_get_referer/)?_

spencerfinnell:
@LisaCee Backwards compatibility from the original proposal in 1996: 

https://en.wikipedia.org/wiki/HTTP_referer#Etymology
https://tools.ietf.org/html/rfc1945

robincornett:
There are two other usages of `wp_lostpassword_url()` in EDD. I'm wondering if we should update those to include this new redirect behavior as well, and if so, should it be consolidated into an `edd_lostpassword_url()` function and used? The usages are in `templates/shortcode-login.php` and `includes/payments/actions.php`.

includes/payments/actions.php
```diff
@@ -534,7 +534,7 @@ function edd_recovery_force_login_fields() {
 			<div class="edd-alert edd-alert-info">
 				<p><?php _e( 'To complete this payment, please login to your account.', 'easy-digital-downloads' ); ?></p>
 				<p>
-					<a href="<?php echo wp_lostpassword_url(); ?>" title="<?php _e( 'Lost Password', 'easy-digital-downloads' ); ?>">
+					<a href="<?php echo edd_get_lostpassword_url(); ?>" title="<?php _e( 'Lost Password', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 537:
```suggestion
					<a href="<?php echo esc_url( edd_get_lostpassword_url() ); ?>" title="<?php esc_attr_e( 'Lost Password', 'easy-digital-downloads' ); ?>">
```

templates/shortcode-login.php
```diff
@@ -29,7 +29,7 @@
 				<input id="edd_login_submit" type="submit" class="edd-submit" value="<?php _e( 'Log In', 'easy-digital-downloads' ); ?>"/>
 			</p>
 			<p class="edd-lost-password">
-				<a href="<?php echo wp_lostpassword_url(); ?>">
+				<a href="<?php echo edd_get_lostpassword_url(); ?>">
```
ashleyfae, line 32:
```suggestion
				<a href="<?php echo esc_url( edd_get_lostpassword_url() ); ?>">
```

includes/login-register.php
```diff
@@ -159,6 +159,62 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+ * @since 2.10
+ * @param object \WP_Error $errors
+ * @param string $redirect
+ * @return void
+ */
+function edd_login_register_error_message( $errors, $redirect ) {
+	$action_is_confirm   = 'confirm' === filter_input( INPUT_GET, 'checkemail', FILTER_SANITIZE_STRING );
+	$edd_action_is_reset = 'confirm' === filter_input( INPUT_GET, 'edd_reset_password', FILTER_SANITIZE_STRING );
+	$redirect_url        = filter_input( INPUT_GET, 'edd_redirect', FILTER_SANITIZE_URL );
+	if ( $action_is_confirm && $edd_action_is_reset ) {
```
ashleyfae, line 175:
```suggestion
	if ( $action_is_confirm && $edd_action_is_reset && $redirect_url ) {
```

No point customizing the error message if we don't have a redirect URL for whatever reason.

includes/login-register.php
```diff
@@ -159,6 +159,62 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		$errors->remove( 'confirm' );
+		$errors->add(
+			'confirm',
+			apply_filters(
+				'edd_login_register_error_message',
+				sprintf(
+					/* translators: %s: Link to the referring page. */
+					__( 'Follow the instructions in the confirmation email you just received, then <a href="%s">return to what you were doing</a>.', 'easy-digital-downloads' ),
+					$redirect_url
```
ashleyfae, line 184:
```suggestion
					esc_url( $redirect_url )
```

ashleyfae (changes requested):

includes/login-register.php
```diff
@@ -159,6 +159,62 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+
+/**
+ * Gets the lost password URL, customized for EDD. Using this allows the password
+ * reset form to redirect to the login screen with the EDD custom confirmation message.
+ *
+ * @since 2.10
+ * @return string
+ */
+function edd_get_lostpassword_url() {
+	$url = wp_get_raw_referer();
```
robincornett, line 203: This function was actually introduced in WP 4.5, so we need something as a fallback for 4.4, since we do still support that. @ashleyfae is it okay if I copy over the [logic from the function](https://developer.wordpress.org/reference/functions/wp_get_raw_referer/)?

ashleyfae: Looking at this some more..

`wp_get_raw_referer()` seems to differ from `wp_get_referer()` for two reasons:

1. `wp_get_raw_referer()` doesn't care if it's the same as the current URL.
2. `wp_get_raw_referer()` does not validate the redirect (using `wp_validate_redirect()` ).

I suspect those are the two reasons why they say:

```
Do not use for redirects, use wp_get_referer() instead.
```

I'm not sure if that's something we should be ignoring just because it was a convenient "fix" for `wp_get_referer()` returning `false`. Maybe we should look more into why that was?

robincornett: I think part of the issue is how and where we are getting and using the redirect URL. I think generally, the redirect retrieval and output are happening in one place, so it makes sense that the referer URL being the same would be problematic. In our situation, however, we are getting the URL in one place, but sending it to a second page (actually onto a third) to be used. So I think it makes sense why the one is not working for our situation and the other is.

Maybe we should also incorporate the `wp_validate_redirect` function before using the URL, though, so there is further checking going on.

includes/login-register.php
```diff
@@ -159,6 +159,73 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+ * Changes the WordPress login confirmation message when using EDD's reset password link.
+ *
+ * @since 2.10
+ * @param object \WP_Error $errors
+ * @param string $redirect
+ * @return void
+ */
+function edd_login_register_error_message( $errors, $redirect ) {
+	$action_is_confirm   = 'confirm' === filter_input( INPUT_GET, 'checkemail', FILTER_SANITIZE_STRING );
+	$edd_action_is_reset = 'confirm' === filter_input( INPUT_GET, 'edd_reset_password', FILTER_SANITIZE_STRING );
```
cklosowski, line 173: Is it possible for any of these inputs to ever be passed in a POST request, instead of a GET? If it is possible or even likely that someone could convert these to a POST via a template or ourselves in the future, then using `INPUT_GET` would fail for them as we process the error handling.

includes/login-register.php
```diff
@@ -159,6 +159,73 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		$url = wp_get_raw_referer();
+	} else {
+		if ( ! empty( $_REQUEST['_wp_http_referer'] ) ) {
+			$url = wp_unslash( $_REQUEST['_wp_http_referer'] );
+		} elseif ( ! empty( $_SERVER['HTTP_REFERER'] ) ) {
+			$url = wp_unslash( $_SERVER['HTTP_REFERER'] );
+		}
+	}
+	if ( empty( $url ) ) {
+		$url = edd_get_checkout_uri();
```
robincornett, line 215: I just saw an issue come up with the auto register extension which pointed out that there is a setting for the login redirect. I think that should be the fallback, instead of the checkout page.

includes/login-register.php
```diff
@@ -159,6 +159,78 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+ *
+ * @since 2.10
+ * @param object \WP_Error $errors
+ * @param string $redirect
+ * @return void
+ */
+function edd_login_register_error_message( $errors, $redirect ) {
+	$action_is_confirm   = ! empty( $_GET['checkemail'] ) && 'confirm' === sanitize_text_field( $_GET['checkemail'] );
+	$edd_action_is_reset = ! empty( $_GET['edd_reset_password'] ) && 'confirm' === sanitize_text_field( $_GET['checkemail'] );
+	$redirect_url        = ! empty( $_GET['edd_redirect'] ) ? esc_url( $_GET['edd_redirect'] ) : false;
```
ashleyfae, line 174:
```suggestion
	$redirect_url        = ! empty( $_GET['edd_redirect'] ) ? $_GET['edd_redirect'] : false;
```

This is being escaped twice -- once here and again further down in the `sprintf()`. I suggest removing this one because it's better to escape later.

includes/login-register.php
```diff
@@ -159,6 +159,78 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		$url = wp_get_raw_referer();
+	} else {
+		if ( ! empty( $_REQUEST['_wp_http_referer'] ) ) {
+			$url = wp_unslash( $_REQUEST['_wp_http_referer'] );
+		} elseif ( ! empty( $_SERVER['HTTP_REFERER'] ) ) {
+			$url = wp_unslash( $_SERVER['HTTP_REFERER'] );
+		}
+	}
+	if ( empty( $url ) ) {
+		$login_redirect_page = edd_get_option( 'login_redirect_page', '' );
```
ashleyfae, line 215: This doesn't really feel right to me as a fallback. My understanding is the login redirect page is the page users are redirected to _after logging in_, meaning they typically have to be logged in to access that page. Thus it doesn't feel right to use this as a possible redirect after a lost password submission, because they won't be logged in after that.

includes/login-register.php
```diff
@@ -159,6 +159,78 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		if ( ! empty( $login_redirect_page ) ) {
+			$url = get_permalink( $login_redirect_page );
+		} else {
+			$url = edd_get_checkout_uri();
+		}
+	}
+	$url      = wp_validate_redirect( $url );
```
ashleyfae, line 222: Should we be passing a default in the event the URL doesn't validate? (second parameter)

ashleyfae (changes requested):

includes/login-register.php
```diff
@@ -159,6 +159,70 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		} elseif ( ! empty( $_SERVER['HTTP_REFERER'] ) ) {
+			$url = wp_unslash( $_SERVER['HTTP_REFERER'] );
+		}
+	}
+	$url      = wp_validate_redirect( $url, edd_get_checkout_uri() );
+	$redirect = add_query_arg(
+		array(
+			'checkemail'         => 'confirm',
+			'edd_reset_password' => 'confirm',
+			'edd_redirect'       => esc_url_raw( $url ),
```
ashleyfae, line 219: This should probably be `urlencode()` instead of `esc_url_raw()`, then when we use it in `edd_login_register_error_message`, we should `urldecode()` it.

includes/login-register.php
```diff
@@ -159,6 +159,70 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		),
+		wp_login_url()
+	);
+
+	return wp_lostpassword_url( $redirect );
```
ashleyfae, line 224: We could add a filter here so people can change this to be a custom lost password page or whatever.

ashleyfae (changes requested):

includes/login-register.php
robincornett: WordPress Core does provide the `lostpassword_url` filter in that function already; do you want a separate one?

ashleyfae: That's true. Let's leave it without for now.

includes/login-register.php
```diff
@@ -159,6 +159,70 @@ function edd_log_user_in( $user_id, $user_login, $user_pass, $remember = false )
...
+		} elseif ( ! empty( $_SERVER['HTTP_REFERER'] ) ) {
+			$url = wp_unslash( $_SERVER['HTTP_REFERER'] );
+		}
+	}
+	$url      = wp_validate_redirect( $url, edd_get_checkout_uri() );
+	$redirect = add_query_arg(
+		array(
+			'checkemail'         => 'confirm',
+			'edd_reset_password' => 'confirm',
+			'edd_redirect'       => urlencode( $url ),
```
ashleyfae, line 219: Follow up from Slack conversation: this should actually be the _current_ URL instead of the referring URL, because at the point this is used, you're currently on the page we should ultimately be sending you back to.

ashleyfae (changes requested):

includes/login-register.php
```diff
@@ -200,17 +200,7 @@ function edd_login_register_error_message( $errors, $redirect ) {
...
-	if ( function_exists( 'wp_get_raw_referer' ) ) {
-		$url = wp_get_raw_referer();
-	} else {
-		if ( ! empty( $_REQUEST['_wp_http_referer'] ) ) {
-			$url = wp_unslash( $_REQUEST['_wp_http_referer'] );
-		} elseif ( ! empty( $_SERVER['HTTP_REFERER'] ) ) {
-			$url = wp_unslash( $_SERVER['HTTP_REFERER'] );
-		}
-	}
+	$url      = edd_get_current_page_url();
```
ashleyfae, line 203: Let's inline this in the below `wp_validate_redirect()` function call. We don't need it in a variable.

ashleyfae (changes requested):

ashleyfae (approved):
@robincornett Can you confirm if unit tests are passing for you locally? I can't run them due to my PHPUnit version.

robincornett:
Everything that ran passed, @ashleyfae, but 7 tests were marked incomplete and it didn't tell me specifically which those were. However, I ran `phpunit` on `master` and also had 7 incomplete tests, which look to be approximately the same ones.

ashleyfae:
> Everything that ran passed, @ashleyfae, but 7 tests were marked incomplete and it didn't tell me specifically which those were. However, I ran `phpunit` on `master` and also had 7 incomplete tests, which look to be the same ones.

I remember that issue existed on 3.0 and I got it fix there, so I suspect they're probably the same tests. Nothing to worry about.

Thanks!

PR #8119, opened 2020-09-22
Update EDD progress bar styling

### LINKED ISSUE #7903: Active spinner can be hidden behind progress bar

robincornett:
Fixes #7903

Proposed Changes:
1. Update EDD `.notice-wrap` styling to be more flexible/responsive.

Branch based on `issue/8094` as the progress bars there are no longer the width of the screen.

To test:
1) run a 2.9 -> 3.0 migration via the UI (there are multiple progress bars)
2) Visit Downloads > Reports > Export and use the export tools.


LisaCee:
I think the margins on the spinner need to be tweaked a little.  I'm getting a flat side.  Do we want both the progress bar and spinner? 
image

robincornett:
I'm not sure why the flat side is happening there, but I don't think it's due to margins. I'm able to get a flat side, but only by increasing my browser zoom significantly, which I doubt is what you are doing. The spinner is actually a background image, and the spinner itself has a set width and height which is supposed to match the image (this is Core WP; all I'm changing on the spinner here is the margin). If there is a mismatch and the image is larger than the spinner `span`, there will be a flat side.

I did think about adjusting the markup of the `.notice-wrap` element--primarily to move the spinner to be the second element of the container, not the first, so I wouldn't have to set the `order`, but hesitated to do so because the progress bar/spinner pattern (or vice versa, technically) occurs quite a lot in the codebase. My understanding is that the bar is used for jobs which are expected to take more time, generally.

LisaCee:
Hmmm.  You are right that I didn't have the screen zoomed.  

I can't get a screenshot, but it looks like it's being cut off.  The spinner image is 20px, but certain screen sizes (480px, 782px, ...) are only giving it 17.92px of space.  I'm not sure how to fix this, but I think that is at least part of of the problem.

ashleyfae:
> Do we want both the progress bar and spinner?

I would be fine with just the progress bar.

robincornett:
I've removed the spinner (note: I did not commit the compiled JS files so you will have to `npm run build` to get the updated JS. Some questions to consider:

- Should I go ahead and remove the `<spinner>` markup which is in various places and not used (export forms, 3.0 migration screen)? It wasn't used anyway, as the notice-wrap has its own spinner, so it was never necessary.
- One thing I'm noticing with just the progress bar present, especially on large jobs (migrating orders, for example), is that it doesn't quite "feel" like something is happening, because the progress bar grows so slowly. The spinner kind of gave the visual cue that something was going on. Wondering if we should consider some kind of animation effect on the progress bar?
- With the grid layout, on forms which are shorter than their neighbors, it would be nice if the progress bar was just at the bottom of the box instead of at the end of the form. With the current markup/behavior (the notice wrap is inside the form, and I think if it was added _after_ the form, this might be more manageable), I'm not sure this is easy to manage without feeling hacky, so I'm ignoring it for now, but it seems like it would be nice.

ashleyfae:
> Should I go ahead and remove the <spinner> markup which is in various places and not used 

If it's not used, then sure!

> it doesn't quite "feel" like something is happening, because the progress bar grows so slowly. 

Another option is to add the spinner inside the disabled button?

robincornett:
I'm not sure I can insert the spinner into the buttons, because they're not actually buttons--they are `<input type="submit">` (both on the migration and export screens). I could re-add the spinner after the "button", I guess, and just update the JS to show that one?

ashleyfae:
Any reason we can't change them to `<button type="submit">` ? That should make it doable, I think.

LisaCee:
I had wondered why they were not buttons while I was working on #8094.  I _may_ have to tweak a rule in that PR, but buttons with spinners make more sense to me.

robincornett:
I feel like I had asked that on the migrations screen at some point and was told no because of backwards compatibility, but it could have been something else entirely (like links that are really buttons), so I'll try it.

robincornett:
@ashleyfae @LisaCee I think this is ready for review now. I was working on something else over the weekend and in the process of installing a plugin, watched how WP Core handles spinner elements within buttons--just adds a class which enables a pseudo-element. So I've used that.

I don't know if we are open to it, but it would be nice if the progress bar was a little prettier--maybe round the edges off a bit, and I would be interested in considering animating it somehow--not candy striping, but maybe some kind of subtle gradient effect if we could.

LisaCee:
Maybe this is just my computer being wonky, but I'm still seeing the spinners getting cut off.

> , watched how WP Core handles spinner elements within buttons-

Also, should I be seeing the spinners within the button elements?  Or are they alongside the progress bars still?

robincornett:
Sorry, @LisaCee, I should have said to run `npm run build` to get the scripts--I did not commit any of the compiled scripts, but you'll need them to actually see the changes. The original PR wasn't changing scripts but now we are.

Also noting that there are likely extensions which make use of the export pattern for their own exports or updates (Software Licensing has an export tool, and I just ran into an Activation count update which appears to use it as well), so the scripts may need to be finessed to work with extensions using `input` elements instead of `button`....they may not get the new spinners, but we need to make sure the input is still disabled, and I think right now it won't be.

ashleyfae:
I'm digging this! :raised_hands: Still need to do some testing on the upgrade UI, but Reports > Export is looking slick.

LisaCee:
I love this.  So far my only little nitpick is that on smaller screens (<782px) the check mark and spinner are a little high.  I don't think it's any of our CSS rules, but maybe we could tweak it to make it more centered.

image

robincornett:
Yes, that bothered me on the mobile buttons as well--it does affect all buttons in the admin, not just ours. I could fix it, I think, by changing the `vertical-align` and `margin` on the pseudo-element, even if it is just for our buttons.

ashleyfae (approved):

LisaCee (approved):
This looks so good!

robincornett:
Updated with the latest grid classes added and tweaked the button icon placement so it's a little more centered on all screen sizes.

ashleyfae (approved):

PR #8120, opened 2020-09-22
#8068 "Download Requirements" Toggle

### LINKED ISSUE #8068: 3.0 - Discount "Download Requirements" conditions toggle with each product added
### Expected behavior
Once a download/product has been added to the "Download Requirements" on the Discounts screens, I expect the product conditions field to show, and to persist.

### Actual behavior
When I add one product to the "Download Requirements", the product conditions select input shows. When I add a second one, the input is toggled to hide. With each new product added, the conditions toggle between show/hide.

1) Create a new discount.
2) Add a product to the "Download Requirements" field. The product conditions will show.
3) Add another product; the conditions input will disappear.

PHP Version: 7.4.1
EDD Version (or branch): release/3.0 (this does not toggle in 2.9.x)
WordPress Version: 5.4.2

It looks like [the script is toggling the conditions input every time the download requirements input changes](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/js/admin/discounts/index.js#L19), not just when the first/only item is added/removed.

LisaCee:
Fixes #8068 

Proposed Changes:
1.  Changed event to `show` rather than `toggle` when items are added to list

I would have liked to have these fields revert back to `hidden` when all items are removed from the list, but I haven't been able to figure that out with jQuery yet.

### DIFF
--- a/assets/js/admin/discounts/index.js
+++ b/assets/js/admin/discounts/index.js
@@ -16,10 +16,6 @@ jQueryReady( () => {
 	 * Show/hide conditions based on input value.
 	 */
 	products.change( function() {
-		if (products.val() !== null) {
-			$( '#edd-discount-product-conditions' ).show( products.val() );
-		} else if (products.val() === null ) {
-			$( '#edd-discount-product-conditions' ).hide( products.val() );
-		}
+		$( '#edd-discount-product-conditions' ).toggle( null !== products.val() );
 	} );
 } );
--- a/assets/js/admin/discounts/index.js
+++ b/assets/js/admin/discounts/index.js
@@ -17,6 +17,6 @@ jQueryReady( () => {
 	 * Show/hide conditions based on input value.
 	 */
 	products.change( function() {
-		$( '#edd-discount-product-conditions' ).toggle( products.val() );
+			$( '#edd-discount-product-conditions' ).show( products.val() );
 	} );
 } );
--- a/assets/js/admin/discounts/index.js
+++ b/assets/js/admin/discounts/index.js
@@ -8,7 +8,6 @@ import { jQueryReady } from 'utils/jquery.js';
  */
 jQueryReady( () => {
 	const products = $( '#edd_products' );
-
 	if ( ! products ) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


robincornett (commented):
Is there a way to set it up to show once the value is not empty, but then hide again if I remove all of the products? The original issue is fixed--the input is no long disappearing on me, but if it could hide if there are no products, I think that would be nice.

assets/js/admin/discounts/index.js
```diff
@@ -17,6 +16,6 @@ jQueryReady( () => {
 	 * Show/hide conditions based on input value.
 	 */
 	products.change( function() {
-		$( '#edd-discount-product-conditions' ).toggle( products.val() );
+		$( '#edd-discount-product-conditions' ).show( products.val() );
```
spencerfinnell, line 19: @LisaCee I believe the original implementation meant to use `.toggle()` via the `display` argument: https://api.jquery.com/toggle/#toggle-display

However `products.val()` is returning `null` or an array here and jQuery seems to consider both truthy in this case. Instead the value should be checked to not be null and that be used to determine the `display` boolean.

spencerfinnell (changes requested):

assets/js/admin/discounts/index.js
```diff
@@ -16,6 +16,10 @@ jQueryReady( () => {
 	 * Show/hide conditions based on input value.
 	 */
 	products.change( function() {
-		$( '#edd-discount-product-conditions' ).show( products.val() );
+		if (products.val() !== null) {
+			$( '#edd-discount-product-conditions' ).show( products.val() );
+		} else if (products.val() === null ) {
```
ashleyfae, line 21: Could this just be `else` ?

spencerfinnell: @LisaCee The check for null can be done inside of `.toggle()`, like:

```js
$( '#edd-discount-product-conditions' ).toggle( null !== products.val() );
```

If condition side of `.toggle()` is `true` then the selector will to show, if it is false it will hide.

spencerfinnell (changes requested):

assets/js/admin/discounts/index.js
LisaCee: Much more elegant.  Thanks

spencerfinnell (approved):

robincornett (approved):
Working nicely for me!

ashleyfae (approved):

PR #8122, opened 2020-09-25
Docblock: $downloads can be int edd_has_user_purchased()

zackkatz:
This is supported functionality, but not reflected in the docBlocks.

Proposed Changes:
1. Fix the docblocks for edd_has_user_purchased()


ashleyfae (approved):
Thank you @zackkatz !

cklosowski:
@ashleyfae This seems like something that can go in the next minor or in 2.10, so we should just be sure and update the base branch before merging.

PR #8127, opened 2020-09-28
Use wp_safe_redirect with EDD's login

### LINKED ISSUE #8087: EDD Login process should use wp_safe_redirect
## Bug Report
Processing of the EDD Login form should use wp_safe_redirect. See conversation here:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8085#discussion_r471810627

robincornett:
Fixes #8087

### DIFF
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -111,11 +111,11 @@ function edd_process_login_form( $data ) {
 			if ( has_filter( 'edd_login_redirect' ) ) {
 				$user_id = $user instanceof WP_User ? $user->ID : false;
 				/**
-				 * Filters the redirect URL for EDD.
+				 * Filters the URL to which users are redirected to after logging in.
 				 *
 				 * @since 1.0
-				 * @param string $default_redirect_url  the URL to which to redirect
-				 * @param int|bool                      user ID or false, if no ID is available
+				 * @param string $default_redirect_url The URL to which to redirect after logging in.
+				 * @param int|false                    User ID. false if no ID is available.
 				 */
 				wp_redirect( apply_filters( 'edd_login_redirect', $default_redirect_url, $user_id ) );
 			} else {
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -108,7 +108,7 @@ function edd_process_login_form( $data ) {
 		$errors = edd_get_errors();
 		if ( ! $errors ) {
 			$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
-			wp_redirect( $redirect );
+			wp_safe_redirect( $redirect );
 			edd_die();
 		}
 	}
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -109,14 +109,15 @@ function edd_process_login_form( $data ) {
 		if ( ! $errors ) {
 			$default_redirect_url = esc_url_raw( $data['edd_redirect'] );
 			if ( has_filter( 'edd_login_redirect' ) ) {
+				$user_id = $user instanceof WP_User ? $user->ID : false;
 				/**
 				 * Filters the redirect URL for EDD.
 				 *
 				 * @since 1.0
 				 * @param string $default_redirect_url  the URL to which to redirect
-				 * @param int                           deprecated as of 2.9.24; used to be user ID
+				 * @param int|bool                      user ID or false, if no ID is available
 				 */
-				wp_redirect( apply_filters( 'edd_login_redirect', $default_redirect_url, 0 ) );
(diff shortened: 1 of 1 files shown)


includes/login-register.php
```diff
@@ -108,7 +108,7 @@ function edd_process_login_form( $data ) {
 		$errors = edd_get_errors();
 		if ( ! $errors ) {
 			$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
-			wp_redirect( $redirect );
+			wp_safe_redirect( $redirect );
```
ashleyfae, line 111: While we're here, let's also do this:

```suggestion
			wp_safe_redirect( esc_url_raw( $redirect ) );
```

ashleyfae (changes requested):

ashleyfae (approved):

ashleyfae (approved):
Confirmed that:

- [x] Without a filter, `wp_safe_redirect()` is used.
- [x] With a filter, `wp_redirect()` is used and an "unsafe" redirect can be set.

includes/login-register.php
```diff
@@ -107,8 +107,12 @@ function edd_process_login_form( $data ) {
 		// Check for errors and redirect if none present.
 		$errors = edd_get_errors();
 		if ( ! $errors ) {
-			$redirect = apply_filters( 'edd_login_redirect', $data['edd_redirect'], 0 );
-			wp_redirect( $redirect );
+			$default_redirect_url = esc_url_raw( $data['edd_redirect'] );
+			if ( has_filter( 'edd_login_redirect' ) ) {
+				wp_redirect( apply_filters( 'edd_login_redirect', $default_redirect_url ) );
```
spencerfinnell, line 112: Probably a good time to add a docblock for this filter: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

spencerfinnell (changes requested):

includes/login-register.php
robincornett: @ashleyfae this filter has actually been in EDD since 1.0, and the `0` we removed used to be a [$user_ID](https://github.com/easydigitaldownloads/easy-digital-downloads/blame/8c3591e346407af71df7d9e857c27b971ed990fc/includes/login-register.php#L98). Should I add it back and note that it's irrelevant? Or can we safely remove the parameter anyway?

ashleyfae: @robincornett I wonder why it was even removed. :thinking: The user ID is still available via the `$user` object above. Maybe we should just pass that in.

includes/login-register.php
```diff
@@ -107,8 +107,19 @@ function edd_process_login_form( $data ) {
...
+			$default_redirect_url = esc_url_raw( $data['edd_redirect'] );
+			if ( has_filter( 'edd_login_redirect' ) ) {
+				/**
+				 * Filters the redirect URL for EDD.
+				 *
+				 * @since 1.0
+				 * @param string $default_redirect_url  the URL to which to redirect
+				 * @param int                           deprecated as of 2.9.24; used to be user ID
+				 */
+				wp_redirect( apply_filters( 'edd_login_redirect', $default_redirect_url, 0 ) );
```
ashleyfae, line 119:
```suggestion
				wp_redirect( apply_filters( 'edd_login_redirect', $default_redirect_url, $user->ID ) );
```

The `$user` object is still available further up so I see no reason we can't add that back in. (Docs will need to be updated further up.)

ashleyfae (changes requested):

includes/login-register.php
```diff
@@ -107,8 +107,20 @@ function edd_process_login_form( $data ) {
...
+			$default_redirect_url = esc_url_raw( $data['edd_redirect'] );
+			if ( has_filter( 'edd_login_redirect' ) ) {
+				$user_id = $user instanceof WP_User ? $user->ID : false;
+				/**
+				 * Filters the redirect URL for EDD.
+				 *
+				 * @since 1.0
+				 * @param string $default_redirect_url  the URL to which to redirect
+				 * @param int|bool                      user ID or false, if no ID is available
+				 */
```
spencerfinnell, line 119:
```suggestion
				/**
				 * Filters the URL to which users are redirected to after logging in.
				 *
				 * @since 1.0
				 * @param string $default_redirect_url The URL to which to redirect after logging in.
				 * @param int|false                    User ID. false if no ID is available.
				 */
```

spencerfinnell (changes requested):

spencerfinnell (approved):

ashleyfae (approved):

PR #8129, opened 2020-09-29
Use payment ID when recounting store earnings

### LINKED ISSUE #8125: 3.0 - EDD_Tools_Recount_Store_Earnings can cause errors
### Expected behavior
I expect the recount tools to all work.

### Actual behavior
In two different sites, when I try to run the "Recount Store Earnings and Stats" tool, I get PHP errors (or a 500 error on the ajax request) and the tool fails to run in its entirety.

Note: so far I am the only one encountering this, but I think there is an actual issue, namely that this tool uses `edd_get_payment_amount()` incorrectly.

1) In `release/3.0`, visit Downloads > Tools > General.
2) Select "Recount Store Earnings and Stats" and click Submit.
3) If you are lucky, or me, you will get errors and the job will not complete.

PHP Version: 7.4.1
EDD Version (or branch): 3.0
WordPress Version: 5.5.1

This tool passes a payment object to the `edd_get_payment_amount`, but that function expects an integer (ID). That being the case, I am not sure why the tool is failing for me only, but changing the tool to pass the payment ID to the function resolves the issue and generates correct data, or closer to correct data. Scanning our code, all other instances of `edd_get_payment_amount` are passing a payment ID, not an object.

robincornett:
Fixes #8125

Proposed Changes:
1. Use `$payment->ID` instead of entire `$payment` object for retrieving payment amount.


ashleyfae:
I'd kind of prefer to redo this whole method to use our new `edd_get_orders()` function (thus bypass backwards compatibility code that would be running with `edd_get_payments()`), then also update `edd_get_payment_amount()` to accept an `EDD\Orders\Order` object. It seems silly to me that we already have the object, but then have to fetch it again.

robincornett:
PR updated accordingly--uses `edd_get_orders()` and `edd_get_gross_order_statuses()`, which is what we used in updating the customers recount tool.

ashleyfae:
@robincornett This looks good, but I believe the `get_percentage_complete()` function also needs updating.

1. We should use `edd_count_orders()` instead of `edd_count_payments()`.
2. We'll want to make sure to use the same arguments we're using for `edd_get_orders()` (minus the number/offset stuff, of course).

ashleyfae (approved):
This fundamentally works and fixes the original issue so I'm approving it. However, I think improvements can be made, but I will open a new issue to address that.

PR #8135, opened 2020-09-30
Issue/8131 - Call get_item() directly instead of using related _by() functions.

### LINKED ISSUE #8131: 3.0 - Update edd_get_ functions to access Berlin methods directly
Based on a discussion with @JJJ, we are suggesting that `edd_get_` functions, which currently call `edd_get_*_by`, be updated to access and use Berlin's `get_item` method directly.

### Justification or use case
John started looking into this based on #8125, where I discovered that the store recount tool was sending a payment object, rather than an ID, to the `edd_get_payment_amount` function. The doc block for that function says that the parameter should be an integer, not an object, but it doesn't check for the parameter type, so an object can be passed to it without an obvious failure.

`edd_get_payment_amount` in turn calls `edd_get_order`, which also anticipates an ID, and in turn calls `edd_get_order_by`.

`edd_get_order_by` initializes `EDD\Database\Queries\Order()` and finally uses that.

John's suggestion is to allow `edd_get_order` to access the Berlin method directly, removing the last layer of abstraction, to something like this:

```php
**
 * Get an order by ID.
 *
 * @since 3.0
 *
 * @param int $order_id Order ID.
 * @return EDD\Orders\Order|false Order object if successful, false otherwise.
 */
function edd_get_order( $order_id = 0 ) {
	$orders = new EDD\Database\Queries\Order();
	// Return order
	return $orders->get_item( $order_id );
}
```

He says that with this change, the `$order_id` variable no longer has to be an integer, but could also be an array or object, because "`get_item()` internally shapes the ID from an object or an array if need be."

There are 14 `edd_get_` functions which could be updated similarly.

JJJ:
This commit updates all of the "edd_get_" functions (and "edd_fetch_" where necessary) to call get_item() directly, rather than relying on a related global function _by() call.

This allows those functions to use the lower-level shape_item_id() call inside BerlinDB, so that they may accept int/string/array/object types of values.

This makes these functions work more like "get_post()" in that no matter what type of variable is passed in, the lower level API will just figure-it-out rather than throw an error because of an unexpected type of value was passed in.

This is marginally more developer friendly for folks using these functions, as well as marginally easier to understand for the folks maintaining EDD into the future.

Fixes #8131.


robincornett (commented):
If all of these functions now accept more than just an `int`, should the variables and doc blocks also be updated to reflect this?

includes/customer-functions.php
```diff
@@ -807,12 +813,6 @@ function edd_add_customer_email_address( $data ) {
  *                   false on error.
  */
 function edd_delete_customer_email_address( $customer_email_address_id = 0 ) {
```
ashleyfae, line 815:
```suggestion
function edd_delete_customer_email_address( $customer_email_address_id ) {
```

Since a customer ID is required, let's remove the `0` default.

includes/customer-functions.php
```diff
@@ -842,12 +842,6 @@ function edd_delete_customer_email_address( $customer_email_address_id = 0 ) {
  * @return int|false Number of rows updated if successful, false otherwise.
  */
 function edd_update_customer_email_address( $customer_email_address_id = 0, $data = array() ) {
```
ashleyfae, line 844:
```suggestion
function edd_update_customer_email_address( $customer_email_address_id, $data = array() ) {
```

ID is required.

includes/customer-functions.php
```diff
@@ -863,13 +857,10 @@ function edd_update_customer_email_address( $customer_email_address_id = 0, $dat
  *                                                     successful, false otherwise.
  */
 function edd_get_customer_email_address( $customer_email_address_id = 0 ) {
```
ashleyfae, line 859:
```suggestion
function edd_get_customer_email_address( $customer_email_address_id ) {
```

ID is required.

ashleyfae (changes requested):

JJJ:
> If all of these functions now accept more than just an `int`, should the variables and doc blocks also be updated to reflect this?

That sounds like a welcome change.

Do you think `int|array|object` is adequate? Some of the more specific literal object names start to get long if we include their namespaces.

```
EDD\Database\Row\Customer_Email_Address
```

It's also not immediately obvious with the docs that the `array` type needs to contain an `id` key, and the `object` type would need to contain an `id` variable. That starts to be a lot to explain across each of these functions.

Thoughts?

robincornett:
I know Ashley prefers to include the specific object when it's known--it may be long but with the abstractions, I prefer having more specific information when I can. I think adding the id requirements into the general comment above the `@param` line would be reasonable, as those are often more verbose. Just adding a sentence to the current description, which is pretty terse:

> Gets a[n adjustment] by ID, array, or object. IDs are numeric; arrays must contain an `id` key; objects must include an `id` variable.

ashleyfae:
> > If all of these functions now accept more than just an `int`, should the variables and doc blocks also be updated to reflect this?
> 
> That sounds like a welcome change.
> 
> Do you think `int|array|object` is adequate? Some of the more specific literal object names start to get long if we include their namespaces.
> 
> ```
> EDD\Database\Row\Customer_Email_Address
> ```

I think you now know I prefer full class names. :D We can import the namespaces to reduce the length, then just use `Customer_Email_Address`, etc.

> It's also not immediately obvious with the docs that the `array` type needs to contain an `id` key, and the `object` type would need to contain an `id` variable. That starts to be a lot to explain across each of these functions.

I'm fine with either:

1. Including `array` with no other explanation.
2. Not including `array` at all, because although it is valid, it does require explaining and frankly this seems like a not-super-common use case to me anyway. :shrug:

ashleyfae (approved):
Going to merge this for now. We can update any docs separately.

PR #8136, opened 2020-09-30
Issue/8132 - Database: update BerlinDB to latest 1.x nightly.

### LINKED ISSUE #8132: 3.0 - Update to lastest BerlinDB
Related to #8131, @JJJ has updated [BerlinDB](https://github.com/berlindb/core/pull/72) and EDD's BerlinDB files should be updated as well.

### Justification or use case
BerlinDB's `get_item` method has been updated to be more flexible and developer friendly.

JJJ:
This commit updates all of the BerlinDB classes to their latest versions, which includes approximately 12 months worth of updates in some instances.

It also includes 3 new classes for specific types of Queries: Date, Meta, and Compare. These Queries are necessary abstractions away from what is built into WordPress to remove several hard-coded assumptions about how they're used.

It also changes all of the @ since tags in the PHPDoc, as these are now part of BerlinDB and not derived from EDD any longer. This will make merging future versions of Berlin have smaller diffs now that it is its own project.

Fixes #8132.


robincornett (approved):

robincornett:
Note: this all appears to be working for me locally, but I see Travis doesn't agree. Also, the Recount Store Earnings tool still fails (#8125) for me with this PR, but it's a silent failure--I know this was not supposed to fix that, just noting as that's where the discussion started.

includes/database/engine/class-column.php
```diff
@@ -767,19 +723,24 @@ private function sanitize_validation( $callback = '' ) {
...
+	 * @since 1.0.0
+	 * @param string $value Default '0000-00-00 00:00:00'. A datetime value that needs validating
 	 * @return string A valid datetime value
 	 */
 	public function validate_datetime( $value = '0000-00-00 00:00:00' ) {
 
-		// Fallback for empty values
+		// Handle "empty" values
 		if ( empty( $value ) || ( '0000-00-00 00:00:00' === $value ) ) {
-			$value = ! empty( $this->default ) || ( is_null( $this->default ) && $this->allow_null )
```
ashleyfae, line 778: I think we want to keep the null check.

includes/database/engine/class-column.php
```diff
@@ -767,19 +723,24 @@ private function sanitize_validation( $callback = '' ) {
...
+			$value = ! empty( $this->default )
 				? $this->default
-				: '';
+				: '0000-00-00 00:00:00';
```
ashleyfae, line 741: We might want to keep this as an emptys tring.

includes/database/engine/class-query.php
```diff
@@ -599,120 +626,131 @@ protected function set_found_items( $item_ids = array() ) {
...
-		// Remove all valid columns filters
-		remove_filter( 'date_query_valid_columns', array( $this, '__filter_valid_date_columns' ), 2 );
+	/**
+	 * Pass-through method to return a new Meta object.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param array $args See Queries\Meta
+	 *
+	 * @return Meta
```
ashleyfae, line 668:
```suggestion
	 * @return Queries\Meta
```

includes/database/engine/class-query.php
```diff
@@ -599,120 +626,131 @@ protected function set_found_items( $item_ids = array() ) {
...
-		// Return the date
-		return $date_query;
+	/**
+	 * Pass-through method to return a new Compare object.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param array $args See Queries\Compare
+	 *
+	 * @return Compare
```
ashleyfae, line 681:
```suggestion
	 * @return Queries\Compare
```

includes/database/engine/class-query.php
```diff
@@ -1451,10 +1507,10 @@ protected function shape_items( $items = array() ) {
 		 *
 		 * Looks like `edd_get_customers`
 		 *
-		 * @since 3.0
+		 * @since 1.0.0
 		 *
 		 * @param array  $retval An array of items.
-		 * @param Query &$this  Current instance of Query, passed by reference.
+		 * @param object &$this  Current instance of Query, passed by reference.
```
ashleyfae, line 1513:
```suggestion
		 * @param Query &$this  Current instance of Query, passed by reference.
```

I prefer to use explicit class names instead of generic `object`.

includes/database/engine/class-query.php
```diff
@@ -1532,33 +1596,44 @@ protected function shape_item_id( $item = 0 ) {
...
+	 * preferably with a cache group for that column. See: get_item().
 	 *
-	 * @param string $column_name  Name of database column
-	 * @param string $column_value Value to query for
-	 * @return object|false False if empty/error, Object if successful
+	 * @since 1.0.0
+	 *
+	 * @param string     $column_name  Name of database column
+	 * @param int|string $column_value Value to query for
+	 * @return mixed False if empty/error, Object if successful
```
ashleyfae, line 1636:
```suggestion
	 * @return object|false False if empty/error, Object if successful
```

includes/database/engine/class-query.php
```diff
@@ -1883,23 +1968,29 @@ protected function shape_item( $item = 0 ) {
 	/**
 	 * Validate an item before it is updated in or added to the database.
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 *
 	 * @param array $item
-	 * @return array|false False on error, Array of validated values on success
+	 * @return mixed False on error, Array of validated values on success
```
ashleyfae, line 1974:
```suggestion
	 * @return array|false False on error, Array of validated values on success
```

includes/database/engine/class-query.php
```diff
@@ -1934,14 +2030,14 @@ protected function validate_item( $item = array() ) {
 	 * any type, and also resets the key values. It looks weird, but is
 	 * currently by design to protect the integrity of the return value.
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 *
 	 * @param string $method select|insert|update|delete
-	 * @param array|object  $item   Object|Array of keys/values to reduce
+	 * @param mixed  $item   Object|Array of keys/values to reduce
```
ashleyfae, line 2036:
```suggestion
	 * @param array|object  $item   Object|Array of keys/values to reduce
```

includes/database/engine/class-query.php
```diff
@@ -1934,14 +2030,14 @@ protected function validate_item( $item = array() ) {
...
 	 *
-	 * @return array|object Object|Array without keys the current user does not have caps for
+	 * @return mixed Object|Array without keys the current user does not have caps for
```
ashleyfae, line 2038:
```suggestion
	 * @return array|object Object|Array without keys the current user does not have caps for
```

includes/database/engine/class-query.php
```diff
@@ -2078,13 +2175,13 @@ protected function transition_item( $new_data = array(), $old_data = array() ) {
...
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 *
 	 * @param int    $item_id
 	 * @param string $meta_key
 	 * @param string $meta_value
 	 * @param string $unique
-	 * @return int|false The meta ID on success, false on failure.
+	 * @return mixed
```
ashleyfae, line 2184:
```suggestion
	 * @return int|false
```

includes/database/engine/class-query.php
```diff
@@ -2109,12 +2206,12 @@ protected function add_item_meta( $item_id = 0, $meta_key = '', $meta_value = ''
...
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 *
 	 * @param int     $item_id
 	 * @param string  $meta_key
-	 * @param boolean $single
-	 * @return mixed Single metadata value, or array of values
+	 * @param bool    $single
+	 * @return mixed
```
ashleyfae, line 2214:
```suggestion
	 * @return mixed Single metadata value, or array of values
```

includes/database/engine/class-query.php
```diff
@@ -2171,13 +2267,13 @@ protected function update_item_meta( $item_id = 0, $meta_key = '', $meta_value =
...
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 *
 	 * @param int    $item_id
 	 * @param string $meta_key
 	 * @param string $meta_value
 	 * @param string $delete_all
-	 * @return bool True on successful delete, false on failure.
+	 * @return mixed
```
ashleyfae, line 2276:
```suggestion
	 * @return bool True on successful delete, false on failure.
```

includes/database/engine/class-query.php
```diff
@@ -2202,46 +2298,29 @@ protected function delete_item_meta( $item_id = 0, $meta_key = '', $meta_value =
...
 
 		// Get the object type
 		$object_type = $this->apply_prefix( $this->item_name );
 
-		// Use function if it exists
-		if ( function_exists( 'get_registered_meta_keys' ) ) {
-			return get_registered_meta_keys( $object_type, $object_subtype );
-		}
-
-		// This code is for versions of WordPress older than 4.6.0 and can be
```
ashleyfae, line 2225: Just double checking.. I think EDD3 requires 4.6+ so removing this is okay, yes?

includes/database/engine/class-table.php
```diff
@@ -2,133 +2,146 @@
...
 
 	/**
-	 * @var string Passed directly into register_activation_hook()
+	 * Passed directly into register_activation_hook()
+	 *
+	 * @since 1.0.0
+	 * @var   string
 	 */
-	protected $file = EDD_PLUGIN_FILE;
+	protected $file = '';
```
ashleyfae, line 73: Do we not still need `EDD_PLUGIN_FILE` here?

includes/database/engine/class-table.php
```diff
@@ -337,13 +350,13 @@ public function exists() {
...
 
 	/**
-	 * Check if table already exists
+	 * Create the table.
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 *
-	 * @return bool|void
+	 * @return bool
```
ashleyfae, line 357:
```suggestion
	 * @return bool|void
```

due to:

```php
// Bail if no database interface is available
if ( empty( $db ) ) {
	return;
}
```

includes/database/engine/class-table.php
```diff
@@ -720,50 +836,31 @@ private function set_db_version( $version = '' ) {
...
-		if ( version_compare( $this->db_version, $this->version, '<=' ) || 12 !== strlen( $this->db_version ) ) {
-			return;
-		}
-
-		// Parse the new version number from the existing. Converting from {YYYY}{mm}{dd}{xxxx} to {YYYY}{mm}{dd}{x}
-		$date             = substr( $this->db_version, 0, 8 );
-		$increment        = substr( $this->db_version, 8, 4 );
-		$this->db_version = intval( $date . intval( $increment ) ); // Trims off the three prefixed zeros.
-
-		$this->set_db_version( $this->db_version );
```
ashleyfae, line 749: This needs to be added back in.

includes/database/engine/class-table.php
```diff
@@ -720,50 +836,31 @@ private function set_db_version( $version = '' ) {
...
+	 * Delete the table version from the database.
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 */
 	private function delete_db_version() {
 		$this->db_version = $this->is_global()
-			? delete_network_option( get_main_network_id(), $this->db_version_key )
-			:         delete_option(                        $this->db_version_key );
+			? delete_network_option( get_main_network_id(), $this->db_version_key, false )
```
ashleyfae, line 856:
```suggestion
			? delete_network_option( get_main_network_id(), $this->db_version_key )
```

Incorrect number of parameters.

includes/database/engine/class-table.php
```diff
@@ -720,50 +836,31 @@ private function set_db_version( $version = '' ) {
...
+			:         delete_option(                        $this->db_version_key, false );
```
ashleyfae, line 857:
```suggestion
			:         delete_option(                        $this->db_version_key );
```

Incorrect number of parameters.

ashleyfae (changes requested):
Need some changes made and unit tests need looking at.

JJJ:
@ashleyfae I'll push your feedback upstream, and update this PR as needed. 👍

includes/database/engine/class-table.php
JJJ: We do. I missed this in the merge.

includes/database/engine/class-table.php
JJJ: Now that you point it out, I don't like that they return null/void. They could return `false` which is both more predictable and results in simpler documentation.

includes/database/engine/class-query.php
JJJ: Yup! Currently requires 4.7.0. 👍

includes/database/engine/class-table.php
JJJ: OK 👍

includes/database/engine/class-table.php
JJJ: I did more digging into `$file` usage in Berlin, and it appears like this variable can be removed completely.

1. Activation hooks are an unreliable way to perform database upgrades
2. In this instance `EDD_PLUGIN_FILE` does not have a `maybe_upgrade()` method to call anyways

ashleyfae: Alright, fine by me.

includes/database/engine/class-table.php
ashleyfae: Yes that makes more sense.

includes/database/engine/class-column.php
```diff
@@ -536,7 +505,7 @@ private function validate_args( $args = array() ) {
 			'zerofill'      => 'wp_validate_boolean',
 			'binary'        => 'wp_validate_boolean',
 			'allow_null'    => 'wp_validate_boolean',
-			'default'       => array( $this, 'sanitize_default' ),
```
ashleyfae, line 539: We need this back in order for `null` values to be saved properly. Right now they're getting converted to empty strings and that's breaking a bunch of the unit tests.

ashleyfae (changes requested):

includes/database/engine/class-column.php
JJJ: Added this back.

ashleyfae:
@JJJ I've pushed up some changes that fix the remaining failing unit tests. They were all related to the same thing:

`Tests_API::test_get_customers` - The arguments for this query ended up being:

```
array(3) {
  'number' =>
  int(10)
  'offset' =>
  int(0)
  'email' =>
  NULL
}
```

In the previous version of BerlinDB, I believe that `'email' => null` would have been filtered out. But the new version fixed explicitly checking null values, which isn't actually what we want to happen here. We want to get all emails in this case; not those with email = null.

`Tests_Process_Download::test_order_with_complete_item_should_return_true` - We have a default argument in the `edd_order_grants_access_to_download_files()` function:

`'price_id' => ''`

Again, in the previous version this was being filtered out and not being used in the query. In this new version it was converting that empty string to `0` and checking for `price_id = 0`, which isn't what we want.

tl;dr - just some of our "blank" default values getting pushed into the query, which wasn't our desired behaviour.

---

Also found another issue that I'll leave a separate note on.

includes/database/engine/class-query.php
```diff
@@ -50,68 +54,66 @@ class Query extends Base {
...
 	/**
-	 * Name of class used to setup the database schema
+	 * Name of class used to setup the database schema.
 	 *
-	 * @since 3.0
+	 * @since 1.0.0
 	 * @var   string
 	 */
-	protected $table_schema = '\\EDD\\Database\\Schemas\\Base';
+	protected $table_schema = __NAMESPACE__ . '\\Schema';
```
ashleyfae, line 80: For this and `$item_shape` I'm seeing this error:

> expression is not allowed as field default value

ashleyfae (changes requested):

includes/database/engine/class-query.php
JJJ: Aww. That's too bad. I've reverted these and pushed them up here and in Berlin.

I wonder if these errors are related to specific PHP versions. I'm not seeing them anywhere on `7.4.10`.

ashleyfae: Not sure. PhpStorm was warning me about it. Usually if it's version-specific it says so, but this time it didn't say anything. :woman_shrugging:

ashleyfae:
This is looking good to me now - thanks @JJJ ! Just pending unit tests being fixed. :crossed_fingers:

ashleyfae (approved):

Issue #8124, opened 2020-09-25
3.0 - Update the Import/Export Tools styling


robincornett:
### Explain your enhancement (please be detailed)
The Downloads > Reports > Export screen looks amazing after #8094 has been addressed. I've been working on a related issue and when I look at the Downloads > Tools > Import/Export screen, I think it would benefit from a similar treatment.

### Justification or use case
The reports exports look so much more polished now. This screen could as well.

Note: extensions can add tools to this screen. Software Licensing has one (and a request for a second).

robincornett:
I don't know if this would fall under this umbrella, but I'm almost wondering about updating the Tools output to use the vertical tabs/navigation, similar to Reports? It may be way out there to consider, and no idea how easy it might be to implement, if at all, but it would make the two screens more unified in appearance, and it is a pretty setup.

ashleyfae:
> I don't know if this would fall under this umbrella, but I'm almost wondering about updating the Tools output to use the vertical tabs/navigation, similar to Reports? It may be way out there to consider, and no idea how easy it might be to implement, if at all, but it would make the two screens more unified in appearance, and it is a pretty setup.

I'm open to that, but don't think we need to do it for 3.0.

LisaCee:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8106#issuecomment-700771835 about creating multipurpose class for grid container.

PR #8138, opened 2020-09-30
Started converting spaces to tabs, escaping #8124

LisaCee:
Fixes #

Proposed Changes:
1.
2.
3.

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -69,6 +69,7 @@ function edd_tools_page() {
 		</nav>
 
 		<div class="metabox-holder">
+
 			<?php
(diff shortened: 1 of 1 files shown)


robincornett:
One hiccup I've just run into that I didn't think of with the Import tools is that when you do the initial click to upload the file you want to import, it creates/outputs a table of the columns in the CSV to map, which ends up causing some mayhem with the screen:

image

I'm wondering two things:
1. would it be possible, possibly, to output the preview table into a modal/thickbox so that it is outside of the grid? This would have to work with extensions as well as what's in EDD Core.
2. or does this mean we cannot grid the import/export tools after all?

PR #8140, opened 2020-10-01
Check if total_spent is set before using #8128

### LINKED ISSUE #8128: Undefined index: total_spent when no Customer exists
### Expected behavior
to not have any notices in the error logs

### Actual behavior
Customer reports:
Since your last plugin update the bug below shows up in your logs. Could you have a look at this bug:
[29-Sep-2020 10:28:31 UTC] PHP Notice: Undefined index: total_spent in /home/_redacted_/public_html/wp-content/plugins/easy-digital-downloads/includes/user-functions.php on line 353

I was able to replicate this via the Debug Bar console by running the code below but i deleted the EDD Customer record before running it.
```
<?php
	$stats = edd_get_purchase_stats_by_user( 2 );

	print $stats['total_spent'];
```

PHP Version:

EDD Version (or branch): 2.9.25

WordPress Version: latest

perhaps returning an empty array but with the indexes set to '' by default if no Customer exists here https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/user-functions.php#L316 ?

ashleyfae:
Fixes #8128 

Proposed Changes:
1. Check if array key exists before using.

### DIFF
--- a/includes/user-functions.php
+++ b/includes/user-functions.php
@@ -350,7 +350,7 @@ function edd_purchase_total_of_user( $user = null ) {
 
 	$stats = edd_get_purchase_stats_by_user( $user );
 
-	return $stats['total_spent'];
+	return isset( $stats['total_spent'] ) ? $stats['total_spent'] : 0.00;
 }
 
 /**


robincornett (approved):

mihaijoldis:
@ashleyfae looks good here!

mihaijoldis (approved):

PR #8143, opened 2020-10-05
Add customer name column to orders export

### LINKED ISSUE #8142: 3.0 - Export Orders can be missing customer information
### Expected behavior
When I export a CSV of my store's orders, I expect to have all order information in the CSV.

### Actual behavior
If the order has no physical address, no customer name is exported as part of the order.

1) In 3.0, generate some orders via CLI, or otherwise create orders which are associated with a customer, but without requiring an address, so no address is saved as part of the order.
2) Visit Downloads > Reports > Export.
3) Export Orders.
4) View the CSV. There will be no first/last name for those orders.

PHP Version: 7.4.1
WordPress Version: 5.5.1

Discovered in the process of researching #8137, wondering why the import had no customer names. Finally checked the CSV and realized that there were no names for the orders (all generated via CLI). The export tool is trying to populate the first/last name columns from the order address, but if there is no address, then there will be no first/last name.

I think we should be using `edd_get_customer( $order->customer_id )` instead to get the customer name for the order, as we do elsewhere (the orders list table is one example), but the customer object combines the first/last name:

```
[02-Oct-2020 14:26:08 UTC] EDD_Customer Object
(
    [id] => 1
    [purchase_count] => 1
    [purchase_value] => 51
    [email] => Tobi.Lenahan@example.invalid
    [emails:protected] => 
    [name] => Tobi Lenahan
    [status] => active
    [date_created] => 2020-10-02 14:19:24
    [payment_ids:protected] => 
    [user_id] => 0
    [notes:protected] => 
    [prefix:protected] => edd
    [last_error:protected] => 
)
```

So then the question is if the export CSV should be updated to merge the first/last names into one column. It should be possible to update the import tool to handle either separate columns or one together.

robincornett:
Fixes #8142

Proposed Changes:
1. Create a "Name" column for the export CSV.
2. Get the customer for the order and use that name for the new column.
3. Remove the first/last name columns.


includes/admin/reporting/export/class-batch-export-payments.php
```diff
@@ -195,8 +195,7 @@ public function get_data() {
 				'seq_id'       => $order->get_number(),
 				'email'        => $order->email,
 				'customer_id'  => $order->customer_id,
-				'first'        => isset( $address->first_name ) ? $address->first_name : '',
-				'last'         => isset( $address->last_name ) ? $address->last_name : '',
+				'name'         => $customer->name,
```
ashleyfae, line 198:
```suggestion
				'name'         => ! empty( $customer->name ) ? $customer->name : '',
```

In case `$customer` didn't actually have a result.

ashleyfae (changes requested):

ashleyfae (approved):

PR #8146, opened 2020-10-07
Update EDD_Customer get_address

### LINKED ISSUE #8039: 3.0 - get_address method needs to be updated
### Expected behavior
If a logged in user has a stored address, it should be retrieved as part of the checkout process, and `edd_get_customer_address` should work.

### Actual behavior
`EDD_Customer->get_address()` uses `$type = 'primary'` as its default query, so no address is returned, since `billing` is now the default address type, and `is_primary` is the new "primary" address value.

Additionally, when checking out for the first time, an address record is created, but is not assigned as the customer's primary address.

1) On a storefront which collects addresses, place an order as a new customer, filling out address data, and creating a new account so that the user will be logged in.
2) Look in the `edd_customer_addresses` table for the `type` and `is_primary` values.
3) Place a second order as this logged in customer. On the checkout screen, the address is not retrieved.

PHP Version: 7.4.1
WordPress Version: 5.4.2

Looking at the `EDD_Customer->get_address()` method, I'm thinking that code which uses it needs to be checked and updated to account for the possible returned values (it can return either an `object`, if the type is set correctly, or an `array` of objects). The [checkout template](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/checkout/template.php#L324) is one example--it also does not pass a user ID in, so no address will be retrieved there.

Related: #7254, but creating this as a new issue due to discussion and code/database changes.

robincornett:
Fixes #8039

Proposed Changes:
1. Updates the EDD_Customer `get_address` method to use `is_primary` instead of `type`.
2. Updates customer actions which use the `get_address` method.
3. Updates EDD's user meta compatibility functions to make the user meta address compatible with the new customer address.


includes/class-edd-customer.php
```diff
@@ -1082,29 +1082,27 @@ public function get_emails() {
 	 *
 	 * @since 3.0
 	 *
-	 * @param string $type Address type. Default primary.
+	 * @param boolean $is_primary Whether the address is the primary address. Default true.
 	 *
 	 * @return array|\EDD\Customers\Customer_Address Object if primary address requested, array otherwise.
```
ashleyfae, line 1087:
```suggestion
	 * @return array|\EDD\Customers\Customer_Address|null Object if primary address requested, array otherwise. Null if no result for primary address.
```

includes/compat/class-customer.php
```diff
@@ -274,28 +276,34 @@ public function update_user_meta( $check, $object_id, $meta_key, $meta_value, $p
...
+				)
+			);
+		} else {
+			$customer = edd_get_customer_by( 'user_id', absint( $object_id ) );
+
+			if ( $customer ) {
+				edd_add_customer_address(
+					array(
+						'customer_id' => $customer->id,
+						'type'        => 'primary',
```
ashleyfae, line 297:
```suggestion
```

(Removing it will have it use the default, `billing`)

ashleyfae (changes requested):
The following other updates are also needed to remove/update `type` parameter usage:

## `class-edd-customer.php`

On line 286, this should also be updated to remove `type` : 

```php
// Add the primary email address for this customer
edd_add_customer_email_address( array(
	'customer_id' => $customer_id,
	'email'       => $args['email'],
	'type'        => 'primary'
) );
```

Same on line 387 / 395, and on line 511.

## `class-data-migrator.php`

Around line 617 / 623.

spencerfinnell:
@robincornett A new address is being added for a new Customer but it's not being set as primary by default, causing the next purchase to not be prefilled with the address data. I tried with Braintree and Stripe.

robincornett:
@ashleyfae I think the code you're referencing in `class-edd-customer.php` and `class-data-migrator.php` is for email addresses, which do still use the `type` parameter, instead of `is_primary`. But I fixed the other bits.

@spencerfinnell yes, I haven't made any changes yet to how we update addresses in the database, so right now this only works if there is a "primary" physical address. I'm reading back through the comments in the issue to make sure I understand how we want to proceed if a customer's address on an order is different than what we had previously. I think, if I'm reading this correctly, it should be:

- on the first order, the physical address should be set to primary, because it's the only one
- on subsequent orders, if the address is the same, it should be left alone
- on subsequent orders, if the customer enters a new address, that should be set to primary, and the previous primary one should have `is_primary` set to `false`
- on data migration, all physical addresses should be `false` for `is_primary` except for the one associated with the most recent order.
- same for importing orders/customers?

Does this sound correct and am I missing places where this should be checked/updated?

robincornett:
I looked at this again and the `edd_maybe_update_customer_primary_address` function was not properly checking for the primary address or adding it. So I think the issue of the primary address not being created on the first order (at least in the front end) is now resolved, but what happens on subsequent orders is still a question.

In my current storefront, when I view a customer with multiple orders/addresses, I see all of their addresses. I tried adding `edd_maybe_update_customer_primary_address` into `edd_build_order` (after `edd_maybe_add_customer_address`), which is maybe incorrect, but it did update the customer's primary address. However, the previous address information was just gone, which is maybe accurate, but not what I was expecting.

Is the goal to _replace_ (by updating) the existing primary address (the previous one would presumably still be associated with the order), or to create a new one and make it primary, and convert `is_primary` for the original address to false? I feel like replacing would change some behaviors, such as the address selector in the manual orders workflow.

ashleyfae:
> it should be:
> 
> * on the first order, the physical address should be set to primary, because it's the only one
> * on subsequent orders, if the address is the same, it should be left alone
> * on subsequent orders, if the customer enters a new address, that should be set to primary, and the previous primary one should have `is_primary` set to `false`
> * on data migration, all physical addresses should be `false` for `is_primary` except for the one associated with the most recent order.
> * same for importing orders/customers?
> 
> Does this sound correct and am I missing places where this should be checked/updated?

Yes that sounds right to me.



> I guess I am also not sure if _setting_ the primary address should be handled in a separate issue--although I mentioned it when creating this issue, the issue was primarily about _getting_. So I can go either way--should that be a separate issue or should I handle it here

I'm fine with handling it here.



> Is the goal to _replace_ (by updating) the existing primary address (the previous one would presumably still be associated with the order), or to create a new one and make it primary, and convert `is_primary` for the original address to false? I feel like replacing would change some behaviors, such as the address selector in the manual orders workflow.

Let's create a new one. I think there's still a lot of work we can do to improve this whole address flow and management, but that would be for something after 3.0. For 3.0 I think the bare minimum goal would be:

- A customer always has one primary address.
- The primary address is always the most recently used address.
- If a customer adds a different address at checkout, we never _delete_ any information; always create a new record.

In the future we'll look at proper address management on the customer-facing side, including allowing the _customer_ to choose a primary address. In that end result I imagine we would never automatically set `is_primary` (except on first address creation), but rather let the customer dictate that.

robincornett:
@ashleyfae I think I'm ready for you to look at this again, but it's raising an issue with the Stripe extension. With this current update, the `edd_maybe_update_customer_primary_address` function is not working properly, but I don't think it can be, if the idea is that we should not be overwriting the address, but creating a new one and changing the `is_primary` key instead.

So if it is necessary to keep that function (I've removed use of it in EDD itself, but it currently is being used in Stripe 2.8.0), I would suggest updating it to this:

```php
function edd_maybe_update_customer_primary_address( $customer_id = 0, $data = array() ) {

	// Bail if nothing passed.
	if ( empty( $customer_id ) || empty( $data ) ) {
		return false;
	}

	return edd_maybe_add_customer_address( $customer_id, $data );
}
```

because at this point, we are always considering adding a physical address, and whatever is newest is automatically set to the primary one. So if we get to the point in `edd_maybe_add_customer_address` where we _are_ adding an address, it should be set to primary, and the previously retrieved one should be reset to not primary.

This makes the `edd_maybe_update_customer_primary_address` function unnecessary, but if it ships in Stripe 2.8.0, we maybe cannot remove it (opening a related issue there).

includes/customer-functions.php
```diff
@@ -655,7 +655,21 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
...
+		array(
+			'fields'      => 'ids',
+			'customer_id' => $customer_id,
+			'is_primary'  => true,
+			'number'      => 1,
+		)
+	);
+
+	// Primary address exists, so update it.
+	if ( ! empty( $previous_primary_address ) ) {
```
ashleyfae: Let's explicitly check if `$previous_primary_address[0]` is set since we use it on the next line.

ashleyfae (changes requested):

robincornett:
Update: have reworked both the `edd_maybe_add_customer_address` and `edd_maybe_update_customer_primary_address` functions:
- when adding a new address, if the address already exists, but isn't the primary address, it should be updated to the primary address (if its `type` is `billing`).
- when adding a new address, if the address doesn't exist, it should be made the primary address and any previous primary address should be unset from primary.

The updated `edd_maybe_update_customer_primary_address` function now expects an object (`\EDD\Customers\Customer_Address`), which is what I want, but maybe needs to be updated to allow for an array to have been passed instead, or maybe should be reset to always expect an array, and get the addresses from that. Pushing for discussion.

includes/class-edd-customer.php
```diff
@@ -1082,29 +1082,27 @@ public function get_emails() {
...
-				'customer_id' => $this->id,
-				'type'        => $type,
-			) );
+	public function get_address( $is_primary = true ) {
+		$args = array(
+			'customer_id' => $this->id,
+			'is_primary'  => $is_primary,
+		);
+		if ( $is_primary ) {
+			$args['number'] = 1;
```
ashleyfae, line 1095:
```suggestion
			$args['number']  = 1;
			$args['orderby'] = 'date_created';
			$args['order']   = 'desc';
```

This way if a customer does somehow end up with more than one primary address we will always return the most recently created primary.

ashleyfae (changes requested):
I don't love the whole concept of the `edd_maybe_update_customer_primary_address()` function. This would seem a bit cleaner to me:

1. Do not manually call `edd_maybe_update_customer_primary_address`.
2. Instead hook into an action that runs when new addresses are created. (BerlinDB might have one built in, would have to double check.)
3. Whenever an address is created with `is_primary` set to `true`, do a query similar to what `edd_maybe_update_customer_primary_address()` is doing, but with `id__not_in` and pass in the ID of the new address. (To find any addresses for that customer that are primary, but are not the address that was just added). If there are any results, change `is_primary` to `false`.

The takeaways being:

1. We don't manually call a function at all; it just runs via an action hook.
2. There's no updating the new address after it's been created to set `is_primary`; we only ever update the old address(es) we need to change.

Also, related: I wonder if we should modify this index on the `wp_edd_customer_addresses` table:

```mysql
KEY customer (customer_id)
```

to:

```mysql
KEY customer_is_primary (customer_id, is_primary)
```

That will improve all our queries that search for a customer's primary address.

includes/database/queries/class-customer-address.php
```diff
@@ -160,5 +160,35 @@ class Customer_Address extends Query {
 	 */
 	public function __construct( $query = array() ) {
 		parent::__construct( $query );
+		add_action( 'edd_transition_customer_address_type', array( $this, 'demote_primary_addresses' ), 10, 3 );
```
ashleyfae, line 163: Can we move this to `includes/customers/customer-actions.php` instead? (Doesn't yet exist so we'll need to create a new file)

Also, separately, perhaps it would be better to transition on a different column instead. If you edit `includes/database/schemas/class-customer-addresses.php` and find the array for `is_primary`, you can add:

```php
'transition' => true
```

to the array. That will have BerlinDB "listen" on `is_primary`. Then you can hook into `edd_transition_customer_address_is_primary`. Then in the callback function, you can bail if `$new_value` isn't true immediately, without querying the address again.

ashleyfae (changes requested):
I'm definitely preferring this approach! Overall it's looking good, just requesting some organizational changes.

Also we should definitely get some unit tests around this. @robincornett do you want me to handle writing those?

includes/database/queries/class-customer-address.php
robincornett: Ahhh, thank you for the `transition` information--that was really puzzling me!

There is already a `customer-actions.php` file  in `includes/admin/customers/customer-actions.php` ... should I just use that instead of creating a new file?

And yes please on the unit tests, @ashleyfae thank you!

robincornett: Never mind on putting the function in the existing file; that only runs in the admin. Noting though that there is a `customer-functions.php` file in just `includes` and maybe it should be moved to `includes/customers`?

includes/customers/customer-actions.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * Customer Actions.
+ *
+ * This file contains all of the first class functions for interacting with a
```
ashleyfae, line 5: Can you adjust the description?

includes/customers/customer-actions.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+function edd_demote_customer_primary_addresses( $old_value, $new_value, $item_id ) {
+	if ( ! $new_value ) {
+		return;
+	}
+	$customer_addresses = new EDD\Database\Queries\Customer_Address();
+	$address            = $customer_addresses->get_item( $item_id );
+	if ( ! $address->is_primary ) {
+		return;
+	}
```
ashleyfae, line 31: I don't think we should need this query for fetching the whole object. `$new_value` should already be the same as `$address->is_primary` (same value). So querying it again shouldn't be needed.

includes/customers/customer-actions.php
```diff
@@ -0,0 +1,47 @@
...
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       3.0
+ */
+
+/**
+ * When a new primary address is added to the database, any other primary addresses should be demoted.
+ *
+ * @param string $old_value
+ * @param string $new_value
+ * @param int $item_id
```
ashleyfae, line 20: Can you add some descriptions for what these values are?

`$old_value` should be the previous value of `is_primary` and `$new_value` should be the new value.

includes/customers/customer-actions.php
```diff
@@ -0,0 +1,47 @@
...
+			'fields'      => 'ids',
+			'customer_id' => $address->customer_id,
+			'is_primary'  => true,
+		)
+	);
+	if ( empty( $previous_primary_addresses ) ) {
+		return;
+	}
+	foreach ( $previous_primary_addresses as $previous ) {
+		$customer_addresses->update_item( $previous, array( 'is_primary' => false ) );
```
ashleyfae, line 44:
```suggestion
		edd_update_customer_address( $previous, array( 'is_primary' => false ) );
```

`edd_update_customer_address()` uses `EDD\Database\Queries\Customer_Address()` so there isn't any actual difference between them, but it's a better idea to use the wrapper in case we end up expanding the wrapper function at a later date.

ashleyfae (changes requested):

includes/database/queries/class-customer-address.php
ashleyfae: > Noting though that there is a `customer-functions.php` file in just `includes` and maybe it should be moved to `includes/customers`?

While I'd love to do that for organizational purposes, I think I recall a discussion a while ago about not moving existing files around in case people are referencing them directly. :woman_shrugging: So let's leave it at least for now.

ashleyfae: Also I'll handle the unit tests once the code is finalized.

includes/customers/customer-actions.php
robincornett: I agree and thought about removing it, but the object is required to get the `customer_id` parameter for the following query, so I left it in. I do agree that the `$address->is_primary` check is redundant.

robincornett (commented):
Just noting that all unit test pass on this branch locally with PHPUnit 5.7.27.

includes/compat/class-customer.php
```diff
@@ -274,28 +276,33 @@ public function update_user_meta( $check, $object_id, $meta_key, $meta_value, $p
 		if ( is_array( $addresses ) && ! empty( $addresses ) ) {
```
ashleyfae, line 276: I know this isn't part of your changes, but while we're here...

```suggestion
		if ( is_array( $addresses ) && ! empty( $addresses[0] ) ) {
```

Because `$address[0]` is used on the next line.

ashleyfae (changes requested):
One more teeny tiny change, then I'm happy with it! :tada:

ashleyfae (approved):
Just need to wait for unit tests.

includes/customers/customer-actions.php
```diff
@@ -0,0 +1,43 @@
...
+ * @param int $item_id      The address ID in the edd_customer_addresses table.
+ * @return void
+ */
+function edd_demote_customer_primary_addresses( $old_value, $new_value, $item_id ) {
+	if ( ! $new_value ) {
+		return;
+	}
+	$customer_addresses         = new EDD\Database\Queries\Customer_Address();
+	$address                    = $customer_addresses->get_item( $item_id );
+	$previous_primary_addresses = $customer_addresses->query(
```
spencerfinnell, line 28: Similar to [Ashley's other suggestion](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8146/files#discussion_r516671612), could these use `edd_get_customer_addresses()` and `edd_fetch_customer_address()`?

includes/customer-functions.php
```diff
@@ -653,73 +653,21 @@ function edd_maybe_add_customer_address( $customer_id = 0, $data = array() ) {
 	// Check if this address is already assigned to the customer.
 	$address_exists = $customer_addresses->query( $address_to_check );
 	if ( ! empty( $address_exists ) ) {
+		if ( 'billing' === $address_to_check['type'] ) {
+			$customer_addresses->update_item( $address_exists[0], array( 'is_primary' => true ) );
```
spencerfinnell, line 657: Similar as before, I agree that we should stick to using the helper functions like `edd_update_customer_address()` and `edd_add_customer_address()`.

spencerfinnell (changes requested):
@robincornett Nothing that changes behavior, but a few suggestions similar to @ashleyfae that I think would be good to have in now.

spencerfinnell:
My last final thing that I'm not sure is something to worry about, but I do think is something to consider: 

I think the function name `edd_maybe_add_customer_address()` is a little misleading. The function can do three things:

1. Nothing
2. Update an existing address
3. Add a new address

I think if it just did nothing _or one other thing_ the `_maybe_` would make sense.

So perhaps just something like `edd_add_or_update_customer_address()`, but I don't love that either.

... WordPress core has some functions like `update_*_meta()`, `update_option()` that handle adding, updating, or nothing, but then others like `wp_update_user()` only perform updates.

ashleyfae:
> My last final thing that I'm not sure is something to worry about, but I do think is something to consider:
> 
> I think the function name `edd_maybe_add_customer_address()` is a little misleading. The function can do three things:
> 
> 1. Nothing
> 2. Update an existing address
> 3. Add a new address
> 
> I think if it just did nothing _or one other thing_ the `_maybe_` would make sense.
> 
> So perhaps just something like `edd_add_or_update_customer_address()`, but I don't love that either.
> 
> ... WordPress core has some functions like `update_*_meta()`, `update_option()` that handle adding, updating, or nothing, but then others like `wp_update_user()` only perform updates.

I think `edd_maybe_add_customer_address()` is better than `edd_add_or_update_customer_address()` just because the only update that's happening is making the address primary. The actual contents of the address are never updated. (if we're doing an update)

I know `edd_maybe_add_customer_address()` not totally perfect, but I can't think of anything I like better.

spencerfinnell (approved):

PR #8147, opened 2020-10-08
Convert CSS to use webpack

### LINKED ISSUE #7977: Use Webpack to Process Admin CSS

Use Webpack to process `edd-admin.css`

### Justification or use case

At the very least this would allow running the current CSS through [`postcss/autoprefixer`](https://github.com/postcss/autoprefixer) which allows removal of browser prefixes in the file.

We could also:

- Add SASS processing
- Split/organize file structure
- Try setting up [`purgecss`](https://github.com/FullHuman/purgecss) to find unused CSS.

robincornett:
Fixes #7977

@ashleyfae per Spencer's suggestion, marking this ready for review as-is (with CSS for Chosen and tax rates only updated). He suggests that the main admin CSS file can be dealt with more gradually, in the hopes that eventually it can be broken into smaller files which will be loaded as needed, rather than as one large file.


assets/css/admin/chosen/_colors.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+@include wordpress-admin-schemes();
+
+body.admin-color-evergreen {
+	@include admin-scheme(#36533f);
+}
+
+body.admin-color-mint {
+	@include admin-scheme(#4f6d59);
+}
```
spencerfinnell, line 15: Perhaps we should make our own mixin that ensures the fallback default colors are set for CSS variable support, includes the core color schemes, as well as our additional bbPress color scheme support.

assets/css/variables/_colors.scss
```diff
@@ -0,0 +1,8 @@
+
```
spencerfinnell, line 1: Let's document these colors as being current as of WordPress 5.5.1.

spencerfinnell (changes requested):

assets/css/admin/chosen/_colors.scss
robincornett: Just created `@mixin edd-admin-colors` along with adding bbPress colors to the color variables. New color theme implementation is now just:

```css
:root {
	@include edd-admin-colors();
}
```
Which maybe should be in the main EDD admin stylesheet, rather than each individual one, but it is in the Chosen one right now.

spencerfinnell (approved):
I think this is a good starting point that lets us start extracting components out over time.

ashleyfae (approved):

ashleyfae:
The merge conflicts need resolving before we can merge.

ashleyfae (approved):
Feel free to merge when unit tests pass :+1:

Issue #7385, opened 2019-08-28
Can't Decrease Tax With `modify_cart_item()`


spencerfinnell:
### Expected behavior

When using 

```
$payment->modify_cart_item( 123, array(
  'tax' => 0,
) );
```

I expect the tax amount for the payment amount to be adjusted by the difference.

### Actual behavior

The line item amount is _doubled_ and added to the total.

### Steps to reproduce the behavior
1) Not implemented anywhere in core

### Information (if a specific version is affected):
PHP Version: 7.x
EDD Version (or branch): master
WordPress Version: 5.2

Any other relevant information:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/payments/class-edd-payment.php#L1443

`increase_tax()` is always called instead of `decrease_tax()`. If the initial tax amount is $1 it will add $2 ($1 - 0 = $1 which then gets added to the initial $1)

The test case doesn't account for decreases: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/tests/tests-payment-class.php#L886-L898

ashleyfae:
Confirmed. Here's how I reproduced:

Started with an order that had $0 tax. Ran this to set tax to $5:

```php
$payment = new EDD_Payment( 334 );
$result = $payment->modify_cart_item( 0, [
	'tax' => 5
] );
$payment->save();
```

Tax was increased to $5.

Then I ran this to decrease it to $1:

```php
$payment = new EDD_Payment( 334 );
$result = $payment->modify_cart_item( 0, [
	'tax' => 1
] );
$payment->save();
```

The tax for the individual order item became $1, but the overall tax became $6.

PR #8150, opened 2020-10-15
Issue 7385 - Fixed modify_cart_item() when decreasing tax and updated…

nickmond:
Fixed Issue 7385 - https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7385

Proposed Changes:
1. When using modify_cart_item() to modify the tax value, fixed bug where increase_tax() was being used for both increases/decreases. Instead use decrease_tax() when new value is lower than existing value.
2. Updated modify_cart_item_tax() unit test to check for both tax increases and decreases.


ashleyfae (approved):
This looks good - thank you @nickmond !

spencerfinnell (approved):

Issue #7701, opened 2020-04-02
Labels for input fields are rendered inconsistently


robincornett:
### Expected behavior
When I add a label to an input, it should be properly associated with the input (eg: clicking on the label should move focus to the input).

### Actual behavior
* When I add a label to a text field in a repeater row, the use of `edd_sanitize_key` in the `text` method strips out the `[]` and the label is disconnected from the field.
* The `textarea` method uses `$args['name']` for the label; `text` uses `$args['id']`
* 3.0 adds a label to the `checkbox`, but it more properly uses `esc_attr` so that it works in repeaters
* some uses of the `html->` methods define an `id`, but quite a few do not, and just use `name`
* if an `id` is not defined for a text field, for example, the text field will end up with no `id`

### Steps to reproduce the behavior
1) Edit a download and enable variable pricing
2) Inspect the input fields in the Price ID row.

For the Option Name field:

* There is no label associated with the input field
* The `span` id is `edd-edd_variable_prices1name-wrap` because of the use of `edd_sanitize_key`
* To add a label, two parameters must be added to the [field defintion](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/admin/downloads/metabox.php#L454-L459): both a `label` and an `id`, or else the `for` will be blank and the label will not be associated with the field

Price and Default have similar issues.

### Information (if a specific version is affected):
PHP Version: 7.3.5
EDD Version (or branch): master, 3.0
WordPress Version: 5.4

Any other relevant information:
As I'm working with the SL licensing addon, I'm seeing inconsistencies with how labels are added to fields--visually, most fields have some kind of visual label, but in repeatable rows especially, they're not present, which makes the inputs less accessible. This looks to be something that needs to be worked on both in Core and in extensions which use the `EDD->html` functions.

LisaCee:
Can I add a tweak to a class method in this issue, too?  Can the `span` ID in the `EDD()->html->date_field` class method be changed to use a `class` instead (or alternatively, to create unique `id`s) as suggested [here](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8151#discussion_r507957074) in this issue?

robincornett:
@LisaCee I went ahead and changed that to a `class`, and brought this up to date with `release/3.0`. I think anything we do with this class needs to be checked with lots of extensions, as it likely touches each and every one of them (part of the reason it's still a draft).

ashleyfae:
[...]

PR #8151, opened 2020-10-16
Issue/8112 Accessibility on Export->Reports

### LINKED ISSUE #8112: 3.0 - Improve Exports screen accessibility, markup
The Downloads > Reports > Export screen should be updated to improve overall markup and accessibility.

### Justification or use case
Running an automated accessibility scan reveals multiple issues with the Exports screen:
- Form elements are missing labels (I think no inputs on this screen have labels)
- ID attributes are not unique
- Heading levels increase incorrectly (h1 > h3)

We started formalizing form/input markup with #7739 and can resolve many of the accessibility issues while updating the markup ([reference for markup](https://github.com/easydigitaldownloads/easy-digital-downloads/wiki/Form-Input-Markup)).

LisaCee:
Fixes #8112

Proposed Changes:
1. Add labels to all form fields and add ids where needed
2. Update headings to be in hierarchical order
3. Convert spaces to tabs
4. Wrap date elements in `fieldset` with `legend`
5. Create unique id for month and year dropdowns


includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+									echo EDD()->html->date_field(
+										array(
+											'id'          => 'edd-order-export-start',
+											'class'       => 'edd-export-start',
+											'name'        => 'start',
+											'placeholder' => _x( 'From', 'date filter', 'easy-digital-downloads' ),
+										)
+									);
+									?>
+								<label for="edd-order-export-end" class="screen-reader-text">Export Sales and Earnings End Date</label>
```
robincornett, line 2602: This label text (and all following) should still be in `__()` so it can be translated please.

includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+						<p><?php esc_html_e( 'Download a CSV of all sales or earnings on a day-by-day basis.', 'easy-digital-downloads' ); ?></p>
+						<form id="edd-export-sales-earnings" class="edd-export-form edd-import-export-form" method="post">
+							<div class="edd-from-to-wrapper">
+								<label for="edd-order-export-start" class="screen-reader-text">Export Sales and Earnings Start Date</label>
 									<?php
+									echo EDD()->html->date_field(
+										array(
+											'id'          => 'edd-order-export-start',
+											'class'       => 'edd-export-start',
+											'name'        => 'start',
```
robincornett, line 2597: It looks like the `name` parameter should be made unique for each of these `EDD()->html->date_field()` uses, as that method calls the `EDD()->html->text()` method, which takes the `name` parameter and generates an `id` for a wrapping `span` (`'<span id="edd-' . edd_sanitize_key( $args['name'] ) . '-wrap">'`). So these maybe need to be changed to something like `export-start`, etc. to avoid the span ID duplication (I do not know why the `span` needs an ID at all, but that's a separate question).

includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+									array(
+										'name'        => 'download_id',
+										'id'          => 'edd_orders_export_download',
+										'chosen'      => true,
+										/* translators: the plural post type label */
+										'placeholder' => sprintf( __( 'All %s', 'easy-digital-downloads' ), edd_get_label_plural() ),
+									)
+								);
+								?>
+							<label for="edd_order_export_customer" class="screen-reader-text">All Customers or Select Customer</label>
```
robincornett, line 2627: Just noting, and not sure it's a real issue, but this label/input pair uses `_` and some of the others use `-` ... it looks like there is a mix of both, so not sure what is correct. I believe the HTML `select` method forces `_`, but it doesn't look like the text method does. Not a deal breaker but I wonder if we should update them all to be consistent?

robincornett (changes requested):

includes/admin/reporting/reports.php
LisaCee: I totally blanked on that.  Thanks.

includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+
+				<?php do_action( 'edd_reports_tab_export_content_top' ); ?>
+
+				<div class="postbox edd-export-earnings-report">
+					<h2 class="hndle"><span><?php esc_html_e( 'Export Earnings Report', 'easy-digital-downloads' ); ?></span></h2>
+					<div class="inside">
+						<p><?php esc_html_e( 'Download a CSV giving a detailed look into earnings over time.', 'easy-digital-downloads' ); ?></p>
+						<form id="edd-export-earnings" class="edd-export-form edd-import-export-form" method="post">
+							<div class="edd-to-and-from-container">
+							<label for="start_month" class="screen-reader-text">__( Start Month )</label>
```
robincornett, line 2564: These strings need to be echoed as well, and use inline PHP (see the h2 above for an example), and include the text domain please:

```php
<label for="start_month" class="screen-reader-text"><?php esc_html_e( 'Start Month', 'easy-digital-downloads' ); ?></label>
```

includes/admin/reporting/reports.php
LisaCee: Do you mean that each of these `date_field()`s should have a unique `id`?  I tried that and it breaks this rule in the CSS:
```
.edd-from-to-wrapper input[name="start"],
.edd-from-to-wrapper input[name="start-date"],
.edd-from-to-wrapper input[name="filter_from"] {
	border-top-right-radius: 0;
	border-bottom-right-radius: 0;
}
```
Do you think it would be better to tweak the CSS (maybe use a wildcard so it's not too verbose) or keep the duplicate ids created for the span?  Was another accessibility issue flagged with the duplicate ids?

robincornett: On the one hand, it may not be worth fixing, because the export screen will still have so many duplicate IDs (the hidden fields are all `#edd_ajax_export`). But it would be nice to fix what we could.

I wonder if we could replace the input styles for the `edd-from-to-wrapper` with something that won't rely on the name element. This _seems_ to work, but would need to be tested to make sure it doesn't break the filter bar and list tables that use it:

```css
.edd-from-to-wrapper span input {
	margin-left: -1px;
	border-top-left-radius: 0;
	border-bottom-left-radius: 0;
}

.edd-from-to-wrapper span:first-of-type input {
	border-top-right-radius: 0;
	border-bottom-right-radius: 0;
	margin-left: 0;
}
```

Alternatively, an even better option would be to fix the `span` ID in the `HTML` class method to either just be a class, which would make more sense to me in general, or to force it to use the `id` attribute if that's supplied, which it often is not, but I think that would more properly fall into a different issue (#7701 might be appropriate).

includes/admin/reporting/reports.php
LisaCee: D'oh

includes/admin/reporting/reports.php
LisaCee: I added this as a comment in #7701, as it seems to address the use of `html->` methods using the `name` attribute

assets/css/edd-admin.css
```diff
@@ -2344,6 +2344,11 @@ body.download_page_edd-reports {
 	right: 6px;
 }
 
+.edd-reports-wrapper .postbox h2 {
```
robincornett, line 2347: Can we combine these two rules together? Sorry, should have asked for that before. The comment is helpful (there are likely more extensions than SL which use `h3` too), so I would keep that.

includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+									);
+
+									?>
+							</div>
+							<label for="edd_orders_export_download" class="screen-reader-text"><?php esc_html_e( 'All Downloads or Select Download', 'easy-digital-downloads' ); ?></label>
+								<?php
+								echo EDD()->html->product_dropdown(
+									array(
+										'name'        => 'download_id',
+										'id'          => 'edd-orders-export-download',
```
robincornett, line 2620: Since EDD forces this to `_` instead of `-`, I would put it back to that so that it's clear without having to refer to the `product_dropdown` (I think actually you have to dig through to `select`) method. Anything which creates a dropdown/select input, I think EDD will do this to, so we may as well set the `id` parameter to match what's output.

includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+										array(
+											'id'          => 'edd-taxed-orders-export-end',
+											'class'       => 'edd-export-end',
+											'name'        => 'end',
+											'placeholder' => _x( 'To', 'date filter', 'easy-digital-downloads' )
+										)
+									);
+									?>
+							</div>
+							<label for="edd_taxed_orders_export_status" class="screen-reader-text"><?php esc_html_e( 'All Statuses or Select Status', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 2726: I know this `select` field came with `-`, but I would change it to `_` for the same reasons above.

includes/admin/reporting/reports.php
```diff
@@ -2548,377 +2547,454 @@ function display_export_report() {
...
+							$taxonomies = array();
+
+							if ( $results ) {
+								foreach ( $results as $r ) {
+									$t = get_taxonomy( $r->taxonomy );
+									$taxonomies[ absint( $r->term_id ) ] = $t->labels->singular_name . ': ' . esc_html( $r->name );
+								}
+							}
+							?>
+							<label for="taxonomy" class="screen-reader-text"><?php esc_html_e( 'All Taxonomies or Select Taxonomy', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 2807: I would change this to be just `'Select Taxonomy'` (same for select download), as "All" is the first option in the list.

includes/class-edd-html-elements.php
```diff
@@ -510,6 +510,7 @@ public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5,
 
 		$output = $this->select( array(
 			'name'             => $name,
+			'id'               => $name,
```
robincornett, line 513: I'm not 100% sure about this one, but it is breaking some tests. It would be easy enough to update the tests so they no longer expect an empty `id` attribute, which I think is the issue for at least two of the three, but just setting the `$name` to be the `id` as well may still result in duplicate IDs. Although it's not an issue currently on the exports screen so maybe this is fine. At least with the `id` set, we can associate the label with it, so that is definitely an improvement.

robincornett (commented):
Some more fiddly things, and some things I'm not sure of--but looking good!

includes/admin/reporting/reports.php
LisaCee: What do you think about the Country, Region and Status ones?  I didn't realize that a screen reader would read the default choice?

LisaCee: Should I do this for all of the dropdowns (Countries, Statuses...)?  I didn't realize the screen reader would read the default.

robincornett: Yes, it will read the label, and then the options. So you'd end up with "Select Taxonomy. All Taxonomies. Taxonomy 1" vs. "All Taxonomies or Select Taxonomy. All Taxonomies. Taxonomy 1".

I am not sure, but I think that the same rules will apply to both regular `select` inputs and ones implementing Chosen. But looking at these, I think they may have their own...not accessibility, but usability, which isn't really that different, issue. I'm finding that the Chosen inputs don't actually have an "all" option, at least not the ones I'm checking, so if I select a download, or a country, there is no way for me to get back to "All Downloads" or "All Countries" unless I refresh the screen. So I think maybe we need to double check that, to make sure that the "all" options are available, even after a selection is made?

So under "Export Customers", I can select a taxonomy and then go back to "all" (standard `select`), but not downloads (which uses Chosen).

But yes, I would update the labels for the select inputs, and we can dig into the "all" as well, I hope.

robincornett:
@LisaCee something I ran into this week going through the Exports screen is the abbreviated months under "Export Earnings Report". I'm guessing that was originally done to make those select fields closer to the same size as the year dropdown, but the abbreviations are weird when they're being read out loud, and the size isn't an issue anymore. Are these options defined somewhere else, or could we change them, do you think? I am not sure offhand if this is used elsewhere.

LisaCee:
> @LisaCee something I ran into this week going through the Exports screen is the abbreviated months under "Export Earnings Report". I'm guessing that was originally done to make those select fields closer to the same size as the year dropdown, but the abbreviations are weird when they're being read out loud, and the size isn't an issue anymore. Are these options defined somewhere else, or could we change them, do you think? I am not sure offhand if this is used elsewhere.

Thoughts on this @ashleyfae, @spencerfinnell or @cklosowski?  As far as I can tell, this is only used for reports (and tests) and it would literally just be changing a letter in the `edd_month_num_to_name` function.

ashleyfae:
Hmm I'm hesitating a little because that function's purpose is defined as:

> Takes a month number and returns the name three letter name of it.

Although we only use the function in a couple places, other people might be using it too.

Rather than changing it for everyone completely, maybe we introduce a second parameter to that function like `$return_long_name = false`. When false, you get the short name. When true, you get the long name. It defaults to `false` so the behaviour doesn't change for anyone already using that. We can pass that in as `true` when used in the `month_dropdown()` function so we get the long version.

includes/class-edd-html-elements.php
robincornett: Sure--so if you look at one of the [job results](https://travis-ci.org/github/easydigitaldownloads/easy-digital-downloads/jobs/737472910), you can see that two of the tests fail asserting that two strings are equal. The expected string has an empty id (`id=""`), and with the new parameter being passed, it's actually coming out as `id="year"`. [Here's the expected string in the actual test](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8112/tests/tests-html-elements.php#L138), and it should be similar for the `test_year_dropdown_variable` test. The `test_month_dropdown` is not comparing two strings, but I think it should be the same idea, changing out the empty `id` in line 186.

includes/admin/reporting/reports.php
```diff
@@ -2545,377 +2544,454 @@ function display_export_report() {
...
-										/* translators: the plural post type label */
-										'placeholder' => sprintf( __( 'All %s', 'easy-digital-downloads' ), edd_get_label_plural() ),
-									)
-								); ?>
-								<div class="edd-from-to-wrapper">
-									<?php
-
-									echo EDD()->html->date_field( array(
-										'id'          => 'edd-file-download-export-start',
+										'id'          => 'edd_file_download_export_start',
```
spencerfinnell, line 2960: @LisaCee I think this needs to be switched back to hypens so the `label` is linked properly.

spencerfinnell (changes requested):

includes/misc-functions.php
```diff
@@ -531,13 +531,14 @@ function edd_get_currency_name( $code = 'USD' ) {
...
+ * @param integer $n The number of the month.
+ * @param bool    $return_long_name Optional. Return full name of month if true. Default false.
  * @return string Short month name
  */
-function edd_month_num_to_name( $n ) {
+function edd_month_num_to_name( $n, $return_long_name = false ) {
 	$timestamp = mktime( 0, 0, 0, $n, 1, 2005 );
 
-	return date_i18n( "M", $timestamp );
+	return ! $return_long_name ? date_i18n( 'M', $timestamp ) : date_i18n( 'F', $timestamp );
```
ashleyfae, line 541: We can reduce this code a little by using the conditional _only_ for the first parameter value (instead of repeating the whole function twice). Example:

```suggestion
	$date_format = $return_long_name ? 'F' : 'M';
	
	return date_i18n( $date_format, $timestamp );
```

(I also flipped the `$return_long_name` logic just because I find that easier to read.)

includes/class-edd-html-elements.php
ashleyfae: I think we should add a new parameter for the ID value (same for the `month_dropdown()`). Two reasons:

1. It will allow us to set unique values (fixing the issue Robin brought up above).
2. I think `year` and `month` are too vague for us to use as ID values. We'd generally be prefixing IDs and class names with `edd-` and right now we don't have that. It seems very likely that our `year` ID could clash with another plugin or possibly even WP core somewhere.

ashleyfae (changes requested):

LisaCee:
What do you think about `	'id'  => $form . '-' . $name,` ?  It seems a bit verbose, but I'm struggling to think of something else unique and descriptive.  @robincornett

includes/class-edd-html-elements.php
LisaCee: What do you think about `	'id'  => $form . '-' . $name,` ?  It seems a bit verbose, but I'm struggling to think of something else unique and descriptive.  @robincornett

robincornett: So it would be something like `edd_export_earnings_start_month`? (I think it needs to be all underscores as the `select` method will make any `-` into `_` in the end--double check me on that but I think that's the way it's done.) It is maybe verbose but I think it's reasonable, @LisaCee.

ashleyfae: I'm fine with that. :+1:

includes/class-edd-html-elements.php
```diff
@@ -496,7 +496,7 @@ public function category_dropdown( $name = 'edd_categories', $selected = 0 ) {
 	 *
 	 * @return string $output Year dropdown
 	 */
-	public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5, $years_after = 0 ) {
+	public function year_dropdown( $form = 'form', $name = 'year', $selected = 0, $years_before = 5, $years_after = 0 ) {
```
ashleyfae, line 499: The new parameter needs to be added to the end. I know it won't be nicely organized, but the order is important for how the function gets used and we need to assume that third party plugins are using this function and we can't break their implementation.

Also, what about just making this an `$id` attribute and letting the user set the exact ID they're looking for? As opposed to building it from a form "name" of sorts. I don't think `$form` is used anywhere else so just letting us set the exact ID we want seems a bit simpler. To try to make it unique for anyone not using the parameter, the default could be `edd-year-dropdown` or something. :woman_shrugging:

includes/class-edd-html-elements.php
```diff
@@ -529,23 +532,26 @@ public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5,
 	 *
 	 * @return string $output Month dropdown
 	 */
-	public function month_dropdown( $name = 'month', $selected = 0 ) {
+	public function month_dropdown( $form = 'form', $name = 'month', $selected = 0 ) {
```
ashleyfae, line 535: Same as above -- the new parameter needs to be added to the end.

ashleyfae (changes requested):

includes/class-edd-html-elements.php
```diff
@@ -489,14 +489,14 @@ public function category_dropdown( $name = 'edd_categories', $selected = 0 ) {
...
-	 *
+	 * @param string $name         Name attribute of the dropdown.
+	 * @param int    $selected     Year to select automatically.
+	 * @param int    $years_before Number of years before the current year the dropdown should start with.
+	 * @param int    $years_after  Number of years after the current year the dropdown should finish at.
+	 * @param string $id A unique identifier for the field.
 	 * @return string $output Year dropdown
 	 */
-	public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5, $years_after = 0 ) {
+	public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5, $years_after = 0, $id = 'edd-year-dropdown' ) {
```
robincornett, line 499: Can the `id` use underscores please, just because the `select` method will change them and it feels easier to set it up as it will be output? Same for the month dropdown.

Also, and no real reason why other than I don't care for "dropdown" as a word in general, what if the default id was just `edd_year` or `edd_year_select`? But that's just me.

includes/class-edd-html-elements.php
```diff
@@ -524,28 +527,31 @@ public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5,
...
 	 */
-	public function month_dropdown( $name = 'month', $selected = 0 ) {
+	public function month_dropdown( $name = 'month', $selected = 0, $id = 'edd-month-dropdown' ) {
 		$month    = 1;
 		$options  = array();
 		$selected = empty( $selected ) ? date( 'n' ) : $selected;
 
 		while ( $month <= 12 ) {
-			$options[ absint( $month ) ] = edd_month_num_to_name( $month );
+			$options[ absint( $month ) ] = edd_month_num_to_name( $month, true );
```
robincornett, line 541: Should the `true` parameter be added as a parameter to this method as well after the new `$id` parameter? So that anyone who has been using the `month_dropdown` method expecting it to output three letter month names will still have that?

tests/tests-html-elements.php
```diff
@@ -135,7 +135,7 @@ public function test_category_dropdown() {
 	 */
 	public function test_year_dropdown() {
 		$current_year = date( 'Y' );
-		$expected     = '<select name="year" id="" class="edd-select " data-placeholder="">';
+		$expected     = '<select name="year" id="edd-year-dropdown_year" class="edd-select " data-placeholder="">';
```
robincornett, line 138: Because the html `select` method changes `-` to `_`, this needs to be changed to use `_` for the tests to pass the expected output (same for month).

includes/admin/reporting/reports.php
```diff
@@ -2545,335 +2544,424 @@ function display_export_report() {
...
+									array(
+										'name'        => 'download_id',
+										'id'          => 'edd_orders_export_download',
+										'chosen'      => true,
+										/* translators: the plural post type label */
+										'placeholder' => sprintf( __( 'All %s', 'easy-digital-downloads' ), edd_get_label_plural() ),
+									)
+								);
+								?>
+							<label for="edd_order_export_customer" class="screen-reader-text"><?php esc_html_e( 'All Customers or Select Customer', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 2633: Can this label just say "Select Customer"?

includes/admin/reporting/reports.php
```diff
@@ -2545,335 +2544,424 @@ function display_export_report() {
...
+										'id'               => 'edd_taxed_orders_export_status',
+										'name'             => 'status',
+										'show_option_all'  => __( 'All Statuses', 'easy-digital-downloads' ),
+										'show_option_none' => false,
+										'selected'         => false,
+										'options'          => edd_get_payment_statuses(),
+									)
+								);
+								?>
+							<label for="edd_reports_filter_taxed_countries" class="screen-reader-text"><?php esc_html_e( 'All Countries or Select Country', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 2751: Can this label just say "Select country"?

includes/admin/reporting/reports.php
```diff
@@ -2545,335 +2544,424 @@ function display_export_report() {
...
+										'options'          => edd_get_country_list(),
+										'chosen'           => true,
+										'selected'         => false,
+										'show_option_none' => false,
+										'placeholder'      => __( 'All Countries', 'easy-digital-downloads' ),
+										'show_option_all'  => false,
+									)
+								);
+								?>
+							<label for="edd_reports_filter_regions" class="screen-reader-text"><?php esc_html_e( 'All Regions or Select Region', 'easy-digital-downloads' ); ?></label>
```
robincornett, line 2766: Same label question here please.

robincornett (changes requested):

includes/class-edd-html-elements.php
LisaCee: I have mixed feelings on this, since the primary function doesn't use that parameter.  Is there a way I could add it in the `while` loop?

Something more like this but done in the proper PHP way?

```
while ( $month <= 12 ) {
        @since 3.0
        @param bool $return_long_name Optional. Return full name of month if true. Default false.
        $options[ absint( $month ) ] = edd_month_num_to_name( $month, true );
	$month ++;
}
```

includes/admin/reporting/reports.php
```diff
@@ -2545,335 +2544,424 @@ function display_export_report() {
...
+							<legend class="screen-reader-text">
+								<?php esc_html_e( 'Export Sales and Earnings Dates', 'easy-digital-downloads' ); ?>
+							</legend>
+								<label for="edd-order-export-start" class="screen-reader-text"><?php esc_html_e( 'Set start date', 'easy-digital-downloads' ); ?></label>
 									<?php
+									echo EDD()->html->date_field(
+										array(
+											'id'          => 'edd-order-export-start',
+											'class'       => 'edd-export-start',
+											'name'        => 'start',
```
robincornett, line 2603: Can the `name` attributes for all of the `date_field` calls be updated to be unique as well? It can match the ID for each, but the `date_field` method calls the `text` method, which uses the `name` to create the `id` for the wrapper `span` (which I need to update the issue where we mentioned maybe changing that to a class).

includes/admin/reporting/reports.php
```diff
@@ -2545,335 +2544,424 @@ function display_export_report() {
...
+							<fieldset class="edd-from-to-wrapper">
+							<legend class="screen-reader-text">
+								<?php esc_html_e( 'Export Taxed Customer Dates', 'easy-digital-downloads' ); ?></legend>
+							<label for="edd-taxed-customers-export-start" class="screen-reader-text"><?php esc_html_e( 'Set start date', 'easy-digital-downloads' ); ?></label>
+								<?php
+									echo EDD()->html->date_field(
+										array(
+											'id'          => 'edd-taxed-customers-export-start',
+											'class'       => 'edd-export-start',
+											'name'        => 'taxed-orders-export-start',
```
robincornett, line 2866: This is actually taxed customers, not orders (creating a duplicate ID).

```suggestion
											'name'        => 'taxed-customers-export-start',
```

robincornett (changes requested):

assets/css/edd-admin.css
```diff
@@ -57,14 +57,14 @@ a.edd-delete:hover {
 	z-index: 1;
 }
 
-.edd-from-to-wrapper input[name="start"],
+.edd-from-to-wrapper input[name*="export-start"],
```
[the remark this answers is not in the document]
LisaCee: Nice suggestion. I had done it so that I wouldn't accidentally trigger any changes, but I did some digging and was even able to shorten this CSS block by a selector.

includes/class-edd-html-elements.php
robincornett: Sorry, I think I asked my question unclearly. I don't see it used in EDD core outside of the Exports screen, but I'm thinking we should update the month_dropdown parameters to include the `$return_long_name` parameter that you've already added to the `edd_month_num_to_name` function, and then use it in the exports screen with the new parameter to get the long name. So the method would be updated like this:

```php
	/**
	 * Renders an HTML Dropdown of months
	 *
	 * @since 1.5.2
	 *
	 * @param string  $name             Name attribute of the dropdown.
	 * @param int     $selected         Month to select automatically.
	 * @param string  $id               A unique identifier for the field.
	 * @param boolean $return_long_name Whether to use the long name for the month.
	 *
	 * @return string $output Month dropdown
	 */
	public function month_dropdown( $name = 'month', $selected = 0, $id = 'edd_month_select', $return_long_name = false ) {
		$month    = 1;
		$options  = array();
		$selected = empty( $selected ) ? date( 'n' ) : $selected;

		while ( $month <= 12 ) {
			$options[ absint( $month ) ] = edd_month_num_to_name( $month, $return_long_name );
			$month ++;
		}
...
```

And we would use it:

```php
echo EDD()->html->month_dropdown( 'start_month', 0, 'edd_export_earnings', true );
```

That way we won't be making unexpected changes to any extensions using the `month_dropdown` method ([ref](https://github.com/search?q=org%3Aeasydigitaldownloads+html-%3Emonth_dropdown&type=code)).

LisaCee: That makes way more sense!  :joy:

includes/class-edd-html-elements.php
```diff
@@ -489,14 +489,14 @@ public function category_dropdown( $name = 'edd_categories', $selected = 0 ) {
...
-	 * @param string $name         Name attribute of the dropdown
-	 * @param int    $selected     Year to select automatically
-	 * @param int    $years_before Number of years before the current year the dropdown should start with
-	 * @param int    $years_after  Number of years after the current year the dropdown should finish at
-	 *
+	 * @param string $name         Name attribute of the dropdown.
+	 * @param int    $selected     Year to select automatically.
+	 * @param int    $years_before Number of years before the current year the dropdown should start with.
+	 * @param int    $years_after  Number of years after the current year the dropdown should finish at.
+	 * @param string $id A unique identifier for the field.
```
ashleyfae, line 496: Can you add some spaces between `$id` and `A` to line up the description with the others?

includes/class-edd-html-elements.php
```diff
@@ -524,28 +527,33 @@ public function year_dropdown( $name = 'year', $selected = 0, $years_before = 5,
 	 *
 	 * @since 1.5.2
 	 *
-	 * @param string $name     Name attribute of the dropdown
-	 * @param int    $selected Month to select automatically
+	 * @param string  $name     Name attribute of the dropdown.
+	 * @param int     $selected Month to select automatically.
+	 * @param string  $id A unique identifier for the field.
+	 * @param boolean $return_long_name Whether to use the long name for the month.
```
ashleyfae, line 533: Same as above -- let's line up the descriptions with each other. It makes it a little easier to read.
(For both `$id` and `$return_long_name`)

robincornett (approved):

ashleyfae:
Added a unit test covering the new long month names functionality: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8151/commits/5cb602e09412d69c85587ef658299cc9dbd5dfee

ashleyfae (approved):

PR #8152, opened 2020-10-16
Issue/7796 Accessibility on Orders Screens

### LINKED ISSUE #7796: 3.0 - Order Screens Headings/Buttons Should Be Updated for Accessibility
### Expected behavior
For semantic and accessible HTML markup, headings should only increase by one level, and buttons must have discernible text.

### Actual behavior
On the edit and add orders screens, the headings in the postbox elements are `<h3>`, and `<h2>` is skipped. We've addressed this on other screens; just need to do the same here.

On the add order screen, the "delete x item" buttons do not have discernible text (example):

```html
<button class="button-link delete">
	<span class="dashicons dashicons-no"></span>
</button>
```

1) Add a new order
2) Add an item and inspect the red X button, no text
3) Inspect the postbox headings

PHP Version: all
WordPress Version: all

LisaCee:
Fixes #7796 

Proposed Changes:
1. Add ~aria-label~ screen reader label to empty button
2. Update headings to not skip levels

### DIFF
--- a/includes/admin/views/tmpl-order-item.php
+++ b/includes/admin/views/tmpl-order-item.php
@@ -20,7 +20,7 @@ $view_url = add_query_arg(
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<button class="button-link delete">
+		<button class="button-link delete" aria-label="removeItem">
 			<span class="dashicons dashicons-no"></span>
 		</button>
 		<# } #>
--- a/includes/admin/views/tmpl-order-item.php
+++ b/includes/admin/views/tmpl-order-item.php
@@ -20,8 +20,9 @@ $view_url = add_query_arg(
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<button class="button-link delete" aria-label="removeItem">
-			<span class="dashicons dashicons-no"></span>
+		<label for="remove-item" class="screen-reader-text">__( Remove Item )</label>
+		<button class="button-link delete" aria-label="remove item">
+			<span class="dashicons dashicons-no" id="remove-item"></span>
 		</button>
 		<# } #>
 
--- a/includes/admin/views/tmpl-order-item.php
+++ b/includes/admin/views/tmpl-order-item.php
@@ -20,9 +20,9 @@ $view_url = add_query_arg(
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<label for="remove-item" class="screen-reader-text">__( Remove Item )</label>
 		<button class="button-link delete" aria-label="remove item">
 			<span class="dashicons dashicons-no" id="remove-item"></span>
+			<span class="screen-reader-text"><?php esc_html_e( 'Remove Item', 'easy-digital-downloads' ); ?></span>
 		</button>
 		<# } #>
 
--- a/includes/admin/views/tmpl-order-item.php
+++ b/includes/admin/views/tmpl-order-item.php
@@ -20,9 +20,9 @@ $view_url = add_query_arg(
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<button class="button-link delete" aria-label="remove item">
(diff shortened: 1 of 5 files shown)


includes/admin/views/tmpl-order-item.php
```diff
@@ -20,7 +20,7 @@
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<button class="button-link delete">
+		<button class="button-link delete" aria-label="removeItem">
```
robincornett, line 23: I'm curious about why you chose to use an `aria-label` instead of adding text to the button and hiding it with `screen-reader-text`? I believe your approach is valid, so I don't think it needs to change, but is a string which should be translatable and which may be read by screen reader users (as an example), so it should be internationalized and a phrase, please.

robincornett (changes requested):

includes/admin/views/tmpl-order-item.php
LisaCee: I did a bunch of Googling and never came across anyone suggesting using a label, though it seems obvious now.  As that is what we use for other hidden labels, I'm going to change it!

LisaCee: It didn't even dawn on me to use a label.  I Googled a bunch and `aria-label` seemed to be the most standard and [this](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html) really decided for me.  I can and will, however, change it to a string that can be translated.

includes/admin/views/tmpl-order-item.php
```diff
@@ -20,8 +20,9 @@
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<button class="button-link delete">
-			<span class="dashicons dashicons-no"></span>
+		<label for="remove-item" class="screen-reader-text">__( Remove Item )</label>
```
robincornett, line 23: The text string can be in just a span after the dashicons one--the `label` is not needed for a `button`, which can accept text (just in this case, it won't be visible because of the screen-reader-text class).

Also, you'll need to use inline PHP to use the translation function with the text domain, and echo, so the whole span may end up looking something like this (not tested):

```php
<span class="screen-reader-text"><?php esc_html_e( 'Remove Item', 'easy-digital-downloads' ); ?></span>
```

includes/admin/views/tmpl-order-item.php
```diff
@@ -20,8 +20,9 @@
 <td class="has-row-actions column-primary">
 	<div class="removable">
 		<# if ( true === data.state.isAdding ) { #>
-		<button class="button-link delete">
-			<span class="dashicons dashicons-no"></span>
+		<button class="button-link delete" aria-label="remove item">
```
robincornett, line 23: Now that the button has text, the `aria-label` is redundant (it overrides the button text). If it were necessary, it should be added as a translation ready string, but I think in this case, it can be removed.

I think the only thing I am left wondering is whether "Remove Item" is enough, especially since it's _before_ the product name. I wonder if it should be "Remove [Name of Product]" instead--I think what would work would be something like:

```php
<?php printf( __( 'Remove %s', 'easy-digital-downloads' ), '{{{ data.productName }}}' ); ?>
```

because I think the JS is replacing the `{{{}}}` placeholder? This is feeling a little bit like the checkboxes in the post list tables, is why I'm wondering this now, where each checkbox label says, "Select {{{ Hello World! }}}". Using the placeholder in the string gives this button the same kind of effect.

robincornett (changes requested):

includes/admin/views/tmpl-order-item.php
LisaCee: I think this will work now.  Is there a difference between `{{{}}}` and `{{}}`?  They seemed to work the same, but I went with `{{{}}}` as suggested.

robincornett (commented):
Thanks, Lisa, I think this is looking good!

Sorry, I just was building an order this morning and realized that the order adjustments have the same kind of button with no text (`includes/admin/views/tmpl-order-adjustment.php`). Could we give them the same treatment, and just have the text be "Remove Order Credit" (or fee) as needed?

robincornett (approved):

ashleyfae (approved):

PR #8156, opened 2020-10-21
Prevent negative fees from getting applied twice #6837

### LINKED ISSUE #6837: Negative Fees are applied again in PayPal Standard
### Expected behavior
When a discount is applied to total amount, it’s calculated in EDD, and then the discounted amount is passed to PayPal as is. Total amount is the same number between EDD and PayPal.

### Actual behavior
When I add discounts via `EDD()->fees->add_fee()`, everything is shown correctly in cart and Purchase History, but then `PayPal Standard` calculates and adds the discount again. So total of $80 with a discount of -$20 shows as total $60 in cart, but then becomes $40 in PayPal.

This started happening recently when I upgraded EDD from `2.2.8` to `2.9.3`. If I comment out the line 

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/74905fe92666c7b79e65d9db6b8e60e71912a7e3/includes/gateways/paypal-standard.php#L292-L294

everything works as expected / as it used to. I’m hoping you can recognize the issue and tell me what I might be doing wrong, if this is not a bug.

Thank you for EDD and for the help!

PHP Version: 5.6.32-5.6.32+mh2

EDD Version (or branch): 2.9.3

WordPress Version: 4.9.8

ashleyfae:
Fixes #6837

Proposed Changes:
1. Prevents negative fees from being applied twice.

---

This is going to need a bunch of testing because it touches payment amounts.

Testing steps for the original issue are outlined here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6837#issuecomment-713580310

We should also test with:

1. Positive fees.
2. Actual discount codes (not negative fees, but proper coupons).
3. Anything else we can think of?

### DIFF
--- a/includes/gateways/paypal-standard.php
+++ b/includes/gateways/paypal-standard.php
@@ -289,12 +289,21 @@ function edd_process_paypal_purchase( $purchase_data ) {
 			}
 		}
 
+		$price_before_discount = $purchase_data['price'];
 		if ( $discounted_amount > '0' ) {
 			$paypal_args['discount_amount_cart'] = edd_sanitize_amount( $discounted_amount );
+
+			/*
+			 * Add the discounted amount back onto the price to get the "price before discount". We do this
+			 * to avoid double applying any discounts below.
+			 * @link https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6837
+			 */
+			$price_before_discount += $paypal_args['discount_amount_cart'];
 		}
 
-		if( $paypal_sum > $purchase_data['price'] ) {
-			$difference = round( $paypal_sum - $purchase_data['price'], 2 );
+		// Check if there are any additional discounts we need to add that we haven't already accounted for.
+		if( $paypal_sum > $price_before_discount ) {
+			$difference = round( $paypal_sum - $price_before_discount, 2 );
 			if( ! isset( $paypal_args['discount_amount_cart'] ) ) {
 				$paypal_args['discount_amount_cart'] = 0;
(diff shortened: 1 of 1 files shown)


robincornett (approved):
- RCP Member Discounts: I am able to replicate the issue and confirm that the fix works with the RCP Member Discounts extension.
- Discounts Pro: I'm not able to replicate the issue with Discounts Pro. The purchase amount is the same with and without this branch.
- Manual Fee (discount): I do experience the duplicated discount and can confirm that it's fixed when manually adding a discount using `EDD()->fees->add_fee()`.
- Manual Fee (fee): the fee is not duplicated with the master branch; it is correctly applied with the issue branch.
- Standard EDD Discount is correct in both branches.
- With a positive manual fee and a standard discount, the purchase total is correct.

Will keep testing some combinations, but the singletons are working for me.

ashleyfae:
> * Discounts Pro: I'm not able to replicate the issue with Discounts Pro. The purchase amount is the same with and without this branch.

I wasn't able to reproduce it with Discounts Pro either. We had an issue for it because customers were having the issue, but seems like reproducing it was hit or miss. :thinking:

LisaCee (commented):
I haven't yet had a chance to test with RCP discounts, but so far:

- [x] Discounts (percentage and fixed) have no duplicates in either branch
- [x] Positive fee has no duplicates in either branch
- [x] Negative fee has duplicate fee in 2.10, fixed here

PR #8158, opened 2020-10-22
Issue/8157 X Buttons Visible on All Screens

### LINKED ISSUE #8157: 3.0 - Orders - X Buttons Off Screen for Mobile
### Expected behavior
I expect all buttons to be visible on all screen sizes

### Actual behavior
On the New Order screen, the buttons to remove Discounts and Adjustments are off the screen

1) Create a new order with Discounts and/or Adjustments
2) Make the screen <= 782px 
3) See the red X move off the screen 
etc, etc

EDD Version (or branch):
3.0

LisaCee:
Fixes #8157 

Proposed Changes:
1. Add media query to offset buttons only on larger screens
2. Tweak media query to match when table responds

I'm not crazy about this solution, but it works for now.  

In the future I'd love to see this screen styled differently for mobile, with tablet+ (782px) looking more like the larger screens.  IMO, the layout is too spread out, but that's a different issue having to do with when the table responds.

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -688,7 +688,11 @@ table.edd-order-overview-summary--refund {
 	color: #a00;
 }
 
-@media screen and (min-width: 782px) {
+@media screen and (min-width: 783px) {
+
+	.edd-order-overview-summary__adjustments .removable .delete {
+		margin-left: -40px;
+	}
 
 	.edd-order-overview-summary__totals .total {
 		font-size: 150%;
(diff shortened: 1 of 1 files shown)


assets/css/edd-admin.css
```diff
@@ -688,7 +688,11 @@ table.edd-order-overview-summary--refund {
 	color: #a00;
 }
 
-@media screen and (min-width: 782px) {
+@media screen and (min-width: 783px) {
+
+	.edd-order-overview-summary__adjustments .removable .delete {
+		margin-left: -40px;
```
spencerfinnell, line 694: @LisaCee Let's do `-50px` so the text lines up with the other items below it.

spencerfinnell (changes requested):
I like this much simpler solution @LisaCee.

robincornett (approved):
This looks good to me. I can't test it with the adjustments row because I updated my local to the 5.6 beta and adjustments are not working, but assume someone else can check adjustments?

spencerfinnell (approved):

ashleyfae (approved):

PR #8160, opened 2020-10-22
issue/8159 - WordPress 5.6 Compatibility Adding Adjustments/Discounts to Manual Orders

### LINKED ISSUE #8159: 3.0 - Orders - WP 5.6 - Can't Add Adjustment/Discount
### Expected behavior

When using WordPress 5.6-alpha I should be able to add an Adjustment or Discount when creating an Order manually.

### Actual behavior

`VM1118 jquery.js:5762 Uncaught TypeError: Cannot read property 'originalEvent' of undefined`

1) `wp core update --version=nightly --force`
2) Create an Order manually.
3) Attempt to add an Adjustment.

PHP Version:

EDD Version (or branch):

WordPress Version: `5.6-alpha`

https://github.com/WordPress/WordPress/commit/5cee7eced0b691089a875d7e918d4380bac408af

spencerfinnell:
Fixes #8159

Proposed Changes:
1. Ensure WordPress 5.6 compatibility when adding Adjustments and Discounts to manually created Orders.


ashleyfae (approved):
Fixes the issue. :+1:

robincornett (approved):
This is fixing the issue for me.

Noting a change in the modal close button: the "Close" button text is visible and I do not think it was before. Could this be something which changed in 5.6? I see that the CSS is in Core:

image

spencerfinnell:
@robincornett Pushed a few changes that should stop the "Close" text from appearing.

robincornett (commented):
@spencerfinnell that fixed the text showing on the order creation modals, but not on the refund modal--and even after running `npm run build` I do not see the `!` that's in the refunds.php file.

But I'm wondering if this is a change we will need to add with every place a jQuery dialog is being used?

spencerfinnell:
@robincornett Pushed a fix for the Refund modal, thanks for catching that. It uses the same string now. 

These are the only places we are using jQuery UI Dialog as far as I am aware.

robincornett (approved):

PR #8161, opened 2020-10-23
issue/7967 Checkbox Width

### LINKED ISSUE #7967: 3.0 - Checkbox input is enlarged on mobile screens (reports screen)
### Expected behavior
Viewing the admin on small screens, I expect inputs to remain recognizable.

### Actual behavior
On screens below 480px wide, the "Exclude Taxes" checkbox is stretched:

image

1) Visit Downloads > Reports in 3.0
2) Open your browser's dev tools and emulate a phone sized screen
3) View any report with a checkbox filter (eg Exclude Taxes)

PHP Version: 7.3.5
EDD Version (or branch): 3.0
WordPress Version: 5.4.1

The report inputs, especially on the Reports > Exports screen, feel unfinished on mobile devices (borders missing where two inputs are expected to be next to each other, for example, and no padding between inputs). Related: #7966

LisaCee:
Fixes #7967 

Proposed Changes:
1. Update specificity in CSS selector

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -3488,7 +3488,6 @@ span.edd-to-and-from--separator {
 		margin: 2px 0;
 	}
 
-	#edd-filters a,
 	.download_page_edd-reports .button {
 		text-align: center;
 	}
@@ -3512,13 +3511,7 @@ span.edd-to-and-from--separator {
 	.download_page_edd-reports .inside input[type="text"],
 	.download_page_edd-reports .inside select,
 	.download_page_edd-reports .inside input[type="submit"],
-	.download_page_edd-reports .inside .button,
-	#system-info-textarea,
-	#edd-add-discount input[type="text"],
-	#edd-edit-discount input[type="text"],
-	#edd-filters > input:not([type="radio"]),
-	#edd-filters select,
-	#edd-filters a {
+	.download_page_edd-reports .inside .button {
 		width: 100%;
 	}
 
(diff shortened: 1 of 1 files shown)


assets/css/edd-admin.css
```diff
@@ -3516,7 +3516,7 @@ span.edd-to-and-from--separator {
 	#system-info-textarea,
 	#edd-add-discount input[type="text"],
 	#edd-edit-discount input[type="text"],
-	#edd-filters input:not([type='radio']),
+	#edd-filters > input:not([type="radio"]),
```
ashleyfae, line 3519: I'm not sure that change is applying to any inputs anymore. I wonder if this would be better:

```suggestion
	#edd-filters input:not([type="radio"]):not([type="checkbox"]),
```

However, this also kind of has me wondering... what's that line for in the first place? Do we have any input types other than text that we're _actually_ trying to capture here? `number` maybe? Instead of targeting all inputs except radio/checkbox, maybe it's better to only target the ones we actually want.

robincornett: I do think that can safely be removed, as well as the `#edd-filters a` line possibly as well (doesn't do anything since the mobile app link is inside of a `span`).

The `#edd-filters` are also used on the Downloads > Orders screen, and there are some mobile issues there as well (specifically with `#edd-filters #edd-advanced-filters.open .inside`; I think the relative position on `#edd-filters #edd-advanced-filters` is causing problems there), but only one `number` input which might be affected by removing this.

spencerfinnell: I'd like to see the rules removed if possible. There is a lot of CSS fighting itself around these filters, so the more we can strip back the better.

robincornett (approved):

spencerfinnell (approved):

ashleyfae (approved):

PR #8166, opened 2020-10-26
Prevent errors when importing payments with non-existing discounts

### LINKED ISSUE #8145: 3.0 - Error getting property `id` of non-object when importing orders
### Expected behavior
When I import orders, I expect to not have errors.

### Actual behavior
Testing the PR for #8137, I exported a CSV from a 2.9 site. The import stopped when it encountered an issue with one of the orders:

```
PHP Notice:  Trying to get property 'id' of non-object in \wp-content\plugins\easy-digital-downloads.30\includes\payments\class-edd-payment.php on line 890
PHP Fatal error:  Uncaught Error: Call to a member function get_discounted_amount() on bool in \wp-content\plugins\easy-digital-downloads.30\includes\payments\class-edd-payment.php:893
```

I'm guessing this is related to discounts which do not exist on the new site.
1) Run an export on a site which has discounts registered and applied to some orders.
2) Import those orders into a separate site (use `issue/8137`) which does not have those discounts registered.
3) When the import hits an order with a discount which does not exist in the new site, the import will error/stall.

PHP Version: 7.4.1
WordPress Version: 5.5.1

This could be fixed by not using `edd_add_order_adjustment` if the discount object is not found, but I'm thinking that is likely to result in inaccurate order information--my guess is that we probably need to allow for a discount object to not exist, and be added to the order as if it were a manual discount, which also would not have an ID.

This could theoretically be part of the PR for #8137, but I thought it worth reporting as a separate issue as this code affects more than just imports.

robincornett:
Fixes #8145

Proposed Changes:
1. Make the early return logic consistent when `edd_get_discount_by` is used.
2. In `EDD_Payment`, if a discount object does not exist, but a discount does, add it as a fee instead (discounts are expected to exist in the `edd_adjusments` table).

<del>@ashleyfae although the checks for the discounts and taxes are happening in the `EDD_Payment` class, I feel uncomfortable automatically adding the fee/tax rate (haven't started on that one yet) there because of potential ripple effects. However, I'm not sure if I can duplicate the checks in the import class, or if that would be more appropriate at all. So setting this up as a draft to ask for feedback (the non-object error is resolved, at least, but imported orders don't feel correct).</del> Currently discounts and taxes are being added to the order (minimally) so going with this.

Note: if a fee is added to the order using the fees class, it does not appear to be accounted for at all in the CSV file, so I'm not sure how to account for that, but I'm not sure that's an issue specific to 3.0.

**To Test:**
1. This is best done with two different stores, an export store and an import store.
2. In the export store, generate some orders which include taxes.
3. Also generate some orders which use a discount code _which does not exist_ in the import store.
4. Export a CSV of orders.
5. Import the CSV into the store. If you want to see the breakage, do the import with `release/3.0` ... the import will error out because the discount code does not exist. Reset the store and then import using this branch; all orders should import, and the discounts/taxes should show on the orders.

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -2198,6 +2198,10 @@ class EDD_Payment {
 							/** @var EDD_Discount $discount */
 							$discount = edd_get_discount_by( 'code', $discount );
 
+							if ( false === $discount ) {
+								continue;
+							}
+
 							$adjustments = $this->order->adjustments;
 
 							$found_discount = array_filter( $adjustments, function( $adjustment ) use ( $discount ) {
@@ -2216,9 +2220,12 @@ class EDD_Payment {
 									'amount' => $this->subtotal - $discount->get_discounted_amount( $this->subtotal ),
 								) );
 
+								continue;
+							}
+
 							// Add the discount as an adjustment.
-							} else {
-								edd_add_order_adjustment( array(
+							edd_add_order_adjustment(
+								array(
 									'object_id'   => $this->ID,
 									'object_type' => 'order',
 									'type_id'     => $discount->id,
@@ -2226,8 +2233,8 @@ class EDD_Payment {
 									'description' => $discount->code,
 									'subtotal'    => $this->subtotal - $discount->get_discounted_amount( $this->subtotal ),
 									'total'       => $this->subtotal - $discount->get_discounted_amount( $this->subtotal ),
-								) );
-							}
+								)
+							);
 						}
 					}
 
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -658,7 +658,9 @@ class EDD_Payment {
 			);
 
 			foreach ( $order_meta as $key => $value ) {
-				edd_add_order_meta( $order_id, $key, $value );
+				if ( ! empty( $value ) ) {
+					edd_add_order_meta( $order_id, $key, $value );
+				}
 			}
 
 			$this->new = true;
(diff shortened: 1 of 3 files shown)


includes/payments/class-edd-payment.php
```diff
@@ -884,6 +886,20 @@ public function save() {
...
+									array(
+										'object_id'   => $this->ID,
+										'object_type' => 'order',
+										'type'        => 'fee',
+										'description' => $discount,
+										'subtotal'    => floatval( $this->total - $cart_subtotal - $this->tax ),
+										'total'       => floatval( $this->total - $cart_subtotal - $this->tax ),
+									)
+								);
+								continue;
```
ashleyfae, line 900: Instead of `continue`, can you convert the below `edd_add_order_adjustment()` call into an `else`? I think having an if/else here provides more clarity than a continue.

Or another option would be to build up the shared arguments for `edd_add_order_adjustment()` (object_id, object_type, description). Then use an if/else to build up the conditional arguments. Then have just one `edd_add_order_adjustment()` call at the end.

includes/payments/class-edd-payment.php
```diff
@@ -2212,18 +2232,21 @@ public function update_meta( $meta_key = '', $meta_value = '', $prev_value = ''
 									'amount' => $this->subtotal - $discount->get_discounted_amount( $this->subtotal ),
 								) );
 
+								continue;
+							}
+
 							// Add the discount as an adjustment.
-							} else {
-								edd_add_order_adjustment( array(
+							edd_add_order_adjustment(
```
ashleyfae, line 2239: I prefer the if/else approach. I find it easier to read. This whole method is so massive that it's easy to lose track of what the `continue` is actually for.

ashleyfae (changes requested):

ashleyfae (approved):

PR #8168, opened 2020-10-26
Support returning an integer when counting payments via EDD_Payments_Query #7056

### LINKED ISSUE #7056: 3.0: EDD_Payments_Query should respect the edd_get_orders() 'count' argument
### Expected behavior

When I attempt to query for a count of payments, I expect to get an integer.
### Actual behavior
When I attempt to query for a count of payments, I get an array of results, a warning and a notice:

```log
[30-Nov-2018 01:23:29 UTC] PHP Notice:  Array to string conversion in /path/to/edd-recurring/templates/shortcode-subscriptions.php on line 67
[30-Nov-2018 01:23:29 UTC] PHP Warning:  Invalid argument supplied for foreach() in /path/to/easy-digital-downloads/includes/payments/class-payments-query.php on line 208
```

```php
<?php
$count = edd_get_payments( array( 'count' => true ) );

or

$payments = new EDD_Payments_Query( array( 'count' => true ) );
$count = count( $payments->payments );
?>
```

EDD Version (or branch): 3.0

Looks like we just need to check if `$this->items` is numeric and if it is, return that. We could also explicitly check for the presence of the 'count' argument.

Related: https://github.com/easydigitaldownloads/edd-recurring/issues/1018

ashleyfae:
Fixes #7056

Proposed Changes:
1. Support the `count` argument and return an integer when passed through.
2. Add unit test covering this scenario.

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -155,7 +155,7 @@ class EDD_Payments_Query extends EDD_Stats {
 	 * @since 1.8
 	 * @since 3.0 Updated to use the new query classes and custom tables.
 	 *
-	 * @return EDD_Payment[]|EDD\Orders\Order[]
+	 * @return EDD_Payment[]|EDD\Orders\Order[]|int
 	 */
 	public function get_payments() {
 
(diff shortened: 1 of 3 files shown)


robincornett (approved):

spencerfinnell (approved):

LisaCee (commented):
The logic makes sense, but I'm not sure how to test it.

Issue #7435, opened 2019-11-01
Unit tests not compatible with PHPUnit 7


ashleyfae:
### Expected behavior
I expect to be able to run PHPUnit tests locally with PHPUnit version 7.

### Actual behavior
Unit tests fail with the following error:

> Fatal error: Declaration of SpeedTrapListener::addError(PHPUnit_Framework_Test $test, Exception $e, $time) must be compatible with PHPUnit\Framework\TestListener::addError(PHPUnit\Framework\Test $test, Throwable $t, float $time): void in C:\wamp\www\edd3\wp-content\plugins\easy-digital-downloads\tests\phpunit\speed-trap-listener.php on line 7

### Steps to reproduce the behavior
1) Use PHPUnit version 7.
2) Run EDD's unit tests.
3) See above error.

### Information (if a specific version is affected):
PHP Version: 7.3.7

EDD Version (or branch): both `master` and `release/3.0`

WordPress Version: 

Any other relevant information: It looks like our Travis config uses PHPUnit 5.7, and the unit tests were presumably written to be compatible with that. My guess is there's a compatibility issue and some stuff would need to change to be compatible with PHPUnit version 7. At some point it's worth updating our Travis config and tests to use PHPUnit version 7, as version 5 is no longer supported.

https://phpunit.de/supported-versions.html

PHPUnit version 7 EOL is February 7 2020, however WordPress core does not yet support version 8.

cklosowski:
As a note, one of the reasons was WP core support. Since we run a matrix
from WP 4.4 (4.7 for 3.0), those older WP versions have limited PHPUnit
support.

spencerfinnell:
Trac ticket for PHPUnit 8.x support: https://core.trac.wordpress.org/ticket/46149

Handbook recommends PHPUnit 7.x: https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/

If 3.0 is bumped to require WordPress 4.9.x is there a need to support the older versions? Is there an easy way to see which versions of WP are compatible with which versions of PHPUnit?

cklosowski:
If we can move to PHPUnit 7.x that'd be awesome, would just have to rewrite a bunch of our framework I think.

ashleyfae:
PR https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8180 makes tests pass with PHPUnit 7.x if you manually remove the speed trap listener.

cklosowski:
[...]

PR #8180, opened 2020-10-29
Get unit tests usable with PHPUnit 7.0

ashleyfae:
The goal is to get the unit tests _mostly_ working with PHPUnit 5.x and 7.0.

Related: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7435

This PR doesn't entirely fix that issue; you still have to manually remove the speed trap listener. But the goal is to make it so if you _do_ remove that speed trap, the unit tests pass.

---

Update: With these changes the tests are now passing for me locally. I've added some notes throughout the code to explain why I made each change.

@robincornett Please run these tests on your end and let me know if there are any failures.

### DIFF
--- a/tests/orders/tests-trash.php
+++ b/tests/orders/tests-trash.php
@@ -125,8 +125,4 @@ class Trash_Tests extends \EDD_UnitTestCase {
 		$this->assertSame( 'trash', $refunds[0]->status );
 	}
 
-	public function test_restore_order_with_refunds() {
-
-	}
-
 }
--- a/tests/phpunit/class-edd-unittestcase.php
+++ b/tests/phpunit/class-edd-unittestcase.php
@@ -124,7 +124,11 @@ class EDD_UnitTestCase extends WP_UnitTestCase {
 		}
 
 		foreach ( $actual as $item ) {
-			\PHPUnit_Framework_Assert::assertThat( $item, $constraint );
+			if ( class_exists( '\PHPUnit\Framework\Assert' ) ) {
+				\PHPUnit\Framework\Assert::assertThat( $item, $constraint );
+			} else {
+				\PHPUnit_Framework_Assert::assertThat( $item, $constraint );
+			}
 		}
 	}
 
--- a/tests/tests-api.php
+++ b/tests/tests-api.php
@@ -293,9 +293,9 @@ class Tests_API extends EDD_UnitTestCase {
 		$this->assertArrayHasKey( 'earnings', $out['products'][0]['stats']['monthly_average'] );
 
 		$this->assertEquals( '60', $out['products'][0]['stats']['total']['sales'] );
-		$this->assertEquals( '229.43', $out['products'][0]['stats']['total']['earnings'] );
+		$this->assertEquals( '229.430000', $out['products'][0]['stats']['total']['earnings'] );
 		$this->assertEquals( '60', $out['products'][0]['stats']['monthly_average']['sales'] );
-		$this->assertEquals( '229.43', $out['products'][0]['stats']['monthly_average']['earnings'] );
+		$this->assertEquals( '229.430000', $out['products'][0]['stats']['monthly_average']['earnings'] );
 	}
 
 	public function test_get_products_pricing() {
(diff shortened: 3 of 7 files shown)


tests/orders/tests-trash.php
```diff
@@ -125,8 +125,4 @@ public function test_trash_order_with_refunds() {
 		$this->assertSame( 'trash', $refunds[0]->status );
 	}
 
-	public function test_restore_order_with_refunds() {
```
ashleyfae, line 128: Removed this because it wasn't doing anything.

tests/phpunit/class-edd-unittestcase.php
```diff
@@ -124,7 +124,11 @@ public function assertContainsOnlyType( $type, $actual ) {
 		}
 
 		foreach ( $actual as $item ) {
-			\PHPUnit_Framework_Assert::assertThat( $item, $constraint );
+			if ( class_exists( '\PHPUnit\Framework\Assert' ) ) {
```
ashleyfae, line 127: Allows compatibility with version 7.

tests/tests-api.php
```diff
@@ -169,6 +169,8 @@ public static function wpSetUpBeforeClass() {
 	public function setUp() {
 		parent::setUp();
 
+		wp_set_current_user( self::$user_id );
```
ashleyfae, line 172: In the parent `tearDown()` function, the current user was being reset to `0` by WordPress core. This meant that all the `EDD()->api->update_key( self::$user_id );` calls were failing, due to `current_user_can()` returning false.

tests/tests-api.php
```diff
@@ -187,6 +189,9 @@ public function setUp() {
 	public function tearDown() {
 		parent::tearDown();
 
+		// Revoke key to ensure `update_key()` will generate a new one.
+		EDD()->api->revoke_api_key( self::$user_id );
```
ashleyfae, line 193: Added this due to numerous calls in tests to `EDD()->api->update_key( self::$user_id );` . The issue was that if the API key already existed when `EDD()->api->update_key( self::$user_id );` was run, the key was actually just being revoked and not generated.

tests/tests-api.php
```diff
@@ -253,12 +258,14 @@ public function test_get_queried_version() {
 
 		try {
 			self::$api->process_query();
-			$this->assertEquals( 'v1', self::$api->get_queried_version() );
+		} catch ( WPDieException $e ) {}
```
ashleyfae, line 261: All tests that have this change were due to:

`self::$api->process_query();` includes a `wp_die()` at the end, hence catching the exception. However, all the assertions were being made inside the `try { }` block. But the flow is:

- `process_query()`
- Triggers exception
- Therefore following assertions were never being made.

Moving the assertions outside the try/catch block ensures they actually do something. All these were "risky" tests on my end due to no assertions actually being made.

tests/tests-api.php
```diff
@@ -293,9 +300,9 @@ public function test_get_product_stats() {
 		$this->assertArrayHasKey( 'earnings', $out['products'][0]['stats']['monthly_average'] );
 
 		$this->assertEquals( '60', $out['products'][0]['stats']['total']['sales'] );
-		$this->assertEquals( '229.43', $out['products'][0]['stats']['total']['earnings'] );
+		$this->assertEquals( '229.430000', $out['products'][0]['stats']['total']['earnings'] );
```
ashleyfae, line 303: Numerous rounding issues like this were fixed. From what I gather, PHPUnit 5.x was passing these, possibly due to less strict requirements on decimal places. It was treating `229.43` as equal to `229.430000` and thus passing the test. PHPUnit 7.x fails these. So this is (hopefully) just updating the tests to check against the actual true values rather than rounded ones.

tests/tests-scripts.php
```diff
@@ -101,6 +101,9 @@ public function test_register_styles_checkout_ssl() {
 		edd_register_styles();
 
 		$this->go_to( '/' );
+
+		$this->assertTrue( wp_style_is( 'edd-styles', 'registered' ) );
```
ashleyfae, line 105: I wasn't sure what this test was actually trying to accomplish. I just added this in to make it so at least one assertion was happening here. (Without assertions it was a "risky" test.)

robincornett (approved):
All tests pass locally with PHPUnit 5.7.27.

PR #8187, opened 2020-11-04
Validate order country against global list during migration #8028

### LINKED ISSUE #8028: 3.0 - Sanitize address during migration to remove invalid country values
### Expected behavior

I expect to not have any customer addresses with invalid data. If a certain address column is not known, then I expect it to be blank.

### Actual behavior

I have 8 old order addresses (from 2013 - 2014) being imported with `country` set to `*`. They were saved as `*` in `_edd_payment_meta` so that's why they exist like that in 3.0 post-migration. But perhaps this is a good opportunity to fix that data.

Here's the serialized payment meta:

```
{s:5:"line1";s:0:"";s:5:"line2";s:0:"";s:4:"city";s:0:"";s:3:"zip";s:0:"";s:7:"country";s:1:"*";s:5:"state";s:2:"AL";}
```

So everything is empty, but `state` is set to `AL`.

Not really an exact way to reproduce this because it's due to my data set. But you can spoof it by manually editing `_edd_payment_meta` in 2.9, setting the country on the address to `*`, then running an import.

PHP Version:

EDD Version (or branch): Discovered while testing `issue/7254`, but it's not actually related to that branch. Just a `release/3.0` thing in general.

WordPress Version:

ashleyfae:
Fixes #8028

Proposed Changes:
1. Only store the address country if it exists in `edd_get_country_list()`

To test:

1. Start on EDD 2.9.
2. Manually update the address on an order to be something invalid. Can be done by manually running this PHP:
```php
$order_id = 123; // Order/payment ID number
$meta = get_post_meta( $order_id, '_edd_payment_meta', true);
$meta['user_info']['address']['country'] = '*'; // Or set to something totally random
update_post_meta( $order_id, '_edd_payment_meta', $meta );
```
3. Upgrade to 3.0.
4. After upgrading, check the `wp_edd_order_addresses` value for that order ID. The `country` should be blank.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -585,7 +585,9 @@ class Data_Migrator {
 			'address2'     => isset( $user_info['address']['line2'] )   ? $user_info['address']['line2']   : '',
 			'city'         => isset( $user_info['address']['city'] )    ? $user_info['address']['city']    : '',
 			'region'       => isset( $user_info['address']['state'] )   ? $user_info['address']['state']   : '',
-			'country'      => isset( $user_info['address']['country'] ) ? $user_info['address']['country'] : '',
+			'country'      => isset( $user_info['address']['country'] ) && array_key_exists( strtoupper( $user_info['address']['country'] ), edd_get_country_list() )
+				? $user_info['address']['country']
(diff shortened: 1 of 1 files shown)


robincornett (approved):
Followed the testing steps and the migrated order country was blank.

PR #8194, opened 2020-11-09
Set description to discount code #8186

### LINKED ISSUE #8186: 3.0 - Migration - "wpdb::prepare was called incorrectly" when migrating payments
## Bug Report
### Expected behavior

I expect to not get any errors/notices when running the migration.

### Actual behavior

I got the following notice when running the migration via CLI:

```
Migrating Payments  0  % [=>                                                                                                                                                                                                                                               ] 0:01  / 20:04
Notice: wpdb::prepare was called <strong>incorrectly</strong>. Unsupported value type (object). Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 4.8.2.) in /var/www/eddexport/wp-includes/functions.php on line 5229
```

Unfortunately I don't have a full stack trace. Just logging this for now so I can investigate it more later.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8186 

Proposed Changes:
1. Set description to discount code property.

To test:

Do a migration from 2.9 to 3.0. Your data set just needs to include an order that used a discount code.


robincornett (approved):
Ran a 2.9-3.0 migration and discount codes are showing properly in the `edd_order_adjustments` table; no errors.

PR #8201, opened 2020-11-11
Update customer stats via ajax when recalculating single customer

### LINKED ISSUE #8004: 3.0 - Customer lifetime stats do not appear to be updated when they have
### Expected behavior
When I use the "Recount Stats" button (under Tools) for a single customer, and then click the Profile link to see the profile, I expect to see the correct/updated information.

### Actual behavior
Because the entire customer profile is loaded, and then shown selectively using JavaScript, the customer lifetime stats will not show the updated data until the page is reloaded.

It helps if you update a customer's `purchase_value` and `purchase_count` directly in the database to something obviously wrong first.

1) Visit the customer details screen for that customer.
2) Under Tools, click the "Recount Stats" button.
3) Click on Profile. The numbers are not updated.

PHP Version: 7.3.5
EDD Version (or branch): this may need to be done with issue/7947, which is updating customer lifetime stats
WordPress Version: 5.4.1

robincornett:
Fixes #8004

Proposed Changes:
1. Adds a new class `$data` variable to `EDD_Tools_Recount_Single_Customer_Stats`.
2. On success, the new purchase count and value are sent to the JavaScript.
3. The `span` which match the data keys are updated with the new values.

My thought on how the script is updating the matching `span` is that this is something that other export classes could use if needed--the array of data returned would just have to set the keys to match an element with a matching class (prepended with `edd_`). In that case, it might be better to define the new `$data` variable in `EDD_Batch_Export`, but this is a start.

To test:
1. Visit the Customer Details screen for any customer.
2. Use your browser's dev tools to change the numbers for the "Completed Sales" and "Lifetime Value". (You could also do this directly in the database, but this felt easier.)
3. Go to Tools for the customer.
4. Click the Recount Stats button.
5. When the success message appears, click on the customer Profile link.
6. The stats should be updated to the new/correct values.


includes/admin/customers/customers.php
```diff
@@ -597,12 +597,12 @@ function edd_customers_view( $customer = null ) {
 			<li>
 				<a href="<?php echo admin_url( 'edit.php?post_type=download&page=edd-payment-history&customer=' . $customer->id ); ?>">
 					<span class="dashicons dashicons-cart"></span>
-					<?php printf( _n( '%s Completed Sale', '%s Completed Sales', $customer->purchase_count, 'easy-digital-downloads' ), esc_html( number_format_i18n( $customer->purchase_count, 0 ) ) ); ?>
+					<?php printf( _n( '<span class="edd_purchase_count">%s</span> Completed Sale', '<span class="edd_purchase_count">%s</span> Completed Sales', $customer->purchase_count, 'easy-digital-downloads' ), esc_html( number_format_i18n( $customer->purchase_count, 0 ) ) ); ?>
```
spencerfinnell, line 600: The HTML should wrap around the placeholder so translators do not accidentally remove it.

```suggestion
					<?php
					printf(
						_n(
							'%s Completed Sale',
							'%s Completed Sales',
							$customer->purchase_count,
							'easy-digital-downloads'
						),
						(
							'<span class="edd_purchase_count">' .
							esc_html( number_format_i18n( $customer->purchase_count, 0 ) ) .
							'</span>'
						)
					);
					?>
```

includes/admin/customers/customers.php
```diff
@@ -597,12 +597,12 @@ function edd_customers_view( $customer = null ) {
...
 				</a>
 			</li>
 			<li>
 				<span class="dashicons dashicons-chart-area"></span>
-				<?php echo esc_html( edd_currency_filter( edd_format_amount( $customer->purchase_value ) ) ); ?> <?php esc_html_e( 'Lifetime Value', 'easy-digital-downloads' ); ?>
+				<?php echo '<span class="edd_purchase_value">' . esc_html( edd_currency_filter( edd_format_amount( $customer->purchase_value ) ) ) . '</span>'; ?> <?php esc_html_e( 'Lifetime Value', 'easy-digital-downloads' ); ?>
```
spencerfinnell, line 605: Would be nice to update this similarly to the "Completed Sale" string for better translatability.

assets/js/admin/tools/export/index.js
```diff
@@ -59,6 +59,11 @@ const EDD_Export = {
 					} else if ( response.success ) {
 						const success_message = response.message;
 						notice_wrap.html( '<div id="edd-batch-success" class="updated notice"><p>' + success_message + '</p></div>' );
+						if ( response.data ) {
+							$.each( response.data, function ( key, value ) {
+								$( '.edd_' + key ).empty().append( value );
```
spencerfinnell, line 64:
```suggestion
								$( '.edd_' + key ).html( value );
```

`.html()` will replace any of the existing content.

spencerfinnell (changes requested):

includes/admin/customers/customers.php
```diff
@@ -597,12 +597,40 @@ function edd_customers_view( $customer = null ) {
 			<li>
 				<a href="<?php echo admin_url( 'edit.php?post_type=download&page=edd-payment-history&customer=' . $customer->id ); ?>">
 					<span class="dashicons dashicons-cart"></span>
-					<?php printf( _n( '%s Completed Sale', '%s Completed Sales', $customer->purchase_count, 'easy-digital-downloads' ), esc_html( number_format_i18n( $customer->purchase_count, 0 ) ) ); ?>
+					<?php
+					printf(
+						_n(
+							/* translators: the customer's lifetime number of sales */
```
spencerfinnell, line 603:
```suggestion
						/* translators: the customer's lifetime number of sales */
						_n(
```

spencerfinnell (changes requested):

spencerfinnell (approved):

ashleyfae (approved):

Issue #8118, opened 2020-09-22
3.0 - Reloading migration screen during upgrade is not recoverable


robincornett:
### Expected behavior
If I do something I should not do during the UI migration process--namely, refresh the migration screen or leave it--I would hope that I could still finish the migration.

### Actual behavior
While checking something else, I refreshed the migration screen while it was in the middle of migrating the orders (yes, I know I should not have done it and there is a message, but I had reasons). So now my migration screen looks like this:

![image](https://user-images.githubusercontent.com/4853066/93895291-aaa24080-fcbd-11ea-8bd8-32c93bf00244.png)

Everything after "Update Orders" says, "Please complete the previous steps..." but I cannot interact with anything on the screen, so as a user, I'm stuck unless I switch to using CLI. Note also that the orders message says, "Your orders database has been upgraded." (The new database tables have been created and are partially populated, but the migration is not complete--not all orders have been migrated.)

### Steps to reproduce the behavior
1) Update a site from EDD 2.9 to 3.0.
2) Visit the migration screen (`/wp-admin/index.php?page=edd-upgrades&edd-upgrade=v30_migration`).
3) Begin the migration.
4) Interrupt the migration by refreshing the page (I did it during step 3, Orders).

### Information (if a specific version is affected):
PHP Version: 7.4.1
EDD Version (or branch): release/3.0
WordPress Version: 5.5.1

Related: #6880

ashleyfae:
A JavaScript alert to not leave the page is another (or additional option).

PR #8202, opened 2020-11-12
Update 3.0 migration UI #6880

### LINKED ISSUE #6880: 3.0: Upgrade Screen is Confusing

### LINKED ISSUE #7885: 3.0 - Migration - Mention CLI command on the upgrade UI page

I'd like the `wp edd v30_migration` CLI command to be mentioned on the migration UI. Maybe somewhere in between the existing admin notice and `Step 1`

Screenshot from 2020-05-13 14-14-23

### Justification or use case

The command is _only_ mentioned inside the admin notice if you click on `Learn more about this upgrade`. If you just `click here to start the upgrade`, you'd never know there even was a CLI command.

ashleyfae:
Fixes #6880, fixes #7885 

Proposed Changes:
1. Updates the v3 migration UI to make it a bit simpler. Steps aren't revealed until after the button is pressed, steps have been simplified and descriptions removed, progress bars have been removed in favour of a simple numeric percentage, legacy data removal is completely hidden until all other migrations are complete.
2. The actual JavaScript has been decoupled from the export JS to allow us to customize it a bit more.
2. Mentions the WP-CLI command on the upgrade page ( #7885 )
3. Adds in some step recovery logic for #8118 . This is commented out for now though because the upgrades themselves will need updating to ensure duplicate data does not get migrated.

Screenshot from 2020-11-12 11-13-03

This isn't 100% done yet. I'll probably tweak the CSS a little to make the checkboxes green and to actually spin the spinner. **Right now I'm looking for overall concept feedback.** Do we like this UI better? Are there any major changes we should make about it? Once we have the overall concept nailed down we can polish off details like CSS and wording.

**Important testing note:** Right now I have this set up so it doesn't actually migrate any data. From these two lines: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/6880/includes/admin/upgrades/v3/upgrade-actions.php#L67-L68 I've just done that so the process can easily be tested, then you can refresh the page to do it over again.


spencerfinnell:
@ashleyfae This is a huuuuge improvement. I love everything about this. I absolutely think this is the direction the screen should be going.

assets/js/admin/upgrades/v3/index.js
```diff
@@ -0,0 +1,182 @@
+const EDD_v3_Upgrades = {
```
spencerfinnell, line 1: It doesn't make a difference in the end, but this object pattern is not needed with Webpack. Everything is taken out of the global namespace automatically with the module imports so you can just use normal functions if you want.

robincornett (commented):
Loving this so far, especially removing redundancies (and chances of errors).

My vote would be to wrap everything starting with, "Easy Digital Downloads needs to perform upgrades..." to the end in `postbox/inside` div elements to get the "box" appearance visually.

I kind of prefer the progress bar to the percentages changing, but it's a lot of progress bars and real estate involved, so I could be talked out of it for sure. What might be fun (for lack of a better word), but maybe not possible, would be to get the initial counts for all steps before running anything, and just having one big progress bar for the entire migration (how this would work with an interrupted migration, I have no idea, plus this would entail more math to update the individual percentages and the overall percentage). The taxes/discounts go so much more quickly than things like orders and then notes...it might be nice to have kind of an overview of the whole process. If it would even be possible.

But overall I think this is a fantastic direction to take this.

spencerfinnell:
I disagree about the metabox usage -- I think the current style matches the core "Updates" screen well.

assets/js/admin/upgrades/v3/index.js
ashleyfae: Thanks for letting me know! We'll see if I can be bothered to refactor it. :D

robincornett:
Noting here that the [data migrator uses `edd_add_customer_address`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/6880/includes/admin/upgrades/v3/class-data-migrator.php#L63) and it should be using `edd_maybe_add_customer_address` instead (Ashley noted in a meeting).

It may not matter for most standard addresses (which get set to `billing` anyway) once the PR for #8039 is finalized. Actually, it may not matter at all, as this happens when migrating the customer addresses--unless a storefront has more than one address saved to `_edd_user_address` in the user meta? Migrating order addresses does already, I think, use the `maybe` function.

ashleyfae:
> Noting here that the [data migrator uses `edd_add_customer_address`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/6880/includes/admin/upgrades/v3/class-data-migrator.php#L63) and it should be using `edd_maybe_add_customer_address` instead (Ashley noted in a meeting).
> 
> It may not matter for most standard addresses (which get set to `billing` anyway) once the PR for #8039 is finalized. Actually, it may not matter at all, as this happens when migrating the customer addresses--unless a storefront has more than one address saved to `_edd_user_address` in the user meta? Migrating order addresses does already, I think, use the `maybe` function.

I'll update that when I focus on https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8118, which will likely be after this PR is eventually merged.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1426,13 +1445,16 @@ function edd_upgrade_render_v30_migration() {
 	<p>
 		<!-- @todo wording -->
 		<?php esc_html_e( 'Easy Digital Downloads needs to perform upgrades to your WordPress database. Your store data will be migrated to custom database tables to improve performance and efficiency. This process may take a while.', 'easy-digital-downloads' ); ?>
-		<strong><?php esc_html_e( 'Please create a full back up of your website before proceeding.', 'easy-digital-downloads' ); ?></strong>
+		<strong><?php esc_html_e( 'Please create a full backup of your website before proceeding.', 'easy-digital-downloads' ); ?></strong>
 	</p>
 
 	<p>
-		<?php printf( esc_html__( 'This migration can also be run via WP-CLI with the following command: %s', 'easy-digital-downloads' ), '<code>wp edd v30_migration</code>' ); ?>
+		<?php printf( esc_html__( 'This migration can also be run via WP-CLI with the following command: %s This is the recommended method for large sites.', 'easy-digital-downloads' ), '<code>wp edd v30_migration</code>' ); ?>
```
robincornett, line 1452:
```suggestion
		<?php printf( esc_html__( 'This migration can also be run via WP-CLI with the following command: %s. This is the recommended method for large sites.', 'easy-digital-downloads' ), '<code>wp edd v30_migration</code>' ); ?>
```

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,498 +1328,227 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
+					<input type="checkbox" id="edd-v3-remove-legacy-data-confirmation" class="edd-v3-migration-confirmation" name="backup_confirmation" value="1">
+					<?php esc_html_e( 'I have confirmed my store is operational and I have a backup of my website data.', 'easy-digital-downloads' ); ?>
+				</label>
+			</p>
+			<input type="hidden" name="_wpnonce" value="<?php echo esc_attr( wp_create_nonce( 'edd_process_v3_upgrade' ) ); ?>">
+			<input type="hidden" name="upgrade_key" value="v30_legacy_data_removed">
+			<button type="submit" class="button button-primary disabled" disabled>
+				<?php esc_html_e( 'Permanently Remove Legacy Data', 'easy-digital-downloads' ); ?>
+			</button>
+			<!-- @todo align better with button -->
```
robincornett, line 1537: Would it be possible to wrap the button and this span in a container div and then use some flexbox to align them vertically?

includes/admin/upgrades/v3/upgrade-actions.php
```diff
@@ -60,10 +64,10 @@ function edd_process_v3_upgrade() {
...
-	$was_processed       = false; // @todo remove                                                                                                                                                                                                                                                                                                                                                                                                       // @todo remove
-	$percentage_complete = $export->get_percentage_complete();
-	$percentage_complete = 100; // @todo remove
+	$was_processed       = $export->process_step();
+	//$was_processed       = false; // @todo remove
+	$percentage_complete = round( $export->get_percentage_complete(), 2 );
```
robincornett, line 69: This is resulting in percentages like `3.88%` ... is that intentional?

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,498 +1328,227 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
+				<br /><br />
+				<?php esc_html_e( 'You may now leave this page.', 'easy-digital-downloads' ); ?>
+			</p>
+		</div>
+		<?php
+		return;
+	}
+	?>
 	<div id="edd-migration-nav-warn" class="notice notice-info">
 		<p><?php echo wp_kses( __( '<strong>Important:</strong> Do not navigate away from this page until all upgrades complete.', 'easy-digital-downloads' ), array( 'strong' => array() ) ); ?></p>
```
robincornett, line 1442: Although I think this is technically correct wording, it feels a bit grammatically stilted, maybe? Could we change it to something like:

> **Important:** Do not navigate away from this page until all upgrades have completed.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,498 +1328,227 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
-								 <span class="message complete"><?php esc_html_e( 'Data migration complete.', 'easy-digital-downloads' ); ?></span>
-						</span>
-					<?php else: ?>
-						<button type="submit" disabled id="migrate-customer-notes-submit" class="button button-secondary"><?php esc_attr_e( 'Upgrade Database', 'easy-digital-downloads' ); ?></button>
-						&mdash; <?php esc_html_e( 'Legacy data has already been removed, migration is not possible at this time.', 'easy-digital-downloads' ); ?>
-					<?php endif; ?>
-
-					<input type="hidden" name="edd-export-class" value="EDD\Admin\Upgrades\v3\Customer_Notes" />
+	<div id="edd-v3-remove-legacy-data" <?php echo ! $can_remove_legacy_data ? 'class="edd-hidden"' : ''; ?>>
+		<h2><?php esc_html_e( 'Remove Legacy Data', 'easy-digital-downloads' ); ?></h2>
```
robincornett, line 1521: After I've removed the legacy data, I can reload the page and still get this section telling me to remove it. Could we add a check and say something like "Legacy data has already been removed; you're good to go!"?

includes/admin/upgrades/v3/upgrade-actions.php
ashleyfae: Are you referring to the two decimal places? Or what, exactly?

robincornett: Yes, sorry--I guess I was expecting all whole numbers for the percentages. So `4%` instead of `3.88%`.

ashleyfae: I think we need the decimals given that some migrations can be really huge. For example, on my test site when I delete legacy data, I'm below `0%` for quite a while. The decimals help reassure that it's actually processing.

includes/admin/upgrades/upgrade-functions.php
ashleyfae: Yeah still working on fine tuning that logic.

includes/admin/upgrades/upgrade-functions.php
ashleyfae: For sure. Was waiting to do any styling until the CSS webpack PR was merged.

robincornett:
Watching this run on a site with taxes (and therefore addresses) and was surprised to see a nearly instantaneous run on the "Customer Addresses" migration step. Wondering if the addresses being migrated during the orders process before this step means the addresses are already taken care of, or if this is handling addresses which aren't part of the orders and maybe my dataset just doesn't have them, and they just exist as part of the orders?

ashleyfae:
> Watching this run on a site with taxes (and therefore addresses) and was surprised to see a nearly instantaneous run on the "Customer Addresses" migration step. Wondering if the addresses being migrated during the orders process before this step means the addresses are already taken care of, or if this is handling addresses which aren't part of the orders and maybe my dataset just doesn't have them, and they just exist as part of the orders?

@robincornett Is it the same in release/3.0? Just asking because although this PR does address the UI and execution of migrations, it doesn't touch the actual steps themselves. So it shouldn't be changing what the customer address migration does or how quickly it runs.

ashleyfae:
Think this is ready for some more thorough testing/review/feedback.

assets/css/admin/upgrades/_v3.scss
```diff
@@ -0,0 +1,15 @@
+#edd-migration-progress {
+	.dashicons-yes {
```
robincornett, line 2: Could we tweak the color on the ellipses dashicon to be more of an "inactive" color? Just a lighter grey?

assets/css/admin/upgrades/_v3.scss
```diff
@@ -0,0 +1,15 @@
...
+		color: green;
+	}
+	.dashicons-update:before {
+		animation: rotation 2s infinite linear;
+		display: block;
+	}
+}
+
+#edd-v3-migration-remove-legacy-data-submit-wrap {
```
robincornett, line 11: On smaller screens, the `.wp-core-ui .button` has a `margin-bottom: 4px;` which is throwing off the vertical alignment slightly. If we could set that to have a 0 margin it would work on all screens.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,498 +1328,230 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
-					<?php endif; ?>
-
-					<input type="hidden" name="edd-export-class" value="EDD\Admin\Upgrades\v3\Customer_Notes" />
-				</span>
-				</form>
-			</div><!-- .inside -->
-		</div><!-- .postbox -->
+	<div id="edd-v3-migration-complete" <?php echo ! $can_remove_legacy_data ? 'class="edd-hidden"' : ''; ?>>
+		<p>
+			<?php esc_html_e( 'The data migration has been successfully completed. You may now leave this page or proceed to remove legacy data below.', 'easy-digital-downloads' ); ?>
```
robincornett, line 1516: I'm not sure what I would prefer, but I feel like I want this to be worded slightly differently. Trying to think of what I want without it being redundant with the explanation below.

robincornett (commented):
Overall, this looks awesome! A couple minor CSS quibbles/requests.

And you are correct, the release branch is the same timing on migrating customer addresses, so that must just be data I don't have.

includes/admin/upgrades/upgrade-functions.php
ashleyfae: Is there a specific part that's bothering you? Just so I can help think about what to tinker with.

ashleyfae: > Also wondering--and being lazy because I'm on a different branch and can't run the migration again right this moment--but if the migration part is complete, but the legacy data is still present, does the admin notice outside of the migration screen still persist? Mainly I'm wanting to make sure that if someone leaves this screen and dutifully checks their site data, that they can easily get back, and know that they do _need_ to get back and remove the legacy data.

The admin notice does not persist. However, you can return back to the same page and you'll still have the option to remove legacy data.

I don't think I want to persist the admin notice, because although removing legacy data is good for performance, it's not a required step. Also, some people might understandably want to wait a few days before removing it to make sure their store is perfectly operational. We shouldn't persist a notice that long.

I'd be more inclined to add something to the Tools page or similar that helps make it easier to find the "Remove Legacy Data" page again.

robincornett: That would be fine by me. I'm concerned that if a user navigates away (for example, to let it wait a few days, which totally makes sense), they're going to have a hard time finding that screen again to remove the legacy data and never be able to find it again (and never know either).

So maybe add a notice with a link to the tools screen if there is still legacy data (or are you saying to add it as a tool, because that would work as well)? And update this message to say something like:

> The data migration has been successfully completed. You may now remove the legacy data, or you can leave this page. If you leave this page without removing the legacy data, you can use the Legacy Data Removal tool anytime in the future on the Downloads > Tools screen.

ashleyfae: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8202/commits/f75667172c2f89e828e8046ad18171ddcdcc6450 adds the same migration tool to the tools page. It only appears (hopefully) if you've completed all migration steps but have not yet removed legacy data.

I made it exactly the same in both places (mostly) to try to reuse as much code as possible.

spencerfinnell (commented):
Nothing that requires changing, but a few UI/UX things I thought might help:

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,499 +1328,294 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
+						if ( true === $upgrade_statuses[ $upgrade_key ] ) {
+							?>
+							<span class="dashicons dashicons-yes"></span>
+							<span class="screen-reader-text"><?php esc_html_e( 'Upgrade complete', 'easy-digital-downloads' ); ?></span>
+							<?php
 						} else {
-							processNext( next_step_wrapper );
+							?>
+							<span class="dashicons dashicons-ellipsis"></span>
+							<span class="screen-reader-text"><?php esc_html_e( 'Upgrade pending', 'easy-digital-downloads' ); ?></span>
```
robincornett, line 1497: This text should change once the upgrade has completed, I think.

Wondering if it would be possible to have it set initially to "pending" or "waiting", and then "in progress", and then "complete"?

OR...wondering if this is necessary at all, since the percentage is read by screen readers...so it could just be removed?

robincornett: Noting that if I reload the page, the percentage is gone, and this span does read "Upgrade complete", so maybe we just need to adjust the script behavior a bit.

robincornett (commented):
This is looking really good, just some more quibbles.

I know it's not part of this PR, but I'm wondering about reducing the number per step for migrating orders--it feels like the numbers tick upwards more slowly for that step than the others; I'm guessing it's just because there is a ton of data to adjust for each order.

ashleyfae:
> I'm wondering about reducing the number per step for migrating orders--it feels like the numbers tick upwards more slowly for that step than the others; I'm guessing it's just because there is a ton of data to adjust for each order.

I'm not inclined to reduce it because some hosts can also get suspicious about more frequent hits to the ajax endpoint and can start rate limiting it or even blocking requests.

includes/admin/upgrades/upgrade-functions.php
robincornett: The legacy data tool does not show on the tools screen after I have used it to remove the legacy data.

However, it is showing on a "new" 3.0 site...not new, but it's not had 2.9 on it at all, and I did just reset the store data this morning. (It is a multisite setup and the other child sites are _not_ showing the tool, so ??)

includes/admin/upgrades/upgrade-functions.php
ashleyfae: I've pushed up a change that updates the screen reader text when the step completes.

The percentage is gone when you reload the page because I'm not going to make that work perfectly until I focus on step recovery. I'm going to do that in a separate issue once this PR gets merged.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,499 +1328,294 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
+	</form>
+	<?php endif
+
+	/*
+	 * Progress is only shown immediately if the upgrade is in progress. Otherwise it's hidden by default
+	 * and only revealed via JavaScript after the process has started.
+	 */
+	?>
+	<div id="edd-migration-progress" <?php echo count( array_filter( $upgrade_statuses ) ) ? '' : 'class="edd-hidden"'; ?>>
+		<ul>
```
robincornett, line 1478: I think this could reasonably be changed to an ordered list, although I'm not sure it matters significantly--since the WP admin does show the numbers, but not standard bullets. I was just thinking it might be nice to show the information such as "Step 1 of 8" and so on.

ashleyfae: I think I'm going to leave it. I don't think it's hugely important one way or another, but adding numbers will add a bit more "business" and ultimately the spirit of this PR is to keep it really simple.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,499 +1328,294 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
+		}
+	}
 
+	if ( $migration_complete ) {
+		?>
+		<div id="edd-migration-ready" class="notice notice-success">
+			<p>
+				<?php echo wp_kses( __( '<strong>Database Upgrade Complete:</strong> All database upgrades have been completed.', 'easy-digital-downloads' ), array( 'strong' => array() ) ); ?>
+				<br /><br />
+				<?php esc_html_e( 'You may now leave this page.', 'easy-digital-downloads' ); ?>
```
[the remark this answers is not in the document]
ashleyfae: I've added a link. I think the message on this page might be easier for support to work with rather than a redirect.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -1328,499 +1328,297 @@ function edd_include_file_download_log_migration_batch_processor( $class = '' )
...
+			</p>
+		</div>
 
-	<div id="edd-migration-ready" class="notice notice-success" style="display: none;">
 		<p>
-			<?php echo wp_kses( __( '<strong>Database Upgrade Complete:</strong> All database upgrades have been completed.', 'easy-digital-downloads' ), array( 'strong' => array() ) ); ?>
-			<br /><br />
-			<?php esc_html_e( 'You may now leave this page.', 'easy-digital-downloads' ); ?>
+			<a href="<?php echo esc_url( admin_url() ); ?>" class="button button-primary">Return to the dashboard</a>
```
robincornett, line 1439:
```suggestion
			<a href="<?php echo esc_url( admin_url() ); ?>" class="button button-primary"><?php esc_html_e( 'Return to the dashboard', 'easy-digital-downloads' ); ?></a>
```

robincornett (approved):

LisaCee (commented):
I don't understand most of what's going on under the hood, but from a user's perspective this is lovely.  The progression of steps makes sense, it looks far more sleek, and the legacy data notice and function work well.

PR #8212, opened 2020-11-16
Remove backwards compatibility script

### LINKED ISSUE #7136: Remove Legacy JS code for Software Licensing Metabox
## Enhancement Request
The newer markup for EDD Metaboxes from https://github.com/easydigitaldownloads/easy-digital-downloads/issues/2704 is being added to Software Licensing here: https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1179

This means that we can remove the legacy handler from EDD core here:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/admin-backwards-compatibility.js#L16

@pippinsplugins and @cklosowski [had said ~2 years ago](https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1179#issuecomment-335594128) that they wanted to leave it in in order to support older versions of EDD SL. Is that still the case?

robincornett:
Fixes #7136

Proposed Changes:
1. Removes the `edd-admin-backwards-compatibility.js` script.
2. Updates the `.edd-custom-price-option` CSS.

To test:
Check out the PR branch for a related extension (currently only Software Licensing). The variable price options (Advanced Settings) should render with the extension title and appropriate labels and markup.

Not marking this as a draft, but it probably should not be finalized until we've made headway on the affected extensions:
- [ ] Software Licensing ([1179](https://github.com/easydigitaldownloads/edd-software-licensing/issues/1179))
- [ ] Recurring ([1367](https://github.com/easydigitaldownloads/edd-recurring/issues/1367))
- [ ] Simple Shipping ([97](https://github.com/easydigitaldownloads/edd-simple-shipping/issues/97))
- [ ] Purchase Limit ([41](https://github.com/easydigitaldownloads/edd-purchase-limit/issues/41))
- [ ] All Access ([371](https://github.com/easydigitaldownloads/edd-all-access/issues/371), not listed in the compatibility script)


assets/css/admin/downloads/_custom-prices.scss
```diff
@@ -0,0 +1,41 @@
...
+			display: flex;
+			gap: 12px;
+			margin-bottom: 6px;
+
+			.edd-form-group {
+				display: inline-flex;
+				flex-direction: column;
+				justify-content: flex-end;
+				margin-bottom: 0;
+			}
```
spencerfinnell, line 34: Is there a specific use case for this? Not that it will always be avoidable, but any time we need to override or redefine anything related to the [form input markup](https://github.com/easydigitaldownloads/easy-digital-downloads/wiki/Form-Input-Markup) we should ensure there isn't something we can fix on a more global level. 

If not, then changes should be commented inline so we know why it was overwritten.

assets/css/admin/downloads/style.scss
```diff
@@ -0,0 +1 @@
+@import "custom-prices";
```
spencerfinnell, line 1: The markup references `custom-prices` but I think calling this `variable-prices` or just `prices` would make mores sense. There is an [EDD Custom Prices](https://github.com/easydigitaldownloads/edd-custom-prices) extension this could be confused with.

spencerfinnell (changes requested):

assets/css/admin/downloads/style.scss
robincornett: I would be fine with renaming all of the related classes, maybe to something which matches the text, like `.edd-variable-price-option-advanced-settings` and so on. Since none of the extensions have been updated yet, I think we could be pretty flexible here, but if a third party extension is counting on these styles, which were added in 2.8, then that might be problematic.

robincornett: Update: @spencerfinnell it turns out that All Access is also an affected plugin (it's using the "new" markup, but not the correct hook), so it may be safest to keep the classes as they are, but we could certainly rename the partial file.

assets/css/admin/downloads/_custom-prices.scss
robincornett: It's actually copying (mostly) from the CSS for the same in [bundled/repeatable rows](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/css/admin/style.scss#L1575-L1582), so I should just get the class added there, I guess.

spencerfinnell: @robincornett I think the issue here is that we still aren't providing a generic enough way for extensions to create similar "columns" in these areas -- instead we are overriding the default behavior of `.edd-form-group` in very specific areas which can cause unpredictability/confusion if you aren't intimately familiar with those two additional classes. 

An extension would need to provide their own CSS or add a wrapper of `.edd-repeatable-row-standard-fields` to their output which isn't something we should encourage.

Instead, expanding our form input markup to include something like `.edd-form-row` that turns `.edd-form-group` children in to columns is something we could document and make predictable. This PR isn't the place for that, but it would be a good time to document why `.edd-repeatable-row-standard-fields .edd-form-group, .edd-bundled-product-row .edd-form-group` needs to exist for the time being.

assets/css/admin/downloads/style.scss
spencerfinnell: Yes I think renaming the partial will help avoid confusing the contents with CSS specific to the Custom Prices integration.

assets/css/admin/downloads/_custom-prices.scss
robincornett: I would be fine to set that up; it totally makes sense to me. I think we actually could argue for putting it in here so we can make sure the linked extensions add the class in their PRs as well.

spencerfinnell: @robincornett Good point -- that will help with the changes required in the extensions when the script is removed. Let's  go ahead and add it here.

robincornett: @spencerfinnell what about just adding a BEM variation class, so the form group class would be `"edd-form-group edd-form-group--row"`? I've added that in the [PR for All Access](https://github.com/easydigitaldownloads/edd-all-access/pull/372), but without styling. Then we could replace the multiple class styling with that, and get the core CSS updated.

spencerfinnell: @robincornett Does that still rely on `.edd-custom-price-option-section-content` being present so the parent has `display: block;`?

I think we should still provide a generic parent:

```html
<div class="edd-form-row">
  <div class="edd-form-group is-column" />
</div>
```

or

```html
<div class="edd-form-row">
  <div class="edd-form-row__column edd-form-group" />
</div>
```

I'm not sure either is more right than the other. If using the variant/modifier class it should just use the `is-` syntax [that we already use](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/css/admin/style.scss#L4074-L4077) and is recommended by the [WordPress CSS coding standards](https://developer.wordpress.org/block-editor/contributors/develop/coding-guidelines/#naming)

robincornett: I've opted for the first: `edd-form-row` and `edd-form-group is-column`. Started a "forms" directory for these classes but so far have only put the rows in.

If we want to go this route, we should update the wiki and add a new issue to update the core markup as needed.

assets/css/admin/forms/_form-row.scss
```diff
@@ -0,0 +1,12 @@
+.edd-form-row {
+	display: flex;
+	flex-wrap: wrap;
+	gap: 12px;
+
+	.is-column {
```
ashleyfae, line 6: Should this have an `edd-` prefix?

spencerfinnell: This should be moved inside `.edd-form-group` so if can only be used with that class which is already prefixed.

ashleyfae: I'm mostly concerned about another plugin that's not prefixing having an `is-column` class and those styles affecting ours.

spencerfinnell: @ashleyfae Good point. I don't like how the naming _looks_ but it's a safer route to take.

@robincornett I still believe it should be nested under `.edd-form-group`.

robincornett: I guess that's why I was wondering about using the `--` in the class--I've seen that recommended to represent variations ([ref](https://medium.com/@davehouse_80809/bem-for-everyone-else-89ccc8ad66f2)).

Totally fine with moving it, too--working on that now.

spencerfinnell: @robincornett Yep, you were correct with the more verbose name the first time around.

I di think the variation name makes more sense being `.edd-form-group--is-column` as that is a variation of the current block where `.edd-form-group--row` makes it seem like its a variation of the parent block, which goes against BEM -- even though it's super loose BEM to begin with.

spencerfinnell: @robincornett if we are going with verbose names to help avoid conflicts I'm swinging more towards using `.edd-form-row__column`. The naming aligns more with BEM vs. WordPress' loose BEM for variants and it also helps make it clearer that the parent wrapper is needed for the columns to work correctly.

```html
<div class="edd-form-row">
  <div class="edd-form-row__column" />
</div>
```

That leaves the contents of the columns completely arbitrary which leaves more flexibility if we find that we need something in addition to `.edd-form-group` in the future.

robincornett:
Per conversation with @spencerfinnell, proposed markup will be:

```html
<div class="edd-form-row">
  <div class="edd-form-row__column edd-form-group">
    // normal form group stuff in the column only
  </div>
</div>
```
or

```html
<div class="edd-form-row">
  <div class="edd-form-row__column">
    Something unrelated or outside of a form group, but still in the column.
    <div class="edd-form-group" />
  </div>
</div>
```

(Adding it here as I think it disappears from above when I mark the conversation as resolved.)

LisaCee:
What am I expecting to see?  I found this old [comment](https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1179#issuecomment-456885765) about how to test, but I'm not sure if it's still relevant.

robincornett:
@LisaCee that comment is out of date; you will want to start from the opening comment on this PR (for the issue/PR branches of affected extensions).

So if you check out this branch (7136-2), and activate the master branch for any of the first four plugins listed, the advanced settings for variable pricing should look awful. I don't think the SL issue has a screenshot because it was an existing issue, but Recurring is a good one to check out. With `release/3.0`, the shim script is still in place to "fix" that. Then you will want to check out the PR for Recurring with this branch, and the advanced settings should look correct.

So mainly, the end result of this PR should be to _break_ the current releases of the first four plugins, but then with their PR branches, you'll see the new CSS we are adding here, so it will look correct again. It's all very codependent. :D

ashleyfae (approved):

assets/css/admin/forms/_form-group.scss
```diff
@@ -0,0 +1,6 @@
+.edd-form-group {
+	&__control--is-inline {
+		display: inline-flex;
+		align-items: flex-end;
+	}
+}
```
spencerfinnell, line 6: @robincornett _I_ know what's going on here, but I still find this a bit hard to immediately understand. I wonder if we need to nest these child classes and use the `&` shorthand? It might save some keystrokes typing but I don't think it saves time when understanding how the HTML is constructed based on the CSS.

```suggestion
.edd-form-group__control {

	&--is-inline {
		display: inline-flex;
		align-items: flex-end;
	}
}
```

That seems more obvious and lends itself to eventually being merged in with our [existing styles](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/css/admin/style.scss#L4070-L4081) in the future.

spencerfinnell (approved):

LisaCee (approved):
- [x] Software Licensing (1179)
- [x] Recurring (1367)
- [x] Simple Shipping (97)
- [x] Purchase Limit (41)
- [x] All Access (371, not listed in the compatibility script)

Issue #8193, opened 2020-11-06
Product Dropdown: no empty option


robincornett:
### Explain your enhancement (please be detailed)
It would be nice to be able to "reset" a product dropdown to have no value.

### Justification or use case
I've noticed this on the Exports screen (in both 2.9 and 3.0), and @LisaCee also noted it affecting the upgrade paths metabox in Software Licensing. Basically, once a download has been selected, it is not possible to "reset" the dropdown to have no value, because the `product_dropdown` method in the HTML class [forces the `show_option_none` parameter to be `false`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8145/includes/class-edd-html-elements.php#L233)--which also means that anyone using that method cannot override the behavior.

On the exports screen, this can be managed by reloading the screen. On the download edit screen, in 2.9, clearing/removing the row does not _appear_ to remove the download, because there is no empty option, but when the download is saved, the upgrade path is cleared. So it does work, but the UX is confusing. In 3.0, due to #8192, it's not possible to clear out/reset the upgrade paths. I'm marking that issue as a regression and this as an improvement, but fixing either one would, I think, resolve the issue with Software Licensing.

robincornett:
Update: allowing the product dropdown to use the `show_option_none` doesn't resolve the issue for Software Licensing, because the value for the "none" option is `-1` ... but that seems to be true for any select, so I'm wondering how that works with the JS in general.

PR #8213, opened 2020-11-16
Do not hide repeatable row remove link

### LINKED ISSUE #8192: 3.0 - Impossible to "clear" a repeatable row when there is only one
### Expected behavior
I expect to be able to remove a repeatable row, even if it is the only row (in 2.9, clicking "Remove" clears the contents of the first row).

### Actual behavior
In 3.0, due to CSS hiding the "Remove" link when there is only one row, it is not possible to clear the data out of a repeatable row with one click if there is only one row ([CSS added in commit c30cc01](https://github.com/easydigitaldownloads/easy-digital-downloads/commit/c30cc01ef071a589cfb45140e9682127baad825c)).

1) Create/edit a download in 3.0.
2) Scroll to the Download Files metabox and add a file.
3) Click "Add New File". Note that both rows show a "Remove" link.
4) Remove one of the rows. The "Remove" link no longer shows.

PHP Version: 7.4.1
WordPress Version: 5.5.3

In some cases, it's possible to manually clear out a row, but for an extension like Software Licensing, it is currently not possible to clear out the download once it has been selected (which is a separate issue).

robincornett:
Fixes #8192


LisaCee:
If there are multiple rows, this removes the row as expected.

There is different behavior between the remove on Download Files and License Upgrade Paths, though.  If there is only one row in Download Files and you click remove, it clears the data.  In the same scenario in License Upgrade Paths, the Download select and Prorate check don't get cleared.  Price Option and Discount are both cleared.

robincornett:
The download select is because of #8193. I'll have to look into the JavaScript to see why the checkbox isn't clearing--my initial scan suggests that the script is resetting `input` and `select` but not anything else, which would explain why, but I need to see tomorrow if that is the case.

robincornett:
Update: the checkbox not clearing also happens in 2.9; I think the script does just clear out `input` and `select` elements and ignores checkboxes.

@ashleyfae is that something we would want to change?

ashleyfae:
> Update: the checkbox not clearing also happens in 2.9; I think the script does just clear out `input` and `select` elements and ignores checkboxes.
> 
> @ashleyfae is that something we would want to change?

@robincornett I'm fine with leaving it as-is for now.

ashleyfae (approved):

LisaCee (approved):

ashleyfae:
@robincornett Can you sort out the conflicts? Then we can get this merged!

ashleyfae:
Sorry @robincornett -- there are merge conflicts again. :laughing:

PR #8216, opened 2020-11-17
Add deprecation notices for variable pricing hooks

### LINKED ISSUE #8214: Formally deprecate variable price hooks
Formally deprecate the `edd_download_price_table_head` and `edd_download_price_table_row` hooks, which were replaced in 2.8 by `edd_download_price_option_row`, but without a notice.

### Justification or use case
We are working to remove the backwards compatibility shims for these hooks, but should give any third party developers who may still be using them warning that they need to update.

Related: 7136 ([comment](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7136#issuecomment-728329220))

robincornett:
Fixes #8214

To test:
1. Activate this branch and an extension which adds options to the variable prices. Software Licensing may be easiest.
2. Make sure debugging is enabled.
3. Edit a download. PHP Notices should be logged.
4. Switch to the [SL branch 1182](https://github.com/easydigitaldownloads/EDD-Software-Licensing/pull/1182). Notices should be gone!

### DIFF
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -422,7 +422,7 @@ function edd_render_price_row( $key, $args = array(), $post_id, $index ) {
 	ob_end_clean();
 
 	ob_start();
-	if ( has_action( 'edd_download_price_table_row' ) ) {
+	if ( -1 !== has_action( 'edd_download_price_table_row', 'edd_hijack_edd_download_price_table_head' ) ) {
 		do_action_deprecated( 'edd_download_price_table_row', array( $post_id, $key, $args ), '2.8', 'edd_download_price_option_row' );
 	}
 	$show_advanced = ob_get_clean();
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -422,8 +422,13 @@ function edd_render_price_row( $key, $args = array(), $post_id, $index ) {
 	ob_end_clean();
 
 	ob_start();
-	if ( -1 !== has_action( 'edd_download_price_table_row', 'edd_hijack_edd_download_price_table_head' ) ) {
-		do_action_deprecated( 'edd_download_price_table_row', array( $post_id, $key, $args ), '2.8', 'edd_download_price_option_row' );
+	$found_fields = isset( $wp_filter['edd_download_price_table_row'] ) ? $wp_filter['edd_download_price_table_row'] : false;
+	if ( ! empty( $found_fields->callbacks ) ) {
+		if ( 1 !== count( $found_fields->callbacks ) ) {
+			do_action_deprecated( 'edd_download_price_table_row', array( $post_id, $key, $args ), '2.8', 'edd_download_price_option_row' );
+		} else {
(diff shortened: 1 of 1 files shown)


includes/admin/downloads/metabox.php
```diff
@@ -416,11 +416,15 @@ function edd_render_price_row( $key, $args = array(), $post_id, $index ) {
...
 	ob_start();
-	do_action( 'edd_download_price_table_head', $post_id );
+	if ( has_action( 'edd_download_price_table_head' ) ) {
+		do_action_deprecated( 'edd_download_price_table_head', array( $post_id ), '2.8', 'edd_download_price_option_row' );
+	}
 	ob_end_clean();
 
 	ob_start();
-	do_action( 'edd_download_price_table_row', $post_id, $key, $args );
+	if ( -1 !== has_action( 'edd_download_price_table_row', 'edd_hijack_edd_download_price_table_head' ) ) {
```
ashleyfae, line 425: I'm not sure this is right, and this notice isn't triggering for me on SL master (above one is).

When the second parameter is included, the return value is the priority of that function. `edd_hijack_edd_download_price_table_head` is using priority `-1`, which means, for me, this function is always returning `-1`.

includes/admin/downloads/metabox.php
```diff
@@ -416,11 +416,15 @@ function edd_render_price_row( $key, $args = array(), $post_id, $index ) {
 	// Run our advanced settings now, so we know if we need to display the settings.
 	// Output buffer so that the headers run, so we can log them and use them later
 	ob_start();
-	do_action( 'edd_download_price_table_head', $post_id );
+	if ( has_action( 'edd_download_price_table_head' ) ) {
+		do_action_deprecated( 'edd_download_price_table_head', array( $post_id ), '2.8', 'edd_download_price_option_row' );
```
ashleyfae, line 420: I wonder if we should change the version to `2.10`. Although it was technically intended to be deprecated in `2.8`, we didn't actually enforce that until now (`2.10`).

includes/admin/downloads/metabox.php
robincornett: Yes, I'm trying to figure out how to fix this now. Ideally, I would like to show the notice if an extension is using the hook, but allow it to ignore the hijacking function--so that notices only show for extensions.

ashleyfae: Let me know if you want help on it.

robincornett: That was painful, but I _think_ I got it.

ashleyfae (approved):

LisaCee:
I'm still seeing notices, so I want to verify I'm testing the right branches/versions.  I'm starting on this branch, EDD `issue/8214` and SL `release/3.7`.  After seeing notices, I switch to SL `issue/1179`.  Is this right?

robincornett:
Yes, with `issue/8214` and SL `issue/1179`, you should not see the notices, unless you have one of the other affected plugins is active (Recurring, Product Limits, All Access, or Simple Shipping).

LisaCee (approved):

PR #8218, opened 2020-11-20
add alt to credit card icons

### LINKED ISSUE #8217: Missing alt tag on payment icons
The payment icons on the checkout template do not have an alt attribute.   Adding an alt here will allow user using assistive technology, such as a screen reader to be informed about the types of payment that are excepted.

philwp:
Fixes #8217 

Proposed Changes:
Add alt tag containing payment type to the payment icon.


ashleyfae (approved):
Thanks @philwp! This works for now - in the future I think we can improve this by having proper display values for each option.

PR #8226, opened 2020-11-27
Fix user fallback if user is not in initial batch for user_dropdown

### LINKED ISSUE #8224: EDD_HTML_Elements::user_dropdown() assumes `selected` value isn't an array
### Expected behavior

I expect to be able to use `user_dropdown()` and set `selected` to an array of user IDs. I expect this because it supports the `multiple` parameter, which allows you to select multiple users.

### Actual behavior

When you pass in an array of user IDs you get this notice:

```
array_key_exists(): The first argument should be either a string or an integer
```

This is due to using `array_key_exists()` with `$args['selected']` here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-html-elements.php#L343-L357 This fails if `$args['selected']` is an array.

1) Install Discounts Pro.
2) Create a discount and assign it to a user.
3) Save.
4) Notice triggers when the page refreshes.

PHP Version:

EDD Version (or branch):

WordPress Version:

robincornett:
Fixes #8224

Proposed Changes:
1. Updates the user dropdown function to accept an array of selected users when making sure that a selected user will appear in the initial batch.

To test:
1. Check out the [1.5-master-merge branch](https://github.com/easydigitaldownloads/edd-discounts-pro/tree/release/1.5-master-merge) for Discounts Pro (the issue is not present in `master`).
2. With `release/2.10`, create a discount for a specific user and see the error indicated in the related issue.
3. Switch to this branch and the error should be resolved.

Notes: I referred to the `product_dropdown` method to set up the logic for this, as it also can be set to single/multiple.

### DIFF
--- a/includes/class-edd-html-elements.php
+++ b/includes/class-edd-html-elements.php
@@ -339,21 +339,21 @@ class EDD_HTML_Elements {
 			$options[0] = __( 'No users found', 'easy-digital-downloads' );
 		}
 
+		$selected = $args['selected'];
+		if ( ! is_array( $selected ) ) {
+			$selected = array( $selected );
+		}
 		// If a selected user has been specified, we need to ensure it's in the initial list of user displayed
-		if( ! empty( $args['selected'] ) ) {
+		if ( ! empty( $selected ) ) {
+			foreach ( $selected as $selected_user ) {
+				if ( ! array_key_exists( $selected_user, $options ) ) {
+					$user = get_userdata( $selected_user );
 
-			if( ! array_key_exists( $args['selected'], $options ) ) {
-
-				$user = get_userdata( $args['selected'] );
-
-				if( $user ) {
-
-					$options[ absint( $args['selected'] ) ] = esc_html( $user->display_name );
-
+					if ( $user ) {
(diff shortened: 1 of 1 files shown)


includes/class-edd-html-elements.php
```diff
@@ -339,21 +339,21 @@ public function user_dropdown( $args = array() ) {
...
-				if( $user ) {
-
-					$options[ absint( $args['selected'] ) ] = esc_html( $user->display_name );
+		if ( ! empty( $selected ) ) {
+			foreach ( $selected as $selected_user ) {
+				if ( ! array_key_exists( $selected_user, $options ) ) {
+					$user = get_userdata( $selected_user );
 
+					if ( $user ) {
+						$options[ absint( $args['selected'] ) ] = esc_html( $user->display_name );
```
ashleyfae, line 353:
```suggestion
						$options[ absint( $user->ID ) ] = esc_html( $user->display_name );
```

ashleyfae (changes requested):

ashleyfae (approved):

LisaCee (approved):
Thanks for the instructions on how to replicate the error.

PR #8227, opened 2020-11-29
Fix time gmt offset issues

AkramiPro:
- Fix gmt offset not include in start & expire time check.
- Add Support for set exact discount code expire time (H:i:s).


ashleyfae:
Thank you @AkramiPro , however this is already sorted in version 3.0 (`release/3.0` branch). The start and end times are _displayed_ and entered in local time, but then stored in UTC, which is preferable. The times are then compared against UTC when checking validity.

PR #8231, opened 2020-11-30
Define default/fresh admin color scheme

### LINKED ISSUE #8229: 3.0 - Chosen item hover colour is white text on white background
### Expected behavior

I expect items in a Chosen dropdown to still be visible when I hover over them.

### Actual behavior

When I hover over an item in a Chosen dropdown, the text becomes white. It's a white background, so it essentially becomes invisible.

1) Go to any page containing a Chosen dropdown. Example: Downloads > Reports > Export. Any country or downloads dropdown.
2) Click on the dropdown to open it.
3) Hover over an item in the dropdown.

PHP Version:

WordPress Version:

robincornett:
Fixes #8229

To test:
1. Visit a screen such as Exports.
2. Make sure that any select inputs using Chosen highlight a selection properly using the default (fresh) admin color scheme.
3. Make sure that the color is still correct when using an alternative color scheme.


LisaCee (changes requested):
This works well, but the contrast between the default scheme and white font is really low on the Chosen dropdowns.  I feel there's probably a greater WP theme rule in play, but could we change the color to something darker like `#0073aa`?

ashleyfae:
I agree with @LisaCee . Also, what GitHub has done with that hex value is super cute.

robincornett:
Sorry, I had pulled a blue hover on black--will update to the default background color for the primary buttons, which is `#007cba`...not quite as dark as your suggestion, but closer.

ashleyfae (approved):

LisaCee (approved):

PR #8239, opened 2020-12-03
Update how discounts are displayed in the cart

### LINKED ISSUE #8236: 3.0 - Percentage discounts show confusingly on checkout
### Expected behavior
When I'm viewing my shopping cart, I expect the data I see to make sense.

### Actual behavior
When I use a percentage based discount on an order, the row showing the discount shows the currency value of the discount applied, but with a `%` sign after. Example for a cart using a 20% discount:

image

1) In `release/3.0`, create a percentage based discount.
2) Add an item to your cart and apply the discount.

PHP Version: 7.4.1
WordPress Version: 5.6-RC1

robincornett:
Fixes #8236

Proposed Changes: initially I was just going to fix the output to show the proper discount rate, but after reading through the code with changes from #7993, it seemed important to figure out how to show the calculated discount amount, as that appeared to be the goal of the updated code (2.9.x shows just the discount rate, not the calculated/applied amount):

1. Show the calculated discount amount after the discount code. This has been updated to honor the storefront currency placement settings.
2. Additionally, for percentage discounts, the percentage is also displayed.
3. Prefixed the row with "Discount(s):" to be consistent with Subtotal, Tax, and Total rows.
4. Added link text to the `edd_discount_remove` link (hidden with `screen-reader-text`) for accessibility.

To test:
- Create multiple discounts: one flat rate, one percentage rate.
- Add items to the cart and apply one or both discounts (you may have to enable multiple discounts in settings). The prefix should properly show Discount(s) as singular or plural.
- Check the discounts with the currency position set to "after".

This adjusts the display only of the discounts--the calculations were already handled by Spencer in the other issue.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1288,7 +1288,7 @@ function edd_get_cart_discounts_html( $discounts = false ) {
 			}
 		}
 
-		$rate = edd_format_discount_rate( edd_get_discount_type( $discount_id ), $discount_amount );
+		$rate = edd_format_discount_rate( edd_get_discount_type( $discount_id ), edd_get_discount_amount( $discount_id ) );
 
 		$remove_url  = add_query_arg(
 			array(
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1275,7 +1275,7 @@ function edd_get_cart_discounts_html( $discounts = false ) {
 		return;
 	}
 
-	$html = '';
+	$html = _n( 'Discount', 'Discounts', count( $discounts ), 'easy-digital-downloads' ) . ':&nbsp;';
 
 	foreach ( $discounts as $discount ) {
 		$discount_id     = edd_get_discount_id_by_code( $discount );
@@ -1288,7 +1288,8 @@ function edd_get_cart_discounts_html( $discounts = false ) {
 			}
 		}
 
-		$rate = edd_format_discount_rate( edd_get_discount_type( $discount_id ), edd_get_discount_amount( $discount_id ) );
+		$type = edd_get_discount_type( $discount_id );
+		$rate = edd_format_discount_rate( $type, edd_get_discount_amount( $discount_id ) );
 
 		$remove_url  = add_query_arg(
 			array(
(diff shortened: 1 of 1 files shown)


LisaCee (approved):
Tested with flat-rate, percentage, and currency with the symbol after the amount.

includes/discount-functions.php
```diff
@@ -1299,10 +1300,19 @@ function edd_get_cart_discounts_html( $discounts = false ) {
...
+		$discount_html   = '';
+		$discount_html  .= "<span class=\"edd_discount\">\n";
+		$discount_amount = edd_currency_filter( edd_format_amount( $discount_amount ) );
+		$discount_html  .= "<span class=\"edd_discount_total\">{$discount}&nbsp;&ndash;&nbsp;{$discount_amount}</span>\n";
+		if ( 'percent' === $type ) {
+			$discount_html .= "<span class=\"edd_discount_rate\">($rate)</span>\n";
+		}
+		$discount_html .= sprintf(
+			'<a href="%s" data-code="%s" class="edd_discount_remove"><span class="screen-reader-text"%s</span></a>',
+			$remove_url,
```
ashleyfae, line 1312:
```suggestion
			esc_url( $remove_url ),
```

includes/discount-functions.php
```diff
@@ -1299,10 +1300,19 @@ function edd_get_cart_discounts_html( $discounts = false ) {
...
+			$discount,
```
ashleyfae, line 1313:
```suggestion
			esc_attr( $discount ),
```

includes/discount-functions.php
```diff
@@ -1299,10 +1300,19 @@ function edd_get_cart_discounts_html( $discounts = false ) {
...
+			__( 'Remove discount', 'easy-digital-downloads' )
```
ashleyfae, line 1314:
```suggestion
			esc_attr__( 'Remove discount', 'easy-digital-downloads' )
```

ashleyfae (changes requested):

ashleyfae:
Tested:

- [ ] Flat rate only
    - [ ] Currency position before
    - [ ] Currency position after
- [ ] Percentage only
    - [ ] Currency position before
    - [ ] Currency position after
- [ ] Flat rate and percentage
    - [ ] Currency position before
    - [ ] Currency position after

ashleyfae (approved):

PR #8245, opened 2020-12-04
Change selector of upload field on admin download page to take precedence over WordPress CSS

napoleaowingert:
# Fixes

1. I think the CSS selector in EDD needs to be changed like this:
from: .edd_upload_field
to: input[type="text"].edd_upload_field

Then it will work just fine as it takes precedence over WordPress CSS.

### DIFF
--- a/assets/css/edd-admin.css
+++ b/assets/css/edd-admin.css
@@ -468,7 +468,7 @@ label.edd_prices_shipping { /* remove once Simple Shipping markup is improved */
 	top: 3px;
 	right: 7px;
 }
-.edd_upload_field {
+input[type="text"].edd_upload_field {
 	padding-right: 8em;
 }
 textarea#edd-payment-note {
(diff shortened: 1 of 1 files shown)


assets/css/edd-admin.css
```diff
@@ -468,7 +468,7 @@ label.edd_prices_shipping { /* remove once Simple Shipping markup is improved */
 	top: 3px;
 	right: 7px;
 }
-.edd_upload_field {
+input[type="text"].edd_upload_field {
 	padding-right: 8em;
```
ashleyfae, line 472: I actually think I'd prefer an `!important` here and reverting the selector back to `.edd_upload_field` . That way we won't get caught out if we ever change the input type for whatever reason (not that I'm planning to, but you never know :shrug: ).

ashleyfae (changes requested):
Thank you @napoleaowingert! I requested a minor change if you don't mind. :smiley:

napoleaowingert:
Hi @ashleyfae, thank you for checking this! The requested change has already been made.

ashleyfae (approved):
Thank you! :raised_hands:

PR #8249, opened 2020-12-07
Add support for variable pricing when getting item discount amount post-checkout #8246

### LINKED ISSUE #8246: Discount amount is incorrect in /sales/ API endpoint when using variable pricing
### Expected behavior

The `/sales/` endpoint includes a `discounts` array, which is meant to contain the total discount amount (flat dollar amount) for that order. I expect this number to be accurate.

### Actual behavior

When using variable prices and purchasing _not the cheapest price option_, the discount amount is incorrect.

1) Create a variable price product with at least two price options. For example: $3 and $10.
2) Create a discount code. In my tests I used 25%, but setting isn't relevant.
3) Purchase the _higher priced_ option. So, $10.
4) View the output of the `/sales/` API endpoint. Look at the `discounts` array. This amount is incorrect.

In my example:

- I purchased a $10 product with a 25% off discount.
- The `discounts` array shows: `array( 'discounts' => 0.75 )`. That 0.75 would be correct if I'd purchased the cheapest ($3) price option. But since I purchased $10, it should have been `2.5`

PHP Version:

WordPress Version:

The issue is that the `options` array in cart details isn't being properly passed through to the discount functions. In the API we use `edd_get_cart_item_discount_amount()` here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/api/class-edd-api-v2.php#L451 

The structure of `$item` looks like this:

```
array(10) {
  ["item_number"]=>
  array(3) {
    ["id"]=>
    int(157)
    ["quantity"]=>
    int(1)
    ["options"]=>
    array(2) {
      ["quantity"]=>
      int(1)
      ["price_id"]=>
      string(1) "2"
    }
  }
}
```

Note how `options` is nested inside `item_number`.

But when you dig through `edd_get_cart_item_discount_amount()`, it looks for `options` inside `$item['options']` directly https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/cart/class-edd-cart.php#L645 Which means it's never getting picked up and falls back to using the `edd_price` meta, which, with variable pricing enabled, is defaulted to the cheapest option.

ashleyfae:
Fixes #8246 

Proposed Changes:
1. Add support for options inside `item_number` for variable pricing.

For testing, consult steps in the original issue: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8246

We should both 1) reproduce those steps; and 2) test with a non-variable-priced product to ensure the discount amount is still valid for that.

### DIFF
--- a/includes/cart/class-edd-cart.php
+++ b/includes/cart/class-edd-cart.php
@@ -639,6 +639,14 @@ class EDD_Cart {
 
 		if ( ! isset( $item['options'] ) ) {
 			$item['options'] = array();
+
+			/*
+			 * Support for variable pricing when calling `edd_get_cart_item_discount_amount()`
+			 * @link https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8246
+			 */
+			if ( isset( $item['item_number']['options'] ) ) {
+				$item['options'] = $item['item_number']['options'];
+			}
 		}
 
 		$amount           = 0;
--- a/tests/tests-api.php
+++ b/tests/tests-api.php
@@ -545,4 +545,29 @@ class Tests_API extends EDD_UnitTestCase {
 		$this->assertEquals( 'Purchase Notes', $out['products'][0]['notes'] );
 	}
 
+	/**
+	 * Ensures the correct discount amount is included in the recent sales endpoint.
+	 *
+	 * @link https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8246
+	 *
+	 * @covers EDD_API_V2::get_recent_sales
+	 * @covers EDD_Cart::get_item_discount_amount
+	 */
+	public function test_recent_sales_contains_correct_discount_amount() {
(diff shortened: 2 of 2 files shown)


robincornett (approved):
Tested:
- [x] `release/2.10`: a variably priced product purchased at the non-default/lowest level shows an incorrect discount amount when accessed via the API
- [x] `issue/8246`: the same purchase shows the correct discount amount via the API
- [x] both branches: the same discount on a standard product shows the correct amount
- [x] `release/2.10`: a purchase with a standard product and the variable product shows the incorrect discount amount
- [x] `issue/8246`: the same purchase shows the correct discount amount

LisaCee (approved):
- [x] Successfully replicated incorrect discount on `2.10`.  
- [x] Confirmed correct discount from API on this branch. 
- [x] Non-variable priced items were discounted correctly on both branches. 
- [x] Mixed order (variable and non) discount was incorrect in `2.10` but correct here.

PR #8252, opened 2020-12-07
Create country/region select methods and update related code

### LINKED ISSUE #8053: 3.0 - Regions do not update for Reports > Export when the country is changed
### Expected behavior
When I change the country from the United States to "Export Taxed Orders", I expect the regions select field to update to use the new country's regions.

### Actual behavior
When I change the country, the regions do not change.

1) Visit Downloads > Reports > Export.
2) Under "Export Taxed Orders", select a country other than the United States.

PHP Version: 7.4.1
WordPress Version: 5.4.2

robincornett:
Fixes #8053

Proposed Changes:
1. Creates two new HTML elements methods, `country_select` and `region_select`.
2. Moves the JavaScript to listen/update the dropdowns to `edd-admin.js`.

Currently, the `select` elements do not have any general classes (such as `edd-select` or `edd-form-group__input`) added to the defaults, but the classes required for the JS to work are checked for and added.

Any screen using these inputs needs to be checked:

- Reports (filter bar)
- Orders (filter bar)
- New/Edit order
- Adding tax rates
- Customers (individual customer)

I evaluated every country/region select which was using the `EDD()->html->select()` method and put together a [spreadsheet](https://docs.google.com/spreadsheets/d/1CHXXv0joSjPnJHpyWRrwHgjClhZjobFgszQZQNMcZx4/edit?usp=sharing) to establish what parameters should be set as defaults and which should be overridden in specific scenarios.

This currently doesn't change any behavior of existing selects--specifically, and I don't remember where we ran across it, if you select a country or region on the Reports > Export screen, you cannot reset it to "All" without reloading the screen. I will happily fix that here if it's okay, because I think it shouldn't be that way, but that's how it currently is in 3.0.


ashleyfae:
Love this change: :raised_hands: 

It looks like the individual customer page wasn't updated -- was that an oversight?

---

Tested:

- [ ] Admin orders list table
- [ ] Admin single order view
- [ ] Admin reports
- [ ] Tax rate settings
- [ ] Customer page

robincornett:
For the profile editor, no, I intentionally did not update that one, because it does not use the `EDD()->html->select()` method (I do not know why but it doesn't). Same for the country/region settings on the main downloads settings page (note that there is a `edd_shop_states_callback` function, but not a corresponding one for the country...looks like the country is set up to use a standard settings select callback and the state uses its own custom one).

There is also a country/state select in the shortcode profile editor and on the checkout screen, at least--but I'm not sure that we use the HTML elements methods on front end fields.

ashleyfae:
I'm fine with this as-is for now and we can always address more thorough replacements later. :+1:

LisaCee (commented):
In `3.0` and here there is an oddity with `region_select` that I wonder if could be fixed in here.  In new/edit order, when a country without regions listed is selected, the region input changes from a select to a text field.  I've also noticed this behavior when setting Tax rates.  On the other screens, if a country doesn't have regions listed, the select changes to `All Regions`.  I'm not sure if this was an intentional move, but if not can it be addressed here?

Also, outside of the scope of this PR but since I found it while testing this, the Reports->Taxes filter doesn't seem to be doing anything.  I think this has been noted elsewhere, but just in case I wanted to mention it.

robincornett:
I'm going to guess that the text input is intentional on screens where data is being input (new orders, tax rates)--so that the region can be entered if it exists but we don't have a list for the particular country?

The taxes filter is working for me, but I'm noticing that if I select a Region without selecting a Country (which I appear to be able to do only once?) that although the filter works, the Region is not set to the one I selected. And I can't "clear" out the filters/reset that to the default--I can go back to All Countries, but then it's persistently selected instead of being set to the store default.

LisaCee (approved):
- [x] Reports > Exports > Export Taxed Orders
- [x] Reports >  Taxes (filter bar)
- [x] Orders > More
- [x] Orders (new and edit)
- [x] Settings > Tax Rates
- [x] Individual Customer
  
This works really well.  I got zero errors and I couldn't find any missed instances.  🎉

ashleyfae (approved):

PR #8253, opened 2020-12-08
Add tax_rate_id to orders table #8060

### LINKED ISSUE #8060: 3.0 - Review tax rates data for order adjustments table
### Expected behavior
I feel like data in the `edd_order_adjustments` table should be consistently stored, with some variations as needed.

### Actual behavior
The way that tax rates are stored in the `edd_order_adjustments` table feels inconsistent and incomplete:
- The only numeric data stored for tax rates is the rate amount itself (as `total`). The `subtotal` and `tax` amounts are left to the default.
- On migration, tax rates are applied to/stored with the order.
- On subsequent/new orders, tax rates are applied to each order item, regardless of whether the item is taxable or not.
- When reviewing the database table, because the `subtotal` and `tax` are empty, and the `total` is the same, regardless of item cost or taxability, there is no ready way to scan and differentiate between items in the order.
- Tax rates are the only order adjustment items with no `subtotal` stored.
- It seems that discounts and tax rates could be considered to be similar kinds of data adjustments, because neither apply to free items, and order items can be excluded from a discount or tax. Additionally, when processing a refund, discounts and taxes are both prorated to calculate the refund amount. However, discounts are applied to the order as a whole ([historical and accounting reasons](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8014#issuecomment-655160388)), while tax rates are applied to each item. Should tax rates be applied to the order as a whole instead?

PHP Version: 7.4.1
WordPress Version: 5.4.2

Related: #8012 (order adjustments in general, original issue), #8056 (fees). Opening this issue to more specifically discuss tax rates behavior.

ashleyfae:
Fixes #8060

Proposed Changes:
1. Add `tax_rate_id` column to orders table. This references a tax rate (stored in adjustments).
2. Stop storing tax rates in order_adjustments table.
3. Fix a number of backwards compatibility issues with regards to tax rate percentage formatting (2.9 stored as decimals - e.g. `0.2`; 3.0 stores as percentages - e.g. `20`).
4. Allow for arbitrary tax rates to be set on payments (another 2.9 backwards compatibility thing, ensures compatibility with EU VAT add-on). https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8253/commits/2fca9a9660181f1bf204d2d7f97ff3b086adefbf

### Core plugin

- [ ] Tax rates copied over and stored correctly via migration from 2.9.
- [ ] Tax rates saved correctly when making orders.
- [ ] `tax_rate_id` set as `null` when making a purchase without tax.

### EU VAT add-on

Each test ensures the VAT rate is both _displayed_ and _saved_ correctly. (Displayed being on purchase receipts and invoices.)

- [x] Using "real" tax rate.
- [x] Using tax rate auto provided by the plugin (no tax rate saved in DB, but purchasing in EU country). `tax_rate_id` should be `null`, but a custom tax rate should be stored in order meta.
- [x] Using "real" tax rate, but with validated VAT number & reverse charge (`tax_rate_id` should have the rate ID, but this should take the tax amount to $0 & invoice should say `reverse charged`).


includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -516,6 +567,7 @@ public static function orders( $data = null ) {
 			'mode'           => $mode,
 			'currency'       => ! empty( $payment_meta['currency'] ) ? $payment_meta['currency'] : edd_get_currency(),
 			'payment_key'    => $purchase_key,
+			'taxs_rate_id'   => $tax_rate_id,
```
robincornett, line 570:
```suggestion
			'tax_rate_id'   => $tax_rate_id,
```

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -498,6 +498,57 @@ public static function orders( $data = null ) {
...
+		 * Previously tax rates were stored as a decimal (e.g. `0.2`) but they're now stored as a percentage
+		 * (e.g. `20`). So we need to convert.
+		 */
+		if ( $tax_rate < 1 ) {
+			$tax_rate = $tax_rate * 100;
+		}
+
+		$set_tax_rate_meta = false;
+
+		if ( ! empty( $tax_rate ) && ! empty( $order_address_data['country'] ) ) {
```
robincornett, line 533: I need to spend more time here digging into why, but this does not seem to be evaluating as true for me, even though it should. Any error logging I'm doing inside this function is not happening, even if the country is set and the tax rate is not empty. I'll dig more tomorrow.

robincornett (changes requested):
Just started migrating and haven't gotten very far but wanted to start giving feedback.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -498,6 +498,57 @@ public static function orders( $data = null ) {
...
+				: '',
+			'postal_code'  => isset( $user_info['address']['zip'] )     ? $user_info['address']['zip']     : '',
+			'date_created' => $date_created_gmt,
+		);
+
+		// Remove empty data.
+		$order_address_data = array_filter( $order_address_data );
+
+		$tax_rate_id = null;
+		$tax_rate = isset( $meta['_edd_payment_tax_rate'][0] )
```
robincornett, line 519: These two lines need to be changed to `_edd_payment_tax` in order to get the tax amount--once that is set then the `$tax_rate_id` can be retrieved.

ashleyfae: Not sure I'm following this. This line is getting the tax rate percentage, why should I get the tax amount instead?

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -498,6 +498,57 @@ public static function orders( $data = null ) {
 			$total   = $meta['_edd_payment_total'][0];
 		}
 
+		// Build up the order address data. Actual insertion happens later, but we need this now to figure out the tax rate.
+		$order_address_data = array(
+			'name'         => trim( $user_info['first_name'] . ' ' . $user_info['last_name'] ),
```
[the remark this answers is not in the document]
ashleyfae: Moved that up in https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8253/commits/b98d70c768e2f6a3782a2c8774caa70029cc9f33

includes/admin/upgrades/v3/class-data-migrator.php
robincornett: Sorry, my mistake, you are right. But somehow I'm ending up with no tax rate object being retrieved so still digging on that. Scratch that suggestion because it was totally wrong!

ashleyfae:
@davidsherlock Are you interested in testing this with your tax plugin to ensure it still works as expected? Or if you want to send me a copy I'd be happy to try it out.

robincornett (approved):
Core plugin

- [x] Tax rates copied over and stored correctly via migration from 2.9.
- Tax rates saved correctly when making orders.
    - [x] Standard order
    - [x] Manual order
    - Note: because the manual order doesn't automatically add the tax to the purchase when the address is added to the order, it is possible to complete an order with a `tax_rate_id` but no tax applied.
- `tax_rate_id` set as null when making a purchase without tax.
    - [x] Standard order
    - [x] Manual order

EU VAT add-on

Each test ensures the VAT rate is both displayed and saved correctly. (Displayed being on purchase receipts and invoices.)

- [x] Using "real" tax rate (note: my tax rate percentage is displayed incorrectly in the checkout but that's not EDD)
- tax rate provided by plugin:
    - [x] `tax_rate_id` is null
    - [x] `tax_rate` is stored in order meta
- Using "real" tax rate, but with validated VAT number & reverse charge
    - [x] tax rate stored in database is used over the one from the plugin (before adding VAT number)
    - [x] `tax_rate_id` is set
    - [x] tax amount is 0
    - [x] no `tax_rate` in order meta
    - [x] VAT information shows in order confirmation, order details, invoice; invoice says "reverse charged"

PR #8254, opened 2020-12-08
Revert visibility changes on BerlinDB query methods

### LINKED ISSUE #8247: 3.0/BerlinDB: Incorrect visibility on Query methods
### Expected behavior
When I filter orders in the WP admin (in this instance, by country/region), I expect it to work.

### Actual behavior
When I attempt this filter, I hit two different errors. The first is a PHP fatal error due to the `get_primary_column_name` method being `private` ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/engine/class-query.php#L743)).

If I change that to `protected`, the crash is less spectacular, but the filter still fails, because of the visibility on the `get_table_name` method ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/engine/class-query.php#L721)). Changing that to `protected` allows the filter to complete.

1) Check out `release/3.0`.
2) Go to Downloads > Orders.
3) Click the More button and filter the orders by country/region (this is what I did, but I think actually using any filter has the same issues).

PHP Version: 7.4.1
WordPress Version: 5.5.3

I saw several methods in that class marked as `private` and it may be worth reviewing all of them to see if that is too restrictive.

robincornett:
Fixes #8247

Proposed Changes:
1. Revert visibility changes on methods made as part of #8132.
2. I have reviewed everything in the class which is still marked `private` (under the `Private Parsers` heading and then another section further down); everything which is not used outside of the class is still `private`.
3. One exception to 2 is `set_query_var_defaults` which is also used by the Orders class as a protected method.

### DIFF
--- a/includes/database/engine/class-query.php
+++ b/includes/database/engine/class-query.php
@@ -359,7 +359,7 @@ class Query extends Base {
 	 *
 	 * @since 1.0.0
 	 */
-	private function set_last_changed() {
+	protected function set_last_changed() {
 		$this->last_changed = microtime();
 	}
 
@@ -370,7 +370,7 @@ class Query extends Base {
 	 *
 	 * @since 1.0.0
 	 */
-	private function set_alias() {
+	protected function set_alias() {
 		if ( empty( $this->table_alias ) ) {
 			$this->table_alias = $this->first_letters( $this->table_name );
 		}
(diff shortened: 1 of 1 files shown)


ashleyfae (approved):
Tests passing locally.

PR #8257, opened 2020-12-09
Change client side validation when a discount is applied

### LINKED ISSUE #7487: Remove Forced Clientside Form Validation When Applying a Discount
## Bug Report

Previously https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4325
Previously https://github.com/easydigitaldownloads/easy-digital-downloads/issues/4645

### Expected behavior

When applying a discount in the Checkout, other checkout inputs should not be modified.

### Actual behavior

The inputs are marked `required`

---

After a discount is applied if the resulting price is not $0.00 then the checkout form forces certain inputs to be `required` before submitting the form (via HTML5 browser validation). This is done to "undo" any changes that would be made if the price did get to $0.00 which allows the purchase to be made without requiring those inputs.

1) Enable EDD Stripe + saved cards. **Disable billing addresses**.
2) Make a purchase.
3) Create another cart and apply a discount that does not result in the cart price being $0.00
4) Checkout with an existing saved card

Observe the `"card_name" is not focusable` error in the console.

While this is currently specific to the Saved Cards feature in Stripe, I am opening a core issue because the way the inputs are targetted in core do not create an easy solution to prevent this behavior in the extension, and it could possibly affect other extensions in the future.

~~The [selectors used](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/edd-checkout-global.js#L177) in core to find inputs would require removing `.edd-input` on the [custom card form](https://github.com/easydigitaldownloads/edd-stripe/blob/master/includes/template-functions.php#L96) in the extension which currently provides help with auto styling the Stripe Elements input.~~ 

See: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7487#issuecomment-737520495

robincornett:
Fixes #7487

Proposed Changes:
1. When a discount is applied, only fields other than `.card-address-2` should be marked as `required`.

@spencerfinnell this is different than what you originally proposed as the `input` was not an object. Making the logic more verbose felt more readable to me.

To test:
1. Make sure taxes are disabled.
2. This may happen with any gateway, but it's easiest to test with Stripe, so enable that.
3. Do not collect address information on checkout.
4. Add something to your cart.
5. Make sure that Stripe/Credit Card is selected as the payment method, and have at least one saved card (this forces the add new card form to be hidden).
6. Apply a discount to the cart.
7. Complete your purchase. In `master`, you will not be able to complete this step due to a JS error, but in this branch, you should be able to.


assets/js/edd-checkout-global.js
```diff
@@ -184,9 +184,11 @@ window.EDD_Checkout = (function($) {
 
 						} else {
 
-							if (!inputs.is('.card-address-2')) {
-								inputs.attr('required','required');
-							}
+							inputs.each( function ( input ) {
+								if ( !$( input ).hasClass( '.card-address-2' ) ) {
```
spencerfinnell, line 188:
```suggestion
								if ( ! $( input ).hasClass( '.card-address-2' ) ) {
```

["When in doubt, space it out."](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#spacing)

spencerfinnell (changes requested):

assets/js/edd-checkout-global.js
```diff
@@ -184,9 +184,11 @@ window.EDD_Checkout = (function($) {
 
 						} else {
 
-							if (!inputs.is('.card-address-2')) {
-								inputs.attr('required','required');
-							}
+							inputs.each( function ( input ) {
```
ashleyfae, line 187:
```suggestion
							inputs.each( function ( index, input ) {
```

[The first parameter is the index in the array.](https://api.jquery.com/each/) The second one is the actual value. This was throwing a JavaScript error when testing with taxes + address fields enabled.

ashleyfae (changes requested):

assets/js/edd-checkout-global.js
spencerfinnell: Another option is to omit `index, input` completely and use `$( this )` inside of the callback to reference the current jQuery object directly vs. recreating it with `$( input )`.

ashleyfae:
This doesn't actually appear to fix the original issue for me :thinking: Can someone else confirm?

spencerfinnell:
@ashleyfae You're right, my over simplified "solution" actually isn't all we need here. 

We could check if the element is visible before adding the required property:

```js
if ( ! $( this ).hasClass( '.card-address-2' ) && $( this ).is( ':visible' ) ) {
	$( this ).prop( 'required', true );
}
```

But that would only work if the discount is applied and "New Card" has already been selected. If you toggle after you've applied the discount we end up in the same situation.

My next thought would be to check for a `.card-name` class and skip that field as well. 

```js
if ( ! $( this ).hasClass( '.card-address-2' ) && ! $( this ).hasClass( '.card-name' ) ) {
	$( this ).prop( 'required', true );
}
```

Two things about this:

1. It's basically a workaround for Stripe, since I believe it's the only gateway that has a similar setup with saved payment methods
2. It relies on gateways validating the input on the server.

I think both of these are acceptable. All inputs should be validated on the server anyway and we already have quite a mix of client and server validation for fields (i.e the Terms of Service and Privacy Policy checkboxes).

I think we could still look at removing all of these forced validations in the future and leave it to the gateways that implement the fields, but this would be a step in that direction.

robincornett:
I took @spencerfinnell's latest code and with taxes enabled, it isn't working for me. Tinkered with it a bit and what I have right now is working (both with and without address fields):

```js
inputs.each( function () {
	let $input = $( this );
	if ( ! $input.is( '.card-name, .card-address, .card-address-2' ) ) {
		$( this ).prop( 'required', true );
	}
} );
```

Note that I had to add the first address input as well, and I switched to `.is()` because the multiple `.hasClass()` checks weren't chaining up properly for me. How does this feel?

assets/js/edd-checkout-global.js
```diff
@@ -184,9 +184,12 @@ window.EDD_Checkout = (function($) {
 
 						} else {
 
-							if (!inputs.is('.card-address-2')) {
-								inputs.attr('required','required');
-							}
+							inputs.each( function () {
+								var $input = $( this );
+								if ( ! $input.is( '.card-name, .card-address, .card-address-2' ) ) {
+									$( this ).prop( 'required', true );
```
spencerfinnell, line 190:
```suggestion
									$input.prop( 'required', true );
```

assets/js/edd-checkout-global.js
```diff
@@ -184,9 +184,12 @@ window.EDD_Checkout = (function($) {
 
 						} else {
 
-							if (!inputs.is('.card-address-2')) {
-								inputs.attr('required','required');
-							}
+							inputs.each( function () {
+								var $input = $( this );
+								if ( ! $input.is( '.card-name, .card-address, .card-address-2' ) ) {
```
spencerfinnell, line 189: @robincornett The default implementation of `.card-address` can be required: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/checkout/template.php#L353

It would still fall back to server validation but I'm not sure if we should add another exclusion here that hasn't explicitly caused an issue -- unless you're seeing similar behavior with this field on a gateway?

spencerfinnell (changes requested):
@robincornett

assets/js/edd-checkout-global.js
robincornett: With taxes enabled, I was getting the JS error on all three of those fields after a discount was applied, which is why I'd added it. That was with Stripe.

spencerfinnell: @robincornett I see that now. Okay, I think this makes sense.

LisaCee (approved):
This worked for me.

I was able to replicate the issue in `release/2.10`.  In this branch the issue is fixed with no errors in my log or console.

spencerfinnell (approved):

ashleyfae (approved):

PR #8261, opened 2020-12-11
Fix order comparison query #8260

### LINKED ISSUE #8248: 3.0 - "More" filter options modal stays open when orders are filtered
### Expected behavior
When I use the "More" button to expose more filters for the Orders table, and apply them, I expect to be able to see the results.

### Actual behavior
The modal stays open even after the screen reloads, so the first four results are obscured.

1) Visit Downloads > Orders.
2) Use the "More" button to apply a filter to the results (note: you will have to change the visibility on some methods as described in #8247 to get it to work at the moment).
3) Click the "Filter" button.
4) The filtered results are displayed, but the modal is open. You can close it by clicking the "More" button.

PHP Version: 7.4.1
WordPress Version: 5.5.3

### LINKED ISSUE #8260: 3.0 - Admin Orders - Filtering by amount doesn't work
### Expected behavior

When I use the "Total is" extra filter on the admin orders page, I expect the results to update accordingly.

### Actual behavior

This doesn't seem to impact results at all.

1) Go to Downloads > Orders.
2) Click "More" to expand more filters.
3) Search for orders where the total is less than a value.
4) On refresh, first note that the "less than" dropdown is now selected on "equal to" instead.
5) You will still see orders with a higher total.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8260

Proposed Changes:
1. Change select values to `eq`, `gt`, and `lt` instead of angle brackets.
2. Use `compare_query` argument instead of `compare` for comparison array. I suspect this was a change in BerlinDB at some point. :shrug: 
3. Exit early out of `query_by_country()` method if there is no country query.

To test:

Try filtering by all "Total is" options to ensure each of them work.

Try combining a "Total is" option with a country filter and ensure it still works.

### DIFF
--- a/includes/admin/payments/class-payments-table.php
+++ b/includes/admin/payments/class-payments-table.php
@@ -229,9 +229,9 @@ class EDD_Payment_History_Table extends List_Table {
 					<legend for="order-amount-filter-type"><?php esc_html_e( 'Total is', 'easy-digital-downloads' ); ?></legend>
 					<?php
 					$options = array(
-						'=' => __( 'equal to', 'easy-digital-downloads' ),
-						'>' => __( 'greater than', 'easy-digital-downloads' ),
-						'<' => __( 'less than', 'easy-digital-downloads' ),
+						'eq' => __( 'equal to', 'easy-digital-downloads' ),
+						'gt' => __( 'greater than', 'easy-digital-downloads' ),
+						'lt' => __( 'less than', 'easy-digital-downloads' ),
 					);
 
 					echo EDD()->html->select( array(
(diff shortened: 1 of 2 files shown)


robincornett (commented):
The comparison queries are working nicely for me. I would think there would be benefit to having "greater/less than or equal to" comparisons available as well.

It is possible for me to set the region filter to Tennessee without adjusting the country filter, and that is not working. It seems like it should be possible to query by just the region.

ashleyfae:
> The comparison queries are working nicely for me. I would think there would be benefit to having "greater/less than or equal to" comparisons available as well.

Let's leave it as-is for now. We can always add those in at a later date.

> It is possible for me to set the region filter to Tennessee without adjusting the country filter, and that is not working. It seems like it should be possible to query by just the region.

I'll think on this, but I might punt that to post 3.0 depending on how I feel come Monday. :stuck_out_tongue:

LisaCee (approved):
This works really well and the logic makes sense as much as it can to me.

LisaCee:
> It is possible for me to set the region filter to Tennessee without adjusting the country filter, and that is not working. It seems like it should be possible to query by just the region.

I'm not as sure about being able to query just by region, but having the region populated before the country is selected is confusing.  Since the regions are available in the list, it does seem like one could filter just based on that which, as Robin mentioned, isn't working.

ashleyfae:
> > It is possible for me to set the region filter to Tennessee without adjusting the country filter, and that is not working. It seems like it should be possible to query by just the region.
> 
> I'll think on this, but I might punt that to post 3.0 depending on how I feel come Monday.

Yes I've decided we can re-examine that post-3.0. It doesn't have to be resolved right now as there is still a way around it by using both filters.

robincornett (approved):

PR #8267, opened 2020-12-14
Show success message after manually creating order #8191

### LINKED ISSUE #8191: 3.0 - No success message after manually creating order
### Expected behavior

After saving a new manually-created order, I expect to see a success message.

### Actual behavior

You get taken to the `view-order-details` page, but no success message is shown. It doesn't make it super clear that the order was in fact created.

1) Go to Downloads > Orders > Add New and create a new order.
2) After saving, you will be on the `view-order-details` page for the new order, but there's no success message visible.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8191

Proposed Changes:
1. Move all order related notices out of the "Shop reports errors" section and into the order-related notices. I think they were just in the wrong spot.
2. Create a new notice for `order_added` and include this notice in the page redirect after making an order.
3. Correct return type in PHPDocs for `edd_add_manual_order()`.

To test:

Manually create a new order and ensure you get a success message after doing so.

### DIFF
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -601,7 +601,7 @@ class EDD_Notices {
 				case 'payment_deleted' :
 					$this->add_notice( array(
 						'id'      => 'edd-payment-deleted',
-						'message' => __( 'The payment has been deleted.', 'easy-digital-downloads' )
+						'message' => __( 'The order has been deleted.', 'easy-digital-downloads' )
 					) );
 					break;
 				case 'email_sent' :
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -610,12 +610,6 @@ class EDD_Notices {
 						'message' => __( 'The purchase receipt has been resent.', 'easy-digital-downloads' )
 					) );
 					break;
-				case 'refreshed-reports' :
-					$this->add_notice( array(
-						'id'      => 'edd-refreshed-reports',
-						'message' => __( 'The reports have been refreshed.', 'easy-digital-downloads' )
-					) );
-					break;
 				case 'payment-note-deleted' :
 					$this->add_notice( array(
 						'id'      => 'edd-note-deleted',
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -514,40 +514,10 @@ class EDD_Notices {
 		// Shop reports errors
 		if ( current_user_can( 'view_shop_reports' ) ) {
 			switch( $notice ) {
-				case 'order_trashed' :
-					$this->add_notice( array(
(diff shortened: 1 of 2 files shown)


includes/admin/class-edd-notices.php
```diff
@@ -610,6 +580,48 @@ private function add_user_action_notices( $notice = '' ) {
...
+				case 'order_restored' :
+					$this->add_notice( array(
+						'id'      => 'edd-order-restored',
+						'message' => __( 'The order has been restored.', 'easy-digital-downloads' )
+					) );
+					break;
+				case 'payment_deleted' :
+					$this->add_notice( array(
+						'id'      => 'edd-payment-deleted',
+						'message' => __( 'The payment has been deleted.', 'easy-digital-downloads' )
```
robincornett, line 604:
```suggestion
						'message' => __( 'The order has been deleted.', 'easy-digital-downloads' )
```
Just suggesting that we update this, and possibly the "payment note" message below, while these lines are being updated, so that the user facing language we are using is consistent.

includes/admin/class-edd-notices.php
```diff
@@ -610,6 +580,48 @@ private function add_user_action_notices( $notice = '' ) {
...
+						'message' => __( 'The order has been deleted.', 'easy-digital-downloads' )
+					) );
+					break;
+				case 'email_sent' :
+					$this->add_notice( array(
+						'id'      => 'edd-payment-sent',
+						'message' => __( 'The purchase receipt has been resent.', 'easy-digital-downloads' )
+					) );
+					break;
+				case 'refreshed-reports' :
```
robincornett, line 613: This `refreshed-reports` wasn't removed from the `class-edd-notices` file; should it be removed from here (I think this is a duplicate)?

includes/admin/class-edd-notices.php
```diff
@@ -610,6 +580,48 @@ private function add_user_action_notices( $notice = '' ) {
...
+					$this->add_notice( array(
+						'id'      => 'edd-refreshed-reports',
+						'message' => __( 'The reports have been refreshed.', 'easy-digital-downloads' )
+					) );
+					break;
+				case 'payment-note-deleted' :
+					$this->add_notice( array(
+						'id'      => 'edd-note-deleted',
+						'message' => __( 'The payment note has been deleted.', 'easy-digital-downloads' )
```
robincornett, line 622:
```suggestion
						'message' => __( 'The order note has been deleted.', 'easy-digital-downloads' )
```

includes/admin/class-edd-notices.php
ashleyfae: Yes that was an oversight and should be removed here. Updated.

robincornett (approved):

PR #8268, opened 2020-12-14
Check array keys exist before using & change empty dates to use null #8207

### LINKED ISSUE #8207: 3.0 - Fails to parse time string when creating discount codes
### Expected behavior
When creating discount codes and setting any of the date input fields, Carbon fails due to being passed an invalid date format such as 2020-11-30 ::00. 

This happens when the hour/minute fields are empty and the following conditional logic fails:

`$start_date_hour   = (int) $data['start_date_hour'] >= 0 && (int) $data['start_date_hour'] <= 23 ? sanitize_text_field( $data['start_date_hour'] ) : '00';`

`$start_date_minute = (int) $data['start_date_minute'] >= 0 && (int) $data['start_date_minute'] <= 59 ? sanitize_text_field( $data['start_date_minute'] ) : '00';`

This can be fixed by adding a `null` check at the start, which resolves this issue and allows 00 to be an accepted value.

### Actual behavior
Results in the following error:

_Exception: DateTime::__construct(): Failed to parse time string (2020-11-30 ::00) at position 11 (:)_

1) Add a discount code and do not set the hour/minute fields, but set a date.
2) Try adding the discount.
3) It will fail with the error above.

PHP Version:

EDD Version (or branch): 3.0

WordPress Version: latest

ashleyfae:
Fixes #8207 

Proposed Changes:
1. Add extra `isset()` conditional before using array keys.
2. Replace `sanitize_text_field()` with `intval()` because we're only interested in integers anyway.
3. When editing a discount, use `null` to represent no start/end date instead of `0000-00-00 00:00:00`. This was changed in a different PR and this instance clearly just got missed.

To test:

### Adding a discount

Create a brand new discount code with these steps:

1. Set a start date, but no start time.
2. Set an end date, but no end time.
3. Save.
4. Ensure you get no PHP errors/notices, that the start date is saved in the database with the time `00:00:00`, and that the end date is saved in the database with the time `23:59:59`.

Create a second discount code with these steps:

1. Set a start date and time. Set the time to anything other than `00:00`
2. Set an end date and time. Set the time to anything other than `23:59`
3. Save.
4. Ensure the times are saved correctly.

### Editing a discount

Edit an existing discount code with these steps:

1. Set a start date, but no time. If you're editing a discount code that already has a time filled out, then make sure you wipe out those values so they're completely empty.
2. Set an end date, but no time. If you're editing a discount code that already has a time filled out, then make sure you wipe out those values so they're completely empty.
3. Save.
4. Ensure you get no PHP errors/notices, that the start date is saved in the database with the time `00:00:00`, and that the end date is saved in the database with the time `23:59:59`.

After editing, edit again with these steps:

1. Set the start time to anything other than `00:00`
2. Set the end time to anything other than `23:59`
3. Save.
4. Ensure the times are saved correctly.

After editing, edit one more time with these steps:

1. Wipe out the start date and time completely. Both should be blank.
2. Wipe out the end date and time completely. Bot should be blank.
3. Save.
4. Ensure the `start_date` and `end_date` in the database is set to `null`.

### DIFF
--- a/includes/admin/discounts/discount-actions.php
+++ b/includes/admin/discounts/discount-actions.php
@@ -98,7 +98,7 @@ function edd_admin_add_discount( $data = array() ) {
 		$end_date_hour   = isset( $data['end_date_hour'] ) && (int) $data['end_date_hour'] >= 0 && (int) $data['end_date_hour'] <= 23
 			? intval( $data['end_date_hour'] )
 			: '23';
-		$end_date_minute = isset( $data['end_date_hour'] ) && (int) $data['end_date_minute'] >= 0 && (int) $data['end_date_minute'] <= 59
+		$end_date_minute = isset( $data['end_date_minute'] ) && (int) $data['end_date_minute'] >= 0 && (int) $data['end_date_minute'] <= 59
 			? intval( $data['end_date_minute'] )
 			: '59';
 
--- a/includes/admin/discounts/discount-actions.php
+++ b/includes/admin/discounts/discount-actions.php
@@ -80,11 +80,11 @@ function edd_admin_add_discount( $data = array() ) {
 	// Start date.
 	if ( ! empty( $data['start_date'] ) ) {
 		$start_date        = sanitize_text_field( $data['start_date'] );
-		$start_date_hour   = (int) $data['start_date_hour'] >= 0 && (int) $data['start_date_hour'] <= 23
-			? sanitize_text_field( $data['start_date_hour'] )
+		$start_date_hour   = isset( $data['start_date_hour'] ) && (int) $data['start_date_hour'] >= 0 && (int) $data['start_date_hour'] <= 23
+			? intval( $data['start_date_hour'] )
 			: '00';
-		$start_date_minute = (int) $data['start_date_minute'] >= 0 && (int) $data['start_date_minute'] <= 59
-			? sanitize_text_field( $data['start_date_minute'] )
+		$start_date_minute = isset( $data['start_date_minute'] ) && (int) $data['start_date_minute'] >= 0 && (int) $data['start_date_minute'] <= 59
+			? intval( $data['start_date_minute'] )
 			: '00';
 
 		// The start date is entered in the user's WP timezone. We need to convert it to UTC prior to saving now.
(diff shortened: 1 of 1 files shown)


includes/admin/discounts/discount-actions.php
```diff
@@ -204,35 +204,35 @@ function edd_admin_edit_discount( $data = array() ) {
...
+		$start_date_hour   = isset( $data['start_date_hour'] ) && (int) $data['start_date_hour'] >= 0 && (int) $data['start_date_hour'] <= 23
+			? intval( $data['start_date_hour'] )
 			: '00';
-		$start_date_minute = (int) $data['start_date_minute'] >= 0 && (int) $data['start_date_minute'] <= 59
-			? sanitize_text_field( $data['start_date_minute'] )
+		$start_date_minute = isset( $data['start_date_minute'] ) && (int) $data['start_date_minute'] >= 0 && (int) $data['start_date_minute'] <= 59
+			? intval( $data['start_date_minute'] )
 			: '00';
 
 		// The start date is entered in the user's WP timezone. We need to convert it to UTC prior to saving now.
```
robincornett, line 214: @ashleyfae testing this with a new discount and the start/end dates for which I've left the times empty, have been converted to:
- `2020-12-15 05:00:00`
- `2020-12-19 04:59:00`
Which I believe is intentional as I read the code, but I'm also seeing that existing discounts are set to `00:00:00` and `23:59:59` respectively. The existing discounts were migrated from 2.9.x, so if this conversion is correct, maybe it needs to be applied on migrated discounts as well?

ashleyfae: Sorry my steps are actually kind of deceptive.

In the database times are always stored in UTC. My steps were deceptive because my time zone is currently the same as UTC, which means what I enter is exactly how it gets saved. **This is not the case for other time zones.**

In your case, here's the expected result:

1. It defaults to start at midnight _your time zone_ and end at 23:59 _your time zone_. 
2. It then gets converted to UTC when it gets saved in the database. So that means it will add 5 hours.

Therefore what you're seeing is correct behaviour; my testing notes were just kind of crappy.

---

Regarding the imported ones -- that depends on how you entered them to begin with. We've always stored discounts in UTC, but until now, we never did the time zone conversion. That means technically in 2.9 you should be entering start and end times _in UTC_ -- not in your local time. So if you entered them as 00:00 in 2.9, then seeing them saved now as 00:00 is correct.

robincornett: Thanks, @ashleyfae, that makes sense. This is technically outside the scope of this PR, but I'm wondering if it might make more sense to _save_ the user entered time and convert to the site time on use/retrieval display. I can think of some instances where the current scenario could result in inaccurate times:
- When I set up a discount with specific times before a time change happens, but the discount dates are after the time change (so what I enter will be an hour off, as I understand it).
- If I change my site time zone (unusual/unlikely, but if I discover that my site has incorrectly been running in the default time zone and I change it, then the discount will have been saved in a different time than what's being used).

ashleyfae: Saving in UTC is pretty standard for all time-related things and I don't think we're going to deviate from that.

LisaCee (approved):
- [x] Replicated original issue in `release/3.0`
All of these cases saved in UTC (my timezone + 8)
- [x] Add discount with no start time
- [x] Add discount with start time
- [x] Edit discount, blanking out time
- [x] Edit discount with time other than default
- [x] Edit discount to have no start time or date (null)

robincornett (approved):

Issue #8230, opened 2020-11-30
3.0 – Unable to search discount codes (or names)


davidsherlock:
### Expected behavior
When viewing the discount codes table I expect to be able to search the discounts by either name or code. However, neither of these return any results.

### Actual behavior
No results are returned for any search terms.

### Information (if a specific version is affected):
PHP Version:

EDD Version (or branch): 3.0-beta

WordPress Version: Latest

Any other relevant information:

ashleyfae:
@davidsherlock Can you test this again in the latest `release/3.0` branch? I'm unable to reproduce this behaviour.

If you still have the issue, then can you please reply with:

1. Your exact search term.
2. Screenshot of the settings on the discount that _should_ be appearing in results but isn't.

robincornett:
@ashleyfae I'm getting a database error when doing any search. Here's the error when searching for a specific code (copied from the list table):

>[14-Dec-2020 14:04:48 UTC] WordPress database error Incorrect DATETIME value: '%PASTDUE15%' for query SELECT  edd_a.id FROM mce_edd_adjustments edd_a WHERE edd_a.type = 'discount' AND (parent LIKE '%PASTDUE15%' OR name LIKE '%PASTDUE15%' OR code LIKE '%PASTDUE15%' OR status LIKE '%PASTDUE15%' OR amount LIKE '%PASTDUE15%' OR description LIKE '%PASTDUE15%' OR start_date LIKE '%PASTDUE15%' OR end_date LIKE '%PASTDUE15%' OR date_created LIKE '%PASTDUE15%' OR date_modified LIKE '%PASTDUE15%') ORDER BY edd_a.id DESC LIMIT 30 made by require_once('wp-admin/admin.php'), do_action('download_page_edd-discounts'), WP_Hook->do_action, WP_Hook->apply_filters, edd_discounts_page, edd_adjustments_page, do_action('edd_adjustments_page_discount'), WP_Hook->do_action, WP_Hook->apply_filters, edd_discounts_page_content, EDD_Discount_Codes_Table->prepare_items, EDD_Discount_Codes_Table->get_data, edd_get_discounts, EDD\Compat\Discount_Query->query, EDD\Database\Query->query, EDD\Database\Query->get_items, EDD\Database\Query->get_item_ids

ashleyfae:
@robincornett Maybe it's MySQL version specific. I think you're on a newer version than I am.

[...]

PR #8269, opened 2020-12-14
Limit searchable to text based columns #8266

### LINKED ISSUE #8266: Audit "searchable" columns on all tables

We should audit the custom table columns that have `searchable` enabled and ensure only necessary ones are actually enabled.

### Justification or use case

Every additional column that we search in is going to impact performance. Here's an example of what the query looks like when you search for a discount code (adjustments table):

```mysql
SELECT edd_a.id
FROM wp_edd_adjustments edd_a
WHERE edd_a.type = 'discount'
  AND (parent LIKE '%flat%'
    OR name LIKE '%flat%'
    OR code LIKE '%flat%'
    OR status LIKE '%flat%'
    OR amount LIKE '%flat%'
    OR description LIKE '%flat%'
    OR start_date LIKE '%flat%'
    OR end_date LIKE '%flat%'
    OR date_created LIKE '%flat%'
    OR date_modified LIKE '%flat%')
ORDER BY edd_a.id DESC
LIMIT 30
```

IMO, there's no reason to be searching in `date_created` or `date_modified` in particular. I'm also inclined to remove:

- parent
- amount
- start_date
- end_date

When performing a text search, I think we should limit to text-based columns. In a discount, for example, I think there's a reasonable expectation that when you search it's going to look in name, code, and description for the text you're searching for. If we feel it would be beneficial to search in `start_date` or `end_date`, then that's something we should add additional filters for, separate from regular search (kind of like our additional orders filters). Then we can query on that separately instead of doing a wildcard search.

ashleyfae:
Fixes #8266 

Proposed Changes:
1. Remove all `'searchable' => false` parameters for consistency. The default is false, so omitting it completely is essentially already `false`. I did this to improve scanability so you can more easily search for `searchable` to see all the searchable fields.
2. In general, `searchable` has been removed for all non-text fields. This includes numeric columns (like `object_id` or `parent` etc. and amount fields like order totals) and date / datetime columns.

To clarify point 2, by removing searchable we can still query on those columns. We can still find all adjustments with `object_id` `4`; it just has to be done deliberately in the query arguments:

```php
array(
    'object_id' => 4
)
```

We can also still search orders for all orders with a total greater than $x. That remains unaffected.

What this change prevents is _arbitrary wildcard searching_ in those fields. So something like this will not search in `object_id` : 

```php
array(
    'search' => 4
)
```

## To test:

Perform searches in the following interfaces and ensure you still get results you expect:

- Orders
- Customers
    - Email Addresses
    - Physical Addresses
- Discounts - see #8230 
- Logs (Downloads > Tools > Logs)

### DIFF
--- a/includes/database/schemas/class-adjustments.php
+++ b/includes/database/schemas/class-adjustments.php
@@ -48,7 +48,6 @@ final class Adjustments extends Schema {
 			'length'     => '20',
 			'unsigned'   => true,
 			'default'    => '0',
-			'searchable' => true,
 			'sortable'   => true,
 			'transition' => true
 		),
@@ -77,7 +76,6 @@ final class Adjustments extends Schema {
 			'type'       => 'varchar',
 			'length'     => '20',
 			'default'    => 'draft',
-			'searchable' => true,
 			'sortable'   => true,
 			'transition' => true
 		),
@@ -117,8 +115,7 @@ final class Adjustments extends Schema {
 			'name'       => 'amount',
 			'type'       => 'decimal',
 			'length'     => '18,9',
-			'default'    => '0',
-			'searchable' => true
+			'default'    => '0'
 		),
 
 		// description
@@ -170,7 +167,6 @@ final class Adjustments extends Schema {
 			'default'    => null,
 			'allow_null' => true,
 			'date_query' => true,
-			'searchable' => true,
 			'sortable'   => true
 		),
 
@@ -181,7 +177,6 @@ final class Adjustments extends Schema {
 			'default'    => null,
 			'allow_null' => true,
 			'date_query' => true,
-			'searchable' => true,
 			'sortable'   => true
 		),
 
@@ -192,7 +187,6 @@ final class Adjustments extends Schema {
 			'default'    => '', // Defaults to current time in query class
 			'created'    => true,
 			'date_query' => true,
-			'searchable' => true,
 			'sortable'   => true
 		),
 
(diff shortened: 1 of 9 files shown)


robincornett (commented):
- [x] Orders: email addresses appear to be searched, but not names (searching by my last name pulls orders for both me and a client, who has my domain name set up as an email alias apparently). Searching by a customer's last name if it is not part of their email address returns no results.
- [x] Customers
- [x] Email Addresses
- [x] Physical Addresses
- [x] Discounts
- [ ] Logs (Downloads > Tools > Logs): I can search by IP address, but not anything else.
    - Searching for part of an order number returns no results.
    - Searching for a customer name shows all logs.
    - Searching for a download name (or file name) shows all logs.

ashleyfae:
For logs, here's a comparison of what fields used to be checked, and what's checked now:

## API request logs

Previously it allowed searching in:

- log ID
- user_id
- api_key
- token
- API version
- request body
- error
- ip
- request duration (seconds, I think)

Now it searches in:

- api_key
- token
- request body
- error
- ip

## File download logs

Previous:

- file_id
- customer_id
- ip
- user_agent

Now:

- ip
- user_agent

So it was never checking in order number, customer name, or product name.

There is some logic to check based on customer email address though: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8266/includes/admin/reporting/class-base-logs-list-table.php#L343-L364 But this is separate from the `search` query because it's setting individual arguments (if it's an email then find a customer and then set `customer_id` parameter).

I did notice that the API log search looks fundamentally messed up because it's building an entire meta query, which is no longer needed: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8266/includes/admin/reporting/class-api-requests-logs-list-table.php#L121-L166 so this will actually completely break a lot of API log searching in general. However, I'm going to open a separate issue for this since it's ultimately a different issue.

robincornett (approved):

PR #8278, opened 2020-12-16
Ensure that a mislinked EDD settings page is an EDD admin screen

### LINKED ISSUE #8209: 3.0 - Settings pages missing CSS/JS & fundamentally don't all work when navigated to via different URL
## Bug Report

I've confirmed this is not an issue in `master` branch, only in `release/3.0`.

I'm going to explain the steps first:

1) Install `edd-fes`.
2) Go to EDD FES > Settings.
3) This takes you to EDD settings, but via a slightly different URL.

First, you'll notice the URL is slightly different. The normal URL is `edit.php?post_type=download&page=edd-settings&tab=general&section=main` . Main thing being the first bit: `edit.php?post_type=download&page=edd-settings`. But the new URL is: `admin.php?page=edd-settings&tab=general`.

You'll then notice there's no styling on the sub-navigation.

Normal:

Screenshot from 2020-11-16 14-56-23

vs via the above steps:

Screenshot from 2020-11-16 14-56-40

The CSS isn't loading (and probably not the JS either).

Finally, if you click on a sub-item (like `Currency`) you get a `wp_die()` with `Cannot load edd-settings`.

PHP Version:

WordPress Version:

I will be making an issue in FES to correct the settings URL used, but we should also look into this in core to see if there's something we can do to fix these problems anyway.

robincornett:
Fixes #8209

Proposed Changes:
1. Adds a check for `admin.php` and the EDD settings page to the `edd_is_admin_page()` function.

To test:
Navigate to `admin.php?page=edd-settings&tab=general` in your site. With this branch, the navigation tabs should be styled correctly and the script (Chosen) needed to change the country/region selector will load--that's the most visible thing to check.

### DIFF
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -447,6 +447,10 @@ function edd_is_admin_page( $passed_page = '', $passed_view = '' ) {
 			// Supported post types
 			} elseif ( edd_is_insertable_admin_page() ) {
 				$found = true;
+
+			// Mislinked EDD settings screen
+			} elseif ( 'admin.php' === $pagenow && 'edd-settings' === $page ) {
+				$found = true;
 			}
 			break;
 	}
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -448,8 +448,8 @@ function edd_is_admin_page( $passed_page = '', $passed_view = '' ) {
 			} elseif ( edd_is_insertable_admin_page() ) {
 				$found = true;
 
-			// Mislinked EDD settings screen
-			} elseif ( 'admin.php' === $pagenow && 'edd-settings' === $page ) {
+			// The EDD settings screen (fallback if mislinked)
+			} elseif ( 'edd-settings' === $page ) {
 				$found = true;
 			}
 			break;
--- a/includes/admin/settings/display-settings.php
+++ b/includes/admin/settings/display-settings.php
@@ -30,6 +30,8 @@ function edd_options_page_primary_nav( $active_tab = '' ) {
 			$tab_url = add_query_arg(
 				array(
 					'settings-updated' => false,
+					'post_type'        => 'download',
+					'page'             => 'edd-settings',
 					'tab'              => $tab_id,
 				),
 				edd_get_admin_base_url()
--- a/includes/admin/settings/display-settings.php
+++ b/includes/admin/settings/display-settings.php
@@ -27,10 +27,13 @@ function edd_options_page_primary_nav( $active_tab = '' ) {
 		<?php
 
 		foreach ( $tabs as $tab_id => $tab_name ) {
-			$tab_url = add_query_arg( array(
-				'settings-updated' => false,
-				'tab'              => $tab_id,
-			) );
+			$tab_url = add_query_arg(
+				array(
+					'settings-updated' => false,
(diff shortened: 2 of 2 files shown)


includes/admin/admin-pages.php
```diff
@@ -447,6 +447,10 @@ function edd_is_admin_page( $passed_page = '', $passed_view = '' ) {
 			// Supported post types
 			} elseif ( edd_is_insertable_admin_page() ) {
 				$found = true;
+
+			// Mislinked EDD settings screen
+			} elseif ( 'admin.php' === $pagenow && 'edd-settings' === $page ) {
```
ashleyfae, line 452: I'm wondering if this part is really even needed:

```php
'admin.php' === $pagenow && 
```

If we have `?page=edd-settings` at all then I think that's a decent assumption it should be considered an EDD admin screen.

(Just thinking that being very specific is what caused this issue in the first place.)

ashleyfae:
This is working to get the CSS loaded, but there's still a `wp_die()` if you click through to any of the settings sections other than the "main" one.

Example:

1. Click FES settings (EDD FES > Settings).
2. You're taken to "Main Settings".
3. Click "Forms/Pages".
4. You'll get an error.

Due to the `&post_type=download` in the URL; it goes away if you remove that.

EDD 2.9 sub-navigation link:

```php
$tab_url = add_query_arg( array(
	'settings-updated' => false,
	'tab' => $active_tab,
	'section' => $section_id
) );
```

3.0 equivalent:

```php
$tab_url = add_query_arg( array(
	'post_type' => 'download',
	'page'      => 'edd-settings',
	'tab'       => $active_tab,
	'section'   => $section_id
) );
```

So 2.9 doesn't include the post type.

Perhaps the best solution is simply to edit the above so that it doesn't parse the current URL and instead we pass it the URL we want. That will force it back to using `edit.php` for all links. Example:

```php
$tab_url = add_query_arg( array(
	'post_type' => 'download',
	'page'      => 'edd-settings',
	'tab'       => $active_tab,
	'section'   => $section_id
), edd_get_admin_base_url() );
```

includes/admin/settings/display-settings.php
```diff
@@ -27,10 +27,13 @@ function edd_options_page_primary_nav( $active_tab = '' ) {
...
 
 		foreach ( $tabs as $tab_id => $tab_name ) {
-			$tab_url = add_query_arg( array(
-				'settings-updated' => false,
-				'tab'              => $tab_id,
-			) );
+			$tab_url = add_query_arg(
+				array(
+					'settings-updated' => false,
+					'tab'              => $tab_id,
```
ashleyfae, line 33:
```suggestion
					'post_type'        => 'download',
					'page'             => 'edd-settings',
					'tab'              => $tab_id,
```

I hope the spacing is right lol. It isn't lining up in the `Write` view but is in the `Preview` :P

ashleyfae (changes requested):

includes/admin/settings/display-settings.php
robincornett: @ashleyfae actually I think we can remove the `post_type` altogether from both of these as it's added in the `edd_get_admin_base_url` function?

ashleyfae (approved):

LisaCee (approved):

Issue #7799, opened 2020-04-29
3.0 - Show tax amounts on orders & refunds if taxes are disabled, but the order did have taxes applied at an earlier date


ashleyfae:
- Enable taxes.
- Make a purchase that charges tax.
- Refund the order.
- You'll see tax amounts on both the order and refund.
- Disable taxes.
- Tax amounts disappear from the order and refund.

I think if the tax amount is > 0 then we should always show those records on the order, regardless of site settings.

If the tax amount is 0, then we conditionally check if taxes are enabled or not.

---

I don't think we necessarily have to change this for 3.0 itself because it's unlikely that a store owner will charge tax, then stop charging tax. But worth addressing later down the line, I think.

spencerfinnell:
Good idea. In theory we can expand this conditional to check the current Order's tax amount: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/payments/orders.php#L667

robincornett:
Noting that this behavior exists in EDD 2.9.x as well.

ashleyfae:
Thanks @robincornett . Going to punt this for now; we can address it shortly after 3.0.

robincornett:
Noting that this is partially addressed in PR #8279, but only for the edit order screen, not the refund screen.

PR #8279, opened 2020-12-17
Show tax rates on order edit screen

### LINKED ISSUE #8242: 3.0 - Tax rate percentage / label not shown on admin orders page
### Expected behavior

When I view an order, I expect to see the tax percentage and the "label" (country / region from the tax rate)

### Actual behavior

The tax _amount_ is displayed, but not the percentage / label.

Screenshot from 2020-12-04 12-49-10

I'm filing this as a bug, because when you are in the middle of manually creating an order, the rate and label is shown:

Screenshot from 2020-12-04 12-52-12

But as soon as you save the order, that label is gone.

1) Make an order that has a tax rate applied. (Either manually or via the front-end; doesn't matter.)
2) View the order in the admin area.
3) Note the lack of tax rate.

PHP Version:

EDD Version (or branch):

WordPress Version:

Related: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8060

robincornett:
Fixes #8242

Proposed Changes:
1. When setting up the localized data for the order add/edit screen, only send an empty array of tax rate data to on the add edit screen, and only if taxes are enabled.
2. On the edit order screen, the tax rate ID, country, and region are localized.
3. Since the tax rate amount uses `floatval`, the template no longer restricts the rate to two decimal places (as this may not be accurate anyway).

This partially addresses #7799 as well--on the edit order screen, the tax rate is always shown if it exists, even if taxes are disabled on the store.


includes/admin/payments/orders.php
```diff
@@ -684,6 +684,29 @@ function edd_order_details_overview( $order ) {
...
+			'region'  => '',
+		);
+	} elseif ( $tax_rate ) {
+		$has_tax = array(
+			'rate'    => $tax_rate,
+			'country' => '',
+			'region'  => '',
+		);
+		if ( $order->tax_rate_id ) {
+			$tax_rate_object = edd_get_adjustment( $order->tax_rate_id );
```
ashleyfae, line 702: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8253/files adds a new `get_tax_rate_object()` method so let's wait until that's merged, then this PR can be updated to use that.

robincornett: Oh, I can go ahead and update this PR--it's based off of that one. I did use the new `get_tax_rate` method in here already.

ashleyfae: Sure! :+1:

includes/admin/payments/orders.php
```diff
@@ -684,6 +684,29 @@ function edd_order_details_overview( $order ) {
 		}
 	}
 
+	$has_tax  = 0;
```
spencerfinnell, line 687: Lets make the default `false` to match the first condition [here](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/views/tmpl-order-totals.php#L32)

includes/admin/payments/orders.php
```diff
@@ -684,6 +684,29 @@ function edd_order_details_overview( $order ) {
...
+	$tax_rate = $order->get_tax_rate();
+	if ( edd_is_add_order_page() && edd_use_taxes() ) {
+		$has_tax = array(
+			'rate'    => 0,
+			'country' => '',
+			'region'  => '',
+		);
```
spencerfinnell, line 694: Code style suggestion: maybe define the fallback outside of the condition so the same value can be used in multiple places.

spencerfinnell (changes requested):

includes/admin/payments/orders.php
robincornett: If I change it to `false`, it fails [this check](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/views/tmpl-order-totals.php#L29) instead (I think `false` is just empty after localization and `0` is read as `false`, at least as far as I can tell?).

spencerfinnell: Let's make it an explicit `false` string check so we don't have to figure out/guess what `wp_localize_script()` is doing.

robincornett: I'm okay with changing it but it does affect multiple templates (`order-totals`, `order-item`, `add-order-item` at least). Can I use `none`? I like that better than `no` for whatever reason.

spencerfinnell: Yep, that works.

cklosowski:
I restarted the  failing tests here. They were just failing due to a timing issue by 1 second.

spencerfinnell (approved):

LisaCee (approved):
- [x] Country-wide tax rate
- [x] State tax rate
- [x] Refund
- [x] Tax rate and area visible after disabling taxes on Edit Order (#7799).

includes/admin/payments/class-refund-items-table.php
```diff
@@ -432,7 +433,10 @@ public function display_rows() {
 			</td>
 		</tr>
 
-		<?php if ( edd_use_taxes() ) : ?>
+		<?php
+		$order = $this->get_order();
+		if ( $order && $order->get_tax_rate() ) :
+			?>
```
ashleyfae, line 439:
```suggestion
		?>
```

Minor indenting issue.

ashleyfae (changes requested):
Tiny indenting issue - then I'll be happy to merge.

ashleyfae (approved):

Issue #8075, opened 2020-07-29
3.0 - Order credits & fees are not managed correctly with refunds


robincornett:
### Expected behavior
When I refund a manually added order which includes a credit, I expect the refund total to match the original order total.

### Actual behavior
When I refund a manually added order which includes a credit, the refund total does not match ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8069#issuecomment-665744426)).

### Steps to reproduce the behavior
1) Generate a manual order and apply a credit.
2) Refund the order.
3) Compare the original order subtotal/total amounts in the database with the refund order subtotal/total amounts.

### Information (if a specific version is affected):
PHP Version: 7.4.1
EDD Version (or branch): release/3.0
WordPress Version: 5.4.2

spencerfinnell:
Illustrating what can happen:

<img width="799" alt="Screen Shot 2020-07-29 at 12 09 20 PM" src="https://user-images.githubusercontent.com/491124/88824733-baf1df80-d194-11ea-9a86-6c578fd68167.png">
<img width="992" alt="Screen Shot 2020-07-29 at 12 09 28 PM" src="https://user-images.githubusercontent.com/491124/88824746-bd543980-d194-11ea-8a85-0f7193b8272b.png">

I originally added the `credit` type to the "Add Adjustment" flow because it existed in the original UI, but I'm not sure if handling of that adjustment type was fully fleshed out, especially with refunds. I believe only the Fees API is used in things like EDD Wallet?

At this point in time I believe removing Credits from the manual Order creation flow to be implemented more fully at a later date might be the best path forward.

ashleyfae:
Copying from https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8203, which was about fees, but are ultimately the same as credits anyway:

## Bug Report
### Expected behavior

I expect to be able to issue a **full** refund, including any fees that were on the order.

### Actual behavior

Fees are not shown in the "Submit Refund" modal, which means you can never select them for refund, which means the most you can ever refund is the total **without** fees. It's impossible to truly do a full refund.

I manually created an order and applied a $50 fee to the whole order.

![Screenshot from 2020-11-11 13-40-55](https://user-images.githubusercontent.com/6324272/98818690-cefbdd00-2423-11eb-955b-080e1fddaa60.png)

The refund modal does not show this fee at all.

[...]

PR #8282, opened 2020-12-21
Show quantities for existing orders which have them

### LINKED ISSUE #8281: 3.0 - Existing orders should show quantities if enabled at the time of order creation
### Expected behavior
If an order contains multiples of an item, I expect the order details screen to reflect that, regardless of the store's current setting.

### Actual behavior
If quantities were enabled, and a purchase was made with multiple copies of an item, and then quantities were disabled, then the order details screen no longer shows the quantity purchased. The total order amount is correct, but the order items do not appear to add up correctly:

image

1) Enable "Cart Item Quantities" on the storefront.
2) Create an order with more than one of a product in the cart.
3) View the order details--the quantity will show.
4) Disable "Cart Item Quantities".
5) View the same order; the quantity no longer shows.

PHP Version: 7.4.1
WordPress Version: 5.6

Related: #7799. The localization code checks for active discounts and whether quantities are enabled, but this should really only be done for new orders. Viewing an existing order or refund should show these items based on whether the order/refund has quantities applied, regardless of the store's current setting.

I thought originally that discounts may also have this issue, but they do still appear on the order details regardless of whether any discounts are active.

robincornett:
Fixes #8281

Proposed Changes:
1. If an existing order has quantities of an item, show that in the order details, refund details, and refund modal, even if quantities have since been disabled.

To test:
1. Enable quantities.
2. Place an order with multiples of at least one item.
3. Disable quantities.
4. The order will not show quantities in `release/3.0`, but in this branch, quantities should show on the order details screen, refund modal, and refund details screen.


spencerfinnell:
Building the Order/Refund script data is already pretty cumbersome and it feels like we are starting to weave a lot of additional logic in to it. My vague [`@todo` notes](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/payments/orders.php#L612) were originally based on the idea that in the future BerlinDB might be able to do this for us, or let us define alternative output formats closer to the database level. 

I think we should try to keep more of the "processing" out of these areas and instead introduce an Order-specific version of `edd_item_quantities_enabled()` that doesn't need to leverage those foreach loops that happen to exist in this context.

Thoughts?

robincornett:
I did think about that and would be perfectly willing for my part. Possibly add it as a public method in the `EDD_Order` class, similar to what @ashleyfae is doing to get the order tax rate and tax rate object? What would you think of that, and just automatically returning `true` if the storefront has quantities enabled, as the additional checks would not be needed?

Thinking about this and thinking maybe the front end receipt needs to be checked too.

spencerfinnell:
Adding it to the public API was a hesitation on my end as well. I'll defer to @ashleyfae on that.

I actually looked at `EDD_Order` to see what kind of helpers existed there already and noticed semi similar tax helpers. But there are also already some helpers [here](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/functions/orders.php#L463).

ashleyfae:
> Possibly add it as a public method in the `EDD_Order` class

I like that :+1:

ashleyfae:
Actually I want to think about this a bit more. The only reason I want to think about it is because this logic isn't entirely accurate. It's our best guess as to whether or not quantities are enabled just based on whether more than one was purchased. But someone may have just bought one when quantities were enabled, which would result in this being a bit misleading. :shrug: It's not like tax rates where we can know for certain if tax was charged at the time. Here we're just guessing, and I don't love introducing a public API for "just a guess".

robincornett:
Yes, I thought of that--if only one of an item was purchased when quantities were enabled, and then quantities are disabled, the order with just one of each item would not appear any different than a new order--it would only affect orders where multiples of an item were purchased, based on the total number of items in the order being greater than the total number of unique items. My feeling was that it is more important to show the accurate order data than to make all orders placed during a certain time appear "correct".

Note: not germane to this issue, but refunding an item when more than one of it was purchased is currently an all or nothing scenario. If three "Sample Download" items are purchased in one order, all three must be refunded; there isn't a way to refund just one of them. The change to partial/freeform refunds may help address this.

spencerfinnell:
Instead of adding up each quantity in the order items couldn't we instead check if any of the order's items have a `quantity` column value greater than 1?

ashleyfae:
> Note: not germane to this issue, but refunding an item when more than one of it was purchased is currently an all or nothing scenario. If three "Sample Download" items are purchased in one order, all three must be refunded; there isn't a way to refund just one of them. The change to partial/freeform refunds may help address this.

This is sorted in #8075 and I pretty much already have that working. That soon-to-be-created-PR also always shows `Quantity` on the refund UI, regardless of settings. Frankly I'm not opposed to implementing that everywhere (always show quantities regardless of settings). I don't necessarily think that's confusing.

spencerfinnell:
> Frankly I'm not opposed to implementing that everywhere (always show quantities regardless of settings). I don't necessarily think that's confusing.

@ashleyfae I was just thinking similar things. It doesn't seem out of place to always show whatever quantity was purchased in the admin, and always make quantities available to admins for manual order creation. The setting `Cart Item Quantities` is already pretty frontend-specific to _me_.

robincornett:
Always showing the quantity makes sense to me as well, and seems less convoluted; I can rework this PR to do so.

spencerfinnell:
Noting that @pippinsplugins found always showing quantities confusing in https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8284#issuecomment-749595172

cklosowski:
Just re-ran the build to make sure it passed, we had the 1 second off error again.

LisaCee (approved):
Quantity showing for:
- [x] Order details
- [x] Refund modal
- [x] Refund details

cklosowski (approved):

robincornett:
@cklosowski re-requesting your review, please. Here's the behavior/changes now:
1. If quantities are disabled, the quantity column is not added to the new order screen. Since quantities aren't shown in the modal for adding downloads to the order, this feels okay.
2. For existing orders and the refund modal, quantities are _always_ shown, regardless of whether the store has quantities enabled or not.
3. The refund validator now includes a check for whether any items/adjustments have been selected for a refund, because it's possible to set the quantity to refund to `0` but still have a value in the subtotal/total, and the refund amount error was confusing.

The net result is that the JS for orders has not actually been updated after all, since the quantities can be checked for new and existing orders both. Just the relevant PHP/view files have been updated. Although I think we _could_ just always set `$has_quantity` to true, I think that would cause problems with creating new orders when quantities are disabled, because the modal would then also show a quantity selector, and I'm not sure we want to allow that?

robincornett:
Updated in 0dad828:
image

cklosowski:
Okay I think have this tuned up @robincornett.

If the quantity is `1`, the input is a text input and is 'readonly', so we can't end up refunding more than the quantity we had oringially.

If the quantity is more than `1` we present a select with the max being the total number to refund.

Recalculations still happen correctly from what I'm seeing, give it a run through please.

I think this is one area we can maybe improve still a bit and that is if you have something like a quantify of 5, and then refund 3, you can submit a subsequent refund and the total available to refund is still 5 in the select. We may need to find a way to set the 'max' refundable quantity based on existing refunds for the order.

robincornett:
> I think, though, that prior to this, we did allow a refund quantity of 0, while still refunding part of the order total--eg if someone meant to use a coupon but did not, we could allow a refund of 10%, say, without kicking off things like licenses being deactivated. So I'm wondering if 0 needs to be allowed to be a value in the select, because of partial refunds and how they can work.

Updating to say that prior to this change that we did _not_ allow a quantity amount of `0` on the refund so I'm not making any further changes to this.

cklosowski:
@robincornett Being that we might have to find a way to support a partial refund, I think `0` might have to be supported. The initial vision was to be able to add a 'credit' to a customer in order to do this, via a refunded order, but with that not in play we'll need to find a way to be able to partially refund an order item, while still keeping the order item in tact and usable with any licenses that are purchased along side it. As it stands, if we fully refund a quantity of an order item, any licenses will be marked as 'revoked'.

robincornett:
@cklosowski we do support partial refunds on an item. As of version 3.7.2, Software Licensing only revokes a license if an order item is marked as fully refunded ([ref](https://github.com/awesomemotive/EDD-Software-Licensing/pull/1788)).

At this moment, I think that accepting `0` as an allowed quantity in a refund would be a significant enough behavior change that I would be reluctant to introduce it at this point. The release branch doesn't support it--if you manually change the quantity to `0` in the release branch the refund is disabled. I think at this point, this PR updates the UX in a helpful and consistent way without introducing structural changes, which I think allowing `0` would have the potential to do.

cklosowski:
That's a fair assessment @robincornett. Let's go ahead and resolve the CSS conflicts with it and merge it.

cklosowski (approved):

PR #8283, opened 2020-12-21
Prevent EDD sessions starting during Site Health loopback requests

### LINKED ISSUE #8274: WordPress site health loopback check fails in WP 5.6 due to session start
### Expected behavior

I expect the loopback request health check test to succeed.

### Actual behavior

When EDD is installed, the loopback test fails with this error:

> The loopback request to your site failed, this means features relying on them are not currently working as expected.
Error: cURL error 28: Operation timed out after 10000 milliseconds with 0 bytes received (http_request_failed)

1) Make sure you have EDD sessions enabled and that you're on WordPress 5.6.
2) Go to Tools > Site Health.
3) Wait for tests to finish running.
4) Note that `Your site could not complete a loopback request`

PHP Version:

EDD Version (or branch):

WordPress Version: 5.6

In 5.6 they changed the tests to use the REST API instead of ajax. This issue occurs when you start a session, don't close that session, then do a curl request. Here's the order of events:

- Request to the REST API endpoint.
- EDD session starts.
- Endpoint callback executes `wp_remote_get()` on the `admin_url()`
- Because the session started but was never closed, the curl request times out.

robincornett:
Fixes #8274

Proposed Changes:
1. Prevents the EDD session from starting during a loopback request.
2. Makes the sanitized `$uri` available to the `edd_start_session` filter.

To test:
1. Running WP 5.6, perform a Site Health check using the `master` branch. The loopback request will fail.
2. Switch to this branch and reload the Site Health screen. The loopback request should no longer fail.

Depending on your server configuration, this fix may not work--running LocalWP here on Windows and it still fails; I personally had to test on a "live" site.

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -364,6 +364,10 @@ class EDD_Session {
 				$start_session = false;
 			}
 
+			// Do not start a session during the loopback request.
+			if ( false !== strpos( $uri, 'wp-site-health/v1/tests/loopback-requests' ) ) {
+				$start_session = false;
+			}
 		}
 
 		return apply_filters( 'edd_start_session', $start_session );
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -339,6 +339,7 @@ class EDD_Session {
 	public function should_start_session() {
 
 		$start_session = true;
+		$uri           = '';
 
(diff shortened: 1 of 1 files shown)


dgoldak (approved):
Tested with `master` and confirmed loopback error occurred. Switched to `issue/8274` and loopback error did not occur when on the heath check page.

cklosowski:
Updating with latest from the 2.10 release

PR #8284, opened 2020-12-22
Rework refunds

### LINKED ISSUE #8075: 3.0 - Order credits & fees are not managed correctly with refunds

ashleyfae:
Fixes #8075 

### Proposed Changes:

Screenshot from 2020-12-22 14-23-57

1. Allow quantities of order items to be refunded. So if an order has 3 quantities of an item, you can choose to refund just 1 or 2 of those. Incrementing the quantity auto-fills the Subtotal and Tax fields, but those are still editable afterwards.
2. Allow partial refunds on order items. So if there's a $10 order item, you can choose to refund just $5 of that.
3. Remove new (in 3.0) unused functions. These were removed because they were not used in core and they seemed like theoretical _"can be done, so let's add them"_ functions. I chose to remove them because I'd rather spend more time discussing and planning how these things should be done. Once we add them we have to maintain them, and given that they're not actually in use, I'd rather just remove them for now.
    - `edd_apply_order_credit()` https://github.com/easydigitaldownloads/easy-digital-downloads/blob/9c618aa87b190db4677756e74599ced2d277b2e0/includes/orders/functions/credits.php#L26
    - `edd_apply_order_item_credit()` https://github.com/easydigitaldownloads/easy-digital-downloads/blob/9c618aa87b190db4677756e74599ced2d277b2e0/includes/orders/functions/credits.php#L163
    - `edd_apply_order_discount()` (retroactively) https://github.com/easydigitaldownloads/easy-digital-downloads/blob/9c618aa87b190db4677756e74599ced2d277b2e0/includes/orders/functions/discounts.php#L25
    - `edd_refund_order_item()` Removed because it seems like almost entirely duplicated code to `edd_refund_order()`.
4. Introduce new `Refundable_Item` trait. This is used by both `Order_Item` and `Order_Item_Adjustment`. The trait comes with two methods:
    1. `get_refunded_items()` - Returns all items that were refunded from this original item. For example: if you have an order item, then you refunded that order item, this method would return the _refunded_ order item that matched the original. Due to partial refunds, there could be more than one result for a single order item.
    2. `get_refundable_amounts()` - This uses the previous function to determine the maximum amounts (subtotal, tax, and total) that can be refunded for that item.
5. `edd_refund_order()` has been reworked.
    - The second parameter is no longer an array of order item IDs, but a multidimensional array, which includes the IDs but also the amounts. 
    - A third parameter has been introduced. It's similar to the second, but instead of an array of order item data it's an array of fee data.
    - `edd_refund_order()` now returns a `WP_Error` object on failure instead of `false`. This helps better convey _why_ it failed.
6. A new `Refund_Validator` class has been introduced. This handles all the logic for validating the items being refunded. It ensures all the items passed through have the required parameters and do not exceed the maximum amounts.
7. Added a new `parent` column to both the `edd_order_items` table and the `edd_order_adjustments` table. This is similar to the `parent` column on the `edd_orders` table; it links a refunded record to the original order record. So if you have an order item with an ID of `3`, the parent for that would be `0` because it has no parent. Then if you refund that order item, the item associated with the refund record will have `parent` set to `3`. This is so we can easily fetch all the refund records associated with an order item or an adjustment (to keep track of the refundable amounts).
8. The 3.0 migration has been updated to utilize this new `parent` column and also to fix a few inconsistencies.
9. Introduced a new hook: `edd_after_submit_refund_table`. This first _after_ the refund table but before the submit button. The idea being that third party plugins could hook in here to add their refund stuff. Like how Recurring would add an option to "Also cancel subscription".

#### Discount amounts

I made a decision regarding discount amounts that I don't really love, but feels like the simplest solution. I'm open to input on this in case I'm wrong and there's a simpler option that I overlooked.

Say you have an order item:

| Unit Price | Subtotal | Discount Amount | Total
| ------------- | ----------- | ----------------------- | ----- |
| 10             | 10          | 2                           | 8    |

When you refund that order item, the refund record gets added to the database like this:

| Unit Price | Subtotal | Discount Amount | Total
| ------------- | ----------- | ----------------------- | ----- |
| 10             | 8            | 0                           | 8    |

Note that the discount amount never gets "refunded".

I made this decision for two reasons:

1. I think it's confusing to reveal the discount amount to the user as something that can even be refunded. I wanted to handle it completely "internally" without exposing it as a field in the UI.
2. "Refunding" that discount amount is simple enough when refunding an entire order item. But, it gets kind of confusing when dealing with partial order item refunds. What do you do if someone only wants to refund $2 from that $8 order? Do you "refund" the discount amount? What does that even mean? How do you keep track of how much of that has been refunded?

In short: dealing with that number at all just felt like an unnecessary complexity. So when calculating the maximum refund amount, it just gets immediately subtracted from the original subtotal. https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8075/includes/orders/classes/class-order-item.php#L297

#### Quantities

When partially refunding an order item, the refund record still records a `-1` quantity. That's not really accurate since the whole item hasn't been refunded, but :shrug: Unfortunately this also means we can't really keep a running tally of the _quantity_ that has been refunded. Therefore, the refund UI will not convey how many quantities have already been refunded. It's a bummer, but I can't think of a great way to do it that.

#### Adjustments

Fees and credits can be refunded. The `discount` adjustment type is never refunded.

The one thing to note is how order item fees are handled.

If an adjustment attached to an order is refunded, then the refund gets an order adjustment record with: 
- object_type = order
- object_id = the ID of the refund

If an adjustment attached to an _order item_ is refunded, then there are two possibilities for what happens:
1. If the associated order item has also been refunded then the refund order adjustment looks like this:
    - object_type = order_item
    - object_id = the ID of the refund's order item record
2. If the associated order item has NOT been refunded, then the adjustment gets converted to an `order` fee.
    - object_type = order
    - object_id = the ID of the refund

It's not really ideal but I can't think of a better way to handle cases where the adjustment is refunded but the item is not.

### Testing

Some testing ideas:

1. Refund an entire order. Ensure order and all order items are `refunded`. 
2. Have an order that contains two or more order items. Entirely refund just one of them. Ensure order status is `partially_refunded`, the status of the refunded item is `refunded`, and the status of the non-refunded item is `complete`.
3. Enable item quantities. Have an order that contains 2 or more of a single item. Refund just one of those items. Ensure the order status is `partially_refunded` and the status of the order item is `partially_refunded`.
4. Enable taxes. Fiddle around with the refund UI. Ensure auto calculated amounts are correct.
5. Manually change the suggested refund amounts. Ensure your amounts are recorded correctly in the refund record.
6. Intentionally go over the listed "maximum" subtotal / tax amount(s). Ensure you get an error when trying to create the refund.
7. Have an order that contains a discount code. Ensure maximum subtotal is the unit price after the discount has been applied. (Combining discounts with quantities is a good test too.)
8. Have an order with an item and a fee. Only refund the fee; do not refund the order item.


includes/admin/payments/class-refund-items-table.php
```diff
@@ -140,11 +136,71 @@ protected function get_primary_column_name() {
...
+						_e( 'Amount of tax to refund', 'easy-digital-downloads' );
+					}
+					?>
+				</label>
+				<?php
+				if ( 'before' === $currency_pos ) {
+					echo esc_html( $this->get_currency_symbol( $order_item->order_id ) );
+				}
+				?>
+				<input type="text" id="edd-order-item-<?php echo esc_attr( $order_item->id ); ?>-refund-<?php echo esc_attr( $column_name ); ?>" class="edd-order-item-refund-<?php echo esc_attr( $column_name ); ?> edd-order-item-refund-input" name="refund_order_item[<?php echo esc_attr( $order_item->id ); ?>][<?php echo esc_attr( $column_name ); ?>]" value="<?php echo esc_attr( edd_sanitize_amount( 0 ) ); ?>" data-original="<?php echo esc_attr( $original_amount ); ?>" data-max="<?php echo esc_attr( $amount_remaining ); ?>" <?php echo 'refunded' === $order_item->status ? ' disabled' : ''; ?> />
```
robincornett, line 187: Should this be `type="number"` instead? If it needs to be `text`, which may be the case for differing currency separators (?), I think it needs to have `small-text` added as a class, although that's potentially too small. Noticing due to issues with responsiveness--we may actually just need to apply custom CSS for this to work with the currency symbol in the table.

pippinsplugins:
@ashleyfae I've been playing with this a bit and have a few thoughts.

1. When first clicking `Initialize Refund`, all of the input fields have `0` or `0.00` pre-populated as actual values. The user has to fully replace those values by focusing into the field, deleting, and then entering the desired numbers. Can we use placeholders instead of values so that users do not need to explicitly delete the zero values?

2. We don't need to worry about it for beta1 but mentioning it just to be sure it's recorded. The input fields are too big on smaller screens. My 11" laptop shows this:
image

3. The quantity field is confusing when a store doesn't use item quantities. Could we hide that field when quantities are not enabled?

4. A 0.00 order cannot be refunded. When trying to issue a refund, the `Initialize Refund` button shows but it's not possible to make the UI issue a refund. This makes sense because, well, there isn't any money to refund. How about disabling the `Initialize Refund` button on 0.00 orders?

image

assets/js/admin/orders/order-overview/_refund.js
```diff
@@ -59,74 +59,104 @@ $( document.body ).on( 'click', '.ui-widget-overlay', function ( e ) {
...
-		new_total    += item_total;
-		can_refund    = true;
-	});
+		let thisItemSubtotal = 0,
+			thisItemTax      = 0,
+			thisItemTotal    = 0;
+
+		if ( thisItemQuantity ) {
+			thisItemSubtotal = parseFloat( thisItemParent.find( '.edd-order-item-refund-subtotal' ).val() ),
+			thisItemTax      = parseFloat( thisItemParent.find( '.edd-order-item-refund-tax' ).val() ),
```
robincornett, line 131: If taxes are not enabled, this line results in `NaN` as the input is not present; it borks the subtotal and total calculations.

robincornett (commented):
Just starting to review, but wanted to go ahead with some comments.

ashleyfae:
Updates:

- Account for taxes being disabled.
- Remove default `0` values and replace with placeholders.
- Amount inputs are disabled by default and when the quantity is `0` to help hint that you need to increase the quantity first before editing amounts.
- Don't show "Initialize Refund" button on $0 orders.

I'm holding off on any UI changes for now. Once the base functionality is nailed down, I'm going to hand that off to @robincornett .

ashleyfae:
Updated to address the above changes.

Also I'm thinking about bringing back the checkboxes ( https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8284/commits/d93d84540fd84db1e23b6cf74f7a31baef653a6f ). Thoughts?

includes/admin/payments/class-refund-items-table.php
```diff
@@ -141,11 +129,71 @@ protected function get_primary_column_name() {
...
+				$original_amount = $order_item->{$column_name};
+				if ( 'subtotal' === $column_name ) {
+					$original_amount -= $order_item->discount;
+				}
+				ob_start();
+				?>
+				<label for="edd-order-item-<?php echo esc_attr( $order_item->id ); ?>-refund-<?php echo esc_attr( $column_name ); ?>" class="screen-reader-text">
+					<?php
+					if ( 'subtotal' === $column_name ) {
+						_e( 'Amount to refund, excluding tax', 'easy-digital-downloads' );
```
robincornett, line 169: Should these both be `esc_html_e` instead?

includes/admin/payments/class-refund-items-table.php
```diff
@@ -219,19 +267,41 @@ public function column_name( $order_item ) {
...
+
 			return sprintf(
-				'<input type="checkbox" name="%1$s[]" id="%1$s-%2$s" value="%2$s" /><label for="%1$s-%2$s" class="screen-reader-text">%3$s</label>',
-				/*$1%s*/
-				'order_item',
-				/*$2%s*/
-				esc_attr( $order_item->id ),
-				/* translators: product name */
-				esc_html( sprintf( __( 'Select %s', 'easy-digital-downloads' ), $order_item->product_name ) )
+					'<input type="checkbox" name="%1$s[]" id="%1$s-%2$s" class="edd-order-item-refund-checkbox" value="%2$s" /><label for="%1$s-%2$s" class="screen-reader-text">%3$s</label>' . $quantity_html,
```
robincornett, line 280: These lines have gotten an extra indent.

robincornett (commented):
Left some comments. I'm wondering whether we should allow taxes to be edited, or if they should be just automatically calculated off of the subtotal (full or partial). I feel like that's a potential place for errors to happen. Maybe only quantity (if applicable) and subtotal should be editable fields?

includes/admin/payments/class-refund-items-table.php
```diff
@@ -71,16 +71,16 @@ public function get_columns() {
 		$columns = array(
 			'cb'       => '<input type="checkbox" />',
 			'name'     => __( 'Product', 'easy-digital-downloads' ),
-			'amount'   => __( 'Amount', 'easy-digital-downloads' ),
+			'amount'   => __( 'Unit Price', 'easy-digital-downloads' )
 		);
 
 		// Maybe add quantity column.
 		if ( edd_item_quantities_enabled() ) {
```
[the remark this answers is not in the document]
ashleyfae: I'm inclined to leave it as-is for now until we figure out #8281 . I'm hoping the scenario where someone had them enabled, then disabled them, and is now refunding a previous order is pretty uncommon.

robincornett (commented):
I like having the checkboxes back if there aren't quantities. I still am not sure about allowing the tax amounts to be adjusted manually.

ashleyfae:
I've updated the PR note with a list of all the new changes. Any comments or questions about the changes? If not, @robincornett do you want to get started on the UI fixes? I figure you could do that then we could do all the testing at the end. (Testing for both UI and functionality at the same time.)

includes/admin/payments/class-refund-items-table.php
```diff
@@ -191,59 +295,88 @@ private function format_currency( $order_item, $column_name ) {
...
+	 * Render the Name Column
+	 *
+	 * @since 3.0
+	 *
+	 * @param Order_Item|Order_Adjustment $item Order Item object.
+	 *
+	 * @return string Data shown in the Name column
+	 */
+	public function column_name( $item ) {
+		return '<span class="row-title">' . esc_html( $this->get_item_display_name( $item ) ) . '</span>';
```
robincornett, line 365: Could this be reverted somewhat to what it is in 3.0? ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/payments/class-refund-items-table.php#L365))

I think the `row-title` class is better than the `strong`, but as I'm testing/revising CSS, I'm remembering that we had intentionally made the refundable items output be a `label` element because I keep being lazy and clicking the order item name instead of the checkbox.

includes/admin/payments/class-refund-items-table.php
```diff
@@ -125,30 +117,140 @@ public function get_sortable_columns() { return array(); }
...
+				 * For subtotals, we actually do subtotal minus discounts for simplicity so that the end user
+				 * doesn't have to juggle that.
+				 */
+				$original_amount = $item->{$column_name};
+				if ( 'subtotal' === $column_name && ! empty( $item->discount ) ) {
+					$original_amount -= $item->discount;
+				}
+				ob_start();
+				?>
+				<label for="edd-order-item-<?php echo esc_attr( $item_id ); ?>-refund-<?php echo esc_attr( $column_name ); ?>" class="screen-reader-text">
```
robincornett, line 220: Could we update this cell markup to use the new form group markup? It would make it more manageable for styling. Something like this:

```php
				<div class="edd-form-group">
					<label for="edd-form-group__label edd-order-item-<?php echo esc_attr( $item_id ); ?>-refund-<?php echo esc_attr( $column_name ); ?>" class="screen-reader-text">
						<?php
						if ( 'subtotal' === $column_name ) {
							esc_html_e( 'Amount to refund, excluding tax', 'easy-digital-downloads' );
						} else {
							esc_html_e( 'Amount of tax to refund', 'easy-digital-downloads' );
						}
						?>
					</label>
					<div class="edd-form-group__control">
						<?php
						if ( 'before' === $currency_pos ) {
							echo '<span class="edd-amount-control__currency is-before">';
							echo esc_html( $this->get_currency_symbol( $item->order_id ) );
							echo '</span>';
						}
						?>
						<input type="text" id="edd-order-item-<?php echo esc_attr( $item_id ); ?>-refund-<?php echo esc_attr( $column_name ); ?>" class="edd-order-item-refund-<?php echo esc_attr( $column_name ); ?> edd-order-item-refund-input" name="refund_<?php echo esc_attr( $object_type ); ?>[<?php echo esc_attr( $item->id ); ?>][<?php echo esc_attr( $column_name ); ?>]" value="<?php echo esc_attr( edd_format_amount( $amount_remaining ) ); ?>" placeholder="<?php echo esc_attr( edd_format_amount( 0 ) ); ?>" data-original="<?php echo esc_attr( $original_amount ); ?>" data-max="<?php echo esc_attr( $amount_remaining ); ?>" disabled />
						<?php
						if ( 'after' === $currency_pos ) {
							echo '<span class="edd-amount-control__currency is-after">';
							echo esc_html( $this->get_currency_symbol( $item->order_id ) );
							echo '</span>';
						}
						?>
					</div>
					<small class="edd-order-item-refund-max-amount">
						<?php
						echo _x( 'Max:', 'Maximum input amount', 'easy-digital-downloads' ) . '&nbsp;';

						echo $this->format_currency( $item, $column_name, $amount_remaining );
						?>
					</small>
				</div>
```

assets/js/admin/orders/order-overview/_refund.js
```diff
@@ -59,120 +63,194 @@ $( document.body ).on( 'click', '.ui-widget-overlay', function ( e ) {
 	$( '#edd-refund-order-dialog' ).dialog( 'close' );
 } );
 
-// Handles including items in the refund.
-$(document.body).on( 'change', '#edd-refund-order-dialog tbody .check-column input[type="checkbox"]', function () {
-	let parent = $(this).parent().parent(),
-		all_checkboxes = $('#edd-refund-order-dialog tbody .check-column input[type="checkbox"]');
+/**
+ * Listen for the bulk action scheckbox, since WP doesn't trigger a change on sub-items.
```
LisaCee, line 67: Teeny-tiny typo

includes/orders/functions/refunds.php
```diff
@@ -307,264 +308,110 @@ function edd_refund_order( $order_id = 0, $order_items = array() ) {
...
-			'object_type' => 'order',
-			'type_id'     => $adjustment->type_id,
-			'type_key'    => $adjustment->type_key,
-			'type'        => $adjustment->type,
-			'description' => $adjustment->description,
-			'subtotal'    => edd_negate_amount( $adjustment->subtotal ),
-			'tax'         => edd_negate_amount( $adjustment->tax ),
-			'total'       => edd_negate_amount( $adjustment->total ),
-		) );
+	foreach( $validator->get_refunded_adjustments() as $adjustment ) {
```
robincornett, line 343:
```suggestion
	foreach ( $validator->get_refunded_adjustments() as $adjustment ) {
```

includes/orders/functions/refunds.php
```diff
@@ -307,264 +308,110 @@ function edd_refund_order( $order_id = 0, $order_items = array() ) {
...
+				/*
+				 * First check in our map, as it avoids DB queries.
+				 * We should hit this if the order item was refunded in this same transaction.
+				 */
+				if ( ! empty( $order_item_id_map[ $adjustment['object_id'] ] ) ) {
+					$adjustment['object_id'] = $order_item_id_map[ $adjustment['object_id'] ];
+					$should_convert   = false;
+				} else {
+					// Otherwise we'll have to do a DB query, which would pick up previously refunded items.
+					$refund_order_item = edd_get_order_item_by( 'parent', $adjustment['object_id'] );
```
robincornett, line 367: I left the comment about the refunded fee not showing properly on a refund and I think, as I read it, that this logic is where it's gotten snarled, because I had already done a partial refund on the related `order_item`. However, in the refund I was creating, I was not refunding the item at all, but I think this logic will have returned the order item anyway.

This may be a case of I was doing it wrong, but I managed it. Can this be accounted for?

includes/admin/payments/class-refund-items-table.php
```diff
@@ -191,59 +249,167 @@ private function format_currency( $order_item, $column_name ) {
...
 	 *
-	 * @return string Displays the order_item status
+	 * @return string
 	 */
-	public function column_status( $order_item ) {
-		return ! empty( $order_item->status )
-			? ucwords( $order_item->status )
+	public function column_status( $item ) {
+		return ! empty( $item->status )
+			? ucwords( $item->status )
```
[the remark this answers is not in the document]
robincornett: Actually I think this function can be removed as there is no status column.

includes/orders/functions/refunds.php
robincornett: Update: I refunded the order item fee, but not the order item, on an order which had no refunds applied, and the fee shows properly on the refunds screen.

includes/admin/payments/class-refund-items-table.php
```diff
@@ -191,59 +249,167 @@ private function format_currency( $order_item, $column_name ) {
...
-	public function column_cb( $order_item ) {
-		if ( 'refunded' !== $order_item->status ) {
+	public function column_cb( $item ) {
+		$item_id            = $this->get_item_unique_id( $item );
+		$object_type        = $this->get_object_type( $item );
+		$item_quantity      = $item instanceof Order_Item ? $item->quantity : 1;
+		$refundable_amounts = $item->get_refundable_amounts();
+		$total_remaining    = array_key_exists( 'total', $refundable_amounts ) ? floatval( $refundable_amounts['total'] ) : 0.00;
+
+		if ( 'refunded' !== $item->status && $total_remaining > 0 ) {
```
robincornett, line 361: This logic is working with credits, discounts, and positive fees, but fails with negative fees. I'm trying this with Discounts Pro and it's failing because the `$total_remaining` is negative.

assets/js/admin/orders/order-overview/_refund.js
```diff
@@ -59,125 +63,197 @@ $( document.body ).on( 'click', '.ui-widget-overlay', function ( e ) {
...
+		newTax        = 0,
+		newTotal      = 0,
+		canRefund     = false,
+		allInputBoxes = $( '#edd-refund-order-dialog .edd-order-item-refund-input' );
 
 	// Set a readonly while we recalculate, to avoid race conditions in the browser.
-	all_checkboxes.prop('readonly', true);
-	$('#edd-refund-submit-button-wrapper .spinner').css('visibility', 'visible');
+	allInputBoxes.prop( 'readonly', true );
+	$( '#edd-refund-submit-button-wrapper .spinner' ).css( 'visibility', 'visible' );
```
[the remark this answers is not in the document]
ashleyfae: Honestly the calculations should be so fast that I'm not sure we really need a spinner.

includes/orders/functions/refunds.php
ashleyfae: Just going to remove that `else` statement completely. Think that will sort it.

ashleyfae:
Great job on the UI @robincornett ! :raised_hands: 

The one thing I'm not really sure about is the right alignment of the amounts:

Screenshot from 2021-03-29 10-40-21

I don't like the large space between the currency symbol and the actual amount. I feel like they should be flush.

assets/js/admin/orders/order-overview/_refund.js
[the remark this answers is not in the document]
robincornett: They are going super fast for me; I would be fine with removing this line then.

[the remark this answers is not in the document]
ashleyfae: Removed!

robincornett:
On the gap, I guess I was envisioning it to end up feeling kind of "mathy" when the form input is active:

image

I can make it not right aligned...if I do that, do you think I should revert the border changes on the input? I'm wondering how the left alignment would feel with a currency positioned after the input. I could left-align the input text for currency before, and right-align for currency after?

I am still struggling with negative fees and credits--mainly I think I am wanting them to behave like discounts, and not have them be part of the refund UI at all. Because if I have an order placed with Discounts Pro and try to refund just the product and not the fee, I get the expected validation error:

image

I know it's expected because technically, it's all behaving as it ought to, but it _feels_ weird to me to have to refund a negative amount. I guess I am wondering if it is possible to just adjust the maximum refundable amount for the order/item when there is a negative fee or credit, and not show the fee/credit at all?

I think one confusing thing is that an order _credit_ is shown as a positive amount, even though it has the same net effect on the order/refund as a negative _fee_, but you see how the negative fee shows above and here is a credit:

image

Also I think it would help for this to be clearly labeled "Credit" in the row if it's going to show (which I am leaning towards it not showing, if that's possible, and treat credits the same way we treat discounts?).

Otherwise, I feel like the general flow of this is working well.

ashleyfae:
> I could left-align the input text for currency before, and right-align for currency after?

I like that idea. :+1: 

> I am still struggling with negative fees and credits--mainly I think I am wanting them to behave like discounts, and not have them be part of the refund UI at all. Because if I have an order placed with Discounts Pro and try to refund just the product and not the fee, I get the expected validation error:
> 
> image
> 
> I know it's expected because technically, it's all behaving as it ought to, but it _feels_ weird to me to have to refund a negative amount. I guess I am wondering if it is possible to just adjust the maximum refundable amount for the order/item when there is a negative fee or credit, and not show the fee/credit at all?

The reason we can easily exclude discounts is because each order item has a `discount` column in the order item record. The discounts have _already_ been split across all items in the order. That logic is handled for us. So when getting the refundable amount, we can just very easily do `subtotal - discount`. It's nowhere near this easy with fees, because fee amounts never get spread evenly across order items.

I don't love how it is right now either, but, frankly, I think it just has to suck. Or at least I don't think we should put in the massive effort to make it not suck for 3.0. Rather than putting in a ton of effort here, I think it would be nice to make negative fees actually turn into proper discounts somehow.

Plus, the number of people using Discounts Pro is rather small, so this isn't going to have a huge impact.

> I think one confusing thing is that an order _credit_ is shown as a positive amount, even though it has the same net effect on the order/refund as a negative _fee_, but you see how the negative fee shows above and here is a credit:

It's done that way because that's how it works when manually adding an order too. You add `10` credit -- not `-10`. Although it's weird, I think keeping it consistent might help.

> treat credits the same way we treat discounts?).

Same as above regarding negative fees. Honestly I just think it's too complicated, at least for now. We can always try to tackle it later, but for now let's keep it simpler and aim for "good enough".

assets/css/admin/orders/_refunds-modal.scss
```diff
@@ -0,0 +1,104 @@
...
+			box-shadow: none;
+			text-align: right;
+			width: 100%;
+
+			&:disabled {
+				border-bottom: none;
+			}
+
+			&:focus {
+				border-bottom: 2px solid var(--wp-admin-theme-color-darker-10);
```
spencerfinnell, line 70: To avoid a jump by adding an additional pixel to the calculated height a `box-shadow` could be used in addition to the `1px border`.

```suggestion
				border-bottom: 1px solid var(--wp-admin-theme-color-darker-10);
				box-shadow: 0 1px 0 var(--wp-admin-theme-color-darker-10);
```

spencerfinnell (commented):
This is working really well so far.

Any thoughts on removing some of the additional padding/borders and adding a separate "Submit" area to closer match the modals for creating the Order?

Screen Shot 2021-03-31 at 11 31 18 AM

Removing the alternate row colors on the totals might also help add some distinction to the submission area.

assets/css/admin/orders/_refunds-modal.scss
robincornett: Holy moly, Spencer, you are a wizard indeed!

robincornett:
I don't love the feel of removing the padding (I like more space than less), but what about ditching the table border? I would also like to not right align the submission button anymore:

image

spencerfinnell:
I prefer it with no borders, or all borders. Probably not worth getting hung up on at this point though. I'll keep testing for functionality.

tests/orders/tests-refunds.php
```diff
@@ -318,4 +233,72 @@ public function test_get_refund_date() {
...
+
+		$this->assertEquals( ( $order->subtotal - $order->discount ), $validator->subtotal );
+		$this->assertEquals( $order->tax, $validator->tax );
+		$this->assertEquals( $order->total, $validator->total );
+
+		$order_item_ids  = wp_list_pluck( $order->items, 'id' );
+		$refund_item_ids = wp_list_pluck( $validator->get_refunded_order_items(), 'parent' );
+
+		sort( $refund_item_ids );
+		sort( $refund_item_ids );
```
LisaCee, line 255: Is this duplicate line intentional?

robincornett: I think you're right--I'll update.

LisaCee (commented):
Odd question, but should `tests-process-dowload.php` have the spelling corrected?  I'm not sure if it was intentional or if updating the file name would have any unexpected consequences.

ashleyfae:
> Odd question, but should `tests-process-dowload.php` have the spelling corrected? I'm not sure if it was intentional or if updating the file name would have any unexpected consequences.

Updating it should be fine, but not in this PR. :+1:

robincornett:
Here is a question about partial refunds: what should a partial refund _mean_ in terms of access to download files, etc.? Should the download files still be available? Right now, they're not, and things are not completely consistent. Here's some things I noticed:

- In the order details screen (admin), the "Copy Download Links" button is missing for a partially refunded download.
- In the customer view of the receipt/payment confirmation, the download file is not available.
- In the purchase history view, a fully refunded order says "Refunded" and there is no link to view the details/downloads (should a user still be able to see the details of a refunded payment? It seems like they should?).
- In the purchase history view, a partially refunded order looks like any other order and shows the links to view details/downloads, but when clicking that link, it shows that the order status is refunded, but not the status of the individual items (something to consider updating in #8508)

I am just wondering what partial refunds should mean in terms of access to files...actually I'm going to create a separate issue for that as it's not technically part of what this PR is tackling, although it's definitely related.

LisaCee:
Leaving this here so I don't forget what I've tested:
- Entire order
- 1 item from multiple item order
- 1 item from multiple qty
- Taxes (with tax enabled and disabled)
- Change the dollar amount
- Go over maximum refund amount
- Use a discount code
- Use a discount code with quantities
- Order with fee, only refund fee
- Order with status of processing

All of these scenarios worked as I had expected.

robincornett:
Some notes on some changes I've made this week:

**Quantities**
I added `quantity` to the `get_refundable_amounts` function, which is working now but took a couple of tries. Mainly this was due to some inconsistencies when doing a partial refund on an order with quantities, and then wanting to do a second refund. With an item with a quantity of more than 1, the behavior is now:
- if 1 or more of the items were refunded, so that the refund quantity does not match the original quantity, then the quantity of items which can be refunded is adjusted. eg: if an order has 4 items, and 3 of them are refunded, and you want to initiate another refund, the quantity is adjusted to 1, to be consistent with the subtotal, etc.
- if all items received an equal partial refund, the quantity is not adjusted.
- Where it gets a little wacky is if you have 4 items in a cart, but then do a partial refund on 3 of them, and want to do another refund. The quantity is adjusted to 1, but the subtotal reflects the amount remaining on the entire order. You can process a refund for the amount remaining, even though it's more than the item price, so it's still possible to do a full refund, and the entire order and the items in it are marked as refunded. So it works but it feels a little weird--not sure if it's something we can fix, as the other values in the database are variable (for example, the subtotal is different for a refund than the original order if a discount is involved, because the discount is included in the refund subtotal). I feel like the quantities should be adjusted and that this last is kind of an edge case, but want to explain my thinking here in case I should be overruled.

**UI**
I adjusted the styling somewhat to make the non-name columns narrower (100px with padding included). If a storefront has all of the columns, this displays a bit awkwardly on smaller screens, before the media query kicks in, but it's an improvement. I don't want to adjust it at this point, but I would suggest that we consider not using the primary column functionality from WordPress, but that would mean dealing with the responsive CSS on our own/completely from scratch. This isn't perfect, but since WP Core is doing the heavy lifting for us, it's totally livable, I think.

**Testing**
- [x] Full refund
- [x] Full refund with discounts
- [x] Full refund with fees
- [x] Full refund with credits
- [x] Full refund with a mix
- [x] Full refund with taxes
- [x] Partial refund
- [x] Partial refund after a partial refund
- [x] Partial refund with taxes
- [x] Partial refund with quantities after a partial refund
- [x] Partial refund with taxes but no tax rate object (VAT): refund doesn't completely work but filed as a separate issue #8546 ... the issue is not with the refund process, but the order meta not getting copied over
- [x] Refunding just one item completely in an order with multiple items
- [x] Partially refunding all items of a quantity (basically 10% off of each)

Basically I'm ready to approve this (as long as the quantity adjustment change I made is acceptable, @ashleyfae), and I'm at the point where I'd like to have it merged so that any further bugs/adjustments can be dealt with individually.

robincornett (approved):

ashleyfae:
The reason I didn't include quantity in the refundable amounts is because it can be fundamentally inaccurate with partial refunds on single order items.

For example:

Say you have an order with 3 quantities of an item.

Screenshot from 2021-04-12 10-55-14

You then want to refund $10 of a single quantity (so $10 refund on one of the 3). This will get recorded as a `-1` quantity, but you didn't actually refund an entire item. This then throws all future quantities out of whack. For example, if I do exactly that and do a refund on `1` for `$10`, then try to do that a second time, it looks like this:

Screenshot from 2021-04-12 10-55-04

The quantity is set to `1`, but the subtotal is now calculated incorrectly. It's set to `$73.50` when it should be `$49`.

But let's say I fix that and do a second refund exactly the same as before (quantity = `1`, subtotal = `$10`). Now, the maximum refundable quantity is set to `1` because it sees that I've already refunded a total of `2`. So the maximum on the quantity is incorrectly set to `1` when technically I haven't even refunded the entirety of 1 item.

Although we could probably fix the above calculation issue, I just chose to omit it due to the entire inaccuracy of it all when it comes to partial refunds on single items.

robincornett:
@ashleyfae I feel like the partial refunds/quantities issues is resolvable--I have some thoughts--but am going to revert the calculations I added and put it on a personal list to ponder/tackle separately, please.

ashleyfae:
Going to go ahead and get this merged. If a bug is discovered, please open a new issue/PR. :+1:

PR #8287, opened 2020-12-29
Fix convert_object_to_array() method

Kubitomakita:
This prevents from "PHP Warning: Invalid argument supplied for foreach() in .../EDD_SL_Plugin_Updater.php on line 356"

Probably will be a better idea to decode JSON directly to the array than converting it here

### DIFF
--- a/includes/EDD_SL_Plugin_Updater.php
+++ b/includes/EDD_SL_Plugin_Updater.php
@@ -353,7 +353,7 @@ class EDD_SL_Plugin_Updater {
 	 */
 	private function convert_object_to_array( $data ) {
 		$new_data = array();
-		foreach ( $data as $key => $value ) {
+		foreach ( (array) $data as $key => $value ) {
 			$new_data[ $key ] = is_object( $value ) ? $this->convert_object_to_array( $value ) : $value;
 		}
 


ashleyfae:
Thanks @Kubitomakita !

@robincornett Something we need to get into Software Licensing?

robincornett:
I think this is good to look into for Software Licensing, which will make its way back to EDD core. I'm a little confused about casting the passed in value to an array, as most of the checks in the class only use the method if the value is _not_ an array ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/EDD_SL_Plugin_Updater.php#L176)). I'm guessing that this warning is being thrown if the value is **set** but not an object (I would be interested to know what `$data` is when this warning occurs--if it's an empty value, rather than an object, then we should modify the logic differently, I think.

Looking at the method, `$data` is expected to be an object.

ashleyfae:
@Kubitomakita Do you know exactly what `$data` is where it's throwing this error? I wonder if it's something other than an object. (Because if it is an object, that shouldn't be throwing the error, as we should be able to iterate the public properties with a foreach)

Kubitomakita:
@ashleyfae unfortunately I don't know what's the exact data type being passed to this method. This issue was brought to my attention by one of my own customers who is using a plugin licensed with EDD.

My thought was that with strict types it may throw this warning, but [it's working anyway](http://sandbox.onlinephpfunctions.com/code/8233c9d9c25a94d938bc1f3867f43c5648cfdf96).

mihaijoldis:
A new report with the same error message https://secure.helpscout.net/conversation/1413115822/176484?folderId=180499
Seems to be visible on the WordFece page

ashleyfae:
Closing this in favour of a new issue in the Software Licensing repo: https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1748

Once we get that class updated, the changes will eventually make their way into EDD core as well.

Issue #7762, opened 2020-04-20
3.0 - Physical Address Table Issues


robincornett:
### Expected behavior/Actual Behavior

- on the address list table, one of the action links is "Delete". If I click that, I'm taken to the Customer edit screen where I have the option to delete the customer entirely, not the address.
- clicking the "View" action link takes me to view the customer record, not just the address
- in the address list table, if I click the linked customer name in the row, I would maybe expect to go to the customer screen, but I get a filtered view of all addresses related to that customer. If I'm in the filtered view, clicking the customer name just reloads the same table view, which is confusing.

### Information (if a specific version is affected):
PHP Version: 7.3.5
EDD Version (or branch): release/3.0
WordPress Version: 5.4

robincornett:
> My initial thought would be to adjust the link in the "Customer" column while viewing the "Email Addresses" or "Physical Addresses" to just link directly to the Customer record. This presents that  Customer's email and physical addresses all in one place. However, the tables in the Customer record would maybe benefit from having "Delete" links there as well.
> 
> _Originally posted by @spencerfinnell in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7771#issuecomment-617305919_

Quoting here because I agree that the links in the rows are confusing and I think this belongs to this issue. One thought I had was to:
- change the customer name link to go to the individual customer record
- change the "action" links underneath to "View all customer addresses" (or something similar, and have this be the search/filter link)
- maybe remove the "delete" link for the moment because all it does is take you to the screen to delete the entire customer record, not just the email address.

PR #8295, opened 2021-01-07
Remove unnecessary meta query & fix email search #8272

### LINKED ISSUE #8272: 3.0 - Searching API logs doesn't work due to legacy meta query being built
### Expected behavior

I expect to be able to search through the API logs. Looking at the code, it looks like the following should be supported:

- Searching for an IP address
- Searching for an email address (matches to a user)
- Searching for an API key (if the search is 32 characters)
- Searching for a token, if the search is prefixed with `token:`
- Searching by user ID number

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/class-api-requests-logs-list-table.php#L121-L166

### Actual behavior

No results are returned. That's because it's actually building up a _meta_ query, which would have been used in 2.9, but is no longer required in 3.0. Searching for an IP results in a query like this:

```mysql
ELECT edd_la.id
FROM wp_edd_logs_api_requests edd_la
INNER JOIN wp_edd_logmeta
ON ( edd_la.id = wp_edd_logmeta.edd_log_id )
WHERE edd_la.ip = '10.0.0.1'
AND ( ( wp_edd_logmeta.meta_key = '_edd_log_request_ip'
AND wp_edd_logmeta.meta_value = '10.0.0.1' ) )
ORDER BY edd_la.id DESC
LIMIT 30
```

All this information (IPs, user_id, and token) is now stored in the table directly and does not need a meta query.

1) Go to Downloads > Tools > Logs and change the dropdown to "API Requests".
1) Make sure you have at least one API log in the table.
2) Search for the IP listed under `Request IP`, so you should definitely get results.
3) You get no results.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8272 

Proposed Changes:
1. Remove entire `get_meta_query()` method, as it's no longer needed.
2. Fix email search to get customer by the email address.
3. When a customer is found via search, also set `user_id` parameter. Some log tables use `customer_id` and others use `user_id` so this ensures compatibility with both.

To test:

Follow original testing steps here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8272

You will need at least one API request. Then try a variety of searches, including:

- Use date parameters.
- Enter an IP address in search.
- Enter an email in search. (This should match the account that performed the API request.)
- Enter a user ID number in search. (This should match the account that performed the API request.)
- Also try some invalid search parameters to make sure you correctly get no results.

### DIFF
--- a/includes/admin/reporting/class-base-logs-list-table.php
+++ b/includes/admin/reporting/class-base-logs-list-table.php
@@ -358,12 +358,18 @@ class EDD_Base_Log_List_Table extends List_Table {
 					$user = get_user_by( 'email', $search );
 					if ( ! empty( $user->ID ) ) {
 						$retval['user_id'] = $user->ID;
+					} else {
+						// This is a fallback to help ensure an invalid email will produce zero results.
+						$retval['search'] = $search;
 					}
 				} else {
 					// All other logs are linked to customers.
 					$customer = edd_get_customer_by( 'email', $search );
 					if ( ! empty( $customer->id ) ) {
 						$retval['customer_id'] = $customer->id;
+					} else {
+						// This is a fallback to help ensure an invalid email will produce zero results.
+						$retval['search'] = $search;
 					}
 				}
 			} elseif ( 'api_requests' === $this->log_type && 32 === strlen( $search ) ) {
@@ -378,6 +384,8 @@ class EDD_Base_Log_List_Table extends List_Table {
 					$user = get_user_by( 'email', $search );
 					if ( ! empty( $user->ID ) ) {
 						$retval['user_id'] = $user->ID;
+					} else {
+						$retval['search'] = $search;
 					}
 				} else {
 					// All other logs are linked to customers.
@@ -386,6 +394,8 @@ class EDD_Base_Log_List_Table extends List_Table {
 						$retval['customer_id'] = $customer->id;
 					} elseif ( 'file_downloads' === $this->log_type ) {
 						$retval['product_id'] = $search;
+					} else {
+						$retval['search'] = $search;
 					}
 				}
 			} else {
--- a/includes/admin/reporting/class-base-logs-list-table.php
+++ b/includes/admin/reporting/class-base-logs-list-table.php
@@ -353,14 +353,18 @@ class EDD_Base_Log_List_Table extends List_Table {
 			if ( filter_var( $search, FILTER_VALIDATE_IP ) ) {
 				$retval['ip'] = $search;
 			} elseif ( is_email( $search ) ) {
-				$customer = edd_get_customer_by( 'email', $search );
-				if ( ! empty( $customer->id ) ) {
-					/*
-					 * File logs users customer_id but API logs use user_id, so we're setting
-					 * both to be compatible with both.
-					 */
-					$retval['customer_id'] = $customer->id;
-					$retval['user_id']     = $customer->user_id;
+				if ( 'api_requests' === $this->log_type ) {
+					// API requests are linked to user accounts, so we're checking user data here.
+					$user = get_user_by( 'email', $search );
+					if ( ! empty( $user->ID ) ) {
+						$retval['user_id'] = $user->ID;
+					}
+				} else {
+					// All other logs are linked to customers.
+					$customer = edd_get_customer_by( 'email', $search );
+					if ( ! empty( $customer->id ) ) {
+						$retval['customer_id'] = $customer->id;
+					}
 				}
 			} elseif ( 'api_requests' === $this->log_type && 32 === strlen( $search ) ) {
 				// Look for an API key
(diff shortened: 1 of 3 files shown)


robincornett (commented):
Currently, an invalid search returns all logs (or all logs matching the applied filter).

Otherwise, searching by:
- [x] customer ID works
- [x] email address works
- [x] IP address works

Still working on things but noting some inconsistencies:
- if no logs exist for a filter (or presumably a search?), the entire filter bar is not loaded, so the only way to try filtering again is to click on the navigation tab/link
- in the table, it's hard to tell whether a link is going to run a filter (clicking on order, customer name--File Downloads, download name, email address) or take me to a new screen (IP address, customer name--Sales)

Those are maybe separate issues, and the links thing is similar to the issues noted in #7762.

ashleyfae:
> * if no logs exist for a filter (or presumably a search?), the entire filter bar is not loaded, so the only way to try filtering again is to click on the navigation tab/link

I'm not sure I'm seeing this behaviour. If I apply filters that result in no results then the _search_ bar doesn't appear, but the rest of the filter bar does. Am I misunderstanding this? Maybe you can take a screenshot?



> * in the table, it's hard to tell whether a link is going to run a filter (clicking on order, customer name--File Downloads, download name, email address) or take me to a new screen (IP address, customer name--Sales)

Let's handle that separately for UX things. The main goal for this issue is to ensure you get the functionality you expect.

robincornett:
You are correct--that behavior actually just happens if I switch to the "License Renewal Notices" filter--so that is a third party log, and it's currently not returning results, and, I think, is unable to do so because of SL issue [1719](https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1719) (the SL logs currently are not migrated to the new database tables).

It looks like there is backwards compatibility code in place which should allow these old logs to show, but they're not. Digging to see if that is an issue in SL or EDD core, or possibly something which happened during the migration. I see that the query for the license logs in SL is returning no results unless I remove the tax query parameter, so that makes me wonder about the migration.

LisaCee:
> Currently, an invalid search returns all logs (or all logs matching the applied filter).

I'm getting weird results with this, too.  I would expect if there is not a match I would get a message that no items are found.  I am getting this occasionally (still figuring out when/why), but more often then not I'm getting all of the results which feels misleading.  I can delete part of an API key and get the same results as when I enter it correctly which doesn't give the user any indication of typos/error handling.

ashleyfae:
@LisaCee and @robincornett Updated the code a bunch. Also changed stuff for the File Download logs so searching there should also be checked.

For file download searches you should be able to:

- Search by IP address
- Search by customer email
- Search by customer ID number
- If you search by an integer that isn't a valid customer ID, then it should check for a product ID with that number.
- Search by _file name_ (see "File" column). This logic is not perfect so paginated results will be wonky.
- Click a download name to filter by that download.
- Click a customer name to filter by that customer.

robincornett (commented):
Sorry, it took me a while to generate some API requests to test.

Everything is working nicely for the file download logs. No results show when I input something invalid, and all of the searches seem to be working.

For the API requests, if I put in something which should return invalid data/no results, I am still getting all results.

However, if I search for an email address which I know should work (API requests), I get no results (and it shows no results, so a different behavior than when I am intentionally searching for something wrong).

ashleyfae:
> Sorry, it took me a while to generate some API requests to test.
> 
> Everything is working nicely for the file download logs. No results show when I input something invalid, and all of the searches seem to be working.
> 
> For the API requests, if I put in something which should return invalid data/no results, I am still getting all results.
> 
> However, if I search for an email address which I know should work (API requests), I get no results (and it shows no results, so a different behavior than when I am intentionally searching for something wrong).

I'm unable to reproduce either of those.

> For the API requests, if I put in something which should return invalid data/no results, I am still getting all results.

What exactly did you search for? For example: was it an invalid email address, or was it an integer, or random text? There are many different search conditions so I want to make sure I'm reproducing in the right way.

> However, if I search for an email address which I know should work (API requests), I get no results 

I don't know if it'll fix the problem, but I've just pushed up an email-related change. Give it another try. :+1:

includes/admin/reporting/class-base-logs-list-table.php
```diff
@@ -346,20 +353,47 @@ protected function get_query_args( $paginate = true ) {
 			if ( filter_var( $search, FILTER_VALIDATE_IP ) ) {
 				$retval['ip'] = $search;
 			} elseif ( is_email( $search ) ) {
-				$customer = edd_get_customer( $search );
-				if ( ! empty( $customer->id ) ) {
-					$retval['customer_id'] = $customer->id;
+				if ( 'api_requests' === $this->log_type ) {
+					// API requests are linked to user accounts, so we're checking user data here.
+					$user = get_user_by( 'email', $search );
+					if ( ! empty( $user->ID ) ) {
```
robincornett, line 359: I think that this and line 365 are exhibiting the same behavior for me, and returning all results on any log search when I input an invalid email address--because the user/customer ID is not added as a parameter to the `$retval` array, it's not factored in to the search results. This feels weird to me because if I enter an invalid email address, I would think I should get no results, but it looks like this behavior is the same in 2.x as well, so it's not a regression, just something I don't like?

The updated email search is working in the API Request logs now, as long as I put in a valid email address.

ashleyfae: Okay just pushed up a change that I believe should fix that.

This function is officially a janky piece of mess and may be worth refactoring at some point. But not now. :stuck_out_tongue:

robincornett (approved):

PR #8300, opened 2021-01-12
Use helper functions to add log metadata

### LINKED ISSUE #8299: 3.0 - undefined method add_meta in EDD_Logging
### Expected behavior
When I do something which will add a log to the database, I expect it to work.

### Actual behavior
Using EDD Message to send an email to a customer, I get the following fatal error:

```
Fatal error: Uncaught Error: Call to undefined method EDD\Logs\Log::add_meta()
in \easy-digital-downloads\includes\class-edd-logging.php on line 228
```

1. Install EDD Message on a site running EDD3.
2. Navigate to a customer, then Messages (for that customer).
3. Send a message to the customer.

PHP Version: 7.4.1
WordPress Version: 5.6

Although I encountered this issue when using EDD Message specifically, I think the issue is with EDD core, as there is no add_meta function defined in the logging class.

Additionally, at this point of the code, `$log` is a database row object, not an instance of a class. Looking at `EDD_Payment` and `EDD_Customer`, both of those call a helper function, which then uses `add_metadata` directly.

I believe that the `update_log` method in this class will have the same issue/error.

robincornett:
Fixes #8299

Proposed Changes:
1. Use the `edd_add_log_meta` (and related file download/api request functions) to add metadata to the log.
2. Use `edd_update_log_meta` (and related file download/api request functions) to update metadata in `update_log`.

### DIFF
--- a/includes/class-edd-logging.php
+++ b/includes/class-edd-logging.php
@@ -270,8 +270,8 @@ class EDD_Logging {
 		}
 
 		// Used to dynamically dispatch the method call to insert() to the correct class.
-		$update_method = 'edd_update_log';
-		$meta_update   = 'edd_update_log_meta';
+		$update_method        = 'edd_update_log';
+		$update_meta_function = 'edd_update_log_meta';
 
 		$type = $args['log_type'];
 		if ( ! empty( $type ) ) {
@@ -286,8 +286,8 @@ class EDD_Logging {
 		);
 
 		if ( 'api_request' === $data['type'] ) {
-			$meta_update = 'edd_update_api_request_log_meta';
-			$legacy      = array(
+			$update_meta_function = 'edd_update_api_request_log_meta';
+			$legacy               = array(
 				'user'         => 'user_id',
 				'key'          => 'api_key',
 				'token'        => 'token',
@@ -306,8 +306,8 @@ class EDD_Logging {
 				}
 			}
 		} elseif ( 'file_download' === $data['type'] ) {
-			$meta_update = 'edd_update_file_download_log_meta';
-			$legacy      = array(
+			$update_meta_function = 'edd_update_file_download_log_meta';
+			$legacy               = array(
 				'file_id'    => 'file_id',
 				'payment_id' => 'payment_id',
 				'price_id'   => 'price_id',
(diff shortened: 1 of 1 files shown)


LisaCee (approved):
Duplicated original error, working here

includes/class-edd-logging.php
```diff
@@ -217,15 +217,15 @@ public function insert_log( $log_data = array(), $log_meta = array() ) {
...
-				$log = edd_get_file_download_log( $log_id );
+				$add_meta = 'edd_add_file_download_log_meta';
 			} else {
-				$log = edd_get_log( $log_id );
+				$add_meta = 'edd_add_log_meta';
 			}
 
 			foreach ( (array) $log_meta as $key => $meta ) {
-				$log->add_meta( sanitize_key( $key ), $meta );
+				$add_meta( $log_id, sanitize_key( $key ), $meta );
```
ashleyfae, line 228: While there's no real risk of this causing any error as-is, just to be super safe.. Can you wrap the whole `foreach` statement in:

```php
if ( is_callable( $add_meta ) ) {
    // foreach in here
}
```

Also, for clarity, maybe rename `$add_meta` to `$add_meta_function`. `$add_meta` feels like it would contain a boolean ("yes, I want to add meta").

includes/class-edd-logging.php
```diff
@@ -334,10 +337,8 @@ public function update_log( $log_data = array(), $log_meta = array(), $log_id =
 
 		// Set log meta, if any
 		if ( 'edd_update_log' === $update_method && ! empty( $log_meta ) ) {
-			$log = edd_get_log( $log_id );
-
 			foreach ( (array) $log_meta as $key => $meta ) {
-				$log->update_meta( sanitize_key( $key ), $meta );
+				$meta_update( $log_id, sanitize_key( $key ), $meta );
```
ashleyfae, line 341: Same as above. 

1. Wrap foreach in `is_callable()`
2. Rename variable to `$update_meta_function`

ashleyfae (changes requested):

ashleyfae (commented):
Code looks good! Just need to do some testing.

ashleyfae (approved):

PR #8302, opened 2021-01-12
#8301 PHP version requirement

### LINKED ISSUE #8301: 3.0 - Add minimum PHP version to plugin header
We should add our minimum PHP version to the main plugin header.

### Justification or use case
It will prevent users who do not have the minimum PHP version from being able to update their copy of the plugin to a version which may break their site (unless they update manually).

We just need to add this to the plugin header information:

```php
Requires PHP: 5.6
```

This is in addition to, but separate from, #6514 and #6715. Adding as an issue after meeting with @cklosowski.

LisaCee:
Fixes #8301 

Proposed Changes:
1. Add PHP version requirement

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         "issues": "https://easydigitaldownloads.com/support"
     },
     "require": {
-        "php": ">=5.3.0",
+        "php": ">=5.6",
         "composer/installers": "~1.0"
     },
     "require-dev": {
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -64,7 +64,7 @@ final class EDD_Requirements_Check {
 
 		// PHP
 		'php' => array(
-			'minimum' => '5.3.0',
+			'minimum' => '5.6',
 			'name'    => 'PHP',
 			'exists'  => true,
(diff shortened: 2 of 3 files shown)


robincornett (approved):

ashleyfae:
@LisaCee Can you also add the same 5.6 requirement here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/97b73a3fc400d00045f8b7282c28d16ceb93818b/easy-digital-downloads.php#L67

and here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/97b73a3fc400d00045f8b7282c28d16ceb93818b/composer.json#L12

robincornett (changes requested):
Sorry, it also needs to be updated in the [readme.txt file](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/readme.txt#L9) too please! (In hindsight I think I erred when creating the issue--I think this is supposed to be in the readme file as opposed to the plugin header, but I'm okay with having it both places personally).

robincornett (approved):

ashleyfae (approved):

PR #8303, opened 2021-01-13
Add a dismissible admin notice for outdated PHP versions

### LINKED ISSUE #6514: Add notice for sites on PHP 5.3 and 5.2 about upcoming EOL for support
## Enhancement Request
We need to add a notice to customers running PHP 5.2 and 5.3 before EDD 3.0 releases that contains information about running PHP 5.2 and 5.3. This goes along with #6290.

### Justification or use case
Since EDD 3.0 will drop support for PHP 5.2 and 5.3 in #6290 we need to let people know in advance of this so they can start the conversations with their hosting provider, server admin, or find a new host.

This should coincide with an official blog post on the main EDD blog about the upcoming changes so the user can get more information as to why the changes are coming.

robincornett:
Fixes #6514

Proposed Changes:
1. For users running a PHP version less than 5.6, display an admin notice which can be dismissed, advising them that they need to update their PHP before EDD3.
2. The notice links to the official WordPress page for updating PHP, and an unpublished recommended hosting page.

To Test:
Probably easiest to open `includes/admin/class-edd-notices.php` and change the `<` to `>` so the notice will display. We need to make sure the language is clear and that the error can be dismissed properly.


includes/admin/class-edd-notices.php
```diff
@@ -103,6 +103,74 @@ public function show_notices() {
...
+					'<a href="https://easydigitaldownloads.com/recommended-wordpress-hosting/" target="_blank" rel="noopener noreferrer">',
+					'</a>',
+					'</p>'
+				)
+			);
+			echo wp_kses_post(
+				sprintf(
+					/* Translators: %1$s - Opening anchor tag, %2$s - The url to dismiss the ajax notice, %3$s - Complete the opening of the anchor tag, %4$s - Open span tag, %4$s - Close span tag */
+					__( '%1$s %2$s %3$s %4$s Dismiss this notice. %5$s', 'easy-digital-downloads' ),
+					'<a href="',
```
LisaCee, line 156: Do we want this to have a text-decoration?  Right now it is inheriting an underline.

includes/admin/class-edd-notices.php
```diff
@@ -103,6 +103,74 @@ public function show_notices() {
...
+				sprintf(
+					/* translators:
+					%1$s Opening paragraph tag, do not translate.
+					%2$s Opening anchor tag, do not translate.
+					%3$s Closing anchor tag, do not translate.
+					%4$s Closing paragraph tag, do not translate.
+					*/
+					__( '%1$sMany web hosts can give you instructions on how/where to upgrade your version of PHP through their control panel, or may even be able to do it for you. If you need to change hosts, please see %2$sour hosting recommendations%3$s.', 'easy-digital-downloads' ),
+					'<p>',
+					'<a href="https://easydigitaldownloads.com/recommended-wordpress-hosting/" target="_blank" rel="noopener noreferrer">',
```
LisaCee, line 147: This link is leading to a Page not Found.  After doing some digging on our site, I have yet to find the correct link.

LisaCee (changes requested):
This looks really good.  Just a few things:

includes/admin/class-edd-notices.php
robincornett: It's not a live link yet, but my instructions were to go ahead and use it because it will be.

includes/admin/class-edd-notices.php
robincornett: Let me add some CSS to remove that--it's because it's a link, and generally those are buttons--but I borrowed the markup from an existing notice we have to get the dismiss functionality and it uses a link.

includes/admin/class-edd-notices.php
```diff
@@ -103,6 +103,74 @@ public function show_notices() {
 
 		}
 
+		/**
+		 * Notice for users running PHP < 5.6.
+		 * @since 2.10
+		 */
+		if ( version_compare( PHP_VERSION, '5.6', '<' ) && ! get_user_meta( get_current_user_id(), '_edd_upgrade_php_dismissed', true ) ) {
```
spencerfinnell, line 110: @robincornett This should likely have the same capability check as the other server-related notices: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/77c88c0e43fb4c74ec2320dd73bd068d529080e1/includes/admin/class-edd-notices.php#L62

Showing this to Subscribers, Authors, etc wouldn't be relevant.

LisaCee (approved):
Looks good to me.
- [x] Notice dismiss-able
- [x] Dismissal honored on refresh
- [x] Only admin user sees notice
- [x] No link underline on button

includes/admin/class-edd-notices.php
```diff
@@ -103,6 +103,74 @@ public function show_notices() {
...
 
+		/**
+		 * Notice for users running PHP < 5.6.
+		 * @since 2.10
+		 */
+		if ( version_compare( PHP_VERSION, '5.6', '<' ) && ! get_user_meta( get_current_user_id(), '_edd_upgrade_php_dismissed', true ) && current_user_can( 'manage_shop_settings' ) ) {
+			echo '<div class="notice notice-warning is-dismissible edd-notice">';
+			printf(
+				'<h2>%s</h2>',
+				esc_html__( 'Get Ready for Easy Digital Downloads 3.0', 'easy-digital-downloads' )
```
ashleyfae, line 114: I wonder if this is clear enough that "you need to do something". My first impression leaned more towards "exciting update is coming! Read about the features" as opposed to "you have to take action".

robincornett: We could change it to something like: "Easy Digital Downloads 3.0: Your Site Is Not Ready" ... "Don't Miss Out On Easy Digital Downloads 3.0" ... ?

ashleyfae: I was thinking something like _"Upgrades Required for Easy Digital Downloads 3.0"_ but then I wondered if that sounds like you've been upgraded to 3.0 and it wants to do a DB upgrade. :thinking:

spencerfinnell: It's tough because something like "Action Required" (or "Upgrades Required") isn't _technically_ true since it doesn't need to be done immediately... but I think it should learn towards being too aggressive vs. too passive.

"Action Required" is a pretty frequent subject line in my mailbox:

> Action required: Changes to the Events API that affect your team's Slack apps

> [Action Required] Automated Certificate Management and compatibility with Android versions prior to v.7.1.1

> Action Required: Migrate to the new SmartThings app

> Action Required: Confirm Your Email Now

So maybe something like:

**Action Required: Upgrade PHP to Prepare for Easy Digital Downloads 3.0**

---

It would be interesting to show an additional notice when the action has actually been completed. 

1. See action require/upgrade required notice
2. Perform action off-site
3. Return to site and see new notice confirming no additional steps are needed

ashleyfae: > Upgrade PHP to Prepare for Easy Digital Downloads 3.0

I like that.

> It would be interesting to show an additional notice when the action has actually been completed.

I'm not necessarily opposed to that, but do question if it's worth the extra effort. I also wonder if that's getting too notice happy. :shrug:

spencerfinnell: > I'm not necessarily opposed to that, but do question if it's worth the extra effort. I also wonder if that's getting too notice happy. 🤷

Worth questioning for sure -- and you may be right. I just like the idea of having confirmation since it's not something we can fully walk them through. But it might not be worth it for the few lingering installs. People should have been getting upgrade notices for years at this point.

ashleyfae: Yeah given that it's the minority of users, it might actually be easier/faster to just handle those in support for anyone who is unsure or wants confirmation. A support person can tell immediately from the system info file if they're good to go.

includes/admin/class-edd-notices.php
```diff
@@ -103,6 +103,74 @@ public function show_notices() {
 
 		}
 
+		/**
+		 * Notice for users running PHP < 5.6.
+		 * @since 2.10
+		 */
+		if ( version_compare( PHP_VERSION, '5.6', '<' ) && ! get_user_meta( get_current_user_id(), '_edd_upgrade_php_dismissed', true ) && current_user_can( 'manage_shop_settings' ) ) {
```
spencerfinnell, line 110:
```suggestion
		if ( version_compare( PHP_VERSION, '5.6', '<' ) && ! get_user_meta( get_current_user_id(), '_edd_upgrade_php_56_dismissed', true ) && current_user_can( 'manage_shop_settings' ) ) {
```

Adding the version number now will let us keep the IDs consistent when we do it again for 7.0

includes/admin/class-edd-notices.php
```diff
@@ -103,6 +103,74 @@ public function show_notices() {
...
+			echo wp_kses_post(
+				sprintf(
+					/* Translators: %1$s - Opening anchor tag, %2$s - The url to dismiss the ajax notice, %3$s - Complete the opening of the anchor tag, %4$s - Open span tag, %4$s - Close span tag */
+					__( '%1$s %2$s %3$s %4$s Dismiss this notice. %5$s', 'easy-digital-downloads' ),
+					'<a href="',
+					esc_url(
+						add_query_arg(
+							array(
+								'edd_action' => 'dismiss_notices',
+								'edd_notice' => 'upgrade_php',
```
spencerfinnell, line 161:
```suggestion
								'edd_notice' => 'upgrade_php_56',
```

spencerfinnell (changes requested):

spencerfinnell (approved):

ashleyfae (approved):

PR #8305, opened 2021-01-14
Move float performed on cart values during migration

### LINKED ISSUE #8304: 3.0 - non-numeric value encountered during migration
### Expected behavior
I should be able to migrate a storefront's orders without errors.

### Actual behavior
Running a migration on a fairly old store, I'm getting PHP warnings like this:

```
PHP Warning:  A non-numeric value encountered in \easy-digital-downloads\includes\admin\upgrades\v3\class-data-migrator.php on line 428
```

I'm not sure what specific orders had non-numeric data, but the affected code appears to be running in an edge case scenario when an order doesn't have cart details established:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/35bef8e3bfa1a53942e3de47bcbab9b0e240b2f8/includes/admin/upgrades/v3/class-data-migrator.php#L420-L431

My thought is that the `float` should not be running on the `isset` check, but instead on the value in the if part of the condition.

PHP Version: 7.4.1
WordPress Version: 5.6

robincornett:
Fixes #8304 

Proposed Changes:
When looping through `$cart_details` during a migration, use `float` when a value is set, not on the `isset` check for the value.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -422,10 +422,10 @@ class Data_Migrator {
 
 			// Loop through the items in the purchase to build the totals.
 			foreach ( $cart_details as $cart_item ) {
-				$subtotal += (float) isset( $cart_item['subtotal'] ) ? $cart_item['subtotal'] : 0;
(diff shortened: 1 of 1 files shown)


LisaCee (approved):
I'm not sure how to test this, but your logic makes perfect sense.

ashleyfae (approved):

PR #8308, opened 2021-01-15
Fix payment error log table output

### LINKED ISSUE #8307: 3.0 - Payment error logs are presented incorrectly
### Expected behavior
I expect the payment error logs to be displayed correctly.

### Actual behavior
In 3.0, the payment error logs table uses post functions such as `get_the_title` and `get_post_field( 'post_content'... )` to display the log title (error) and content, and since the logs are no longer post objects, WordPress is returning unlikely data for that (sample posts on my local site).

1. On 3.0, make sure there are some payment errors logged.
2. Visit Tools > Logs and select Payment Errors as the filter.

PHP Version: 7.4.1
WordPress Version: 5.6

robincornett:
Fixes #8307

Proposed Changes:
1. Use the correct log data for the payment errors table.

### DIFF
--- a/includes/admin/reporting/class-gateway-error-logs-list-table.php
+++ b/includes/admin/reporting/class-gateway-error-logs-list-table.php
@@ -55,6 +55,8 @@ class EDD_Gateway_Error_Log_Table extends EDD_Base_Log_List_Table {
 		switch ( $column_name ) {
 			case 'error' :
 				return $item['error'];
+			case 'payment_id' :
+				return ! empty( $item['payment_id'] ) ? $item['payment_id'] : '&ndash;';
 			default:
 				return $item[ $column_name ];
 		}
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/class-gateway-error-logs-list-table.php
```diff
@@ -54,7 +54,7 @@ protected function get_primary_column_name() {
 	public function column_default( $item, $column_name ) {
 		switch ( $column_name ) {
 			case 'error' :
-				return get_the_title( $item['ID'] ) ? get_the_title( $item['ID'] ) : __( 'Payment Error', 'easy-digital-downloads' );
+				return $item['error'];
```
ashleyfae, line 57: While we're here, can you add this:

```suggestion
				return $item['error'];
			case 'payment_id' :
				return ! empty( $item['payment_id'] ) ? $item['payment_id'] : '&ndash;';
```

I think that looks better when you have a gateway error that doesn't have an associated payment ID. 

Before:

Screenshot from 2021-01-19 13-33-28

After:

Screenshot from 2021-01-19 13-33-16

ashleyfae (approved):

PR #8310, opened 2021-01-18
Fix payment address setup

### LINKED ISSUE #8309: 3.0 - Payment address is set up too early
### Expected behavior
When a subscription is renewed, if an address is associated with the customer, I expect the renewal order to have an address.

### Actual behavior
With Recurring, the address is not saved, per @ashleyfae comment here:

> Perhaps the cause for what you described: Note in `EDD_Payment::setup_payment` :
> 
> ```
> $this->address         = $this->setup_address();
> $this->discounts       = $this->setup_discounts();
> $this->user_info       = $this->setup_user_info();
> ```
> 
> `address` is set up first, but that utilizes `user_info`, which hasn't been set up yet. Maybe we just need to move the address set up below the user info set up.

Per the [discussion in Recurring](https://github.com/easydigitaldownloads/edd-recurring/issues/1375), set up a subscription with Stripe and trigger a renewal invoice.

PHP Version: 7.4.1
WordPress Version: 5.6

robincornett:
Fixes #8309

Proposed Changes:
1. Sets up the payment/order address after the `user_info` has been set up (since it relies on `user_info`)
2. Update usage of `edd_add_order_address` and `edd_update_order_address` to use the correct `name` parameter instead of `first_name` and `last_name`, which are incorrect.
3. Update related doc blocks.

### DIFF
--- a/includes/admin/payments/actions.php
+++ b/includes/admin/payments/actions.php
@@ -193,20 +193,8 @@ function edd_update_payment_details( $data = array() ) {
 	// Address
 	$address = $data['edd_order_address'];
 
-	// Setup first and last name from input values.
-	$name       = $customer->name;
-	$names      = explode( ' ', $name );
-	$first_name = ! empty( $names[0] ) ? $names[0] : '';
-	$last_name  = '';
-
-	if ( ! empty( $names[1] ) ) {
-		unset( $names[0] );
-		$last_name = implode( ' ', $names );
-	}
-
 	edd_update_order_address( absint( $address['address_id'] ), array(
-		'first_name'  => $first_name,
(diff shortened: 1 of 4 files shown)


ashleyfae (commented):
Code looks good. Just need to do some testing.

ashleyfae:
To test:

Have an order record that has an associated address (you can confirm by checking in the admin UI).

In release/3.0 and using debug bar console or similar, run this:

```php
$payment = new EDD_Payment( $your_order_id_here );
var_dump( $payment->address );
```

The resulting array will be empty.

Switch to this issue branch and run the same code again. This time you should get the full address.

ashleyfae (approved):

PR #8318, opened 2021-01-21
Migrate log metadata

### LINKED ISSUE #8317: 3.0 - File download log metadata is not migrated
### Expected behavior
When I migrate data, I expect all of the related data to be migrated.

### Actual behavior
I'm running a migration of a database with a lot of file download logs, and most of them have metadata (I was specifically looking for `_edd_log_all_access_pass_id`), and none of the metadata is being migrated.

1. Start with a 2.x database which definitely has file download logs which definitely have metadata (other than that which matches up to the new database columns). A database with All Access pass related downloads should be sufficient.
2. Migrate the data (at least the logs).
3. Check the `edd_logs_file_downloadmeta` table--it may be empty (mine is), but it will not have any AA metadata.

PHP Version: 7.4.1
WordPress Version: 5.6

Looking at the data migrator class, there is a loop set up to [migrate metadata](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L291-L304), but it's only in the final `else`, so doesn't run at all for download or API logs.

I think we could move it outside of the `else` and fix it, but looking at the `edd_logmeta` table, it looks like post metadata which is mapped to a new database column is being included in the migration to the meta table, so maybe we could consider tweaking it to remove the metadata which has been added to a column already, and just add the remaining data.

robincornett:
Fixes #8317

Proposed Changes:
1. Updates the data migrator code to update the log metadata for all core log types (previously just main logs; file download and API request logs were excluded).
2. Does not migrate metadata which has been migrated to a column in the main/related log table.

Note: a meta key which did exist in 2.x, apparently, is `_edd_log_user_id`. However, looking at all of the rows that include it (on the site I'm migrating), I'm not convinced we need to keep it. The empty value appears to be saved as `-`, so I see a lot of that; additionally, the rows with values are all single digit numbers, and that does not make sense for a very large site. Actually, the newest rows (migration still going) are all the `-` value, so maybe that's a meta value which is no longer valid at all?

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -238,7 +238,8 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
-			edd_add_file_download_log( $log_data );
+			$new_log_id        = edd_add_file_download_log( $log_data );
+			$add_meta_function = 'edd_add_file_download_log_meta';
 		} elseif ( 'api_request' === $data->slug ) {
 			$meta = $wpdb->get_results( $wpdb->prepare( "SELECT meta_key, meta_value FROM {$wpdb->postmeta} WHERE post_id = %d", absint( $data->ID ) ) );
 
@@ -274,9 +275,11 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
-			edd_add_api_request_log( $log_data );
+			$new_log_id        = edd_add_api_request_log( $log_data );
+			$add_meta_function = 'edd_add_api_request_log_meta';
 		} else {
-			$post = \WP_Post::get_instance( $data->ID );
+			$post      = \WP_Post::get_instance( $data->ID );
+			$post_meta = get_post_custom( $data->ID );
 
 			$log_data = array(
 				'object_id'     => $post->post_parent,
@@ -288,19 +291,23 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
-			$meta            = get_post_custom( $data->ID );
-			$meta_to_migrate = array();
+			$new_log_id        = edd_add_log( $log_data );
+			$add_meta_function = 'edd_add_log_meta';
+		}
 
-			foreach ( $meta as $key => $value ) {
-				$meta_to_migrate[ $key ] = maybe_unserialize( $value[0] );
-			}
+		if ( ! is_callable( $add_meta_function ) ) {
+			return;
+		}
 
-			$new_log_id = edd_add_log( $log_data );
+		$meta_to_migrate = array();
 
-			if ( ! empty( $meta_to_migrate ) ) {
-				foreach ( $meta_to_migrate as $key => $value ) {
-					edd_add_log_meta( $new_log_id, $key, $value );
-				}
+		foreach ( $post_meta as $key => $value ) {
+			$meta_to_migrate[ $key ] = maybe_unserialize( $value[0] );
+		}
+
+		if ( ! empty( $meta_to_migrate ) ) {
+			foreach ( $meta_to_migrate as $key => $value ) {
+				$add_meta_function( $new_log_id, $key, $value );
 			}
 		}
 	}
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -238,6 +238,13 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
+			$meta_to_remove    = array(
+				'_edd_log_file_id',
+				'_edd_log_payment_id',
+				'_edd_log_price_id',
+				'_edd_log_customer_id',
+				'_edd_log_ip',
+			);
 			$new_log_id        = edd_add_file_download_log( $log_data );
 			$add_meta_function = 'edd_add_file_download_log_meta';
 		} elseif ( 'api_request' === $data->slug ) {
@@ -275,6 +282,14 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
+			$meta_to_remove    = array(
+				'_edd_log_request_ip',
+				'_edd_log_user',
+				'_edd_log_key',
+				'_edd_log_token',
+				'_edd_log_version',
+				'_edd_log_time',
+			);
 			$new_log_id        = edd_add_api_request_log( $log_data );
 			$add_meta_function = 'edd_add_api_request_log_meta';
 		} else {
@@ -291,6 +306,7 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
+			$meta_to_remove    = array();
 			$new_log_id        = edd_add_log( $log_data );
 			$add_meta_function = 'edd_add_log_meta';
 		}
@@ -307,7 +323,9 @@ class Data_Migrator {
 
 		if ( ! empty( $meta_to_migrate ) ) {
 			foreach ( $meta_to_migrate as $key => $value ) {
-				$add_meta_function( $new_log_id, $key, $value );
+				if ( ! in_array( $key, $meta_to_remove, true ) ) {
+					$add_meta_function( $new_log_id, $key, $value );
+				}
 			}
 		}
 	}
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -218,6 +218,7 @@ class Data_Migrator {
 			return;
 		}
 
+		$meta_to_migrate = array();
 		if ( 'file_download' === $data->slug ) {
 			$meta = $wpdb->get_results( $wpdb->prepare( "SELECT meta_key, meta_value FROM {$wpdb->postmeta} WHERE post_id = %d", absint( $data->ID ) ) );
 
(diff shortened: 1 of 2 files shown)


robincornett:
Just noting that although the new meta rows are being created, the values are not correct, so this still needs adjusting. That may also explain the wonky key I noted above.

robincornett:
This is now working properly and is ready for review. Note: even though `_edd_log_user_info` shows as a meta key and should be added to the meta table (downloads), for some reason it is not being added for me (this may be intentional due to something else going on in the code but I haven't seen it yet; since the user ID is saved as a meta key/value, we can get the same information from that, I think, so it may not be necessary. If that is the case, then we need to check the code base for where that might be used.).

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -274,9 +287,20 @@ public static function logs( $data = null ) {
...
+				'_edd_log_token',
+				'_edd_log_version',
+				'_edd_log_time',
+			);
+			$meta_to_migrate   = $post_meta;
+			$new_log_id        = edd_add_api_request_log( $log_data );
+			$add_meta_function = 'edd_add_api_request_log_meta';
 		} else {
-			$post = \WP_Post::get_instance( $data->ID );
+			$post      = \WP_Post::get_instance( $data->ID );
```
ashleyfae, line 302: I know this isn't new code, but I'm questioning whether this is actually necessary? I thought `$data` already had the full object and `post_parent` could be accessed via `$data->post_parent` :thinking:

ashleyfae:
Code looks solid. Left one question and just holding off on testing until that change is (possibly) made.

includes/admin/upgrades/v3/class-data-migrator.php
robincornett: Checked and confirmed, `$data` does look like a post object to me as well.

ashleyfae:
## Testing

- [x] Confirmed meta added to edd_logs_file_downloadmeta
- [ ] Confirmed meta added to edd_logs_api_requestmeta
- [ ] Confirmed meta added to edd_logs

ashleyfae:
Confirmed that all meta is added successfully.

I also feel like we probably don't need the user ID meta because we store the customer, and the user ID can be retrieved from that.

However, there are a few backwards compatibility issues. They actually exist right now and need updating either way.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8317/includes/compat/class-log.php#L238-L260

The whole compatibility layer assumes that there is a `user_id` column in the file download log table, but there isn't. There's a `customer_id` column instead.

That get post meta compatibility ends up trying to pull `$file_download_log->user_id` which doesn't exist. What actually needs to happen is:

1. Get the customer from `$file_download_log->customer_id`
2. Once we have the customer record, we can then return `$customer->user_id`

We also need to add a `_edd_log_customer_id` to the list of `$meta_keys` and add some backwards compat logic for that.

All of this also applies to the update meta layer: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8317/includes/compat/class-log.php#L290

includes/compat/class-log.php
```diff
@@ -313,6 +321,7 @@ public function file_download_log_update_post_meta( $check, $object_id, $meta_ke
 			case '_edd_key_ip':
 			case '_edd_log_payment_id':
 			case '_edd_log_price_id':
+			case '_edd_log_customer_id';
```
LisaCee, line 324: Should this be a colon rather than a semicolon?

includes/compat/class-log.php
```diff
@@ -251,11 +253,16 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
 						$key = 'api_key';
 						break;
 					case 'user':
-						$key = 'user_id';
+						$key = 'customer_id';
```
ashleyfae, line 256: This should go back to `user_id`. `user` and `user_id` are equivalent, but `customer_id` is separate.

includes/compat/class-log.php
```diff
@@ -251,11 +253,16 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
...
+
 						break;
 				}
 
 				$value = $file_download_log->{$key};
```
ashleyfae, line 261: Let's wrap this in ```if ( isset( $file_download_log->{$key} ) ) {`

includes/compat/class-log.php
```diff
@@ -251,11 +253,16 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
...
+				if ( 'user' === $meta_key ) {
```
ashleyfae, line 262:
```suggestion
				if ( 'user_id' === $key ) {
```

Because in the switch statement above, `$key` `user` is converted to `user_id`.

includes/compat/class-log.php
```diff
@@ -251,11 +253,16 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
...
+					$customer = new EDD_Customer( $value );
```
ashleyfae, line 263:
```suggestion
					$customer = new \EDD_Customer( $value );
```

Because this class is namespaced as `\EDD\Compat` and `EDD_Customer` is in the root namespace. So we need to prefix it with a `\` or import it. (Importing it would be adding `use EDD_Customer;` to the top of the file.

includes/compat/class-log.php
```diff
@@ -251,11 +253,16 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
...
+					$value    = $customer ? $customer->user_id : 0;
```
ashleyfae, line 264:
```suggestion
					$value    = ! empty( $customer->user_id ) ? $customer->user_id : 0;
```

Feels like the better check.

includes/compat/class-log.php
```diff
@@ -331,8 +340,22 @@ public function file_download_log_update_post_meta( $check, $object_id, $meta_ke
 					) );
 				}
 				break;
+
+			case '_edd_log_customer_id':
```
ashleyfae, line 344: I don't think this is needed? Since you already added a `case '_edd_log_customer_id':` on line 324

ashleyfae (changes requested):

ashleyfae:
Backwards compat logic testing:

- [ ] _edd_log_user_info
- [ ] _edd_log_user_id
- [ ] _edd_log_file_id
- [ ] _edd_key_ip
- [ ] _edd_log_payment_id
- [ ] _edd_log_price_id
- [ ] _edd_log_customer_id
- [ ] Custom key

includes/compat/class-log.php
```diff
@@ -252,15 +254,23 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
...
 						break;
 				}
 
-				$value = $file_download_log->{$key};
+				if ( isset( $file_download_log->{$key} ) ) {
+					$value = $file_download_log->{$key};
+				}
+
+				if ( 'user_id' === $meta_key ) {
+					$customer = new \EDD_Customer( $value );
```
ashleyfae, line 266:
```suggestion
					$customer = new \EDD_Customer( $file_download_log->customer_id );
```

ashleyfae (changes requested):

includes/compat/class-log.php
```diff
@@ -223,6 +223,7 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
 			'_edd_key_ip',
```
ashleyfae, line 223:
```suggestion
			'_edd_log_ip',
```

includes/compat/class-log.php
```diff
@@ -241,6 +242,7 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
 			case '_edd_key_ip':
```
ashleyfae, line 242:
```suggestion
			case '_edd_log_ip':
```

includes/compat/class-log.php
```diff
@@ -295,6 +305,7 @@ public function file_download_log_update_post_meta( $check, $object_id, $meta_ke
 			'_edd_key_ip',
```
ashleyfae, line 305:
```suggestion
			'_edd_log_ip',
```

ashleyfae (changes requested):

ashleyfae:
When we have the `switch ( $key )` statement, we need a new case for `payment_id` to convert that to `order_id`

includes/compat/class-log.php
```diff
@@ -252,15 +254,23 @@ public function file_download_log_get_post_meta( $value, $object_id, $meta_key,
...
+
 						break;
 				}
 
-				$value = $file_download_log->{$key};
+				if ( isset( $file_download_log->{$key} ) ) {
+					$value = $file_download_log->{$key};
+				}
+
+				if ( 'user_id' === $meta_key ) {
```
ashleyfae, line 265:
```suggestion
				if ( 'user_id' === $key ) {
```

ashleyfae (changes requested):

ashleyfae:
This is all working well now. However, as one last thing, I propose removing this whole section:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/02aa3fc50dab86f7a3eb94e27a52306237779207/includes/compat/class-log.php#L340-L351

There is no column for `user_id` so this does nothing right now. Given that `_edd_log_user_info` has already been deprecated and removed from 2.9, I think we can also reasonably remove update backwards compatibility logic for it.

ashleyfae (approved):

PR #8323, opened 2021-01-25
Update discounts to accommodate for price variations

### LINKED ISSUE #8322: 3.0 - Discounts - incorrectly able to select variable price option for download requirements/exclusions
### Expected behavior
When I set excluded or required downloads with specific price options on a discount, I expect the price options to be saved correctly.

### Actual behavior

The price options are not honoured and revert to "all options" in all instances. This means specific price options cannot be included (or excluded) from the cart discounts.

1) Add a discount and add a variable price "option" to the excluded (or required) to downloads.
2) Save the discount.
3) When you reload the page, you will it reverts to "All Options" for that price option.

PHP Version: 7.4

EDD Version (or branch): 3.0-beta

WordPress Version: 5.6

robincornett:
Fixes #8322

Proposed Changes:
1. Changes all sanitization on product requirements/exclusions to `esc_attr` instead of casting to an integer.
2. Updates the product in cart check and discount calculation to allow for price IDs, or just download ID even for variable products.

### DIFF
--- a/includes/admin/discounts/discount-actions.php
+++ b/includes/admin/discounts/discount-actions.php
@@ -108,8 +108,8 @@ function edd_admin_add_discount( $data = array() ) {
 	}
 
 	// Meta values.
-	$to_add['product_reqs']      = isset( $data['product_reqs'] ) ? array_unique( array_map( 'esc_attr', $data['product_reqs'] ) ) : '';
-	$to_add['excluded_products'] = isset( $data['excluded_products'] ) ? array_unique( array_map( 'esc_attr', $data['excluded_products'] ) ) : '';
+	$to_add['product_reqs']      = isset( $data['product_reqs'] ) ? array_unique( array_map( 'sanitize_text_field', $data['product_reqs'] ) ) : '';
+	$to_add['excluded_products'] = isset( $data['excluded_products'] ) ? array_unique( array_map( 'sanitize_text_field', $data['excluded_products'] ) ) : '';
 
 	$to_add = array_filter( $to_add );
 
@@ -236,8 +236,8 @@ function edd_admin_edit_discount( $data = array() ) {
 	}
 
 	// Known & accepted core discount meta
-	$to_update['product_reqs']      = isset( $data['product_reqs'] ) ? array_unique( array_map( 'esc_attr', $data['product_reqs'] ) ) : '';
-	$to_update['excluded_products'] = isset( $data['excluded_products'] ) ? array_unique( array_map( 'esc_attr', $data['excluded_products'] ) ) : '';
+	$to_update['product_reqs']      = isset( $data['product_reqs'] ) ? array_unique( array_map( 'sanitize_text_field', $data['product_reqs'] ) ) : '';
+	$to_update['excluded_products'] = isset( $data['excluded_products'] ) ? array_unique( array_map( 'sanitize_text_field', $data['excluded_products'] ) ) : '';
 
 	// "Once per customer" checkbox.
 	$to_update['once_per_customer'] = isset( $data['once_per_customer'] )
--- a/includes/admin/discounts/discount-actions.php
+++ b/includes/admin/discounts/discount-actions.php
@@ -108,8 +108,8 @@ function edd_admin_add_discount( $data = array() ) {
 	}
 
 	// Meta values.
-	$to_add['product_reqs']      = isset( $data['product_reqs']      ) ? wp_parse_id_list( $data['product_reqs']      ) : '';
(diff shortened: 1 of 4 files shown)


includes/admin/discounts/discount-actions.php
```diff
@@ -108,8 +108,8 @@ function edd_admin_add_discount( $data = array() ) {
 	}
 
 	// Meta values.
-	$to_add['product_reqs']      = isset( $data['product_reqs']      ) ? wp_parse_id_list( $data['product_reqs']      ) : '';
-	$to_add['excluded_products'] = isset( $data['excluded_products'] ) ? wp_parse_id_list( $data['excluded_products'] ) : '';
+	$to_add['product_reqs']      = isset( $data['product_reqs'] ) ? array_unique( array_map( 'esc_attr', $data['product_reqs'] ) ) : '';
```
ashleyfae, line 111:
```suggestion
	$to_add['product_reqs']      = isset( $data['product_reqs'] ) ? array_unique( array_map( 'sanitize_text_field', $data['product_reqs'] ) ) : '';
```

That feels like a better choice since we're saving data and not displaying it.

Will need updating in all other areas.

includes/cart/class-edd-cart.php
```diff
@@ -717,11 +717,16 @@ public function display_cart_discount( $echo = false ) {
...
 	 * @return bool
 	 */
 	public function is_item_in_cart( $download_id = 0, $options = array() ) {
-		$cart = $this->get_contents();
+		$cart      = $this->get_contents();
+		$variation = explode( '_', $download_id );
+		if ( ! empty( $variation[1] ) ) {
+			$download_id         = absint( $variation[0] );
+			$options['price_id'] = absint( $variation[1] );
+		}
```
ashleyfae, line 729: Can you help me understand why this change is needed? Just trying to figure out the circumstances in which a `{download_id}_{price_id}` string would be passed into this function.

robincornett: There are use cases where the product ID/price ID string is passed to the function from the discount data ([product requirements](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8322/includes/class-edd-discount.php#L1429) here). So my understanding is that the cart check function needs to know not just if the product is in the cart, but also if the specific variation is in the cart?

ashleyfae: Thank you! I figured there was something -- I just couldn't find it.

I am in two minds about what I'd prefer though.. Part of me is inclined to do the exploding logic at [this point](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8322/includes/class-edd-discount.php#L1429) instead of inside the function itself.

I'll think on it a bit.

robincornett: My thought with putting it where I did is that I would have to do it just the one time in the one place. Otherwise we need to make sure that any time it's used in the context of checking a discount, we need to potentially include the conversion, and that may extend to add-ons and/or third party code. This felt like it would cover multiple bases, I guess, but I'm willing to change.

LisaCee:
This works well for the Discount screen.  If I save a variation, it stays the same after reload.

This PR is only addressing the admin screen, right?  Upon testing this, I may have discovered another new issue (or two) and I want to make sure before reporting elsewhere.

LisaCee:
My previous comment was unintentionally vague.  The issues I _may_ have discovered are that the inclusion/exclusion requirements do not seem to be honored on the frontend.  Also, on the discount overview page, the uses column remains at 0 even after the discount code is used.

ashleyfae:
Am I correct in thinking that showing/allowing variations on downloads is new in 3.0?

robincornett:
It should be working on the front end as well as the back end, but it feels a little shoehorned in, TBH.

In 2.x, you can only select the main/parent product; the variations are not available for independent selection.

ashleyfae:
Yeah and unless I'm missing it, I don't see an issue for "allow variations in discounts". It makes me wonder if including them in the dropdown was actually unintentional when we reworked that?

I'm inclined to just revert that so they can't be selected, because as it is right now, it feels like we're rushing in a feature that was possibly never meant to be there.

robincornett:
One option would be to remove variations from 3.0 so we're not adding in "one more thing", and keep the issue open and milestone for 3.1?

ashleyfae:
> One option would be to remove variations from 3.0 so we're not adding in "one more thing", and keep the issue open and milestone for 3.1?

I'm fine with that for now, but I'd want to think some more about whether or not we _should_ do this. So put it in 3.1 for now, but not guaranteed.

ashleyfae:
Closing in favour of https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8356

PR #8328, opened 2021-01-27
Introduce edd_refundable_order_statuses filter #8324

### LINKED ISSUE #8324: 3.0 - Add filter on statuses checked when determining order refund eligibility

The `edd_is_order_refundable()` function has a list of statuses it checks. If the order's status isn't in that array, the order cannot be refunded.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/functions/refunds.php#L39

```php
// Only completed orders can be refunded.
if ( ! in_array( $order->status, array( 'complete', 'publish', 'partially_refunded' ), true ) ) {
	return false;
}
```

We should add a filter on that list of statuses so other plugins can hook in.

### Justification or use case

Renewal payments made via the Recurring add-on cannot be refunded because Recurring uses a custom `edd_subscription` status.

https://github.com/easydigitaldownloads/edd-recurring/issues/1400

ashleyfae:
Fixes #8324

Proposed Changes:
1. Introduce `edd_refundable_order_statuses` filter to supported refund statuses can be changed.

### DIFF
--- a/includes/orders/functions/refunds.php
+++ b/includes/orders/functions/refunds.php
@@ -35,8 +35,19 @@ function edd_is_order_refundable( $order_id = 0 ) {
 		return false;
 	}
 
-	// Only completed orders can be refunded.
-	if ( ! in_array( $order->status, array( 'complete', 'publish', 'partially_refunded' ), true ) ) {
+	/**
+	 * Filters the order statuses that are allowed to be refunded.
+	 *
+	 * @param array $refundable_order_statuses
+	 *
+	 * @since 3.0
+	 */
+	$refundable_order_statuses = (array) apply_filters( 'edd_refundable_order_statuses', array(
+		'complete', 'publish', 'partially_refunded'
+	) );
+
+	// Only orders with a supported status can be refunded.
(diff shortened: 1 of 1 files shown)


robincornett (commented):
Would you be interested in making the refundable order statuses (and filter) into a function? In 3.0, we've added two new functions for [gross](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/2724f848583d6bf936e53cc27cf5234bbe7373dc/includes/orders/functions/orders.php#L1314) and net order statuses and it seems like this would logically work there.

ashleyfae:
> Would you be interested in making the refundable order statuses (and filter) into a function? In 3.0, we've added two new functions for [gross](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/2724f848583d6bf936e53cc27cf5234bbe7373dc/includes/orders/functions/orders.php#L1314) and net order statuses and it seems like this would logically work there.

I'm good with that. :+1:

robincornett (approved):

PR #8339, opened 2021-02-01
Change default order_item status to pending #8313

### LINKED ISSUE #8313: 3.0 - "inherit" order item status not properly implemented
## Bug Report

The default `edd_order_item` `status` is `inherit` https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/schemas/class-order-items.php#L109 However, this doesn't really feel like it's been properly implemented as a status.

To see one example:

1. Activate the Free Downloads add-on and set up a free download.
2. Download the free product.
3. This will have created a new order record, but the order item status is `inherit`. (Check in DB)
4. Go to your order history on the front-end and click through to "View Details and Downloads". The download links will not be available. That's due to this conditional being used in the receipt template:

```php
'complete' === $item->status
```

In this case, `$item->status` is `inherit` and not `complete`.

Other examples of `inherit` not being considered are with reporting. `EDD\Stats::get_order_item_earnings()` for example.. It only checks if the order item status is `complete`.

I'd actually never seen `inherit` in my database table until I ran this Free Downloads test. I don't think it is possible for it to appear in core at all. Therefore, for simplicity, I suggest we simply change the default to `pending`.

PHP Version:

WordPress Version:

robincornett:
Fixes #8313

### DIFF
--- a/includes/database/tables/class-order-items.php
+++ b/includes/database/tables/class-order-items.php
@@ -38,7 +38,7 @@ final class Order_Items extends Table {
 	 * @since 3.0
 	 * @var int
 	 */
-	protected $version = 202002141;
+	protected $version = 202102010;
 
 	/**
 	 * Array of upgrade versions and methods
@@ -52,6 +52,7 @@ final class Order_Items extends Table {
 		'201807273' => 201807273,
 		'201906241' => 201906241,
 		'202002141' => 202002141,
+		'202102010' => 202102010,
 	);
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/database/tables/class-order-items.php
```diff
@@ -183,4 +184,19 @@ protected function __202002141() {
 
 	}
 
+	/**
+	 * Upgrade to version 202102010.
+	 *  - Change default value for `status` column to 'pending'.
+	 *
+	 * @return void
```
ashleyfae, line 191:
```suggestion
	 * @return bool
```

ashleyfae (changes requested):
Also, in the `edd_add_order_item()` function, can you update the PhpDocs to mention the default `pending` value?

ashleyfae (approved):

PR #8340, opened 2021-02-01
Update query docblocks & log namespaces #8315

### LINKED ISSUE #8315: 3.0 - Review/update doc blocks for correct database column parameters
### Expected behavior
I expect the doc blocks to be correct.

### Actual behavior
Reviewing an issue for All Access, I referred to the [doc block for the file download logs](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/queries/class-log-file-download.php#L116-L118) and mistakenly recommended using the `payment_id__in` parameter for a query. It should be `order_id__in` (same for other columns).

n/a

PHP Version: 7.4.1
WordPress Version: 5.6

It would be a good idea to check all of the log doc blocks, at least, while updating this, I think.

ashleyfae:
Fixes #8315 

Proposed Changes:
1. Review and update all query docblocks for accuracy.
2. Change file download log namespace from `EDD\File_Download_Logs\File_Download_Log\File_Download_Log` to `EDD\Logs\File_Download_Log`
3. Change API request log namespace from `EDD\Api_Request_Logs\Api_Request_Log` to `EDD\Logs\Api_Request_Log`

### DIFF
--- a/includes/admin/reporting/class-file-downloads-logs-list-table.php
+++ b/includes/admin/reporting/class-file-downloads-logs-list-table.php
@@ -10,7 +10,7 @@
  * @since       3.0 Updated to use the custom tables.
  */
 
-use EDD\File_Download_Logs\File_Download_Log\File_Download_Log;
+use EDD\Logs\File_Download_Log;
 
 // Exit if accessed directly
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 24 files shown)


robincornett (changes requested):
A few minor changes requested.

Also a thing I've noticed, particularly in the schema classes, is that we are not adhering to the PHPCS standard "Each array item in a multi-line array declaration must end in a comma". Is that intentional?

ashleyfae:
> Also a thing I've noticed, particularly in the schema classes, is that we are not adhering to the PHPCS standard "Each array item in a multi-line array declaration must end in a comma". Is that intentional?

I don't think we're particularly strict about that company-wise. Going to leave it for now.

robincornett (approved):

PR #8343, opened 2021-02-01
Update the edd_30_migrate_order parameters

### LINKED ISSUE #8342: 3.0 - update edd_30_migrate_order parameters
We should update the parameters for this hook to be the order ID and payment meta.

### Justification or use case
Currently, the hook has two identical parameters:

```php
do_action( 'edd_30_migrate_order', $order_id, $data->ID );
```

The 3.0 migration specifically maps the original payment ID to the new order ID.

In working on hooking into the migration process for Simple Shipping, it's necessary to retrieve the original payment meta, as not all of it is copied over to the new order (which may be a bug), but Ashley suggests we go ahead and make the meta available to this hook as it's already been retrieved:

>I feel like we should actually update the action hook in core (edd_30_migrate_order) to pass through the meta. It's already retrieved it at that point and it seems reasonable that anyone hooking in would want to use it. ([ref](https://github.com/easydigitaldownloads/edd-simple-shipping/pull/104#discussion_r567958034))

robincornett:
Fixes #8342

Proposed Changes:
1. Removes the `$data->ID` parameter from the hook because it's identical to `$order_id`.
2. Adds the original `$meta` to the hook as the second parameter. Wondering if this should really be `$payment_meta` instead as that's been unserialized and is probably what's really needed (or hard to get) for an extension.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -1089,10 +1089,11 @@ class Data_Migrator {
 		 * Now that we're done, let's run a hook here so we can allow extensions to make any necessary changes.
 		 *
 		 * @since 3.0
-		 * @param int   $order_id The order ID.
-		 * @param array $meta     The original post meta for the payment.
+		 * @param int   $order_id     The order ID.
+		 * @param array $payment_meta The `_edd_payment_meta` value for the original payment.
+		 * @param array $meta         The original post meta for the payment.
 		 */
-		do_action( 'edd_30_migrate_order', $order_id, $meta );
+		do_action( 'edd_30_migrate_order', $order_id, $payment_meta, $meta );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


ashleyfae:
> 2\. Wondering if this should really be `$payment_meta` instead as that's been unserialized and is probably what's really needed (or hard to get) for an extension.

That would only cover what's stored in the `_edd_payment_meta` key -- it wouldn't include other custom meta outside that.

I wouldn't mind both though.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -1085,8 +1085,15 @@ public static function orders( $data = null ) {
...
 
-		// Now that we're done, let's run a hook here so we can allow extensions to make any necessary changes
-		do_action( 'edd_30_migrate_order', $order_id, $data->ID );
+		/**
+		 * Now that we're done, let's run a hook here so we can allow extensions to make any necessary changes.
+		 *
+		 * @since 3.0
+		 * @param int   $order_id     The order ID.
+		 * @param array $payment_meta The `_edd_payment_meta` value for the original payment.
+		 * @param array $meta         The original post meta for the payment.
```
ashleyfae, line 1094:
```suggestion
		 * @param array $meta         All post meta associated with the payment.
```

Just provides better clarity that it's **all** meta keys/values.

ashleyfae (approved):

ashleyfae:
Feel free to merge when tests pass. :+1:

Issue #7621, opened 2020-03-04
Subtotal value is calculated differently across different UIs when fees have been added


nosegraze:
## Bug Report

These examples come from fees added by Discounts Pro but I believe the actual fee logic and showing the subtotal is in EDD core, which is why I've made the issue here.

### Expected behavior

When I view the order subtotal in various parts of EDD (checkout screen, confirmation screen, admin UI, etc.), I expect the subtotal value to always be the same.

### Actual behavior

The subtotal value changes between different UIs. Some show the subtotal before the fees are added, some show the subtotal after the fees are added.

**Example on `master` branch:**

![Screenshot from 2020-03-04 13-54-28](https://user-images.githubusercontent.com/6324272/75886746-a47f6c00-5e20-11ea-8f00-0abba1e28152.png)

![Screenshot from 2020-03-04 13-54-37](https://user-images.githubusercontent.com/6324272/75886771-ae08d400-5e20-11ea-9dab-27e3f18c2497.png)

**Example on `release/3.0` branch:**

![Screenshot from 2020-03-04 13-24-18](https://user-images.githubusercontent.com/6324272/75886877-dbee1880-5e20-11ea-8ed5-b903083da8ff.png)

![Screenshot from 2020-03-04 13-26-11](https://user-images.githubusercontent.com/6324272/75886889-dee90900-5e20-11ea-8bfd-128a7331989c.png)

![Screenshot from 2020-03-04 13-27-38](https://user-images.githubusercontent.com/6324272/75886896-e14b6300-5e20-11ea-8eb8-a681e366785d.png)

### Steps to reproduce the behavior
1) Install Discounts Pro and configure a discount.
2) Make a purchase that will have the discount applied. Note the subtotal on the cart page.
3) Complete payment.
4) Note the subtotal on the purchase confirmation page.
5) Note the subtotal on the purchase confirmation page.

### Information (if a specific version is affected):
PHP Version:

EDD Version (or branch): Tested on both `master` and `release/3.0`

WordPress Version:

Any other relevant information:

nosegraze:
Just doing a little more research...

From the 3.0 database description on the `subtotal` column in the `orders` table:

[...]

PR #8346, opened 2021-02-02
Don't add entire order fees if fee has a download_id #8345

### LINKED ISSUE #8345: 3.0 - Fees appear twice on purchase receipt
### Expected behavior

When I view a purchase receipt, I expect each fee to only appear once.

### Actual behavior

Fees appear twice. The correct amount is charged, but they appear twice on the receipt.

download

The issue is that the fee is getting added to both the order and the order_item.

1) Install something that will add fees. (Simple Shipping, Discounts Pro, or custom code.) I _think_ you need the fee applied to a specific download.
2) Purchase an item that will contain a fee.
3) View purchase receipt.

PHP Version:

WordPress Version:

Original report: https://github.com/easydigitaldownloads/edd-simple-shipping/issues/102

ashleyfae:
Fixes #8345 

Proposed Changes:
1. When adding entire order fees, skip the fee if it has a `download_id` specified.
2. Update View Order page to show fees assigned to individual order items.

For testing:

## Checkout / purchase testing

We should test with a variety of scenarios, including Simple Shipping, Discounts Pro, and different combinations of manual fees. Here's a code snippet to aid in testing:

```php
add_action( 'init', function() {
	// Add a $10 fee for download ID 100 only.
	EDD()->fees->add_fee( array(
		'amount'      => 10,
		'id'          => 'specific_download_fee',
		'label'       => 'Specific Download Fee',
		'download_id' => 100, // Change to whatever download ID you want
	) );

	// Add a $20 overall cart fee.
	EDD()->fees->add_fee( array(
		'amount'      => 20,
		'id'          => 'overall_cart_fee',
		'label'       => 'Overall Cart Fee',
	) );
} );
```

In all test we should ensure the following:

- If a fee is assigned to a specific download, an order adjustment should be added with: `object_type` = `order_item`. There should be no adjustment with `object_type` = `order` for this fee.
- If a fee is NOT assigned to a specific download, an order adjustment should be added with `object_type` = `order` . There should be no order_item adjustment for this fee.
- If you have both fee types at the same time, then one of the fees should be for order_item and the other for order. But a single fee should never be given two records (one for each type). This is the issue that was happening in release/3.0

## Admin Order testing

- Manually create an order. Add an order item fee and entire order fee. Upon save, ensure both appear correctly (both in UI and in database).
- View an existing order that was created via the front-end. Ensure fees appear correctly for both order items and entire orders.


robincornett (commented):
I spent some time digging into why this felt familiar and realized that I had toyed with removing the duplicated `order` fees when a fee is applied to an `order_item` in #8057. However, I reverted that because I discovered that it had more ramifications than I was prepared to deal with in that issue.

Specifically, the order totals are calculated off of `order` adjustments, I think, and do not factor in `order_item` adjustments ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8057#issuecomment-660103904)).

So with this current PR and Simple Shipping, the shipping fee is added only to the `order_item`, but the total charged does not include the fee and the fee doesn't show anywhere with the order (not on the front end or in the admin).

ashleyfae:
Good catch on that @robincornett . I'll see about reworking that logic.

ashleyfae:
> the fee doesn't show anywhere with the order (not on the front end or in the admin).

@robincornett Where are you viewing that on the front end? I see it on the purchase confirmation, just wondering if you're looking at something else?

robincornett:
It shows during checkout, but not on the purchase confirmation screen for me (or on the purchase history details screen):

image

The total appears correct on checkout, but the amount in the database is not (the shipping fee is $2 and the total shows as 20).

ashleyfae:
Thanks. I think it only doesn't appear if you don't _also_ have an order fee. (So if you have order + order_item, they both appear. If you only have order_item, it doesn't.)

ashleyfae:
Pushed up some changes. This should hopefully fix the order totals and fees appearing on the front-end.

To-do in this PR: 

- Fees appearing in admin.

To-do later:

- I think I might handle any refund logic in the dedicated refund issue, as all that logic will be changing anyway.

ashleyfae:
Updated to include in the admin UI. For testing this: Make sure fees appear when both adding a manual order and viewing an existing order.

includes/admin/payments/orders.php
```diff
@@ -593,20 +593,23 @@ function edd_order_details_overview( $order ) {
...
-					'type'         => esc_html( $adjustment->type ),
-					'description'  => esc_html( $adjustment->description ),
-					'subtotal'     => esc_html( $adjustment->subtotal ),
-					'tax'          => esc_html( $adjustment->tax ),
-					'total'        => esc_html( $adjustment->total ),
-					'dateCreated'  => esc_html( $adjustment->date_created ),
-					'dateModified' => esc_html( $adjustment->date_modified ),
-					'uuid'         => esc_html( $adjustment->uuid ),
+				$adjustment_args = array(
+						'id'           => esc_html( $adjustment->id ),
```
robincornett, line 597: Why did all of these get an extra tab to indent?

robincornett (commented):
Placing an order from the front end:
- [x] fee applied to the order and fee applied to an item both show correctly in the front end and admin, and the database values all appear correct
- [x] fee applied to an item shows correctly in the front end and admin, and the database values all appear correct

In the admin:
- [x] applying a fee to the order shows correctly
- [ ] applying a fee to an order item does not--in the database, it's applied to the order. The totals are calculated correctly. 
- [x] fee applied to the order shows correctly in the front end and admin, and the database values all appear correct

LisaCee:
In admin the `order_item` fee is added twice in the database, once as `order_item` and once as `order`.  The order, though, still evaluated to the right total.

LisaCee:
So far, on the front end I have tested all of these scenarios and have not seen a duplicate fee in the UI, admin or database:
- Discount Pro item-level fee
- Discount Pro with cart-level fee
- Snippet with item and cart-level fees
- Simple Shipping with item and cart-level fees

LisaCee:
If I create a fee in Discount Pro, either item or cart, should I see that when I create an order in the admin or are they totally separate?

assets/js/admin/orders/order-overview/views/form-add-order-adjustment.js
```diff
@@ -201,10 +201,13 @@ export const FormAddOrderAdjustment = Dialog.extend( {
 			orderItem.get( 'adjustments' ).add( model );
 			// Adding to the Collection doesn't bubble up a change event.
 			orderItem.trigger( 'change' );
+			model.set( 'objectType', 'order_item' );
```
robincornett, line 204: @spencerfinnell if you have a chance, could you put eyes on this? I changed it as the template markup was always setting the adjustment to be `order`, even for `order_item` specific fees, but I don't trust my JS here and may have broken things.

robincornett:
@LisaCee Discounts Pro works by using the EDD_Fees class and none of those are added to manual orders. This also includes the Gateway Fees extension and Simple Shipping.

Not sure why downloads might be missing from the selector, unless it's that you have too many and they're not all loading initially? They should be able to load additional downloads if you search for them.

assets/js/admin/orders/order-overview/views/form-add-order-adjustment.js
spencerfinnell: @robincornett The code itself looks correct. If it's performing as expected, I think you're good to go.

robincornett:
@ashleyfae @LisaCee This should be ready for further testing--the updates I've pushed have mostly been to get the `order_item` fees added properly in `edd_add_manual_order`, as that was never working properly due to how the adjustment data was attached to the order information.

Side benefit was that I've also fixed the order fee display in the admin, so that if a fee is item specific, it will show the item name with the fee description--that was already supposed to be happening but wasn't.

LisaCee (approved):
In admin everything seems to be working now.
- Item-level fee saved as `order_item` in db
- Item-level fee showing item name on admin
- Cart fee saved as `order` db

ashleyfae:
## Tested:

For all tests, confirmed:

- Total price calculation is correct
- Database entries are correct
- Admin order page is correct
- Purchase receipt is correct

- [ ] Manually created order with order item fee
- [ ] Manually created order with entire order fee
- [ ] Front-end checkout with order item fee
- [ ] Front-end checkout with entire order fee
- [ ] Front-end checkout with Discounts Pro

ashleyfae:
I've pushed up two changes.

1. Only show the fee description on the purchase receipt if we have one. Because you can have a fee without a description when adding a manual order. This prevents the `&ndash;` from appearing if there is no description. https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8346/commits/84d92383ed666f3c18c93e5700fb5468ecaee06e
2. The admin UI now shows the fee **subtotal** instead of the total. Previously it was confusing because it was showing the fee TOTAL amount (including tax). But then the tax amount was *also* in the separate tax listing.

Before:

Screenshot from 2021-02-10 10-50-37

After:

Screenshot from 2021-02-10 10-50-50

In both cases, it's a $10 fee and a $20 fee, with 5% tax. In "Before" the fees were listed as $10.50 (fee + tax) and $21 (fee + tax). But then the Tax row further down included the tax for all order items AND fees. It was confusing. This change now makes the admin UI consistent with the purchase receipt, which shows subtotal amounts.

There is still one bug: when manually creating an order, fees never have tax applied. I'm inclined to handle this separately as it's less urgent than the data issues fixed in the rest of this PR.

robincornett:
Two comments on that, one which I think we maybe could address now, and one which may merit its own separate issue:
1. The item specific fees now correctly reference the item in the admin view of the order, but the front end view does not. So your front end receipt for your Book Covers will show only "Specific Download Fee" for the fee amount. Could we go ahead and update the descriptions in the customer's receipt as well so they match? (Note: I'm looking at the admin template and there is a `:&nbsp;` added to the order item name even if there is not a description, which maybe should be tweaked?)
2. Since the subtotal is just the total of the items in the cart, before taxes/fees/discounts, I feel like mentally I want to see it above those items, because it seems like it should be the sum of the things above it. But I think that could be its own issue and definitely not a deal breaker.

ashleyfae:
> 2\. Since the subtotal is just the total of the items in the cart, before taxes/fees/discounts, I feel like mentally I want to see it above those items, because it seems like it should be the sum of the things above it. But I think that could be its own issue and definitely not a deal breaker.

Let's save that for a different issue. In general, EDD has a massive problem with subtotal inconsistency. See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7621

robincornett:
The shortcode receipt now shows the product name for order item level fees.

templates/shortcode-receipt.php
```diff
@@ -84,8 +84,20 @@
...
 					<li>
-						<?php if ( ! empty( $fee->description ) ) : ?>
-							<span class="edd_fee_label"><?php echo esc_html( $fee->description ); ?></span>
+						<?php if ( 'order_item' === $fee->object_type || ! empty( $fee->description ) ) : ?>
+							<span class="edd_fee_label">
+								<?php
+								if ( 'order_item' === $fee->object_type ) {
+									$order_item = edd_get_order_item( $fee->object_id );
+									if ( $order_item ) {
+										echo esc_html( $order_item->product_name ) . ':&nbsp;';
```
ashleyfae, line 93: This is actually giving me a double space. I don't think the `&nsbp;` is needed.

Screenshot from 2021-02-10 14-36-39

ashleyfae (commented):
:+1: from me!

robincornett (approved):

PR #8351, opened 2021-02-02
Keep extra user_info during order migration

### LINKED ISSUE #8350: 3.0 - Order migration should only remove core payment metadata
### Expected behavior
When I migrate orders from posts to the new orders table in 3.0, I expect my data to all be there.

### Actual behavior
If an extension stores any data inside of the `_edd_payment_meta`, specifically adding data to the `user_info`, it will be lost during migration.

1. Set up a 2.9 storefront with Simple Shipping and make some orders with shipping addresses.
2. Update to 3.0 and migrate the orders.
3. The shipping addresses are gone.

PHP Version: 7.4.1
WordPress Version: 5.6

Per @ashleyfae in Slack:
>In the migrator, we could match $user_info against a list of keys that we put there in EDD core. So that's:
id
email
first_name
last_name
discount
address

>I think that's it. If the `$user_info` array has more keys than that, then we save the extra data into meta to preserve it. That way the shipping info (and anything else people have thrown in) is never lost. So that's an EDD core change.

>So if $user_info contains id, email, first_name, and everything else I mentioned.. and also has shipping_address or whatever Simple Shipping names it, then we'd only store shipping_address in order meta (using the key user_info and the value is a serialized array with just array( 'shipping_address' => array() )
Then, we'd need a change at the end of this method: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L3089 to fetch user_info meta. If anything is in there, we have to merge the contents of that with $user_info here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L3163

robincornett:
Fixes #8350

Proposed Changes:
1. During order migration, only removes core `user_info` from the payment meta--if anything is left, it's saved.
2. Adds a backwards compatibility check to the `setup_user_info` method in `EDD_Payment` to get the extra `user_info`.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -737,6 +737,14 @@ class Data_Migrator {
 		// Remove core keys from `user_info`.
 		$remaining_user_info = false;
 		if ( ! empty( $user_info ) ) {
+			/**
+			 * Array keys which are part of the core `user_info` in payment meta which are not needed as part of the order meta.
+			 * Extensions can add their keys to this filter if they use the `user_info` array to store data and have
+			 * established a migration process to keep the data intact with the new order tables.
+			 *
+			 * @since 3.0
+			 * @param array The array of user info keys.
+			 */
 			$core_user_info      = apply_filters( 'edd_30_core_user_info', array( 'id', 'email', 'first_name', 'last_name', 'discount', 'address', 'user_id' ) );
 			$remaining_user_info = array_diff_key( $user_info, array_flip( $core_user_info ) );
 		}
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -726,7 +726,6 @@ class Data_Migrator {
 			'key',
 			'email',
 			'date',
-			'user_info',
 			'downloads',
 			'cart_details',
 			'currency',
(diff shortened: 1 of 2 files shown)


includes/payments/class-edd-payment.php
```diff
@@ -3160,6 +3160,12 @@ private function setup_user_info() {
 			);
 		}
 
+		// Check for old `user_info` meta which may still exist.
+		$old_meta = $this->get_meta( 'payment_meta' );
```
ashleyfae, line 3164: I think we should be able to just use `edd_get_order_meta()` here. That's preferable because it will bypass some of the old filters and backwards compat stuff that's in `$this->get_meta()`. I don't think we need all that here.

ashleyfae (commented):
Ran a migration and this is working very well.

However, one thing I noticed is a ton of user info meta with `user_id` (note `user_id` instead of `id`).

How about we add that to our `$core_user_info` list, if it's not empty and is the same value as the normal `id`.

robincornett:
Updated, but I was thinking about this with Simple Shipping yesterday and left a comment there...is it worth adding a filter to the list of core `user_info` keys so that an extension could add its key(s) to the array and then not create the order meta? So Simple Shipping could add its filter with its migration process and the shipping info would not be passed on to the payment meta, since the extension is taking care of it.

I was just thinking that with this as it stands, any orders with shipping will always have the `payment_meta` created and added, even if it's not necessary. Alternatively, the extension could remove the meta when it hooks into the migration. Just thinking out loud.

ashleyfae:
> Updated, but I was thinking about this with Simple Shipping yesterday and left a comment there...is it worth adding a filter to the list of core `user_info` keys so that an extension could add its key(s) to the array and then not create the order meta? So Simple Shipping could add its filter with its migration process and the shipping info would not be passed on to the payment meta, since the extension is taking care of it.
> 
> I was just thinking that with this as it stands, any orders with shipping will always have the `payment_meta` created and added, even if it's not necessary. Alternatively, the extension could remove the meta when it hooks into the migration. Just thinking out loud.

Let's add a filter. :+1:

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -735,6 +734,20 @@ public static function orders( $data = null ) {
 			'tax',
 		);
 
+		// Remove core keys from `user_info`.
+		$remaining_user_info = false;
+		if ( ! empty( $user_info ) ) {
+			$core_user_info      = apply_filters( 'edd_30_core_user_info', array( 'id', 'email', 'first_name', 'last_name', 'discount', 'address', 'user_id' ) );
```
ashleyfae, line 740: Let's also add some docs to explain why you might use this filter.

ashleyfae (approved):

PR #8371, opened 2021-02-09
Exit early if user's email address didn't change #8370

### LINKED ISSUE #8370: EDD_DB_Customers::update_customer_email_on_user_update() can cause memory issues when bulk running get_password_reset_key()
## Bug Report

Running this on a site with only 60 users causes memory issues:

```php
$users = get_users();

foreach( $users as $user ) {
	$key = get_password_reset_key( $user );
}
```

`get_password_reset_key()` eventually updates the user in question to set the new reset key. That update triggers `EDD_DB_Customers::update_customer_email_on_user_update()` which can very quickly lead to memory problems.

Customer issue here: https://secure.helpscout.net/conversation/1416421601/176850?folderId=3554290

In their use case, they are importing users into WordPress using a plugin, which uses `get_password_reset_key()` to send imported users a reset key. After only 5-7 users they run into memory problems and the process stops working. I was able to reproduce by just running the above snippet on a test site.

1) Run the provided code above.
2) Check your error logs. You'll likely run out of memory.

PHP Version:

EDD Version (or branch): 2.9

WordPress Version:

ashleyfae:
Fixes #8370

Proposed Changes:

The only main change is this extra snippet at the very top:

```php
$user = get_userdata( $user_id );

// Bail if the email address didn't actually change just now.
if ( empty( $user ) || $user->user_email === $old_user_data->user_email ) {
	return;
}
```

We exit ASAP if the user's email address didn't just change. The following code should only execute if the user's email changed to something else during the current update.

Other changes are minimal while in the area:

1. Update PhpDocs to include parameters.
2. Reformat code to exit early instead of a bunch of nested if statements.
3. Change `return false;` to `return;`
4. Add comments to action hook.
5. Remove required parameter before optional (PHP 8 error)

This fixed the memory issue for me. We should verify that this hook does still successfully run when actually changing a user's email address.

### DIFF
--- a/includes/class-edd-db-customers.php
+++ b/includes/class-edd-db-customers.php
@@ -319,45 +319,58 @@ class EDD_DB_Customers extends EDD_DB  {
 	/**
 	 * Updates the email address of a customer record when the email on a user is updated
 	 *
+	 * @param int     $user_id       User ID.
+	 * @param WP_User $old_user_data Object containing user's data prior to update.
+	 *
 	 * @since   2.4
+	 * @return  void
 	*/
-	public function update_customer_email_on_user_update( $user_id = 0, $old_user_data ) {
-
-		$customer = new EDD_Customer( $user_id, true );
-
-		if( ! $customer ) {
-			return false;
-		}
+	public function update_customer_email_on_user_update( $user_id, $old_user_data ) {
 
 		$user = get_userdata( $user_id );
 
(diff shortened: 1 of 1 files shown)


robincornett (approved):
I was having trouble testing this and realized that the issue was actually in FES, looks like `FES_DB_Vendors->update_vendor_email_on_user_update()` will need to be looked at as well.

Once I got that sorted, I was able to reproduce the issue using the code from the original issue post, and this does resolve it.

PR #8377, opened 2021-02-10
Don't use associative array in call_user_func_array() #8358

### LINKED ISSUE #8358: 3.0 – Uncaught Error: Unknown named parameter $endpoint in class-table-endpoint.php on line 217
### Expected behavior
When viewing the reports, I expect to not get a fatal error :-) This may still be in progress though, so mentioning it...

### Actual behavior
When clicking on "Downloads" or "Payment Gates" I get the above error.

ashleyfae:
Fixes #8358

Proposed Changes:
1. Don't use an associative array.

## To test

Visit all reports and ensure everything displays correctly.

@robincornett please test in PHP 8
@LisaCee Please test in non-PHP 8

### DIFF
--- a/includes/class-edd-html-elements.php
+++ b/includes/class-edd-html-elements.php
@@ -698,7 +698,7 @@ class EDD_HTML_Elements {
 		if ( ! empty( $args['show_option_all'] ) ) {
 			if ( $args['multiple'] && ! empty( $args['selected'] ) ) {
 				$selected = selected( true, in_array( 0, $args['selected'] ), false );
-			} else {
+			} elseif ( isset( $args['selected'] ) && ! is_array( $args['selected'] ) ) {
 				$selected = selected( $args['selected'], 0, false );
 			}
 			$output .= '<option value="all"' . $selected . '>' . esc_html( $args['show_option_all'] ) . '</option>';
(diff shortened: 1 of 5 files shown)


robincornett (commented):
In PHP8, I can view all core reports with no fatal errors.

I do get this notice on Downloads (noting here as it seems specific to PHP8):

>usort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero | /easy-digital-downloads/includes/reports/data/downloads/class-earnings-by-taxonomy-list-table.php:147

On Taxes, I'm getting an array to string conversion warning, which appears to be because the country and region `selected` value is an empty array, not a string (no country or region has been selected), by the time it gets to the HTML `select` method. That looks to be an error I introduced when making the `country/region_select` methods, whoops.

ashleyfae:
> I do get this notice on Downloads (noting here as it seems specific to PHP8):
> 
> > usort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero | /easy-digital-downloads/includes/reports/data/downloads/class-earnings-by-taxonomy-list-table.php:147

I wasn't able to reproduce that one -- not sure why. Maybe because I only had one taxonomy in the list.. But I understand why it was happening so I've theoretically fixed it.

robincornett (approved):

LisaCee (approved):

PR #8378, opened 2021-02-10
Re-add discount meta helper functions & update calls to delete/add meta #8372

### LINKED ISSUE #8372: 3.0 - EDD_Discount missing all meta methods
### Expected behavior

In 2.9, EDD_Discount has these methods:

get_meta()
update_meta()

I expect them to still be there in 3.0 and reroute to adjustment meta accordingly.

### Actual behavior

The methods are missing. They're even referenced throughout the code base, but don't actually exist. https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-discount.php#L1128-L1137

PHP Version:

WordPress Version:

Related: this logic is weird: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-discount.php#L1126-L1137

If it's an array, reset the meta, then add the new meta. Otherwise... just delete it? But never actually re-add it?

ashleyfae:
Fixes #8372

Proposed Changes:
1. Re-adds two missing meta methods: `get_meta()` and `update_meta()`. 
2. Replaces uses of `$this->delete_meta()` with `edd_delete_adjustment_meta()` (It appears the `delete_meta()` method does not exist in 2.x so I opted not to add it.)
3. Replaces uses of `$this->add_meta()` with `edd_add_adjustment_meta()`

### DIFF
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -594,7 +594,7 @@ class EDD_Discount extends Adjustment {
 	 * @param string $value      New meta value to set.
 	 * @param string $prev_value Optional. Previous meta value.
 	 *
-	 * @return int|bool Meta ID if the key didn't exist, tru on successful update, false on failure.
+	 * @return int|bool Meta ID if the key didn't exist, true on successful update, false on failure.
 	 */
 	public function update_meta( $key, $value = '', $prev_value = '' ) {
 		$filter_key = '_edd_discount_' . $key;
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -1176,7 +1176,7 @@ class EDD_Discount extends Adjustment {
 
 		if ( isset( $args['product_reqs'] ) ) {
 			// Reset meta
-			edd_delete_adjustment_meta( $this->id, 'product_reqs' );
+			edd_delete_adjustment_meta( $this->id, 'product_requirement' );
 
 			if ( is_array( $args['product_reqs'] ) ) {
 				// Now add each newly required product
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -572,6 +572,46 @@ class EDD_Discount extends Adjustment {
 	}
 
 	/**
+	 * Helper method to retrieve meta data associated with the discount.
+	 *
+	 * @since 2.7
+	 *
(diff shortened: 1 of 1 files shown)


includes/class-edd-discount.php
```diff
@@ -1123,32 +1163,26 @@ public function update( $args = array() ) {
...
 			}
 		}
 
 		if ( isset( $args['product_reqs'] ) ) {
-			if ( is_array( $args['product_reqs'] ) ) {
-
-				// Reset meta
-				$this->delete_meta( 'product_requirement' );
+			// Reset meta
+			edd_delete_adjustment_meta( $this->id, 'product_reqs' );
```
robincornett, line 1179: I think this is not quite right--the $args are `product_reqs`, but the database key is `product_requirement`?
```suggestion
			edd_delete_adjustment_meta( $this->id, 'product_requirement' );
```

includes/class-edd-discount.php
```diff
@@ -571,6 +571,46 @@ private function setup_discount( $discount = null ) {
...
+	/**
+	 * Helper method to update meta data associated with the discount.
+	 *
+	 * @since 2.7
+	 *
+	 * @param string $key        Meta key to update.
+	 * @param string $value      New meta value to set.
+	 * @param string $prev_value Optional. Previous meta value.
+	 *
+	 * @return int|bool Meta ID if the key didn't exist, tru on successful update, false on failure.
```
robincornett, line 597:
```suggestion
	 * @return int|bool Meta ID if the key didn't exist, true on successful update, false on failure.
```

robincornett (commented):
A couple suggestions, but everything appears to be working. However, I'm wondering how I have not been triggering errors when adding/editing discounts in 3.0 until now, as I have been adding product requirements and exclusions--reviewing this code, it seems like that should not have been working?

ashleyfae:
> A couple suggestions, but everything appears to be working. However, I'm wondering how I have not been triggering errors when adding/editing discounts in 3.0 until now, as I have been adding product requirements and exclusions--reviewing this code, it seems like that should not have been working?

I think because EDD core doesn't actually use any of this code. Our logic is completely separate. This stuff is just for backwards compatibility.

robincornett (approved):

PR #8379, opened 2021-02-10
Remove database upgrades from 2018 #8341

### LINKED ISSUE #8341: 3.0 - Remove old 3.0 database upgrades

Many of our custom tables have had numerous upgrade routines since 3.0 started development. For example:

The customers table has 7 upgrades, 5 of which are from 2018.
The orders table has has 7 as well, 4 of which are from 2018.
Most other tables have 3-4.

We should remove some of these old ones before the beta release.

### Justification or use case

Someone upgrading from 2.9 to 3.0 will run through aaaaall these old upgrades. These aren't necessary for brand new upgrades; they've only been needed for people developing 3.0. Running all these upgrades is unnecessary and extremely slow.

I propose we start removing old ones. At minimum I think it's pretty safe to remove those from 2018.

ashleyfae:
Fixes #8341

Proposed Changes:
1. Removes all database upgrades from 2018.

We should test:

1. An upgrade from 2.9.
2. A fresh install of EDD 3.0.

And in both cases ensure there are no database errors.

### DIFF
--- a/includes/database/tables/class-adjustment-meta.php
+++ b/includes/database/tables/class-adjustment-meta.php
@@ -47,9 +47,7 @@ final class Adjustment_Meta extends Table {
      * @since 3.0
      * @var array
      */
-    protected $upgrades = array(
-        '201806142' => 201806142
-    );
+    protected $upgrades = array();
 
 	/**
 	 * Setup the database schema.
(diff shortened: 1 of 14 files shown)


robincornett (approved):
Tested:
- [x] Brand new site, no errors!
- [x] Migration from 2.9, no errors!

PR #8384, opened 2021-02-11
Fix backwards compatibility for log inserts #8383

### LINKED ISSUE #8383: 3.0 - insert_log method missing user_id
### Expected behavior
I am hoping that the [`insert_log`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/class-edd-logging.php#L138) method will work in 3.0 and 2.9 with the same parameters, and it mostly does.

### Actual behavior
When I try to set the `user_id` parameter using this method, there appears to be no way to do it.

Try running this code:

```php
$log_data = array(
	'log_type'     => 'email',
	'post_content' => 'this is the log content',
	'post_title'   => 'subject line',
	'post_author'  => get_current_user_id(),
);
$log_meta = array(
	'to' => 'sample@edd.local',
);
global $edd_logs;
$edd_logs->insert_log( $log_data, $log_meta );
```

EDD converts the other post fields, but not the author, and there is no way in the method to set an author (which would have been automatically set by WP in 2.x).

PHP Version: 7.4.1
WordPress Version: 5.6

Running into this trying to get EDD Message behavior consistent in 2.x and 3.0. I can use the new `edd_add_log` function instead, but then need to rewrite all of the parameters to match the new keys.

ashleyfae:
Fixes #8383 

Proposed Changes:
1. Add support for converting `$log_meta['user']` to `user_id`. If not provided, it falls back to `get_current_user_id()`
2. Set `user_id` to during 3.0 migration.
2. Check if all array keys exist before using.
3. Whenever a piece of `$log_meta` is used in the main `$data` array, ensure it gets unset from `$log_meta`. This ensures that we don't add duplicate meta data unnecessarily. (The data was being added to the main table _and_ the meta table.)
4. For file downloads: get the `order_id` arg from `$log_meta['payment_id']` instead of `$log_meta['order_id']`
5. Add PHPDocs for action hooks.
6. Strip `_edd_log_` meta key prefix during log migration.

### DIFF
--- a/includes/class-edd-logging.php
+++ b/includes/class-edd-logging.php
@@ -146,6 +146,12 @@ class EDD_Logging {
 			'log_type'     => false,
 		) );
 
+		/**
+		 * Triggers just before a log is inserted.
+		 *
+		 * @param array $args     Log entry data.
+		 * @param array $log_meta Log meta data.
+		 */
 		do_action( 'edd_pre_insert_log', $args, $log_meta );
 
 		// Used to dynamically dispatch the method call to insert() to the correct class.
@@ -160,6 +166,7 @@ class EDD_Logging {
 			'object_type' => isset( $args['log_type'] )
 				? $args['log_type']
 				: null,
+			'user_id'     => ! empty( $args['post_author'] ) ? intval( $args['post_author'] ) : 0
 		);
 
 		$type = $args['log_type'];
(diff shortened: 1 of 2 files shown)


ashleyfae:
Updated to set `user_id` during migration: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8384/commits/01844f0e7d617b8e8e9e78098c3e61162d516a96

includes/class-edd-logging.php
```diff
@@ -160,6 +166,7 @@ public function insert_log( $log_data = array(), $log_meta = array() ) {
 			'object_type' => isset( $args['log_type'] )
 				? $args['log_type']
 				: null,
+			'user_id'     => ! empty( $args['post_author'] ) ? intval( $args['post_author'] ) : 0
```
robincornett, line 169: So I'm thinking out loud here and looking at the `insert_log` method in 2.x and the `post_author` is never actually explicitly set in there. Since it uses `wp_insert_post`, the `post_author` was set to the current user ID as the default. So maybe this needs to use the current user ID as the default, skip the `post_author` as I don't think anyone would have been explicitly setting it, but then allow anyone still using this method to explicitly set the `user_id` in a new parameter (because I believe they would have had to be setting this as a new/separate parameter anyway?). I feel like I'm rambling--basically what if we set the default to the current user ID, because I think that would be consistent, but allow a [new] parameter to be passed to any log, not just the api requests. I guess the download log would ignore it if we went that route.

ashleyfae: How about this: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8384/commits/76f11b86fe3d840727fa46beca5875cb03f5d404

I used `$log_meta['user']` as that was already used by API request logs, so it felt consistent.

robincornett (approved):

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8384/commits/e55f82256f616784e1bbb37d315fa108b169ca2f updates the migrator to look for `user_id` in the newly supported meta before falling back to `post_author`.

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8384/commits/3bca8c36f14e979044b4cb2514a60f2d44039dfe strips the `_edd_log_` meta key prefix during migration.

Discovered while testing https://github.com/easydigitaldownloads/edd-message/pull/86#issuecomment-777501294

(Send + log a message in 2.9, upgrade + migrate to 3.0, view logs.)

robincornett (approved):

PR #8393, opened 2021-02-15
Update downloads menu CSS for WP 5.7

### LINKED ISSUE #8360: 3.0 - Admin menu "secondary separators" are a little off in 5.7
### Expected behavior
When I hover over the admin menu for Downloads, and then hover over the submenu items, I expect it to look "right".

### Actual behavior
In WordPress 5.7, it looks like Core is adding a `box-shadow` to basically create a border on the left side of the menu item being hovered. We have some menu items which have an `:after` element styled to give them a little more room, and the `box-shadow` is looking a little wonky with that:

image

1. Set up a site running 5.7 (trunk/beta1) and release/3.0
2. From the dashboard, use your mouse to hover over Downloads, then Reports.

PHP Version: 7.4.1
WordPress Version: 5.7-beta1

[Here's where our CSS to add the extra padding is done.](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/css/admin/menu.scss#L13-L33)

robincornett:
Fixes #8360

Proposed Changes:
This actually reworks some of the CSS entirely as the original approach, which used a combination of `:before` and `:after` pseudo elements, did not seem to play nicely with the box shadow at all.

This should be tested in both WP 5.6 and 5.7.


LisaCee (approved):
Tested with 5.6.1 and 5.7-beta2

ashleyfae (approved):

PR #8399, opened 2021-02-16
Add new indexes to custom tables #8176

### LINKED ISSUE #8176: 3.0 - Audit indexes on new custom tables

We should audit all the indexes we have on the new tables to make sure we have them in the right places. Example:

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8146#pullrequestreview-518726454

> Also, related: I wonder if we should modify this index on the wp_edd_customer_addresses table:
> 
> KEY customer (customer_id)
> 
> to:
> 
> KEY customer_is_primary (customer_id, is_primary)
> 
> That will improve all our queries that search for a customer's primary address.

### Justification or use case

Speed! :sunglasses:

ashleyfae:
Fixes #8176

Proposed Changes:
1. Patches the `index_exists()` method in BerlinDB. The query wasn't working before.
2. **Orders Table:**
    - Removes `searchable` from some columns that do not need to be searched (id, parent, status, etc.). They can still be queried individually, but they do not need `LIKE` queries. This improves the performance of searching orders.
    - Drops old `KEY status (status(20))`
    - Creates new index `KEY status_type (status, type)` . Adding `type` improves the performance of the admin orders table.
3. **Adjustments Table:**
    - Drops the old `code_status_type_scope_amount` index. I couldn't find it being used anywhere. This came closest: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/settings/register-settings.php#L1479-L1488 but funnily enough, that _still_ wouldn't use the index due to `code` not being in use.
    - New index: `KEY type_status (type(20), status(20))` . I put `type` first because I found several cases where we only query on type (such as: get the tax rates) and omit status, but none where it's the reverse.
    - New index: `KEY code (code)` due to removal of the old index that included code. And "get discount by code" is a common usage.

## Testing

- Checkout this branch on an existing 3.0 install. Ensure no database errors.
- Test a fresh EDD 3.0 install. Ensure no database errors.

### DIFF
--- a/includes/database/engine/class-table.php
+++ b/includes/database/engine/class-table.php
@@ -591,10 +591,12 @@ abstract class Table extends Base {
 			return false;
 		}
 
+		$column = esc_sql( $column );
+
 		// Query statement
-		$query    = "SHOW INDEXES FROM {$this->table_name} WHERE %s LIKE %s";
+		$query    = "SHOW INDEXES FROM {$this->table_name} WHERE {$column} LIKE %s";
 		$like     = $db->esc_like( $name );
-		$prepared = $db->prepare( $query, $column, $like );
+		$prepared = $db->prepare( $query, $like );
 		$result   = $db->query( $prepared );
 
 		// Does the index exist?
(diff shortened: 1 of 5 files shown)


robincornett (approved):
Tested:
- [x] new/fresh install of EDD 3.0
- [x] existing install of EDD 3.0
- [x] migration from 2.x to 3.0 just because why not

All good and no errors for me!

PR #8400, opened 2021-02-16
Specify the tax_rate type when retrieving adjustments (compat)

### LINKED ISSUE #8398: 3.0 - Tax backwards compat - include `tax_rate` type in query

We have a query here for checking if a tax rate exists: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/compat/class-tax.php#L100-L105

But it doesn't include `'type' => 'tax_rate'`, and IMO, it should just to be safe.

### Justification or use case

Ensuring we only end up with tax rate objects and don't accidentally end up with another adjustment type that just so happens to have that name/description/scope somehow.

robincornett:
Fixes #8398

### DIFF
--- a/includes/compat/class-tax.php
+++ b/includes/compat/class-tax.php
@@ -98,6 +98,7 @@ class Tax extends Base {
 			// Check if the tax rate exists.
 			} else {
 				$rate = edd_get_adjustments( array(
+					'type'        => 'tax_rate',
 					'fields'      => 'ids',
 					'name'        => $tax_rate['country'],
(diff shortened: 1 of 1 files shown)


ashleyfae (approved):
Confirmed backwards compatibility still works for both adding and updating tax rates. Example:

```php
$rates = array(
	array(
		'country' => 'US',
		'state' => 'TX',
		'rate' => 5
	)
);

update_option( 'edd_tax_rates', $rates );
```

PR #8402, opened 2021-02-16
Update last payment to use edd_get_orders

### LINKED ISSUE #8396: 3.0 - Admin customers page edd_customers_view() - refactor payments query to use edd_get_orders()
## Enhancement Request

This query should be refactored to use `edd_get_orders()` : 

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/customers/customers.php#L347-L352

I can't see any reason it needs to use `edd_get_payments()`. We end up running a completely unnecessary query to get ALL the customer's payments ( `$customer->get_payment_ids()` ) when we could just pass the `customer_id` into `edd_get_orders()`

### Justification or use case

More efficient querying.

robincornett:
Fixes #8396

### DIFF
--- a/includes/admin/customers/customers.php
+++ b/includes/admin/customers/customers.php
@@ -341,24 +341,25 @@ function edd_render_customer_view( $view, $callbacks ) {
 function edd_customers_view( $customer = null ) {
 	$customer_edit_role = edd_get_edit_customers_role();
 
-	$agreement_timestamps = $customer->get_meta( 'agree_to_terms_time',   false );
+	$agreement_timestamps = $customer->get_meta( 'agree_to_terms_time', false );
+	$show_terms           = edd_get_option( 'show_agree_to_terms' );
 	$privacy_timestamps   = $customer->get_meta( 'agree_to_privacy_time', false );
+	$show_privacy         = edd_get_option( 'show_agree_to_privacy_policy' );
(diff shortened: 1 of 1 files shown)


includes/admin/customers/customers.php
```diff
@@ -344,16 +344,19 @@ function edd_customers_view( $customer = null ) {
...
-	$last_payment = edd_get_payments( array(
-		'output'   => 'payments',
-		'post__in' => $customer->get_payment_ids(),
-		'orderby'  => 'date',
-		'number'   => 1
-	) );
+	$last_payment = edd_get_orders(
+		array(
+			'customer_id' => $customer->id,
+			'id__in'      => $customer->get_payment_ids(),
```
ashleyfae, line 350: This shouldn't be necessary because it's already covered by the `customer_id` argument.

ashleyfae (changes requested):

robincornett:
@ashleyfae note my last commit is a more significant change--we only need to run this code if the customer meta doesn't exist but one of the options is enabled, so I reworked the logic to only retrieve the last payment if the necessary conditions are met in the first place.

ashleyfae (commented):
Still need to do some testing, but I love this change. :raised_hands:

PR #8403, opened 2021-02-16
Update CLI migration completed messages

### LINKED ISSUE #8312: 3.0 - Update CLI "Migration complete" messages
After each step of a CLI migration, a "Migration complete." message shows. It would be great if each message showed which migration, or which step of the migration, was complete.

### Justification or use case
I'm running a migration on a storefront which does not collect taxes and when running `wp edd v30_migration`, the first response from the server says "Migration complete", which I assume is related to tax rates (it goes right on into Discounts).

If it just said "Tax rate migration complete", that would feel more helpful, and I would know that the step was at least looked at, even if nothing needed to be migrated.

robincornett:
Fixes #8312

### DIFF
--- a/includes/class-edd-cli.php
+++ b/includes/class-edd-cli.php
@@ -1020,7 +1020,7 @@ class EDD_CLI extends WP_CLI_Command {
 		$upgrade_completed = edd_has_upgrade_completed( 'migrate_order_notes' );
 
 		if ( ! $force && $upgrade_completed ) {
-			WP_CLI::error( __( 'The notes custom table migration has already been run. To do this anyway, use the --force argument.', 'easy-digital-downloads' ) );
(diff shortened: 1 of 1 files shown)


includes/class-edd-cli.php
```diff
@@ -1039,7 +1039,7 @@ public function migrate_order_notes( $args, $assoc_args ) {
 
 			$progress->finish();
 
-			WP_CLI::line( __( 'Migration complete.', 'easy-digital-downloads' ) );
+			WP_CLI::line( __( 'Migration complete: Notes', 'easy-digital-downloads' ) );
```
ashleyfae, line 1042: How about changing this to `Order Notes` ? It would need to be updated in a few other places throughout this method too.

Just thinking we also have `Customer Notes` so it makes sense to prefix this one with `Order` too.

ashleyfae (approved):

PR #8404, opened 2021-02-16
#8374 Redirect After Permanently Deleting

### LINKED ISSUE #8374: 3.0 - Permanently delete trashed order to return on the same page
### Expected behavior
When I click the Permanently Delete link next to an oder in the trash it should return me back to the Trash section.

### Actual behavior
Instead it redirects back to the ALL section which if not being careful and thinking you are back on the Trash section you could delete real orders

1) Trash an order
2) In the Orders filter section click on Trash then Delete permanently a trashed order

PHP Version: 7+

EDD Version (or branch): 3.0

WordPress Version:

LisaCee:
Fixes #8374 

Proposed Changes:
1. Redirect to trashed orders after permanently deleting order
2. Escaping for function
3. Update doc block typo and param

I'm not quite sure the difference between this function and the one right above it, `edd_trigger_purchase_delete`.  I would love to clarify in the descriptions, once it is explained.

### DIFF
--- a/includes/admin/payments/actions.php
+++ b/includes/admin/payments/actions.php
@@ -239,29 +239,6 @@ function edd_update_payment_details( $data = array() ) {
 add_action( 'edd_update_payment_details', 'edd_update_payment_details' );
 
 /**
- * Trigger a Purchase Deletion
- *
- * @since 1.3.4
- * @param array $data Arguments passed
- * @return void
- */
-function edd_trigger_purchase_delete( $data ) {
-	if ( wp_verify_nonce( $data['_wpnonce'], 'edd_payment_nonce' ) ) {
-
-		$payment_id = absint( $data['purchase_id'] );
-
-		if ( ! current_user_can( 'delete_shop_payments', $payment_id ) ) {
-			wp_die( __( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
-		}
-
-		edd_delete_purchase( $payment_id );
-
-		edd_redirect( admin_url( '/edit.php?post_type=download&page=edd-payment-history&edd-message=payment_deleted' ) );
-	}
-}
-add_action( 'edd_delete_payment', 'edd_trigger_purchase_delete' );
-
-/**
  * New in 3.0, permanently destroys an order, and all its data, and related data.
  *
  * @since 3.0
@@ -275,7 +252,7 @@ function edd_trigger_destroy_order( $data ) {
 		$payment_id = absint( $data['purchase_id'] );
 
 		if ( ! current_user_can( 'delete_shop_payments', $payment_id ) ) {
-			wp_die( esc_html_e( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), esc_html_e( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
+			wp_die( esc_html_e( 'You do not have permission to edit this order', 'easy-digital-downloads' ), esc_html_e( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
 		}
 
 		edd_destroy_order( $payment_id );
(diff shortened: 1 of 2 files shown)


includes/admin/payments/actions.php
```diff
@@ -262,24 +262,34 @@ function edd_trigger_purchase_delete( $data ) {
...
 
-		edd_redirect( admin_url( '/edit.php?post_type=download&page=edd-payment-history&edd-message=payment_deleted' ) );
+		$redirect_link = add_query_arg(
+			array(
+				'post_type'   => 'download',
+				'page'        => 'edd-payment-history',
+				'edd-message' => 'payment_deleted',
+				'status'      => 'trash',
+			),
+			admin_url( 'edit.php' ),
```
robincornett, line 290: Although this is totally working as it is, we could use `edd_get_admin_base_url` here instead of the `admin_url` function and remove the `post_type` parameter.

includes/admin/payments/actions.php
```diff
@@ -262,24 +262,34 @@ function edd_trigger_purchase_delete( $data ) {
...
  */
 function edd_trigger_destroy_order( $data ) {
+
 	if ( wp_verify_nonce( $data['_wpnonce'], 'edd_payment_nonce' ) ) {
 
 		$payment_id = absint( $data['purchase_id'] );
 
 		if ( ! current_user_can( 'delete_shop_payments', $payment_id ) ) {
-			wp_die( __( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
+			wp_die( esc_html_e( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), esc_html_e( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
```
robincornett, line 278: Not sure, but would it make sense to go ahead and update the verbiage here to use "order" instead of "payment record"?

robincornett (commented):
This totally works, and is a much better user experience--just a suggestion and a question.

includes/admin/payments/actions.php
LisaCee: I thought about that, but isn't `edd_get_admin_base_url` a 3.0 function?  I wasn't sure about having to add an if/else here.

robincornett: Yes, but it's fine since this is going directly into 3.0 itself instead of an extension--it's the one time we can be sure the function will exist. No condition needed. :)

ashleyfae:
> I'm not quite sure the difference between this function and the one right above it, `edd_trigger_purchase_delete`. I would love to clarify in the descriptions, once it is explained.

From what I can tell, `edd_trigger_purchase_delete()` is the 2.9 equivalent of `edd_trigger_destroy_order()`. In 2.9, `edd_trigger_purchase_delete()` was responsible for handling payment deletions. In 3.0 that's not used at all and `edd_trigger_destroy_order()` has replaced it.

With that in mind, @LisaCee you're welcome to move `edd_trigger_purchase_delete()` (and the `add_action()` hook below it) to `admin/admin-deprecated-functions.php`

includes/admin/payments/actions.php
```diff
@@ -239,47 +239,33 @@ function edd_update_payment_details( $data = array() ) {
...
  */
 function edd_trigger_destroy_order( $data ) {
+
 	if ( wp_verify_nonce( $data['_wpnonce'], 'edd_payment_nonce' ) ) {
 
 		$payment_id = absint( $data['purchase_id'] );
 
 		if ( ! current_user_can( 'delete_shop_payments', $payment_id ) ) {
-			wp_die( __( 'You do not have permission to edit this payment record', 'easy-digital-downloads' ), __( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
+			wp_die( esc_html_e( 'You do not have permission to edit this order', 'easy-digital-downloads' ), esc_html_e( 'Error', 'easy-digital-downloads' ), array( 'response' => 403 ) );
```
robincornett, line 255: These should both be just `esc_html__`, rather than echoing. I would also maybe go ahead and add a period to the end of the sentence--feel like it looks a little bit funny on the die screen without it, but I know it wasn't there before, so I am fine without it.

LisaCee: The missing period was bugging me, too.  :joy:

robincornett (approved):

ashleyfae (approved):

PR #8406, opened 2021-02-17
Introduce new hooks for front-end templates #7759

### LINKED ISSUE #7759: 3.0 - Investigate speed improvements on [purchase_history] & [download_history] shortcodes

We should investigate the performance on `[purchase_history]` and `[download_history]` shortcodes and see if we can improve them. They're super sluggish now. I haven't yet done a true before/after benchmark on the same server, but I can notice a serious difference between a live site on `master` and a local clone of that site on `release/3.0` that has been migrated.

For me, `[purchase_history]` takes 4-5 seconds. EDD is also running a massive number of queries (a little over 300).

`[download_history]` isn't so bad on my live clone where I have 3 purchases. It takes about 2 seconds. However, it's way worse on my general EDD 3 test site where I have a lot of purchases. It ranges from 6-10 seconds and is also running about 380 queries.

I suspect the issue here is related to https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7636 and all the backwards compatibility that's probably running.

ashleyfae:
Fixes #7759

Proposed Changes:
1. `history-downloads.php` : 
    - Replace old `edd_get_payment_x` functions with data directly from the order.
    - Wrap `edd_download_history_files` action in `has_action()` so it only runs if someone has hooked in. (Due to the necessity for passing through payment meta.) Might still rework this.
2. `history-purchases.php` : 
    - Replace all payment queries with order queries.
    - Replace all `edd_x_purchase_history` hooks with new `edd_x_order_history` hooks.
    - In `deprecated-functions.php` (might move this stuff), hook into the new actions to trigger the old ones, only if they've been hooked into.

The goal is that all the old hooks still 100% work, but only get executed if they're actually in use. That way, someone with a fresh EDD install and no extensions will only use new code and reduce the queries on the page from 600+ to more like 50.

I'm keeping this as a draft for now because I don't 100% know how I feel about it. Was just playing around with an idea and want to get the code pushed up.

### DIFF
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -1132,3 +1132,59 @@ function edd_record_sale_in_log( $download_id, $payment_id, $price_id = false, $
 function edd_agree_to_terms_js() {
 	_edd_deprecated_function( __FUNCTION__, '3.0' );
 }
+
+add_action( 'edd_before_order_history', function( $orders ) {
+	if ( ! has_action( 'edd_before_purchase_history' ) ) {
+		return;
+	}
+
+	$payments  = array();
+
+	if ( ! empty( $orders ) ) {
+		$order_ids = wp_list_pluck( $orders, 'id' );
+		$payments  = edd_get_payments( array(
+			'id__in'  => $order_ids,
+			'orderby' => 'date',
+		) );
+	}
+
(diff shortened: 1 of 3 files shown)


robincornett:
Closed; incorporated into #8549.

PR #8413, opened 2021-02-18
Fix order fee backwards compatibility to avoid duplication #8412

### LINKED ISSUE #8412: 3.0 - EDD_Payment - getting meta, then updating meta, can result in duplicate fees being saved
## Bug Report

I ran into a weird scenario while testing EDD Invoices. When viewing an invoice for an order that contained a fee, each refresh was inserting a new fee into the database. This was caused by EDD Invoices fetching payment meta (which included fee data), then immediately updating the meta again. This caused the fees to get updated too, but an error in the setup_fees() logic meant that entirely new fees were being added.

I _think_ this is the correct set of circumstances/steps:

1. Have an order that contains both an `order` fee and an `order_item` fee.
2. Both of the `type_key` values should be `null`.
3. View the invoice for that order.
4. New fees will be added to the database.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8412 

Proposed Changes:
1. `Order::get_fees()` no longer sets the array key. It's not actually needed in new 3.0 code, so I've removed this and we'll only implement this in 2.9 backwards compat layers (`EDD_Payment`).
2. Removes all usage of `download_id` and `price_id` order adjustment meta. These are only set when the adjustment type is `order_item`, and in that case they can be retrieved from the order item itself.
3. `EDD_Payment::setup_fees()` no longer loops over order items. This is where some duplication was coming from, because `Order::get_fees()` already includes the order item fees in the array.
4. `EDD_Payment::setup_fees()` - if we end up modifying the `type_key`, then actually update it in the DB to reflect that. Basically, in my reproduction steps, my `type_key` was getting modified in the setup. But then, when re-saving that meta, it was no longer coming up in this query: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L2366-L2373 (because in the DB the key was `null`, but in the array we passed back to the customer it was like `87_2` or something. So that query had no results, which means it fell into this `else` block to add a new fee: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L2386
This also has me rethinking the fact that `type_key` is null for all new fees in 3.0. I'm thinking we should just set it anyway. :thinking: 
5. (update) Set `type_key` value for adjustments in manually created orders.

### DIFF
--- a/includes/orders/functions/actions.php
+++ b/includes/orders/functions/actions.php
@@ -318,7 +318,7 @@ function edd_add_manual_order( $args = array() ) {
 						}
 
 						$type_key = ! empty( $order_item_adjustment['description'] )
-							? sanitize_text_field( sanitize_title( $order_item_adjustment['description'] ) )
+							? sanitize_text_field( strtolower( sanitize_title( $order_item_adjustment['description'] ) ) )
 							: $index;
 
 						edd_add_order_adjustment( array(
@@ -352,7 +352,7 @@ function edd_add_manual_order( $args = array() ) {
 			}
 
 			$type_key = ! empty( $adjustment['description'] )
-				? sanitize_text_field( sanitize_title( $adjustment['description'] ) )
+				? sanitize_text_field( strtolower( sanitize_title( $adjustment['description'] ) ) )
 				: $index;
 
 			edd_add_order_adjustment( array(
--- a/includes/orders/functions/actions.php
+++ b/includes/orders/functions/actions.php
@@ -310,17 +310,22 @@ function edd_add_manual_order( $args = array() ) {
 				if ( isset( $download['adjustments'] ) ) {
 					$order_item_adjustments = array_reverse( $download['adjustments'] );
 
-					foreach ( $order_item_adjustments as $order_item_adjustment ) {
+					foreach ( $order_item_adjustments as $index => $order_item_adjustment ) {
 
 						// Discounts are not tracked at the Order Item level.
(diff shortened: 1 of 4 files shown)


robincornett (approved):
This is working for me. Another issue I noticed when testing 3.0 and viewing the invoice is that the `order_item` fee was being updated and changed to an `order` fee--that's no longer happening.

ashleyfae:
Updated to set the `type_key` for manually created orders. Although it's not needed in 3.0 code, I think having makes the backwards compatibility layer function better.

It's set to a slug version of the description, if that's set, otherwise falls back to the array key.

includes/orders/functions/actions.php
```diff
@@ -310,17 +310,22 @@ function edd_add_manual_order( $args = array() ) {
...
 
-					foreach ( $order_item_adjustments as $order_item_adjustment ) {
+					foreach ( $order_item_adjustments as $index => $order_item_adjustment ) {
 
 						// Discounts are not tracked at the Order Item level.
 						if ( 'discount' === $order_item_adjustment['type'] ) {
 							continue;
 						}
 
+						$type_key = ! empty( $order_item_adjustment['description'] )
```
robincornett, line 320: It is probably not necessary, but should the description as fallback also be cast to lowercase?

robincornett (approved):

PR #8424, opened 2021-02-22
Do not update past order emails when updating customer's primary email

### LINKED ISSUE #8420: 3.0 - Order emails are being updated when the customer email is changed
### Expected behavior
When I change the primary email address for my account, I expect that the existing orders will retain the email addresses used when the order was placed.

### Actual behavior
When I have more than one email in my account, and use the profile editor to change the primary email, all orders are updated to use that new email address.

1. Place an order.
2. Place a second order and use a different email address.
3. Check the database--each order will have a different email address associated with it in the edd_orders table.
4. Now, use the profile editor to change the primary email address for your account.
5. Refresh the database. Both orders will now be set to use the new primary address.
6. Use the profile editor to select the other email address.
7. Check the orders again.

PHP Version: 7.4.1
WordPress Version: 5.7-beta3

This is happening because of [this code](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/compat/class-customer.php#L184-L193). Removing those lines resolves the issue. The [`update_order_email_addresses` doc block specifically says that previous order email addresses are no longer changed](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/compat/class-customer.php#L184-L193), at least as I read it, and my understanding is that orders are intended to be a static snapshot.

robincornett:
Fixes #8420

### DIFF
--- a/includes/compat/class-customer.php
+++ b/includes/compat/class-customer.php
@@ -138,50 +138,6 @@ class Customer extends Base {
 		add_filter( 'get_user_metadata',    array( $this, 'get_user_meta'    ), 99, 4 );
 		add_filter( 'update_user_metadata', array( $this, 'update_user_meta' ), 99, 5 );
 		add_filter( 'add_user_metadata',    array( $this, 'update_user_meta' ), 99, 5 );
-
-		/** Actions **********************************************************/
-		add_action( 'profile_update', array( $this, 'update_customer_email_on_user_update' ), 10 );
-	}
-
-	/**
-	 * Updates the email address of a customer record when the email on a user is updated.
-	 *
-	 * @since 2.4.0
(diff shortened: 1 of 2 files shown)


ashleyfae (commented):
Confirmed this code fixes the problem.

@robincornett I find it strange that this code still lives in `compat/class-customer.php` . I know it's because it used to be in the `EDD_DB_Customers` class and this file is backwards compatibility for that class, but IMO this whole file should exist to power backwards compatibility logic.. it shouldn't be for "still in use" code.

How about we move this logic to `includes/customers/customer-actions.php` ?

ashleyfae (approved):

PR #8436, opened 2021-02-25
Update block editor content for pages added on initial activation

### LINKED ISSUE #8423: 3.0 - Page setup content should be Gutenberg compatible
### Expected behavior
I expect the page content to be Gutenberg compatible (blocks), instead of the "classic editor".

### Actual behavior
On an initial install of EDD, the pages are created for checkout, payment history, completion, etc. However, the way the content is inserted results in the pages falling back to displaying the "classic editor" block and introduces a few issues:

1. When blocks aren't used, you get an extra <p> tag on the checkout and other pages at the top.
2. The pages display the classic editor, which isn't intuitive as the shortcode and paragraph blocks should be used.

robincornett:
Fixes #8423

Proposed Changes:
1. If the block editor exists, updates the content for pages added on activation to use blocks.

To test:
Do a fresh install of EDD with this branch on a site running WP 5.0 or higher. The pages automatically created by EDD should use paragraph and shortcode blocks rather than the classic editor block.

### DIFF
--- a/includes/install.php
+++ b/includes/install.php
@@ -299,6 +299,9 @@ function edd_install_pages() {
 	// We'll only update settings on change
 	$changed  = false;
 
+	// Check if the block editor exists as that changes the content to add.
+	$block_editor = function_exists( 'register_block_type' );
+
 	// Loop through all pages, fix or create any missing ones
 	foreach ( array_flip( $pages ) as $page ) {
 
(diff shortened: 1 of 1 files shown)


spencerfinnell:
@robincornett I don't believe you need to do checks for the block editor. One of the reasons the block editor decided to delineate blocks with HTML comments is you get automatic fallback to "nothing" when nothing is parsing them. The classic editor should just ignore them on < 5.0.

robincornett:
Note: if I use the code to insert the block editor style content (eg the receipt shortcode) into a post type which doesn't support the classic editor, the shortcode renders but in the editor is wrapped in a paragraph tag:

```
<!-- wp:shortcode -->
<p>[edd_receipt]</p>
<!-- /wp:shortcode -->
```
It doesn't _seem_ to be an issue on the front end but worth nothing.

spencerfinnell:
@robincornett That could cause some styling issues for certain themes. If you remove the `\n` does the same thing happen?

robincornett:
Actually I take it back--the classic editor adds the paragraph tags on its own if you switch to the text editor. Looking at the unedited page in the database, there is not extra markup.

spencerfinnell:
@robincornett I still think you can remove the remaining `\n`. With those in place, for me, there is an extra line above the shortcode:

Screen Shot 2021-02-26 at 2 15 52 PM

Keeping everything in a single line should work because the included `<p>` tags will be correctly stripped by the Classic Editor when switching to the Text tab, but remain correctly in the block editor Paragraph block.

LisaCee (approved):

ashleyfae (approved):

PR #8448, opened 2021-03-04
Fix backwards compatibility for old customers DB properties #8434

### LINKED ISSUE #8434: 3.0 - EDD\Compat\Customer missing $table_name property
### Expected behavior

I expect `EDD\Compat\Customer` to support all old public properties for the `EDD_DB_Customers` class.

### Actual behavior

The `$table_name` property doesn't exist in the new class.

1) Execute `EDD()->customers->table_name`
2) Note PHP notice and table name not returned.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8434

Proposed Changes:
1. Add a magic getter for properties that no longer exist.

### DIFF
--- a/includes/compat/class-customer.php
+++ b/includes/compat/class-customer.php
@@ -10,6 +10,8 @@
  */
 namespace EDD\Compat;
 
+use EDD\Database\Table;
+
 // Exit if accessed directly
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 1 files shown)


robincornett (approved):
I don't know if it's ever going to be an issue but noting that the version number is very different (format) in 3.0 than 2.x (`202006101` vs `1.0`).

Otherwise the values returned are all consistent/identical.

ashleyfae:
> I don't know if it's ever going to be an issue but noting that the version number is very different (format) in 3.0 than 2.x (`202006101` vs `1.0`).
> 
> Otherwise the values returned are all consistent/identical.

Don't know that there's any way around that really. Hey, at least the new one is greater than the old one. :P Can't imagine many people are using that for anything anyway.

PR #8449, opened 2021-03-04
#8437 Prevent PHP warning if no email address

### LINKED ISSUE #8437: 3.0 - Physical Addresses table warning: Attempt to read property "email" on bool in easy-digital-downloads-release-3.0/includes/admin/customers/class-customer-addresses-table.php on line 219
### Expected behavior
I expect to not see PHP warnings on the physical address table. 

### Actual behavior
I see the above warning as per the title.

PHP Version: 8.0

EDD Version (or branch): 3.0-release

WordPress Version: 5.7

LisaCee:
Fixes #8437

Proposed Changes:
1. Change fallback for name to be address id.

### DIFF
--- a/includes/admin/customers/class-customer-addresses-table.php
+++ b/includes/admin/customers/class-customer-addresses-table.php
@@ -216,7 +216,7 @@ class EDD_Customer_Addresses_Table extends List_Table {
 	 */
 	public function column_cb( $item ) {
 		$customer = edd_get_customer_by( 'id', $item['customer_id'] );
-		$name     = isset( $customer->name ) ? $customer->name : 'Address ID:' . $item['id'];
+		$name     = isset( $customer->name ) ? $customer->name : 'Address ID: ' . $item['id'];
 		return sprintf(
 			'<input type="checkbox" name="%1$s[]" id="%1$s-%2$s" value="%2$s" /><label for="%1$s-%2$s" class="screen-reader-text">%3$s</label>',
 			/*$1%s*/ 'customer',
--- a/includes/admin/customers/class-customer-addresses-table.php
+++ b/includes/admin/customers/class-customer-addresses-table.php
@@ -210,18 +210,21 @@ class EDD_Customer_Addresses_Table extends List_Table {
 	 * @access public
 	 * @since 3.0
 	 *
-	 * @param array Address
+	 * @param array $item Address object.
 	 *
 	 * @return string Displays a checkbox
 	 */
 	public function column_cb( $item ) {
 		$customer = edd_get_customer_by( 'id', $item['customer_id'] );
-		$name     = isset( $customer->name ) ? $customer->name : 'Address ID: ' . $item['id'];
+		$name     = sprintf( __( 'Address ID: ' ) . $item['id'] );
+		if ( ! empty( $customer->name ) ) {
+			$name = $customer->name ? $customer->name : $customer->email;
+		}
 		return sprintf(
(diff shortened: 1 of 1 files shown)


includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -216,7 +216,7 @@ public function column_customer( $item ) {
 	 */
 	public function column_cb( $item ) {
 		$customer = edd_get_customer_by( 'id', $item['customer_id'] );
-		$name     = isset( $customer->name ) ? $customer->name : $customer->email;
+		$name     = isset( $customer ) ? $customer->name : $item['id'];
```
robincornett, line 219: I think you want to leave the `isset` as it was (checking for the name value) and then as I read the issue, I think Ashley was suggesting that the item ID fallback be prefaced with "address ID {ID number here}"? So then the screen reader text will either read "Select John Doe" (which I'm thinking now isn't as clear as it could be for an address) or "Select address ID ##". Maybe the label needs to be rethought as well.

LisaCee: More like this?

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -216,7 +216,7 @@ public function column_customer( $item ) {
 	 */
 	public function column_cb( $item ) {
 		$customer = edd_get_customer_by( 'id', $item['customer_id'] );
-		$name     = isset( $customer->name ) ? $customer->name : $customer->email;
+		$name     = isset( $customer->name ) ? $customer->name : 'Address ID: ' . $item['id'];
```
ashleyfae, line 219: Here's sort of what I was thinking:

Before this entire line, set `$name` to be the address ID (similar to what you've done with `'Address ID: ' . $item['id']`, though it needs to be translation-ready using `__()`, and let's use `sprintf()`).

Then, move the old `$name` assignment inside an if statement that checks `if ( ! empty( $customer ) ) {`

So that way, we have a default using the address ID at the very start, then we only use the customer name or email if the `$customer` variable is not empty.

While we're here: IMO the `isset( $customer->name )` check should be changed to be `! empty()` instead.

ashleyfae (changes requested):

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -210,18 +210,21 @@ public function column_customer( $item ) {
...
 	 *
-	 * @param array Address
+	 * @param array $item Address object.
 	 *
 	 * @return string Displays a checkbox
 	 */
 	public function column_cb( $item ) {
 		$customer = edd_get_customer_by( 'id', $item['customer_id'] );
-		$name     = isset( $customer->name ) ? $customer->name : $customer->email;
+		$name     = sprintf( __( 'Address ID: ' ) . $item['id'] );
```
robincornett, line 219: The text string is missing the placeholder to insert the ID (you won't concatenate with the `.`), and is missing the text domain.

includes/admin/customers/class-customer-addresses-table.php
```diff
@@ -210,18 +210,21 @@ public function column_customer( $item ) {
...
+		if ( ! empty( $customer->name ) ) {
+			$name = $customer->name ? $customer->name : $customer->email;
```
robincornett, line 221: I think this no longer needs to be a ternary, just the assignment since it is within a condition (and then the email can be removed from the comment) ... because if the `$customer->name` is not empty above then this check is redundant.

robincornett (changes requested):

robincornett (approved):

ashleyfae (approved):

PR #8467, opened 2021-03-10
Update filter bar to fix unneeded filters showing on Reports > Exports

### LINKED ISSUE #8095: 3.0 - Reports filter bar is showing on Exports screen, but not used
### Expected behavior
I expect visible controls on a screen to affect the output/behavior of that screen

### Actual behavior
When I visit Downloads > Reports > Exports (`wp-admin/edit.php?post_type=download&page=edd-reports&view=export`), the filter bar is visible above the export forms, but the filters there do not affect any of the export options/forms.

1) Visit Downloads > Reports > Exports.
2) Change anything on the filter bar.
3) For exports which can have modified dates, the date controls for each export are what manage them, not the filter bar.

PHP Version: 7.4.1
WordPress Version: 5.5.0

robincornett:
Fixes #8095

Proposed Changes:
1. Move the mobile link promo function from payments history to reports functions since that's where it's output now.
2. Refactor CSS...this got involved because the `#edd-filters .filter-items` div is always output whether or not there is anything in it. Originally, it had a margin applied to it, so it was always taking up space. Fixing that ended up being more involved than I anticipated.
3. Add an early return to the `filter_items` function if a registered report has errors. Currently, there is no way to have **no** filters for a report, so the export screen has the date and tax filters applied, even though they do nothing. The `export` report has no filters defined, but setting `filters` to an empty array or to false does not work. The export report does have errors, though, because no endpoints have been defined for it.

To test:
Check the filter bars on all reports screens, as well as the Orders table, and logs (under Tools).


includes/reports/reports-functions.php
```diff
@@ -1210,7 +1210,7 @@ function filter_items( $report = false ) {
 	$report_id = $report->get_id();
 
 	// Bail if no report
-	if ( empty( $report_id ) ) {
+	if ( empty( $report_id ) || $report->has_errors() ) {
```
ashleyfae, line 1213: I'm kind of inclined to remove this change in light of the new issue we've created: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8468 That feels like the better fix.

LisaCee (commented):
This works well for me on Reports > Exports, but I checked all of the reports screens and I did not get a filter bar under Taxes or Discounts either.  First I realized that I didn't have any taxes or discounts on this local site, but even after adding and using some, I am still not seeing the filter bar.  The EDD debug log says there is no endpoint for these:
```2021-3-15 18:00:10 - EDD\Utils\Exception: The 'discount_usage_chart' reports endpoint does not exist.
2021-3-15 18:02:11 - EDD\Utils\Exception: The 'total_tax_collected_for_location' reports endpoint does not exist.```

I checked the same screens in 3.0 and it seems like the filter bars there do seem to work and have logical values to filter.

robincornett:
Sorry, @LisaCee, that was the `has_errors` check in the reports that I was supposed to remove--I did but forgot to push up my changes. Merged the latest 3.0 and this branch should be current and correct now, with the error check removed.

ashleyfae (approved):

LisaCee (approved):

PR #8469, opened 2021-03-11
Master master into 3.0

ashleyfae:
Proposed Changes:
1. Merge master into 3.0 and fix conflicts.


ashleyfae:
@spencerfinnell Can you double check everything with Stripe and the build process is okay?

includes/admin/class-edd-notices.php
```diff
@@ -132,6 +132,7 @@ public function add_notices() {
 
 		// User can manage the entire shop
 		if ( current_user_can( 'manage_shop_settings' ) ) {
+			$this->add_requirements_notices();
```
robincornett, line 135: I know this is part of the merge but should we open an issue to remove it in 3.0?

includes/class-edd-license-handler.php
```diff
@@ -5,7 +5,7 @@
  * This class should simplify the process of adding license information
  * to new EDD extensions.
  *
- * @version 3.0
+ * @version 1.1
```
robincornett, line 8: Does it matter that this has changed down to 1.1?

robincornett (commented):
Running unit tests locally and these two are failing with this branch:
- `Tests_Query_Filters::test_edd_block_attachments_no_parent_bail`
- `Tests_Query_Filters::test_edd_block_attachments_no_download_bail`

They are not failing in `release/3.0`. It looks like in this branch that [`$files` is `false` here](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/merge/master-3.0/includes/query-filters.php#L44), which is breaking `wp_list_pluck` on the next line. But I'm not seeing anything obvious that would be affecting it in this branch that isn't in release/3.0.

spencerfinnell (approved):
Pulling in Stripe still works as expected.

tests/tests-query-filters.php
```diff
@@ -37,7 +37,7 @@ public function test_edd_block_attachments_no_attachment_bail() {
 	public function test_edd_block_attachments_no_parent_bail() {
 
 		// Prepare test
-		$filename 		= '../assets/images/loading.gif';
+		$filename 		= EDD_PLUGIN_DIR . '/assets/images/loading.gif';
```
robincornett, line 40: Looking into my failing unit tests more, I'm not seeing this change in `master` or in `release/3.0`, only in this merge branch. Changing it back to the relative path resolves the erroring tests, but I don't know why these two are erroring and the other four are not.

includes/class-edd-license-handler.php
ashleyfae: It was 1.1 in master. Looks like it had been changed to 3.0 in release/3.0. I kind of decided on the spot that the version was meant to designate the version of the class -- not the version of current EDD -- and no major changes had been made to really designate a massive version jump. :shrug:

tests/tests-query-filters.php
ashleyfae: I changed it because without this change, unit tests were actually failing for me. :thinking: I'd also seen it failing in Buildkite and thought this might fix it. Example of a failure: https://buildkite.com/sandhills-development-llc/easy-digital-downloads/builds/339#6269c2a1-3913-4385-91a6-3a2be152d4bb

I don't know why those suddenly started failing when they were fine before. :shrug:

includes/admin/class-edd-notices.php
ashleyfae: I'm fine with just removing it in this branch. LOL I don't know why I even added it. Guess I wasn't really thinking about what I was doing.

tests/tests-query-filters.php
ashleyfae: I'll remove this change and we can move it to a new issue https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8473

robincornett (approved):

ashleyfae:
Note: failing unit tests are just https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8473

PR #8471, opened 2021-03-11
Fix date parameters for report exports

### LINKED ISSUE #8466: 3.0 - Reports' export by date filter is broken
### Expected behavior
When I set a date boundary for an export (eg "Export Orders"), I expect the resulting CSV to have just orders within the date boundaries.

### Actual behavior
All orders (etc) are exported.

1. Go to Downloads > Reports > Export.
2. Export orders, limited to a specific date range.
3. Open the CSV and see that all orders have been exported.

PHP Version: 8
WordPress Version: 5.7

It looks like we broke the date filters when we did the work to make the date range inputs accessible. The export tools which use them are looking for just [`start` and `end` in the request](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/export/class-batch-export-payments.php#L281-L285), but the `name` attribute for the end date in exporting orders is now `orders-export-end`. So either we need to update this method (and any others which have the date range), or the markup needs to be adjusted to keep the `name` as `start` and `end`, while leaving the ID unique for accessibility/labels.

[Reported by a customer.](https://secure.helpscout.net/conversation/1446398148/179384?folderId=2048992)

robincornett:
Fixes #8466

Proposed Changes:
1. Update start/end checks on the `$request` for all exports to match the `name` attributes of the inputs.
2. Remove unused start/end properties on customer export.


ashleyfae (approved):

LisaCee (commented):
This works for honoring the date filter, but I found an unexpected difference between this and 3.0.  In 3.0 if I leave the "To" field blank inside of Export Orders and Export File Download Logs I receive a report with data up to the current date (it is not honoring the date parameters, but it still makes a report). In this issue, though, I receive a message that there is "no data found for export parameters".   I would expect to be able to leave either From or To fields and still get a report (if there is data).

robincornett:
Okay, I need to check this more tomorrow before I commit anything, but both the start and end dates are set to `''` if a date is not selected. That's fine for the start date, because the date query for both uses this code to set the date:

```php
date( 'Y-m-d 00:00:00', strtotime( $this->start ) )
```

Which, with an empty string, results in: `1970-01-01 00:00:00`

Which is fine for a start date, but is going to have logical issues for the end date. I need to see if that should be set to an empty string if it's empty, or just to today's date--just not sure without testing which will work.

robincornett:
Note: I created a helper function in the base export class to generate the date query because of the repetition.

Also discovered that the `orderby` parameter for getting orders was not correct, so that's been fixed.

LisaCee (approved):
I love that helper function.  It makes the rest of the code look so clean.

ashleyfae:
## Tested

- [x] Omitting start, included end.
- [ ] Including start, omitting end.
- [ ] Including start, including end.
- [ ] Omitting start, omitting end.

ashleyfae (approved):

padraigobeirn:
Thanks a million Lisa. I'm sorry I didn't spot that PR. Looking forward to the next 3.0 beta.

PR #8472, opened 2021-03-11
Allow reports to have no filters

### LINKED ISSUE #8468: 3.0 - It is not possible to create a report with no filters
### Expected behavior
I expect to be able to create a report with no filters in the filter bar.

### Actual behavior
When I use the `add_report` method, even if I set the `filters` parameter to `false` or `array()`, the default date/taxes filters are still added. Encountered while working on #8095. Exports are registered as a report, but they're not a true report, so the date/taxes filters do nothing and ideally should not display.

1. Update the [parameters for the export report](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports.php#L2526-L2532) to include `filters` set to `false` or an empty array.
2. Visit Downloads > Reports > Export. The filter(s) for date/taxes show anyway.

PHP Version: 8
WordPress Version: 5.7

robincornett:
Fixes #8468

Proposed Changes:
1. Set the default filters property to an empty array.
2. Allow filters to be empty.

To test: add `'filters' => false,` to the [array of attributes for the export report](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/reporting/reports.php#L2526-L2532). You should still have Export available under Reports, and no filters should show.


includes/reports/data/class-report-registry.php
```diff
@@ -119,7 +119,7 @@ public function add_report( $report_id, $attributes ) {
 		$attributes = array_merge( $defaults, $attributes );
 
 		try {
-			$this->validate_attributes( $attributes, $report_id );
+			$this->validate_attributes( $attributes, $report_id, array( 'filters' ) );
```
ashleyfae, line 122: Can you add a comment above that just says filters are excluded because we allow it to be empty?

ashleyfae (changes requested):
This works to remove filters, but it's revealing that several of our reports are not specifying `dates` as a filter, and as a result, that's disappeared.

We'll need to explicitly set it for all reports that should have it.

robincornett:
Updated with the comment, and added `dates` to all reports that should have it and have something other than the default filters set. Removed `filters` parameter from customers and refunds so they would end up with the default (dates and taxes).

includes/admin/reporting/reports.php
```diff
@@ -2529,6 +2527,7 @@ function edd_register_export_report( $reports ) {
 		    'priority'         => 1000,
 		    'capability'       => 'export_shop_reports',
 		    'display_callback' => 'display_export_report',
+			'filters'          => false,
```
ashleyfae, line 2530: Some wonky spacing here. Looks like it's because the previous arguments use spaces in a place they shouldn't be. Can you adjust that?

ashleyfae (commented):
Noted one tiny formatting issue, but otherwise this looks good!

ashleyfae (approved):

LisaCee (approved):

ashleyfae:
@robincornett Looks like the unit tests need updating. This isn't an issue with your code, but the tests were written to expect `dates` to be the default. We should rework them (and rename them) to expect an empty array instead. Do you feel comfortable digging into that?

ashleyfae:
@robincornett Looks like there's one more failure:

```
There was 1 failure:
--
  |  
  | 1) EDD\Reports\Data\Report_Tests::test_get_filters_with_a_valid_non_dates_filter_should_still_include_dates
  | Failed asserting that an array contains 'dates'.
  |  
  | /tmp/xdeTmbBt0tnJhZiPKzP01AKR7ndptVwW/tests/reports/data/tests-report.php:162
```

ashleyfae (approved):

Issue #8473, opened 2021-03-11
Attachment unit tests suddenly failing


ashleyfae:
## Bug Report

Some of our attachment-related unit tests are suddenly sporadically failing. For example: https://buildkite.com/sandhills-development-llc/easy-digital-downloads/builds/339#8a0a10a5-23b7-42cb-af04-c58af76dea20

```
There were 4 errors:

1) Tests_Query_Filters::test_edd_block_attachments_no_parent_bail

getimagesize(../assets/images/loading.gif): failed to open stream: No such file or directory

/tmp/wordpress/wp-includes/media.php:4985
/tmp/wordpress/wp-admin/includes/image.php:917
/tmp/wordpress/wp-admin/includes/image.php:482
/tmp/PvF7Jx5jvhhLuSD6If66N8W5EYihzFJq/tests/tests-query-filters.php:53


2) Tests_Query_Filters::test_edd_block_attachments_no_download_bail

getimagesize(../assets/images/loading.gif): failed to open stream: No such file or directory

/tmp/wordpress/wp-includes/media.php:4985
/tmp/wordpress/wp-admin/includes/image.php:917
/tmp/wordpress/wp-admin/includes/image.php:482
/tmp/PvF7Jx5jvhhLuSD6If66N8W5EYihzFJq/tests/tests-query-filters.php:95


3) Tests_Query_Filters::test_edd_block_attachments_not_restricted_bail

getimagesize(../assets/images/loading.gif): failed to open stream: No such file or directory

/tmp/wordpress/wp-includes/media.php:4985
/tmp/wordpress/wp-admin/includes/image.php:917
/tmp/wordpress/wp-admin/includes/image.php:482
/tmp/PvF7Jx5jvhhLuSD6If66N8W5EYihzFJq/tests/tests-query-filters.php:157


4) Tests_Query_Filters::test_edd_block_attachments_die

getimagesize(../assets/images/loading.gif): failed to open stream: No such file or directory

/tmp/wordpress/wp-includes/media.php:4985
/tmp/wordpress/wp-admin/includes/image.php:917
/tmp/wordpress/wp-admin/includes/image.php:482
/tmp/PvF7Jx5jvhhLuSD6If66N8W5EYihzFJq/tests/tests-query-filters.php:219
```

I'm not quite sure why because:

1. Those files _do_ exist.
2. The unit tests will pass in some environments, but randomly(?) not a couple others. So it's not failing all the time.
[...]

PR #8474, opened 2021-03-11
#8026 Confirmation to Permanently Delete Order

### LINKED ISSUE #8026: 3.0 - Add JavaScript confirmation when permanently deleting an order

When _permanently deleting_ an order, we should add a JavaScript confirmation.

### Justification or use case

To prevent accidental permanent deletions.

---

Related discussion regarding permanent deletes here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7937

EDD 2.9 has a confirmation.

LisaCee:
Fixes #8026 

Proposed Changes:
1. Add JS confirmation to permanently delete order
2. Import JS script [comment](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8026#issuecomment-795773730)
3. Change wording in message to reflect 3.0 Order object


robincornett (approved):
I was confused by why you were removing the other JS function for the advanced filters, but did some digging and realized it is still in the JS elsewhere. And the button still works. Permanently deleting an order triggers the alert and then successfully deletes when I confirm; not deleted if I hit cancel.

LisaCee:
> I was confused by why you were removing the other JS function for the advanced filters, but did some digging and realized it is still in the JS elsewhere. 

It turns out that this JS file wasn't loaded at all before.  It took me a while to figure out why none of my changes were working.  :joy_cat:

includes/scripts.php
```diff
@@ -466,7 +466,7 @@ function edd_localize_admin_scripts() {
 		'add_new_download'            => __( 'Add New Download', 'easy-digital-downloads' ),
 		'use_this_file'               => __( 'Use This File', 'easy-digital-downloads' ),
 		'quick_edit_warning'          => __( 'Sorry, not available for variable priced products.', 'easy-digital-downloads' ),
-		'delete_payment'              => __( 'Are you sure you want to delete this payment?', 'easy-digital-downloads' ),
+		'delete_order'              => __( 'Are you sure you want to delete this order?', 'easy-digital-downloads' ),
```
spencerfinnell, line 469: The `delete_payment` key should remain the same in case a 3rd party script happens to be referencing it. Although unlikely, it's easy to keep the same. The actual string itself is fine to change.

assets/js/admin/orders/list-table.js
```diff
@@ -7,10 +7,11 @@ import { jQueryReady } from 'utils/jquery.js';
 
 jQueryReady( () => {
 
-	$( '.edd-advanced-filters-button' ).on( 'click', function( e ) {
-		e.preventDefault();
-
-		$( '#edd-advanced-filters' ).toggleClass( 'open' );
-	} );
+	$( '.download_page_edd-payment-history .row-actions .delete a' ).on( 'click', function() {
+		if( confirm( edd_vars.delete_order ) ) {
```
robincornett, line 11: If the key is being reverted to `delete_payment` does this need to be changed also?

LisaCee: 🤦‍♀️ Yep.

robincornett (approved):

ashleyfae (approved):
Nice job @LisaCee ! Sorry about the file confusion.

Note: Failing unit tests are just this issue: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8473

spencerfinnell (approved):

PR #8485, opened 2021-03-17
#8480 3.0 Manual order status change notes

### LINKED ISSUE #8480: 3.0 - Manual order status changes get 2 notes
### Expected behavior

When I change the status of an order, I expect _one_ order note to be logged.

### Actual behavior

Two are logged:

Screenshot from 2021-03-17 12-58-05

This is due to us firing the `edd_update_payment_status` hook twice: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L1993-L2000

The status change notes are added on that hook.

I propose we update our notes logic to use our new transition hook instead: `edd_transition_order_status`.

1) Edit a `Complete` payment.
2) Manually change the status to something else.
3) Check the order notes.

PHP Version:

WordPress Version:

LisaCee:
Fixes #8480 

Proposed Changes:
1. Deprecate `edd_record_status_change()` and replace with `edd_record_order_status_change()` 
2. Hook new function into `edd_transition_order_status`
3. Update test with new function name

### DIFF
--- a/includes/payments/actions.php
+++ b/includes/payments/actions.php
@@ -253,26 +253,31 @@ function edd_process_after_payment_actions( $payment_id = 0, $force = false ) {
 add_action( 'edd_after_payment_scheduled_actions', 'edd_process_after_payment_actions', 10, 1 );
 
 /**
- * Record payment status change
+ * Record order status change
  *
- * @since 1.4.3
- * @param int $payment_id the ID number of the payment
- * @param string $new_status the status of the payment, probably "publish"
- * @param string $old_status the status of the payment prior to being marked as "complete", probably "pending"
+ * @since 3.0
+ * @param string $old_status the status of the order prior to being marked as "complete", probably "pending".
+ * @param string $new_status the status of the order, probably "publish".
+ * @param int    $order_id the ID number of the order.
  * @return void
  */
-function edd_record_status_change( $payment_id, $new_status, $old_status ) {
+function edd_record_order_status_change( $old_status, $new_status, $order_id ) {
 
-	// Get the list of statuses so that status in the payment note can be translated
+	// Get the list of statuses so that status in the payment note can be translated.
 	$stati      = edd_get_payment_statuses();
 	$old_status = isset( $stati[ $old_status ] ) ? $stati[ $old_status ] : $old_status;
 	$new_status = isset( $stati[ $new_status ] ) ? $stati[ $new_status ] : $new_status;
 
-	$status_change = sprintf( __( 'Status changed from %s to %s', 'easy-digital-downloads' ), $old_status, $new_status );
+	$status_change = sprintf(
+		/* translators: %1$s Old order status. %2$s New order status. */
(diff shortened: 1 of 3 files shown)


includes/payments/actions.php
```diff
@@ -253,26 +253,31 @@ function edd_process_after_payment_actions( $payment_id = 0, $force = false ) {
...
 /**
- * Record payment status change
+ * Record order status change
  *
- * @since 1.4.3
- * @param int $payment_id the ID number of the payment
- * @param string $new_status the status of the payment, probably "publish"
- * @param string $old_status the status of the payment prior to being marked as "complete", probably "pending"
+ * @since 3.0
+ * @param string $old_status the status of the order prior to being marked as "complete", probably "pending".
```
ashleyfae, line 259:
```suggestion
 * @param string $old_status the status of the order prior to this change.
```

It could be a change from any status to any status, so let's ditch the "probably" wording.

includes/payments/actions.php
```diff
@@ -253,26 +253,31 @@ function edd_process_after_payment_actions( $payment_id = 0, $force = false ) {
...
+ * @param string $new_status the status of the order, probably "publish".
```
ashleyfae, line 260:
```suggestion
 * @param string $new_status The new order status.
```

Better wording and `publish` is no longer used as a status.

tests/tests-filters.php
```diff
@@ -259,7 +259,7 @@ public function test_comments_clauses() {
 	public function test_edd_update_payment_status() {
 		global $wp_filter;
 		$this->assertArrayHasKey( 'edd_complete_purchase', $wp_filter['edd_update_payment_status'][100] );
-		$this->assertArrayHasKey( 'edd_record_status_change', $wp_filter['edd_update_payment_status'][100] );
+		$this->assertArrayHasKey( 'edd_record_order_status_change', $wp_filter['edd_update_payment_status'][100] );
```
ashleyfae, line 262: Nice job finding this. :raised_hands:

ashleyfae (changes requested):
Code looks perfect. :raised_hands: Just left a couple comments about the PhpDocs that we should update.

tests/tests-filters.php
ashleyfae: Oh one minor thing: that second argument needs to be updated too:

```suggestion
		$this->assertArrayHasKey( 'edd_record_order_status_change', $wp_filter['edd_transition_order_status'][100] );
```

ashleyfae (approved):
Working perfectly!

Screenshot from 2021-03-18 16-32-08

robincornett (approved):
Everything looks great and is working correctly. Just noting that a note is now being generated which wasn't originally:

image

Orders I've previously created are not showing that initial "Status changed from new to Pending" note.

ashleyfae:
I might actually kind of like that new addition.

PR #8492, opened 2021-03-19
Use edd_get_payment_status_label() for payment statuses where possible #8491

### LINKED ISSUE #8491: 3.0 - Custom payment statuses not included in edd_get_status_label() & thus can be displayed incorrecly
### Expected behavior

When I use the `edd_payment_statuses` filter to register a custom payment status, I expect my status name to be used whenever the status is displayed.

### Actual behavior

The "key" for the status is displayed on the admin orders page:

Screenshot from 2021-03-19 11-01-42

This is because that screen uses the new `edd_get_status_label()` function, which does not call the `edd_payment_statuses` filter.

1) Have at least one order that uses a custom status. You can use Fraud Monitor + the `pending_review` status, or Recurring and the `edd_subscription` status.
2) View the admin orders table and look at the status filters at the top.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8491

Proposed Changes:
1. Fetches a status from `edd_get_payment_status_label()` as a first try. This ensures that _payment_ statuses always have a chance to use the `edd_payment_statuses` filter to use a custom status name.

To test:

Follow steps listed in issue, but this time, the status _display name_ should be shown instead of the key.

Screenshot from 2021-03-19 11-13-03

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1515,21 +1515,13 @@ function edd_negate_int( $value = 0 ) {
  * @return string Label for the status
  */
 function edd_get_status_label( $status = '' ) {
-	static $labels = null;
 
-	// Array of status labels
-	if ( null === $labels ) {
+	// If this is a payment label, fetch from `edd_get_payment_status_label()`.
+	if ( array_key_exists( $status, edd_get_payment_statuses() ) ) {
+		$status_label = edd_get_payment_status_label( $status );
+	} else {
+		// Otherwise, fetch from generic array. This covers all other non-payment statuses.
 		$labels = array(
-
-			// Payments
-			'processing'         => __( 'Processing', 'easy-digital-downloads' ),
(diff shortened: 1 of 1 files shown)


robincornett (commented):
This totally works and looks so much better!

However, I was thinking about this and wondering if there would be value in creating new helper functions for the other two kinds of statuses, something like `edd_get_discount_statuses` and `edd_get_common_statuses`, and then using them as needed? Just thinking that it's possible that those lists may need to be added to as well? Then the `$labels` here could merge those three arrays into one and go with more of the original check.

ashleyfae:
I'm inclined to keep it as it is for now and multiple functions is something we can look into at a later date. I'm not yet sure if that's a route I want to go, but it's very easy to move from what we have now to multiple functions later without backwards compatibility concerns.

robincornett (approved):

PR #8500, opened 2021-03-22
Add 'Product Categories' report filter & add filter on all filters #8470

### LINKED ISSUE #8470: 3.0 - Add report filter for "Product Categories" and/or filter hook on the list of filters
## Enhancement Request

EDD Cross-sell / Upsell has a custom report and builds its own product (download) category filter to use with that report. But there is no product category filter in 3.0, and because there's no filter on `EDD\Reports\get_filters()` it cannot add its own.

We should add support for "Product Categories".
It might also be worth adding a filter on the output of that function.

https://github.com/easydigitaldownloads/edd-cross-sell-upsell/issues/32#issuecomment-793978883

ashleyfae:
Fixes #8470 

Proposed Changes:
1. Add `edd_report_filters` filter so add-ons can add their own filters.
2. Add 'Product Categories' filter.

Use these add-ons to test:

1. `edd_report_filters` - https://github.com/easydigitaldownloads/edd-campaign-tracker/pull/9
2. Product Categories - https://github.com/easydigitaldownloads/edd-cross-sell-upsell/pull/39

### DIFF
--- a/includes/reports/reports-functions.php
+++ b/includes/reports/reports-functions.php
@@ -318,34 +318,38 @@ function get_filters() {
 	}
 
 	$filters = array(
-		'dates'     => array(
-			'label'            => __( 'Date', 'easy-digital-downloads' ),
-			'display_callback' => __NAMESPACE__ . '\\display_dates_filter'
-		),
-		'products'  => array(
-			'label'            => __( 'Products', 'easy-digital-downloads' ),
-			'display_callback' => __NAMESPACE__ . '\\display_products_filter'
-		),
-		'taxes'     => array(
-			'label'            => __( 'Exclude Taxes', 'easy-digital-downloads' ),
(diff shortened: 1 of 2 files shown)


includes/reports/reports-functions.php
```diff
@@ -1029,6 +1034,19 @@ function display_products_filter() {
...
 }
 
+/**
+ * Handles display of the 'Products Dropdown' filter for reports.
+ *
+ * @since 3.0
+ */
+function display_product_categories_filter() {
+	?>
+	<span class="edd-graph-filter-optiosn graph-option-selection">
```
robincornett, line 1044:
```suggestion
	<span class="edd-graph-filter-options graph-option-selection">
```

robincornett (approved):

PR #8509, opened 2021-03-23
Reduce round trips to calculating tax rate #8455

### LINKED ISSUE #8455: 3.0 - Retrieving cart total has become a slow process due to tax rate logic
## Bug Report

See this issue on EDD Wallet for a real example: https://github.com/easydigitaldownloads/edd-wallet/issues/65

EDD Wallet uses `edd_get_cart_total()` a few times, and in EDD 3.0 that suddenly results in thousands of extra queries. The main cause is the `edd_get_tax_rate()` function, which is used when determining the tax applicable to the cart. This can run multiple times throughout a single page load, which results in running the same queries to the database.

ashleyfae:
Fixes #8455

Proposed Changes:
1. Adds a new `$tax_rate` parameter to the `edd_calculate_tax()` function. This essentially allows you to pass in a pre-determined tax rate to bypass the `edd_get_tax_rate()` call.
    1. The `EDD_Cart` class now utilizes this. At the start of the setup process, the tax rate is determined once and stored in a property. This tax rate is then passed through via the new parameter each time `edd_calculate_tax()` is called. This reduces the number of trips to `edd_get_tax_rate()`.
2. In `edd_get_tax_rate()` a new `$edd_determined_tax_rates` is introduced. This stores all determined tax rates, keyed by country + region. If a country + region is available in that global, it's returned immediately. This reduces the number of times that the address lookups have to be done within a single request.

This needs extensive testing with everything we can think of that manipulates tax rates.

- EDD Core: switching addresses during checkout to hit a different tax rate region.
- Barn2's VAT plugin.
- TaxJar

@LisaCee and @robincornett , I'm requesting your reviews, but I've slated this for _post internal_ so please do not review it immediately. This is something to do a bit later, non-urgently.


ashleyfae:
Intestingly, most of the failing unit tests are around _global tax rates_ which were removed in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7511

robincornett:
I just did a quick check out of curiosity and on a site with taxes enabled and only the store gateway active, I've got this on the initial page load:
- `release/3.0`: 359 queries
- `issue/8455`: 65 queries

The tax is being applied as the default country/state is the one for the store.

includes/tax-functions.php
```diff
@@ -155,10 +155,19 @@ function edd_active_tax_rates_query_clauses( $clauses ) {
...
- * @return mixed|void
+ * @return float
  */
 function edd_get_tax_rate( $country = '', $region = '', $fallback = true ) {
 
+	// This global is used to avoid excessive DB lookups per request.
+	global $edd_determined_tax_rates;
+
+	// First we only check for non-fallback options, because fallback logic happens later on.
+	$tax_rate_key = sanitize_key( strtolower( $country . '_' . $region ) );
```
robincornett, line 166: I'm reading through the logic of this function and it's confusing me completely on a few points:
- since all three parameters are optional, this `$tax_rate_key` _could_ end up being defined/used as `_`, which feels a little wonky. If the country and region are empty, should this bit be bypassed? Also I suppose you could end up with keys like `US_` if someone sets a coutry wide rate?
- I feel like line 172 is a little weird as we no longer use the option for the tax rate? At least we should not be?

includes/tax-functions.php
```diff
@@ -236,7 +251,12 @@ function edd_get_tax_rate( $country = '', $region = '', $fallback = true ) {
 	 * @param string $city           City.
 	 * @param string $zip            ZIP code.
 	 */
-	return apply_filters( 'edd_tax_rate', $rate, $country, $region, $address_line_1, $address_line_2, $city, $zip );
+	$rate = apply_filters( 'edd_tax_rate', $rate, $country, $region, $address_line_1, $address_line_2, $city, $zip );
```
robincornett, line 254: I was asking this above but then had a second question about this filter so I'll put the two together. I'm curious as to why the address 1/2, city, and zip are being added to this filter, but they're only populated if there is `$_POST` data.

It would seem like in 3.0, adding the `$tax_rate` object as a filter parameter might be useful here? (I know a lot of this was already set and isn't part of this PR but I have questions.)

ashleyfae: I guess it's meant to be providing the full address used to determine the tax rate. :shrug: As you said, this part isn't new to the PR so I can only guess as much as you lol.

I'm fine with adjusting that but I think it should be a separate PR.

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8509/commits/ce1954617a9b854f45c427a0b53560d6f8596648 removes all the global variable stuff for caching tax rate calculations. With the internal cart "caching" in place, this extra bit had no further benefit and only added to the complexity unnecessarily.

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8509/commits/0d6abbbbe0c6b779870410ce35771c256e0d25d5 adds a `set_tax_rate()` method. I didn't really want to add this, but our unit tests needed it in order to break the caching.

The reason why I'm not particularly concerned about needing to edit the unit tests:

1. The unit tests were illustrating a weak point that I do not find particularly realistic (changing the tax rate on the fly within the same page load). In the real world, you can expect the rate to change if you change your country or something, and that still works. But within a single page load, your tax rate shouldn't be changing multiple times.
2. This is a major release anyway.

Provided unit tests all pass now, I think we can proceed with this. (And more thorough testing.)

robincornett:
Tested so far:
- [x] taxable purchase, EDD core
- [x] tax-free purchase, EDD core (different state)
- [x] toggling back and forth between taxable location and non-taxable location, EDD Core
- [x] EU VAT (Barn2): France, 20%
- [x] EU VAT (Barn2): Hungary, 27%
- [ ] EU VAT (Barn2): the VAT number(s) I have are not validating (error message says "The country or VAT number is invalid.")
- [x] EDD Wallet

robincornett (commented):
Tested with Barn2's VAT plugin and a German VAT number. No tax was collected; however, the cart did not update to show the tax being removed, even though it seems like it was trying?

image

Here's what the purchase confirmation shows:
image

ashleyfae:
@robincornett Thanks for checking that. Is the lack of DOM update the same behaviour as in `release/3.0` right now or different? Just want to confirm if this is a _new_ problem from this PR.

robincornett:
The cart does update with release/3.0:
image

They're running a filter on `edd_cart_tax` (adding the "VAT reversed charged" text, not changing the amount) and on `edd_tax_rate`, which should be returning `0.00` when the VAT is reverse charged.

robincornett:
When they run the VAT check via AJAX, they have a `$response` array like this:
```php
		$response = array(
			'html'             => $cart,
			'tax_raw'          => edd_get_cart_tax(),
			'tax'              => html_entity_decode( edd_cart_tax( false ), ENT_COMPAT, 'UTF-8' ),
			'tax_rate_raw'     => edd_get_tax_rate(),
			'tax_rate'         => html_entity_decode( edd_get_formatted_tax_rate(), ENT_COMPAT, 'UTF-8' ),
			'total'            => html_entity_decode( edd_cart_total( false ), ENT_COMPAT, 'UTF-8' ),
			'total_raw'        => edd_get_cart_total(),
			'vat_check_result' => $this->get_vat_check_result(),
		);
```
Logging that array, here are the tax results:
```
    [tax_raw] => 3.80
    [tax] => [VAT reverse charged] $3.80
    [tax_rate_raw] => 0
    [tax_rate] => 0%
    [total] => $23.80
    [total_raw] => 23.8
```
So the rate is correct, but the calculated amount is not.

ashleyfae:
Thanks Robin, I'll look into it a bit more. Might be that my preferred solution is to contact them with a "fix", but I'll take a look.

ashleyfae:
I think I'm just going to recommend they add this to their `Checkout_Handler::ajax_vat_check()` method:

```php
// Wipe tax rate cache so that it gets recalculated.
if ( method_exists( EDD()->cart, 'set_tax_rate' ) ) {
	EDD()->cart->set_tax_rate( null );
}
```

robincornett:
Is that something EDD could add in using the `edd_before_checkout_cart` hook? Because that runs in the `edd_checkout_cart` function, and every ajax function I can think of (EDD, this, Discounts Pro, Gateway Fees, Simple Shipping) all run that function. So we could do, maybe:

```php
add_action( 'edd_before_checkout_cart', function() {
	EDD()->cart->set_tax_rate( null );
} );
```
A quick test shows that this does fix the reverse charge issue. It seems to add one more DB query to the checkout screen, but going from 39 to 40 seems like a fair price to pay.

robincornett:
Tested today with TaxJar and it is getting the correct tax rates when an ajax request calls. I think there is a bug with that extension, though, as it's not calculating a rate initially from the logged in user's saved address. So what I'm experiencing as a logged in user is that no tax is showing in checkout, but what's charged does include tax.

I can replicate this outside of this branch, though, so it's not unique here; I think this is a bug in the TaxJar extension.

ashleyfae:
> Is that something EDD could add in using the `edd_before_checkout_cart` hook? Because that runs in the `edd_checkout_cart` function, and every ajax function I can think of (EDD, this, Discounts Pro, Gateway Fees, Simple Shipping) all run that function. So we could do, maybe:
> 
> ```
> add_action( 'edd_before_checkout_cart', function() {
> 	EDD()->cart->set_tax_rate( null );
> } );
> ```
> 
> A quick test shows that this does fix the reverse charge issue. It seems to add one more DB query to the checkout screen, but going from 39 to 40 seems like a fair price to pay.

The first time I added this, I was seeing unexpected results. But then I tested again and they went away... so that makes me nervous LOL. But I can't make it happen again so maybe it's fine. :thinking: Bloody weird. (The weird stuff I saw was that my tax amount was incorrectly $0.00 when it shouldn't have been.)

robincornett (approved):
- [x] reverse charging VAT removes tax
- [x] tax rate shows for existing address in cart widget (Themedd)
- [x] tax rate shows for existing address when checkout loads
- [x] VAT tax rate updates when country changes, correct amount is charged
- [x] tax rate changes on address change with TaxJar
- [x] tax rate updates to the correct rate for a non-logged in user with TaxJar
- [x] tax rate updates with EDD core functions

It's all feeling well behaved for me so far. I feel like I want to approve it and live with it.

ashleyfae:
> I feel like I want to approve it and live with it.

Can that be our motto?

PR #8525, opened 2021-03-29
Fix fatal error when adding a discount to an order after a fee

### LINKED ISSUE #8164: 3.0 - Manual Orders - Fatal error when adding discount to an order that contains a fee
### Expected behavior

I expect to be able to add a discount code to an order that contains a fee.

### Actual behavior

When manually creating an order, I get a fatal error when trying to apply a discount to an order that already contains a fee. The following error is triggered:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_type() on bool in /var/www/edd3/wp-content/plugins/easy-digital-downloads/includes/ajax-functions.php:1453
```

1) Manually create an order via the admin area.
2) Add a download to the order.
3) Apply an adjustment to the order. I'm using: Type: Fee; Apply to: Entire order; Amount: $50. But I do not think the settings particularly matter.
4) Then attempt to add a discount code. Fatal error is triggered after selecting the discount from the dropdown and clicking "Add Discount".

PHP Version: 7.4

EDD Version (or branch): Using branch `issue/8159` due to the JavaScript error that's currently triggered when on `release/3.0`.

WordPress Version: 5.6-beta1-49280

robincornett:
Fixes #8164

Proposed Changes:
1. If a discount is not retrieved with `edd_get_discount`, continue to the next in the `foreach` loop.
2. Update the JS to send only discounts to the PHP. (Thank you @spencerfinnell )


LisaCee (approved):
I wasn't able to reproduce the original error, but this works as expected.

ashleyfae (approved):
Feel free to merge once unit tests pass.

PR #8527, opened 2021-03-30
Fix status label fallbacks

### LINKED ISSUE #8526: 3.0 - PHP warning for edd_get_payment_status for custom statuses
### Expected behavior
When I view an order in the admin, I expect to not have any PHP warnings.

### Actual behavior
When I view an order with a custom payment status which is no longer registered, I get this warning:

```
PHP Warning:  array_key_exists(): The first argument should be either a string or an integer in \easy-digital-downloads\includes\payments\functions.php on line 564
```

1. Activate Fraud Monitor.
2. Mark an order to be flagged for review (so that the status is `pending_review`.
3. Deactivate Fraud Monitor.
4. View the order in the admin.

PHP Version: 7.4.1
WordPress Version: 5.7

[The relevant code is here](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/functions.php#L563-L564) ... in this case, `$found_key` is `false`, so we can't use `array_key_exists` for that. I'm guessing we just need to add a check to line 564 against `$found_key`.

Related, but worth looking into--in this same instance, the `edd_get_payment_status_label` outputs "Pending_review" as the status where it is used, because it uses `ucwords` to capitalize the words of the status for the default value. No error is thrown, but I'm thinking we could make it look a tiny bit better by replacing the `_` in the status with a space before running `ucwords` so that the fallback status "label" would look a bit better?

robincornett:
Fixes #8526

Proposed Changes:
1. Make sure that `$found_key` has a value before using it in `edd_get_payment_status`.
2. Update the default/fallback label strings in `edd_get_payment_status_label` and `edd_get_status_label` to remove underscores before capitalizing the words.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1516,6 +1516,10 @@ function edd_negate_int( $value = 0 ) {
  */
 function edd_get_status_label( $status = '' ) {
 
+	// Set a default.
+	$status_label = str_replace( '_', ' ', $status );
+	$status_label = ucwords( $status_label );
+
 	// If this is a payment label, fetch from `edd_get_payment_status_label()`.
 	if ( array_key_exists( $status, edd_get_payment_statuses() ) ) {
 		$status_label = edd_get_payment_status_label( $status );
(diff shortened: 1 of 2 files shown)


LisaCee (approved):
I was able to replicate the original error.

PR #8530, opened 2021-03-30
Fix duplicate pages being created on activation

### LINKED ISSUE #8517: 3.0 - on activation, a new checkout page is created even if one already exists
### Expected behavior
When I activate EDD, I expect the pages I have set to use for checkout, etc. to not be duplicated and/or changed.

### Actual behavior
When I activate EDD 3.0, even though I already have all pages set, they're created again and the new pages are used, so I may end up with `mysite.com/checkout-2`, etc.

1. Start with EDD active (I don't think it matters if it is 2.10 or 3.0).
2. Make sure you have pages set for checkout, etc. on Downloads > Settings (this should be automatic).
3. Deactivate EDD.
4. Activate EDD 3.0.
5. Visit Pages and find the duplicated pages. Go through the purchase/checkout process and see that the new checkout page is being used.

PHP Version: 7.4.1
WordPress Version: 5.7

The pages are added in [`edd_install_pages`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/install.php#L273), which appears to be checking for the existence of the pages, but something is evidently getting missed.

robincornett:
Fixes #8517

Proposed Changes:
1. This primarily adjusts the logic of the check for whether a page was considered missing--`$missing_pages` has the `$page` as the array key, not the value, so I think this needed to be a different check.
2. I also set the `WP_Query` to return just IDs, so that the post object would only be retrieved if needed, and an array would not have to be reset.

To Test:
1. Activate EDD 3.0. Make sure that all of the required pages exist and have been set as options.
2. Deactivate EDD, then reactivate it.
3. Check and make sure no duplicate pages were created.
4. Optionally, delete one of the required pages, save the EDD settings (with a missing page), then deactivate and reactivate. The missing page should be created.

### DIFF
--- a/includes/install.php
+++ b/includes/install.php
@@ -288,10 +288,13 @@ function edd_install_pages() {
 	$pages_to_check = array_intersect_key( $current_options, $pages );
 
 	// Query for any existing pages
-	$posts = new WP_Query( array(
-		'include'   => array_values( $pages_to_check ),
-		'post_type' => 'page'
-	) );
+	$posts = new WP_Query(
+		array(
+			'include'   => array_values( $pages_to_check ),
+			'post_type' => 'page',
+			'fields'    => 'ids',
+		)
+	);
 
 	// Default value for checkout page
 	$checkout = 0;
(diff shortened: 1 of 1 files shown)


LisaCee (approved):

ashleyfae (approved):
- [x] Confirmed new pages are created when there are none.
- [x] Confirmed new pages are not created if they are already set.

PR #8531, opened 2021-03-31
Add exception handling to date parsing #8496

### LINKED ISSUE #8496: 3.0 - Fatal Error when upgrading DB, with different date format
## Bug Report

Error Log output: 
PHP Fatal error:  Uncaught Exception: DateTime::__construct(): Failed to parse time string (Januar 15, 2021 09:10:49) at position 0 (J): The timezone could not be found in the database in /src/wp-content/plugins/easy-digital-downloads/includes/libraries/Carbon.php:291
Stack trace:
#0 /src/wp-content/plugins/easy-digital-downloads/includes/libraries/Carbon.php(291): DateTime->__construct('Januar 15, 2021...', Object(DateTimeZone))
#1 /src/wp-content/plugins/easy-digital-downloads/includes/utils/class-date.php(37): Carbon\Carbon->__construct('Januar 15, 2021...', Object(DateTimeZone))
#2 /src/wp-content/plugins/easy-digital-downloads/includes/class-utilities.php(237): EDD\Utils\Date->__construct('Januar 15, 2021...', Object(DateTimeZone))

### Expected behavior
Upgrading databse without any issues.

### Actual behavior
Upgrade stops at 60.37% (of Customer Notes) when the fatal error occurs, databse can not be fully upgraded due the date time format of german orders made with EDD. 

Without uphrading the database further testing and updating other plugins is not possible.

1) have orders with different date time formats

PHP Version: 7.4.2

EDD Version (or branch):  3.0-beta1  

WordPress Version: 5.7

ashleyfae:
Fixes #8496

Proposed Changes:
1. Add `@throws` to docblocks to clarify methods can throw exceptions.
2. Add exception handling in date parsing and set fallback date to an empty string (BerlinDB should then take over and set it to be the current time).

To test:

1. Start on EDD `master` branch and have a customer.
2. Add a note to the customer.
3. Then manually edit the customer's record in the DB (`notes` column) to set the date to something that won't parse. For example:
`I am an invalid date! - This is my note`
Make sure you have at least one ` - ` (space, hyphen, space) in there because the migration code explodes on that.
4. Run the migration.
5. Ensure no errors. Ensure the note gets migrated with a `date_created` of today/now.

You can also change your site language prior to adding the note to get a "real" unparsable date, but I personally think editing the DB manually is a little easier and a better way to _guarantee_ a failure.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -130,9 +130,14 @@ class Data_Migrator {
 
 			if ( ! empty( $notes ) ) {
 				foreach ( $notes as $note ) {
-					$date = isset( $note[0] )
-						? EDD()->utils->date( $note[0], edd_get_timezone_id() )->setTimezone( 'UTC' )->toDateTimeString()
-						: '';
+					try {
+						$date = isset( $note[0] )
+							? EDD()->utils->date( $note[0], edd_get_timezone_id() )->setTimezone( 'UTC' )->toDateTimeString()
+							: '';
+					} catch ( \Exception $e ) {
+						// An empty date will be changed to current time in BerlinDB.
+						$date = '';
+					}
 
 					$note_content = isset( $note[1] )
 						? $note[1]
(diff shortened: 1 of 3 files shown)


robincornett (approved):
Added improper dates to an order and to the customer notes as instructed and confirmed the original issue/crash.

Switched to this branch and the order date has been migrated successfully and is correct; the note was migrated with today's date.

LisaCee (approved):
I was able to recreate the issue (my site just got stuck on notes) and this branch fixed it.

PR #8543, opened 2021-04-06
Replace "start_date" with "end_date" property when building the "before" part of the query inside of "remap_args" function

### LINKED ISSUE #8542: 3.0 - The end_date argument of edd_get_payments function is not working
### Expected behavior
When I send "start_date" and "end_date" arguments to the "edd_get_payments" function, I expect that the retrieved records have their respective date between these two input dates.

### Actual behavior
When I send "start_date" and "end_date" arguments to the "edd_get_payments" function, there are no retrieved records - unless there are some records that have their respective date equal to "start_date".

1) Call the "edd_get_payments" function, sending "start_date" and "end_date" arguments.
2) Check the return value of the function and you will notice that something is wrong.

PHP Version: 7.4.12

EDD Version (or branch): 3.0-beta1

WordPress Version: 5.7

It seems that the source code of "easy-digital-downloads/includes/payments/class-payments-query.php" file, is using "$this->start_date" instead of "$this->end_date" when building the "before" part of the query, inside the "remap_args" function. 

See below:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/c869629d383b736a5d9cde091cc87c2d0f1a887e/includes/payments/class-payments-query.php#L610.

napoleaowingert:
# Fixes

Proposed Changes:
1. Replace "start_date" with "end_date" property, when building the "before" part of the query inside of "remap_args" function.

This closes #8542.

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -609,7 +609,7 @@ class EDD_Payments_Query extends EDD_Stats {
 
 		if ( $this->args['end_date'] ) {
 			if ( is_numeric( $this->end_date ) ) {
-				$this->end_date = \Carbon\Carbon::createFromTimestamp( $this->start_date )->toDateTimeString();
+				$this->end_date = \Carbon\Carbon::createFromTimestamp( $this->end_date )->toDateTimeString();
 			}
 
 			$this->end_date = \Carbon\Carbon::parse( $this->end_date, edd_get_timezone_id() )->setTimezone( 'UTC' )->timestamp;


robincornett (approved):
Confirmed the original issue and the fix resolves it. Thank you @napoleaowingert for submitting the issue and the PR!

LisaCee (approved):
Good catch

ashleyfae (approved):
Nice one - thank you @napoleaowingert ! :raised_hands:

Issue #8539, opened 2021-04-01
3.0 - what should happen with downloads which have been partially refunded?


robincornett:
Here is a question about partial refunds: what should a partial refund _mean_ in terms of access to download files, etc.? Should the download files still be available? Right now, they're not, and things are not completely consistent. Here's some things I noticed:

- In the order details screen (admin), the "Copy Download Links" button is missing for a partially refunded download.
- In the customer view of the receipt/payment confirmation, the download file is not available.
- In the purchase history view, a fully refunded order says "Refunded" and there is no link to view the details/downloads (should a user still be able to see the details of a refunded payment? It seems like they should?).
- In the purchase history view, a partially refunded order looks like any other order and shows the links to view details/downloads, but when clicking that link, it shows that the order status is refunded, but not the status of the individual items (something to consider updating in #8508)

_Originally posted by @robincornett in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8284#issuecomment-811871124_

I posted the above in testing partial refunds, but wanted to move it to its own issue. With partial refunds, we need to decide what should still be available to customers--should the download files still be available or not?

Should a customer still be able to view the details of a refunded/partially refunded order? I feel like the answer there is yes to both, even if download files are not available? And I do think partially refunded items (or if one item in an order is refunded), the receipt/confirmation screen should show specifically which have been affected.

LisaCee:
> Should a customer still be able to view the details of a refunded/partially refunded order? I feel like the answer there is yes to both, even if download files are not available? And I do think partially refunded items (or if one item in an order is refunded), the receipt/confirmation screen should show specifically which have been affected.

I 100% agree with your suggestions.  
- A receipt should always be available (this may have more to do with the `[purchase_history]` than refunds)
- If an order is partially refunded, unaffected items should still have a link to the download
- If an order with quantities is partially refunded (1 of 2 qty refunded) they should still have a link to the download

As for partially refunded items, that's a bit more complicated.  Ideally, I'd like to give the store the option to have a download available or not, either store-wide or per order, but that sounds like it could be really challenging to add to our current refund system.  As a quicker fix, I tend to think we should err on the side of customer: if they paid for a download and it wasn't fully refunded, they should still have access.  It seems like that would catch more edge cases so that people don't accidentally lose access to something they should have.  I'm still thinking this through, though.

kerijacoby:
I'll try to answer some of the questions above from a support perspective :)

[...]

PR #8548, opened 2021-04-08
Update receipt template to use order functions/data

### LINKED ISSUE #8508: 3.0 - update receipt shortcode template to use order object
We should update the receipt shortcode template to use the EDD order object, etc. instead of the payment functions.

### Justification or use case
It will make some things (like showing applied discounts) easier, and give extensions an example to follow.

robincornett:
Fixes #8508

Proposed Changes:
1. Change the payment parameter on the `edd_payment_receipt_before_table` hook to use the `$order` object instead (this hook is new in 3.0, but the others have been in EDD).
2. Use the order data and methods as much as possible.
3. Update discounts output to be itemized because we can.
4. Allow partially refunded items to still be downloaded (per #8539).


templates/shortcode-receipt.php
```diff
@@ -77,62 +75,81 @@
...
 
-		<?php if ( edd_use_taxes() ) : ?>
+		<?php if ( $order->tax > 0 ) : ?>
 			<tr>
 				<td><strong><?php esc_html_e( 'Tax', 'easy-digital-downloads' ); ?>:</strong></td>
 				<td><?php echo esc_html( edd_payment_tax( $order->id ) ); ?></td>
 			</tr>
 		<?php endif; ?>
 
-		<?php if ( filter_var( $edd_receipt_args['price'], FILTER_VALIDATE_BOOLEAN ) ) : ?>
```
ashleyfae, line 125: Seems to me like we might need to keep this? It's from a shortcode argument.

robincornett:
Updated to replace existing hooks which require a `$payment` object:

- `edd_payment_receipt_before_table` => `edd_order_receipt_before_table`
- `edd_payment_receipt_before` => `edd_order_receipt_before`
- `edd_payment_receipt_after` => `edd_order_receipt_after`
- `edd_payment_receipt_after_table` => `edd_order_receipt_after_table`

Hooks which use the payment meta:

- `edd_receipt_files` => `edd_order_receipt_files`
- `edd_receipt_bundle_files` => `edd_order_receipt_bundle_files`
- - `edd_purchase_receipt_after_files` =>`edd_order_receipt_after_files`

On the first two, the new hooks take one fewer parameter--I did not port the `$meta` over as it seems like it would be just as easy to get from the `$order_id`, but I could put it back with `edd_get_order_meta`. On the last, I did a more radical change for the new hook and set it to accept two parameters, the `$item` and `$order` objects instead of getting just bits and pieces.

The other change I had to make to not have to retrieve the entire `$cart` was to update the `edd_receipt_show_download_files` function, and I'm not 100% sure about that. It's expecting an item as an array as it's set up in the payment cart details, but the code comment on that one suggests we are to be moving away from using it, and using the order items directly. But this is a function which has a filter which expects the array, so I think it has to be still used in there somehow.

robincornett:
Noting that originally I just changed the parameter for `edd_payment_receipt_before_table` to `$order` and didn't change the hook name. I would like to change the hook name for consistency and have added it to the deprecated functions, but since it's a new hook in 3.0 anyway and likely not being used, I'm wondering if I can just remove the deprecated functions implementation completely.

robincornett:
@LisaCee based on our discussion yesterday, I've updated the receipt with the following changes:
- The subtotal is now displayed first--immediately after the date.
- Then fees--all fees associated with the order, whether they are order level or order item.
- Then discounts, tax if it exists, and then the total.
- No fees are displayed at the end of the receipt (that is not currently working in 3.0 anyway, and now all fees are displayed up with the financial information).

Additionally, because showing the products required two nested `if` statements and products are the final thing in the template file, I put that in an early return because the nesting was overwhelming me.

includes/checkout/template.php
```diff
@@ -1099,5 +1099,23 @@ function edd_receipt_show_download_files( $item_id, $receipt_args, $item = array
...
 	}
 
-	return apply_filters( 'edd_receipt_show_download_files', $ret, $item_id, $receipt_args, $item );
+	if ( has_filter( 'edd_receipt_show_download_files' ) ) {
+		$cart = array();
+		if ( ! empty( $item->order_id ) ) {
+			$order = edd_get_order_by( 'id', $item->order_id );
+			$cart  = edd_get_payment_meta_cart_details( $order->id, true );
+		}
+		$ret = apply_filters( 'edd_receipt_show_download_files', $ret, $item_id, $receipt_args, $cart[ $item->cart_index ] );
```
ashleyfae, line 1108: I think we need to work in some backwards compatibility logic here. I see that you're now passing in an `Order_Item` object to `edd_receipt_show_download_files()`, but previously `$item` was an array. So someone using an old template could still be passing one in. We need to account for that.

So if `$item` is an array, then that should be passed in as the last argument to this hook. If it's an `Order_Item` object, then we have to run this backwards compat logic.

includes/checkout/template.php
```diff
@@ -1099,5 +1099,23 @@ function edd_receipt_show_download_files( $item_id, $receipt_args, $item = array
...
+	/**
+	 * Modifies whether the receipt should show download files.
+	 *
+	 * @since 3.0
+	 * @param bool                   $ret          True if the download files should be shown.
+	 * @param int                    $item_id      The download ID.
+	 * @param array                  $receipt_args Args specified in the [edd_receipt] shortcode.
+	 * @param \EDD\Orders\Order_Item $item        The order item object.
+	 */
+	return apply_filters( 'edd_order_receipt_show_download_files', $ret, $item_id, $receipt_args, $item );
```
ashleyfae, line 1120: Similar to what I said above, we need to account for `$item` being an array here instead of an `Order_Item` object. If it is an array then we need to turn it into an `Order_Item`.

includes/checkout/template.php
ashleyfae: Also, can you update the docblock for `edd_receipt_show_download_files()` to clarify that `$item` can now also be an `Order_Item` object?

ashleyfae:
I think we should put the discount row _before_ fees, because discounts do not apply to fees.

Screenshot from 2021-04-26 10-39-54

templates/shortcode-receipt.php
```diff
@@ -77,45 +71,67 @@
...
+					<?php
+					foreach ( $discounts as $discount ) {
+						$name   = $discount->description;
+						$amount = edd_currency_filter( edd_format_amount( edd_negate_amount( $discount->total ) ) );
+						if ( 'percent' === edd_get_discount_type( $discount->type_id ) ) {
+							$rate  = edd_format_discount_rate( 'percent', edd_get_discount_amount( $discount->type_id ) );
+							$name .= "&nbsp;({$rate})";
+						}
+						?>
+						<td><?php echo esc_html( $name . ': ' . $amount ); ?></td>
```
LisaCee, line 100: I'm not sure how to remedy this, but right now with multiple discounts the table cells are next to each other rather than stacked in 3.0.  
image

robincornett: That's because I apparently had the `<td>` inside of the discounts `foreach`. I went to fix it and ended up changing discounts and fees completely to match what Invoices does--a separate row for each discount/fee, with the label in the left column and the total in the right column. It feels easier to read to me with this instead of the `<ul>`.

LisaCee: Closer, but I'm still seeing the discounts side by side rather than stacked.  It shouldn't matter, but the 5% discount is from DCG.
image

LisaCee (approved):
I really like the flow of this - it makes more sense to me now.  I can't find a way to break it, so I guess I have to approve it.  :smile:

includes/orders/classes/class-order.php
```diff
@@ -384,8 +384,9 @@ public function get_adjustments() {
...
 		$discounts = array();
 
-		if ( empty( $this->adjustments ) ) {
-			return $discounts;
+		// Ensure adjustments exist.
+		if ( null === $this->adjustments ) {
+			$this->adjustments = $this->get_adjustments();
 		}
 
 		foreach ( $this->adjustments as $adjustment ) {
```
ashleyfae, line 392: This could just be refactored to do:

```suggestion
		foreach ( $this->get_adjustments() as $adjustment ) {
```

And remove the above `null` check. Might be a bit less repetitive.

includes/orders/classes/class-order.php
```diff
@@ -444,6 +445,35 @@ public function get_fees() {
...
+		// Default values
+		$credits = array();
+
+		// Ensure adjustments exist.
+		if ( null === $this->adjustments ) {
+			$this->adjustments = $this->get_adjustments();
+		}
+
+		// Fetch the fees that applied to the entire order.
+		foreach ( $this->adjustments as $adjustment ) {
```
ashleyfae, line 466: This could just be refactored to do:

```suggestion
		foreach ( $this->get_adjustments() as $adjustment ) {
```

And remove the above `null` check. Might be a bit less repetitive.

templates/shortcode-receipt.php
```diff
@@ -77,195 +71,263 @@
...
+		if ( filter_var( $edd_receipt_args['discount'], FILTER_VALIDATE_BOOLEAN ) ) :
+			$discounts = $order->get_discounts();
+			if ( $discounts ) :
+				$label = _n( 'Discount', 'Discounts', count( $discounts ), 'easy-digital-downloads' );
+				?>
+				<tr>
+					<td colspan="2"><strong><?php echo esc_html( $label ); ?>:</strong></td>
+				</tr>
+				<?php
+				foreach ( $discounts as $discount ) {
```
ashleyfae, line 93: Can you rename `$discount` to either `$order_discount` or `$order_adjustment`? I just think that will make it clear that `$discount` is NOT an `EDD_Discount` object.

ashleyfae:
## Testing

- [ ] This PR with new templates.
- [ ] This PR with templates from `release/3.0` branch (testing backwards compat)
- [ ] This PR with templates from `master` branch (testing backwards compat)

includes/checkout/template.php
```diff
@@ -1099,5 +1099,30 @@ function edd_receipt_show_download_files( $item_id, $receipt_args, $item = array
...
+			$order = edd_get_order_by( 'id', $order_item->order_id );
+			$cart  = edd_get_payment_meta_cart_details( $order->id, true );
+			$item  = $cart[ $item->cart_index ];
+		}
+		$ret = apply_filters( 'edd_receipt_show_download_files', $ret, $item_id, $receipt_args, $item );
+	}
+
+	// If the $order_item is an array, get the order item object instead.
+	if ( is_array( $order_item ) && ! empty( $order_item['id'] ) ) {
+		$order_item = edd_get_order_item( $order_item['id'] );
```
ashleyfae, line 1115: This isn't quite right. `$order_item['id']` is actually the product id. https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L3270

We might have to think about deliberately adding a new `order_item_id` property so we can reference it correctly.

If we do something like that, we also need to note to update this: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/functions.php#L883-L901

ashleyfae (commented):
Overall it's looking good. I suggested a few stylistic changes and found one backwards compatibility issue.

includes/checkout/template.php
ashleyfae: I also think we should get a unit test for this.

includes/checkout/template.php
```diff
@@ -1099,5 +1099,30 @@ function edd_receipt_show_download_files( $item_id, $receipt_args, $item = array
 		}
 	}
 
-	return apply_filters( 'edd_receipt_show_download_files', $ret, $item_id, $receipt_args, $item );
+	if ( has_filter( 'edd_receipt_show_download_files' ) ) {
+		$cart = array();
```
ashleyfae, line 1103: This isn't needed, as it's overwritten in the `if` block below, and isn't used outside of there.

tests/orders/tests-edd-payment.php
```diff
@@ -911,6 +911,26 @@ public function test_get_tax_rate_returns_decimal() {
 		$this->assertEquals( 0.2, $payment->tax_rate );
 	}
 
+	public function test_receipt_show_download_files() {
```
ashleyfae, line 914:
```suggestion
	/**
	 * This tests backwards compatibility in `edd_receipt_show_download_files()` when passing
	 * an array as the third parameter instead of the new `Order_Item` object. This test
	 * ensures that the function successfully converts that array to an order item object
	 * to be passed through to the filter `edd_order_receipt_show_download_files`.
	 *
	 * @covers \edd_receipt_show_download_files()
	 */
	public function test_receipt_show_download_files() {
```

Just to help remember what the test is for and exactly what it's covering.

ashleyfae, line 914:
```suggestion
	public function test_receipt_show_download_files_converts_array_to_order_item() {
```

Just a little more descriptive. :smile:

tests/orders/tests-edd-payment.php
```diff
@@ -911,6 +911,26 @@ public function test_get_tax_rate_returns_decimal() {
...
+		$receipt_args = array(
+			'id' => $payment->ID,
+		);
+		$cart         = edd_get_payment_meta_cart_details( $payment->ID, true );
+		$cart_item    = reset( $cart );
+		$download_id  = $cart_item['id'];
+
+		// Test sending a payment item array to the filter, which will convert it to an \EDD\Orders\Order_Item object.
+		add_filter( 'edd_order_receipt_show_download_files', function( $ret, $item_id, $order_receipt_args, $order_item_object ) use ( $cart_item ) {
+			$this->assertTrue( $order_item_object->id === $cart_item['order_item_id'] );
```
ashleyfae, line 925:
```suggestion
			$this->assertInstanceOf( '\\EDD\\Orders\\Order_Item', $order_item_object );
			$this->assertTrue( $order_item_object->id === $cart_item['order_item_id'] );
```

Just to add in one additional check. It's a little redundant, perhaps, but. :shrug:

ashleyfae (commented):
I'm super proud of you for getting that unit test working! :raised_hands: I made a few minor suggestions, but overall this is looking great!

ashleyfae (approved):

PR #8549, opened 2021-04-08
Update purchase_history template

### LINKED ISSUE #8539: 3.0 - what should happen with downloads which have been partially refunded?

robincornett:
Fixes #8539

Branched off of #8508 because of allowing access to partially refunded orders, and that PR controls access to download files based on individual order item status.

Proposed Changes:
1. Use the order functions where possible.
2. Allow user to view details for all sales (including fully refunded).
3. Deprecate the `purchase_history` action hooks (using `EDD_Payment`, etc) and use new ones with the order objects. Based off of deprecating some action hooks in the software licenses table--I just really wanted to do this once I saw the queries change.

If we dropped all of the `$payments` functions, we would see a huge difference in efficiency. For example, a purchase history screen with 15 orders to show:
- release/3.0: 289 queries (pagination is showing incorrectly)
- with this current PR, including use of `EDD_Payment` and `edd_get_users_purchases: 290 queries (no improvement, except for pagination correctly not showing)
- with the payment functions removed (the payment recovery functions do not appear to have an equivalent in orders which is maybe an issue): 38 queries


ashleyfae:
I've also been thinking about deprecating those hooks. Although it kind of sucks to do because they're so old and commonly used, the speed improvements are undeniable.

I _think_ I'm okay with it, but with a few changes.

1. I'd prefer an approach similar to this: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8406 In short, don't trigger the old hooks inside the template files, but have them execute from a separate file by hooking into the new actions. I just like removing that backwards compatibility code from the template itself for better clarity. I want to keep the templates as clear as possible for people who will be looking at and customizing them.
2. Using [this method](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8406/files#diff-ad34587cd389a79323aa6cf312db579295e08de4a974d5e1c6300bdcda2ea903R1144-R1148) to retrieve the payment objects might perform slightly better than `edd_get_users_purchases()`, but should have the same results.
3. I'm not sure I want to _officially_ deprecate those hooks yet (talking about `edd_do_action_deprecated()`). The reason being that it's quite late in the game for us to be deprecating and using those deprecated hooks is still the easiest way to be compatible with both 3.0+ and 2.x. I think I'm more inclined to not run them through `edd_do_action_deprecated()` yet, but maybe schedule to do so in 3.1 or something.

robincornett:
I think that all sounds totally reasonable and _love_ the idea of not having the old hooks/compatibility cluttering up the template. I would be happy to work on this more, but with that PR in draft and this one obviously conflicting with it, not sure how I should proceed on doing so.

ashleyfae:
I'm fine just closing the old PR and you can copy over some of the ideas from it.

Can you connect this new PR to that issue though? ( https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7759 ) Also incorporate the changes from `history-downloads.php` unless that's already covered by your other PR (haven't looked at that one yet).

templates/history-downloads.php
```diff
@@ -9,23 +9,27 @@
...
+	<p class="edd-account-verified edd_success">
+		<?php esc_html_e( 'Your account has been successfully verified!', 'easy-digital-downloads' ); ?>
+	</p>
+	<?php
 endif;
 /**
  * This template is used to display the download history of the current user.
  */
 $customer = edd_get_customer_by( 'user_id', get_current_user_id() );
+$page     = get_query_var( 'paged' ) ? get_query_var( 'paged' ) : 0;
```
ashleyfae, line 22:
```suggestion
$page     = get_query_var( 'paged' ) ? get_query_var( 'paged' ) : 1;
```

This makes the offset logic better, IMO. (see associated `edd_get_orders()` suggestion)

templates/history-downloads.php
```diff
@@ -9,23 +9,27 @@
...
-	$orders = edd_get_orders( array(
-		'customer_id' => $customer->id,
-		'number'      => 20,
-		'type'        => 'sale',
-	) );
+	$orders = edd_get_orders(
+		array(
+			'customer_id' => $customer->id,
+			'number'      => 20,
+			'offset'      => $page ? 20 * ( intval( $page ) - 1 ) : 0,
```
ashleyfae, line 29:
```suggestion
			'offset'      => 20 * ( intval( $page ) - 1 ),
```

When it's the first page it'll just be `20 * 0`, which is just `0` anyway.

templates/history-purchases.php
```diff
@@ -8,64 +8,119 @@
...
-		?>
-		<?php do_action( 'edd_after_purchase_history', $payments ); ?>
-		<?php wp_reset_postdata(); ?>
-	<?php else : ?>
-		<p class="edd-no-purchases"><?php _e('You have not made any purchases','easy-digital-downloads' ); ?></p>
-	<?php endif;
+if ( ! is_user_logged_in() ) {
+	return;
+}
+$page   = get_query_var( 'paged' ) ? get_query_var( 'paged' ) : 0;
```
ashleyfae, line 23:
```suggestion
$page   = get_query_var( 'paged' ) ? get_query_var( 'paged' ) : 1;
```

templates/history-purchases.php
```diff
@@ -8,64 +8,119 @@
...
+$orders = edd_get_orders(
+	array(
+		'user_id' => get_current_user_id(),
+		'number'  => 20,
+		'offset'  => $page ? 20 * ( intval( $page ) - 1 ) : 0,
```
ashleyfae, line 28:
```suggestion
		'offset'  => 20 * ( intval( $page ) - 1 ),
```

templates/history-purchases.php
```diff
@@ -8,64 +8,119 @@
...
+					<a href="<?php echo esc_url( add_query_arg( 'payment_key', $order->payment_key, edd_get_success_page_uri() ) ); ?>"><?php echo esc_html( $link_text ); ?></a>
+					<?php if ( ! in_array( $order->status, array( 'complete' ), true ) ) : ?>
+						| <span class="edd_purchase_status <?php echo esc_attr( $order->status ); ?>"><?php echo esc_html( edd_get_status_label( $order->status ) ); ?></span>
+					<?php endif; ?>
+					<?php if ( in_array( $order->status, array( 'pending', 'abandoned', 'failed' ), true ) ) : ?>
+						<?php
+						$url = add_query_arg(
+							array(
+								'edd_action' => 'recover_payment',
+								'payment_id' => $order->id,
```
ashleyfae, line 81:
```suggestion
								'payment_id' => urlencode( $order->id ),
```

templates/history-purchases.php
```diff
@@ -8,64 +8,119 @@
...
+		)
+	);
+	/**
+	 * Fires after the purchase history.
+	 *
+	 * @since 3.0
+	 * @param array $orders The array of the current user's orders.
+	 */
+	do_action( 'edd_after_order_history', $orders );
+	wp_reset_postdata();
```
ashleyfae, line 121: This is no longer needed, as we're not doing a post query.

ashleyfae (changes requested):

LisaCee:
> 3\. I'm not sure I want to _officially_ deprecate those hooks yet (talking about `edd_do_action_deprecated()`). The reason being that it's quite late in the game for us to be deprecating and using those deprecated hooks is still the easiest way to be compatible with both 3.0+ and 2.x. I think I'm more inclined to not run them through `edd_do_action_deprecated()` yet, but maybe schedule to do so in 3.1 or something.

So, is _unofficially_ deprecating these hooks putting them in the `deprecated_functions` file, but not calling `edd_do_action_deprecated()`?  They will still work, right?  Will users see a notice that they will be deprecated or is that what `edd_do_action_deprecated()` will do?  

Should we add a note explaining that they're not officially deprecated yet, but will be soon?

LisaCee (approved):
Happy to approve as this works really well for me.  The number of queries, size and time have all dropped for my local environment.  I love having the order details still available after a partial or full refund.  

Outside of this PR, what do you think about making invoices available for orders with refunds?  I'm thinking that, especially with partially refunded orders, that businesses may still want invoices for their records.

robincornett:
Can we get an up or down vote about the new file added in the refund hooks PR, because if that is approved, then this should be rebased off of that to use the new file for the deprecated hooks, or updated once that's merged. I think the same would be true for the order receipt PR as that's deprecating hooks also.

We do have an open issue for Invoices about being allowed to [show invoices for refunded orders](https://github.com/easydigitaldownloads/edd-invoices/issues/48).,,,I know I looked into it once but it felt a little complicated at the time--maybe because partial refunds were still pending, or because we have to account for 2.x; I forget which.

includes/deprecated-functions.php
```diff
@@ -1157,3 +1157,87 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
 
 	edd_insert_payment_note( $payment_id, $status_change );
 }
+
+/**
+ * Fires the edd_before_purchase_history hook in the purchase history, if needed.
+ *
+ * @deprecated 3.0
+ * @param array $orders The array of the current user's orders.
```
LisaCee, line 1165: Looking at PR #8556, do we want an `@todo Formally deprecate in EDD 3.1` for these hooks, too?

includes/deprecated-functions.php
```diff
@@ -1158,199 +1158,6 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
 	edd_insert_payment_note( $payment_id, $status_change );
 }
 
-/**
```
robincornett: @LisaCee I think in this last merge that some hook deprecations were lost. These should should still be in this PR and updated, please:
- `edd_before_order_history`
- `edd_order_history_row_start`
- `edd_order_history_row_end`
- `edd_after_order_history`

LisaCee: Sorry, I missed this comment.

includes/deprecated-hooks.php
```diff
@@ -159,3 +159,90 @@
...
+ * @deprecated 3.0
+ * @todo       Formally deprecate in EDD 3.1
+ * @param \EDD\Orders\Order $order The current order object.
+ */
+add_action( 'edd_order_history_row_start', function( \EDD\Orders\Order $order ) {
+	if ( ! has_action( 'edd_purchase_history_row_start' ) ) {
+		return;
+	}
+
+	$payment = edd_get_payment( $order->id );
```
ashleyfae, line 201: Let's bail if `$payment` doesn't exist. It can return `false`.

includes/deprecated-hooks.php
```diff
@@ -159,3 +159,90 @@
...
+ * @deprecated 3.0
+ * @todo       Formally deprecate in EDD 3.1
+ * @param \EDD\Orders\Order $order The current order object.
+ */
+add_action( 'edd_order_history_row_end', function( \EDD\Orders\Order $order ) {
+	if ( ! has_action( 'edd_purchase_history_row_end' ) ) {
+		return;
+	}
+
+	$payment = edd_get_payment( $order->id );
```
ashleyfae, line 218: Early return if `$payment` is `false`.

templates/history-downloads.php
```diff
@@ -116,14 +121,20 @@
...
 	<?php
-		echo edd_pagination(
-			array(
-				'type'  => 'download_history',
-				'total' => ceil( edd_count_purchases_of_customer() / 20 ) // 20 items per page
-			)
-		);
+	$count = edd_count_orders(
+		array(
+			'user_id' => get_current_user_id(),
```
ashleyfae, line 126: This should use the customer ID instead of the current user ID. Otherwise we query the sales above using the customer ID, but base pagination off the user ID. That could potentially lead to unexpected results.

templates/history-purchases.php
```diff
@@ -8,64 +8,86 @@
...
-						<?php if( ! in_array( $payment->status, array( 'complete', 'partially_refunded' ) ) ) : ?>
-							<span class="edd_purchase_status <?php echo $payment->status; ?>"><?php echo $payment->status_nicename; ?></span>
-							<?php if ( $payment->is_recoverable() ) : ?>
-								&mdash; <a href="<?php echo $payment->get_recovery_url(); ?>"><?php _e( 'Complete Purchase', 'easy-digital-downloads' ); ?></a>
-							<?php endif; ?>
-						<?php else: ?>
-							<a href="<?php echo esc_url( add_query_arg( 'payment_key', $payment->key, edd_get_success_page_uri() ) ); ?>"><?php _e( 'View Details and Downloads', 'easy-digital-downloads' ); ?></a>
-						<?php endif; ?>
-					</td>
-					<?php do_action( 'edd_purchase_history_row_end', $payment->ID, $payment->payment_meta ); ?>
```
ashleyfae, line 54: Missing new `edd_order_history_row_end` hook.

ashleyfae (changes requested):

templates/history-purchases.php
```diff
@@ -8,64 +8,88 @@
...
+	</p>
+	<?php
 endif;
 /**
  * This template is used to display the purchase history of the current user.
  */
-if ( is_user_logged_in() ):
-	$payments = edd_get_users_purchases( get_current_user_id(), 20, true, 'any' );
-	if ( $payments ) :
-		do_action( 'edd_before_purchase_history', $payments ); ?>
```
ashleyfae, line 23: Missing the `edd_before_order_history` replacement.

templates/history-purchases.php
```diff
@@ -8,64 +8,88 @@
...
-		</table>
-		<?php
-			echo edd_pagination(
-				array(
-					'type'  => 'purchase_history',
-					'total' => ceil( edd_count_purchases_of_customer() / 20 ) // 20 items per page
-				)
-			);
-		?>
-		<?php do_action( 'edd_after_purchase_history', $payments ); ?>
```
ashleyfae, line 66: Missing the `edd_after_order_history` replacement.

ashleyfae (changes requested):

ashleyfae:
## Testing

- [ ] Template with no add-ons works well; good number of queries.
- [ ] Add-ons using old filters still get their info added to templates.

ashleyfae (approved):

PR #8556, opened 2021-04-12
Add legacy pre/post refund hooks to edd_refund_order

### LINKED ISSUE #8540: 3.0 - Refund hooks have changed
In 2.x we have refund hooks in the payments class-- for example, `edd_post_refund_payment`. It appears that these hooks do not fire during the refund process in EDD 3.0. The comparable hook in EDD 3.0 is `edd_refund_order`, and it has different parameters, so the hooks are not really compatible.

I ran into this working on issue [1315 in Recurring](https://github.com/easydigitaldownloads/edd-recurring/issues/1315) and put together a solution, but wanted to ask, since core has been recommending the use of `edd_post_refund_payment` (and also pre, I think), and it's no longer part of the process, do we need to change anything in EDD 3.0, or just communicate the new hook(s) to use instead?

It looks like these extensions (of ours) use the [post refund payment hook](https://github.com/search?q=org%3Aeasydigitaldownloads+edd_post_refund_payment&type=code):
- Recurring
- Mailchimp
- Software Licensing
- Commissions

robincornett:
Fixes #8540

Marking this as a draft as it results in multiple failed unit tests and I'm not 100% sure how to handle those:

> There were 8 failures:
> 
> 1) EDD\Discounts\Tests_Discounts::test_is_used_case_insensitive
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 2) EDD\Orders\EDD_Payment_Tests::test_refund_payment
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 3) EDD\Orders\EDD_Payment_Tests::test_refund_payment_legacy
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 4) EDD\Orders\EDD_Payment_Tests::test_refund_affecting_stats
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 5) EDD\Orders\Refunds_Tests::test_refund_order
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 6) EDD\Orders\Trash_Tests::test_trash_order_with_refunds
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 7) Tests_Process_Download::test_refunded_item_in_partially_refunded_order_should_return_false
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528
> 
> 8) Tests_Process_Download::test_fully_refunded_order_should_return_false
> Unexpected deprecated notice for edd_pre_refund_payment
> Failed asserting that an array is empty.
> 
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:516
> /tmp/wordpress-tests-lib/includes/abstract-testcase.php:528

Note that the `edd_pre_refund_payment` hook is used in PayPal, and the post hook is not used anywhere in core.

Also noting that these hooks both technically fire _after_ the payment/order has been refunded, but `pre` is before the store stats have been updated, and `post` is afterwards.

### DIFF
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -1157,3 +1157,27 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
 
 	edd_insert_payment_note( $payment_id, $status_change );
 }
+
+/**
+ * Legacy pre-refund hook which fired after a payment status changed, but before store stats were updated.
+ *
+ * @deprecated 3.0
+ * @param int $order_id The original order id.
+ */
+add_action( 'edd_refund_order', function( $order_id ) {
+	if ( has_action( 'edd_pre_refund_payment' ) ) {
+		do_action( 'edd_pre_refund_payment', edd_get_payment( $order_id ) );
+	}
+} );
+
+/**
+ * Legacy post-refund hook which fired after a payment status changed and store stats were updated.
+ *
+ * @deprecated 3.0
+ * @param int $order_id The original order id.
+ */
+add_action( 'edd_refund_order', function( $order_id ) {
+	if ( has_action( 'edd_post_refund_payment' ) ) {
+		do_action( 'edd_post_refund_payment', edd_get_payment( $order_id ) );
+	}
+} );
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -1162,6 +1162,9 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
  * Legacy pre-refund hook which fired after a payment status changed, but before store stats were updated.
  *
  * @deprecated 3.0
+ * @todo       Formally deprecate in EDD 3.1
+ * @link       https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8574
+ *
  * @param int $order_id The original order id.
  */
 add_action( 'edd_refund_order', function( $order_id ) {
@@ -1174,6 +1177,9 @@ add_action( 'edd_refund_order', function( $order_id ) {
  * Legacy post-refund hook which fired after a payment status changed and store stats were updated.
  *
  * @deprecated 3.0
+ * @todo       Formally deprecate in EDD 3.1
+ * @link       https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8574
+ *
  * @param int $order_id The original order id.
  */
 add_action( 'edd_refund_order', function( $order_id ) {
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -1157,33 +1157,3 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
 
 	edd_insert_payment_note( $payment_id, $status_change );
 }
-
-/**
- * Legacy pre-refund hook which fired after a payment status changed, but before store stats were updated.
- *
- * @deprecated 3.0
- * @todo       Formally deprecate in EDD 3.1
- * @link       https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8574
- *
- * @param int $order_id The original order id.
- */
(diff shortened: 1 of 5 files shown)


ashleyfae:
Just noting for reference:

https://github.com/easydigitaldownloads/edd-recurring/pull/1428/files#diff-19d09afffb60ab64a95db87a16c61f304cd71e6ba4f6004ea747d7d98f7e49aeR1233

When EDD Recurring hooks into `edd_post_refund_payment` it checks `$payment->old_status`, which should be the status before the renewal update. We might need to explicitly set that on the payment object before passing through `$payment` to the hook.

ashleyfae:
The problem with the unit tests is that we are hooking into `edd_pre_refund_payment` in `edd_maybe_refund_paypal_purchase()`.

Let's open a separate issue to rework `edd_maybe_refund_paypal_purchase() to use the new hook instead`. We'll also have to add a checkbox to the UI. We'll get that merged first, and having that fix in place should get these unit tests passing again and we can merge this after.

robincornett:
Have updated to load pre/post refund payment hooks in the deprecated functions file. They're marked as deprecated, but will not trigger a notice at this time--the intent is to formally deprecate later, possibly in 3.1.

includes/deprecated-functions.php
```diff
@@ -1157,3 +1157,27 @@ function edd_record_status_change( $payment_id, $new_status, $old_status ) {
 
 	edd_insert_payment_note( $payment_id, $status_change );
 }
+
+/**
+ * Legacy pre-refund hook which fired after a payment status changed, but before store stats were updated.
+ *
```
ashleyfae, line 1163: Can you add to both docblocks something like:

```php
 * @todo Formally deprecate in EDD 3.1
```

ashleyfae: Maybe also include a `@link` to this issue: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8574

robincornett:
@ashleyfae a thought I had last night--what would you think of putting these, and the deprecated receipt/history hooks, into a new file called `deprecated-hooks` instead of this one? That way all of the hooks/closures are together, rather than lumped in with functions which can still be used.

LisaCee (approved):
Tested per Robin's instructions in Slack and everything worked as expected with no errors.

1. with SL master, purchase a renewable licensed download. (not recurring)
2. then purchase a renewal.
3. note the expiration date on the license, which should be twice as long as it was after step 1.
4. then refund the SECOND payment (the renewal), not the first
5. the expiration date should be rolled back to the original term

ashleyfae (approved):

PR #8557, opened 2021-04-12
Update discount legacy args

### LINKED ISSUE #8544: 3.0 - edd_add_discount parameters are inconsistent with 2.x
### Expected behavior
When I use a function which exists in both EDD 2.x and 3.0, I expect them to parse the same parameters.

### Actual behavior
When I pass `products` as an array key to `edd_store_discount`, it's ignored in EDD 3.0.

1. Easiest way is to install the Discount Code Generator extension and add new discounts using that.
2. Alternatively, use `edd_store_discount` with an array of parameters and generate the same discounts in 2.x and 3.0.
3. The key is to have `products` as a parameter with an array of download/products IDs.
4. In 2.x, the products will be added to the Download Requirements for the discount. In 3.0, they will not.

PHP Version: 7.4.1
WordPress Version: 5.7

In 2.x, the `edd_store_discount` function calls the `add` method in the discounts class and ends up parsing the discount parameters as meta values ([products](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-discount.php#L1351)).

In 3.0, `edd_add_discount` [uses `product_reqs` instead of `products`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/discount-functions.php#L26), so the code that works in 2.x to add download requirements no longer works in 3.0.

robincornett:
Fixes #8544

Proposed Changes:
1. Update the old key values in the `convert_legacy_args` method to match the args which were expected in EDD 2.x.
2. Adjust the legacy args scope to always set it to `global` or `not_global`, regardless of whether the parameter exists in the `$args`.
2. Update all use of `min_cart_price` property to `min_charge_amount` as that's the correct property.


ashleyfae (approved):
I ran `edd_store_discount()` manually in EDD 2.x, ensured the discount was added as expected. Then ran the same function with the same arguments in 3.0 to ensure the data was still added in the same way.

Also tested manually adding/editing via the UI due to the input name change.

Issue #8565, opened 2021-04-14
3.0 - update edd_maybe_refund_paypal_purchase to use new refund hook


robincornett:
### Explain your enhancement (please be detailed)
We need to change the PayPal refund function to use the new `edd_refund_order` hook.

### Justification or use case
Currently, the pre refund payment hook does not run at all. Adding it to the new hook, but deprecating it, is breaking unit tests.
> The problem with the unit tests is that we are hooking into `edd_pre_refund_payment` in `edd_maybe_refund_paypal_purchase()`.
> 
> Let's open a separate issue to rework `edd_maybe_refund_paypal_purchase() to use the new hook instead`. 
> 
> _Originally posted by @ashleyfae in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8556#issuecomment-819421325_

ashleyfae:
I'm going to handle this here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8564

PR #8567, opened 2021-04-14
Update PayPal refunds to use new modal/API #8564, #8565

### LINKED ISSUE #8564: 3.0 - PayPal Refunds - Add "Refund Payment in PayPal" checkbox to refund UI
We need to replicate this 2.x behaviour in 3.0:

Screenshot from 2021-04-14 11-45-26

We need to add a checkbox to the new refund UI for that option.

ashleyfae:
Fixes #8564, #8565

Proposed Changes:
1. Deprecates `edd_paypal_refund_admin_js()` and `edd_maybe_refund_paypal_purchase()`. That's the old JavaScript way of handling it.
2. Adds the "Refund Payment in PayPal" checkbox to the new refund UI.
3. Adds processing to actually process the refund.
    - This includes handling for partial refunds. If the refund amount in PayPal should match the amount being refunded via the UI.
    - Upon successful refund, adds a negative transaction to the `edd_order_transactions` table. This includes the refund transaction ID.

For testing, we should test things like:

- Process a refund, but DO NOT check the box to refund in PayPal. Ensure PayPal refund is not processed.
- Refund the entire order. Ensure entire payment is refunded in PayPal.
- Refund only part of the order. Ensure a partial refund is processed in PayPal.
    - Separately, then refund the rest of the order. Ensure the full amount is now refunded in PayPal.

Note: There are a couple UI quirks, but I'm going to open a separate issue to address those, since they're not PayPal-specific. This includes:

- Adding some margin between the table and the checkbox.
- Ensuring the checkbox disappears after the refund is processed.


robincornett (commented):
This worked well for me once I got all of the PayPal bits tied in. Just a couple of finicky tweaks please.

includes/gateways/paypal-standard.php
```diff
@@ -1122,105 +1124,112 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
...
-					$('label[for="edd-paypal-refund"]').remove();
-				}
-			});
-		});
-	</script>
+	<div class="edd-form-group edd-paypal-refund-transaction">
+		<div class="edd-form-group__control">
+			<input type="checkbox" id="edd-paypal-refund" name="edd-paypal-refund" value="1">
+			<label for="edd-paypal-refund">
+				<?php esc_html_e( 'Refund Payment in PayPal', 'easy-digital-downloads' ); ?>
```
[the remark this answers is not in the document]
ashleyfae: The reason I chose payment is I guess because I see an order as different from a payment. You can have an order without an associated payment attached.

Perhaps `transaction` is actually the word I'm looking for.

ashleyfae:
Also updated to include the refunded amount in the note.

includes/gateways/paypal-standard.php
```diff
@@ -1122,105 +1124,119 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
...
+ *                                                    amount is used as the transaction refund amount (used for
+ *                                                    partial refunds), and an EDD transaction record will be
+ *                                                    inserted.
+ *
  * @return void
  */
-function edd_refund_paypal_purchase( $payment ) {
+function edd_refund_paypal_purchase( $payment_id_or_object, $refund_object = null ) {
+	/*
+	 * Internally we want to work with an Order object, but we also need
```
robincornett, line 1223: This comment indicates that the function will be working with the order object, but it's still using the `$payment` object in the `$credentials` array and for the transaction ID (line 1258). Is that intentional?

includes/gateways/paypal-standard.php
```diff
@@ -1122,127 +1124,153 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
...
-				} else {
-					$('#edd-paypal-refund').remove();
-					$('label[for="edd-paypal-refund"]').remove();
-				}
-			});
-		});
-	</script>
+	<div class="edd-form-group edd-paypal-refund-transaction">
+		<div class="edd-form-group__control">
+			<input type="checkbox" id="edd-paypal-refund" name="edd-paypal-refund" value="1">
```
robincornett, line 1151:
```suggestion
			<input type="checkbox" id="edd-paypal-refund" name="edd-paypal-refund" class="edd-form-group__input" value="1">
```

includes/gateways/paypal-standard.php
```diff
@@ -1122,127 +1124,153 @@ function edd_paypal_link_transaction_id( $transaction_id, $payment_id ) {
...
+			<label for="edd-paypal-refund">
```
robincornett, line 1152:
```suggestion
			<label for="edd-paypal-refund" class="edd-form-group__label">
```

robincornett (commented):
Just two finicky class requests and I'm good. This is working really well!

robincornett (approved):

PR #8569, opened 2021-04-14
Add after refund table container, hide on success

### LINKED ISSUE #8568: 3.0 - Refunds Modal - Hide content added via hook after refund is processed
The refund modal has a hook `edd_after_submit_refund_table`. I think we should globally hide all that content after the refund is processed.

After implementing https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8564 ("Refund Payment in PayPal" checkbox), here's how it looks by default after the form is submitted:

Screenshot from 2021-04-14 14-00-58

The checkbox is still visible.

Although we could fix that individually for each item (@robincornett I noticed you had a good solution in Recurring, for example), I think we should have a single, global solution so that we don't have to repeat ourselves for each implementation.

robincornett:
Fixes #8568

Proposed Changes:
1. Wrap the `edd_after_submit_refund_table` hook and refund submit button in a `edd-submit-refund-actions` container which is hidden upon a successful refund.


ashleyfae:
This is perfect, however, I think we can also remove a bit of code now.

This can go: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8568/assets/js/admin/orders/order-overview/_refund.js#L241

Optionally, while we're here, I think we could turn this into CSS too? https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8568/assets/js/admin/orders/order-overview/_refund.js#L240

ashleyfae (approved):

PR #8577, opened 2021-04-21
#8576 Ensure adjustments exists

### LINKED ISSUE #8576: 3.0 - $order->get_discounts can return an empty array
### Expected behavior
When I have a proper `$order` from `edd_get_order()`, I expect that `$order->get_discounts()` will return the discounts applied to the order.

### Actual behavior
When I use `$order->get_discounts()` before otherwise getting the order adjustments, no discounts are returned. This is happening with the PR for #8508.

1. Check out `issue/8508`.
2. Create an order with a discount applied.
3. View the order receipt/confirmation. The discount does not show.

PHP Version: 7.4.1
EDD Version (or branch): release/3.0, but obvious in issue/8508
WordPress Version: 5.7

`get_discounts` [accesses the adjustments property directly](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-order.php#L387), but if they have not been retrieved already, no discounts are returned.

I believe the `get_discounts` method needs to be updated to use the `get_adjustments` method the same way the [`get_fees` method](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-order.php#L414-L417) does.

LisaCee:
Fixes #8576

Proposed Changes:
1. Ensure adjustments exists and then continue with discount logic

### DIFF
--- a/includes/orders/classes/class-order.php
+++ b/includes/orders/classes/class-order.php
@@ -384,8 +384,9 @@ class Order extends Rows\Order {
 	public function get_discounts() {
 		$discounts = array();
 
+		// Ensure adjustments exist.
 		if ( empty( $this->adjustments ) ) {
-			return $discounts;
+			$this->adjustments = $this->get_adjustments();
 		}
 
 		foreach ( $this->adjustments as $adjustment ) {
--- a/includes/orders/classes/class-order.php
+++ b/includes/orders/classes/class-order.php
@@ -385,7 +385,7 @@ class Order extends Rows\Order {
 		$discounts = array();
 
 		// Ensure adjustments exist.
-		if ( empty( $this->adjustments ) ) {
+		if ( null === $this->adjustments ) {
 			$this->adjustments = $this->get_adjustments();
 		}
 


includes/orders/classes/class-order.php
```diff
@@ -384,8 +384,9 @@ public function get_adjustments() {
 	public function get_discounts() {
 		$discounts = array();
 
+		// Ensure adjustments exist.
 		if ( empty( $this->adjustments ) ) {
```
robincornett, line 388: I think the check for this should change to check for `null`, as the fees method does--if the adjustments have not been retrieved yet, that's what it will be, and this will set it, but if they have been retrieved, but there are no adjustments, `$this->adjustments` could still be empty. It _may_ not make a difference (this is working for sure), but I think it would be good to be consistent.

LisaCee: Thanks.  I was going back and forth on this, but I couldn't decide which one was "right".

ashleyfae (approved):
Confirmed issue and fix. @robincornett please merge if you're happy with it.

robincornett (approved):

PR #8581, opened 2021-04-21
issue/8578 - Manual Orders - Match Frontend Subtotal Handling

### LINKED ISSUE #8578: 3.0 - order subtotals are not consistent between the admin and front end
### Expected behavior
When I view an order in the admin, I expect the order details to match what the customer sees in the cart and on their purchase receipt.

### Actual behavior
If a discount is applied to an order, the subtotal shown in the admin includes the discount, but that is not the actual order subtotal.

It's easiest to see this in action with taxes enabled, as subtotals do not show in the cart unless they are. So enable taxes and make sure the order has a tax applied.
1. Create an order in EDD 3.0 with a discount applied.
2. Note that the subtotal shows the amount of the items in the cart before the discount and tax, and the total shows the amount after the discount and tax are applied.
3. View the order in the admin. Note that the subtotal is a different amount.

Examples:
3 0-receipt

3 0-admin

PHP Version: 7.4.1
WordPress Version: 5.7

My feeling is that the admin JS needs to be updated here to use the actual subtotal value that's stored in the database, instead of recalculating it. [I think it's this section here.](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/assets/js/admin/orders/order-overview/models/state.js#L77-L92)

### LINKED ISSUE #8580: 3.0 - manually created orders can have an incorrect subtotal
### Expected behavior
When I manually create an order, I expect the order details to be consistent with an identical order placed from the front end.

### Actual behavior
When I manually create an order with a discount, and do an identical order from the front end, the order subtotals are not consistent.

Manual order:
image

Storefront order:
3 0-receipt

(I know there is tax on the second one and not the first, but the subtotals are where the issue is, sorry.)

1. Create a discount code.
2. Create an order in the front end and use the discount code.
3. Manually create an order with the same discount code (or amount).
4. Compare the two order rows in the database.
5. The order created from the storefront will not have the discount included in the subtotal. This behavior is consistent with EDD 2.10.
6. The manually created order will have the discount applied to the subtotal, making the subtotal incorrect.

PHP Version: 7.4.1
WordPress Version: 5.7

Related: #8578, but I think there is something going wrong here when the order is set up and added to the database.

spencerfinnell:
Fixes #8578 
Fixes #8580 

Proposed Changes:
1. Calculate subtotal only as the total item amounts.
2. Display subtotal above all order adjustments


includes/admin/views/tmpl-order-subtotal.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Order Overview: Subtotal
+ *
+ * @package     EDD
+ * @subpackage  Admin/Views
+ * @copyright   Copyright (c) 2020, Sandhills Development, LLC
```
ashleyfae, line 7:
```suggestion
 * @copyright   Copyright (c) 2021, Sandhills Development, LLC
```

:grin:

includes/admin/views/tmpl-order-subtotal.php
```diff
@@ -0,0 +1,23 @@
...
+ * @since       3.0
+ */
+?>
+
+<tr class="is-expanded">
+	<td></td>
+	<td colspan="{{ data.config.colspan }}" class="column-primary">
+		<?php esc_html_e( 'Subtotal', 'easy-digital-downloads' ); ?>
+	</td>
+	<td class="column-right" data-colname="<?php esc_html_e( 'Amount', 'easy-digital-downloads' ); ?>">
```
ashleyfae, line 18:
```suggestion
	<td class="column-right" data-colname="<?php esc_attr_e( 'Amount', 'easy-digital-downloads' ); ?>">
```

assets/js/admin/orders/order-overview/models/state.js
```diff
@@ -73,49 +73,12 @@ export const State = Backbone.Model.extend(
...
-					( amount, item ) => {
-						return amount += ( item.get( 'subtotal' ) - item.get( 'discount' ) );
-					},
-					0
-				);
-			}
-
-			// Add all subtotals.
-			const subtotal = items.reduce(
+			return items.reduce(
```
robincornett, line 76: I'm confused about why this is calculating the subtotal instead of using the order object `subtotal` similar to how `getTotal()` does (down at line 174).

spencerfinnell: @robincornett Good point. With the new way it's calculated we can add parity here.

robincornett (changes requested):
@spencerfinnell not sure where to track this down to, but something in backbone is breaking refunds when I view them in the admin:

```
Uncaught TypeError: Cannot read property 'replace' of undefined
    at Function.v.template (underscore.min.js?ver=1.8.3:2)
    at n.template (wp-util.js?ver=5.7.1:39)
    at n.render (wp-backbone.js?ver=5.7.1:541)
    at wp.Backbone.Subviews.<anonymous> (wp-backbone.js?ver=5.7.1:433)
    at underscore.min.js?ver=1.8.3:2
    at Function.v.each.v.forEach (underscore.min.js?ver=1.8.3:2)
    at wp.Backbone.Subviews._attach (wp-backbone.js?ver=5.7.1:429)
    at wp.Backbone.Subviews.set (wp-backbone.js?ver=5.7.1:176)
    at wp.Backbone.Subviews.add (wp-backbone.js?ver=5.7.1:219)
    at n.render (edd-admin-orders.js?ver=1619103480:1)
```

Viewing the same refund in release/3.0 is fine.

spencerfinnell:
@robincornett Pushed a fix.

assets/js/admin/orders/order-overview/models/state.js
```diff
@@ -177,7 +145,7 @@ export const State = Backbone.Model.extend(
 				return this.get( 'order' ).total;
 			}
 
-			return this.getSubtotal() + this.getTax();
+			return ( this.getSubtotal() - this.getDiscount() ) + this.getTax();
```
robincornett, line 148: Sorry to keep doing this one thing at a time. Currently, the subtotal is showing the combined amount of the products. Discounts are applied to modify the total.

However, other adjustments are not being calculated (fees/credits) into the total. If I go ahead and create the order, the fees and credits are displayed, but the order total is using what showed when I was creating the order (incorrect):

image

spencerfinnell: @robincornett I see, I think https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8581/files#diff-e8e2a669744f756eb93f2fbecb659dd2eb08b6662460386f46dbe43c6ff17cc5L101-L118 needs to be moved to the total calculation.

Question: are taxes calculated after all adjustments (discount, credit, fee) applied? So your screenshot should have a $3 total, and if taxes enabled a tax added to $3?

robincornett: I think that's complicated. It looks like the tax calculations are handled in the cart and then passed for each order item to `edd_build_order`, so I don't think tax calculations are handled there, just storage. In the cart, we have:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/6a09489eabefde58bf56e0981c139ba0c2677f23/includes/cart/class-edd-cart.php#L240-L258

So I think the math goes like this for each item:
1. item subtotal
2. plus (item) fees
3. minus discount
4. calculate and add tax
5. then subtract tax if taxes are included in the price

Then those are added together for the cart subtotal (which is different in the cart than it is in the order, I think, which is weird), and the total becomes:
1. subtotal of all items (before discounts)
2. get any order level fees
3. add tax to fees if they are taxed
4. subtotal, subtract discounts, add the calculated cart tax, and add the calculated fees (which may be taxed)

Fees can be taxed, or not, so that adds a layer of complexity.

Actually some of this is replicated in `edd_build_order`, at least outside of the individual item tax amount. My assumption is that this stage of the manual order, though, needs to follow the cart logic?

assets/js/admin/orders/order-overview/models/state.js
```diff
@@ -71,51 +71,19 @@ export const State = Backbone.Model.extend(
...
+				return this.get( 'order' ).subtotal;
 			}
 
-			// Add all subtotals.
-			const subtotal = items.reduce(
+			const { models: items } = this.get( 'items' );
+
+			return items.reduce(
 				( amount, item ) => {
 					return amount += +item.get( 'subtotal' );
```
LisaCee, line 83: Can you explain the `+item.get( 'subtotal' )` use here?  I've used reduce before in vanilla JS, but I've never seen this syntax.

assets/js/admin/orders/order-overview/models/state.js
robincornett: Also with credits being totally new in 3.0, I'm not sure on the answer to that. This might be a question for someone financial?

assets/js/admin/orders/order-overview/models/state.js
spencerfinnell: @LisaCee That ensures the value is a number when adding it to the previous value. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

assets/js/admin/orders/order-overview/models/state.js
spencerfinnell: > add tax to fees if they are taxed

For manual Order creation we don't have UI in place to facilitate the [`no_tax`](https://docs.easydigitaldownloads.com/article/543-edd-fees) argument normally used with preregistered fees. Do we need a checkbox that the fee amount shouldn't be taxed?

cc @ashleyfae

spencerfinnell:
I've updated the "Add Adjustment" flow to include a toggle to have a fee be taxed or untaxed:

Screen Shot 2021-04-27 at 1 40 53 PM
Screen Shot 2021-04-27 at 1 44 15 PM
Screen Shot 2021-04-27 at 1 41 14 PM

It also ensures the tax value of the fee is properly stored in the database.

I'm still not really sure how to handle Credits. Are we treating those as just one-off discounts? I don't know how to apply it on the frontend cart to have something to compare it against.

robincornett:
@spencerfinnell I don't think there is a front end equivalent for credits so my understanding is that it's new territory. I checked with @cklosowski and his understanding (to be confirmed, still) is that the order is this:

1. Order Items
2. Discount Codes
3. Fees
4. Credits
5. Taxes

So credits do affect the taxable amount, but are applied after everything else, and do not change any of the order item totals/values.

#8582 is something to keep in mind as well--I think that's a regression but if we change it then it needs to be changed for both admin and front end orders.

spencerfinnell:
I believe I have this working locally where a Credit is applied pre-tax calculation. 

However, I think there is an argument to be made to allow these to act as true arbitrary "_credit toward amount paid_" and keep it outside of tax calculation. I would say if they are applied pre-tax it makes more sense to just use a Discount which will have the same effect and not introduce a completely separate way to adjust the order amounts that essentially does the same thing.

ashleyfae:
> I believe I have this working locally where a Credit is applied pre-tax calculation.
> 
> However, I think there is an argument to be made to allow these to act as true arbitrary "_credit toward amount paid_" and keep it outside of tax calculation. I would say if they are applied pre-tax it makes more sense to just use a Discount which will have the same effect and not introduce a completely separate way to adjust the order amounts that essentially does the same thing.

I agree. I see credit as, _"I am using this to pay towards a total that has already been completely calculated"_.

If they want to affect the tax amount then they can use a discount or manually change the item amounts, which we also support.

Random source: https://smallbusiness.chron.com/gift-card-money-deducted-before-after-sales-tax-62827.html

> Tax is generally not charged when purchasing a gift card, but most likely will be charged on a purchase in which the gift card is used as a payment method. In general, tax will be added to a sale transaction before applying the credit of a gift card at purchase.

I kind of equate what we're doing here to a gift card.

robincornett:
Note: in `release/3.0`, credits do not affect the taxable amount. An order for $10 with a 9.25% tax is $10.93. Applying a $1 credit makes it $9.93. So factoring in credits as the final element would be consistent with what's already in the release branch.

spencerfinnell:
Latest update implements the following:

- Subtracts Credits from calculated total
- Moves Credits under the Total view
- Uses `Order Fee`, `Order Credit`, `Tax`, etc as the small descriptive text vs. primary when listing adjustments
- Code cleanup

Screen Shot 2021-04-28 at 9 43 50 AM

includes/orders/functions/actions.php
```diff
@@ -362,6 +362,7 @@ function edd_add_manual_order( $args = array() ) {
 				'type_key'    => $type_key,
 				'description' => sanitize_text_field( $adjustment['description'] ),
 				'subtotal'    => floatval( $adjustment['subtotal'] ),
+				'tax'         => floatval( $adjustment['tax'] ),
```
robincornett, line 365: `tax` also needs to be added for taxable fees in the `edd_add_order_adjustment` args beginning at line 324. This will fix the issue I'm currently encountering where the tax is not showing in the refund modal for the fee.

robincornett:
Note: order credits do not currently show on the purchase receipt (we can address this in #8508) or in EDD Invoices/PDF Invoices--we will need to open separate issues for those repositories.

assets/css/admin/style.scss
```diff
@@ -669,24 +669,30 @@ table.edd-order-overview-summary--refund {
 }
 
 /** Totals */
-.edd-order-overview-summary__totals .column-primary {
+.edd-order-overview-summary__subtotal .column-primary,
```
robincornett, line 672: I'm wondering if it would be tidier to make a common class which could be used for the subtotal/adjustments/total `tbody` so that we could have just one rule for these.

assets/css/admin/style.scss
```diff
@@ -669,24 +669,30 @@ table.edd-order-overview-summary--refund {
...
+.edd-order-overview-summary__tax tr:first-of-type .column-primary,
+.edd-order-overview-summary__total .column-primary {
 	font-weight: 600;
 }
 
-.edd-order-overview-summary__totals-total td {
+.edd-order-overview-summary__subtotal td,
```
robincornett, line 678: I know the vertical alignment is not new to this PR but I'm thinking I actually prefer it aligned top, except for the total row (because of the difference in font sizes).

includes/admin/views/tmpl-order-form-add-order-adjustment.php
```diff
@@ -97,6 +97,57 @@ class="edd-select"
...
+			'' === data.state.hasTax.country
+		) {
+		#>
+			<div class="notice notice-warning">
+				<p>
+					<strong><?php esc_html_e( 'No tax rate has been set.', 'easy-digital-downloads' ); ?></strong><br />
+					<?php esc_html_e( 'Tax rates are defined by the customer\'s billing address.', 'easy-digital-downloads' ); ?>
+				</p>
+				<p>
+					<button class="button" id="set-address">
```
robincornett, line 144:
```suggestion
					<button class="button button-secondary" id="set-address">
```

robincornett (commented):
So far, this is working well and I like the new order of information in the `__totals` section. Some fiddly adjustments which can be ignored if you like.

One UI thing I'm noticing today because my local site is running a bit slowly because another site is running a big batch job: the "waiting" UI is not consistent. When I select/add a download, for example, an old style EDD spinner is added to the left of the "add download" button. With a discount, there is a pause after clicking the button but before the modal closes, but no similar UI. Same with adding a fee/credit, but those add much faster than discounts, for some reason, so the lack of spinner is not noticeable.

I'm wondering if we could not use the spinner, maybe, and instead add/remove WP classes to the buttons--we are starting to do that elsewhere in EDD 3.0 (Reports > Export, for example):
- `updating-message` shows a spinner within the button (this may only be a good option with `button-primary` due to the spinner color though)
- `updated-message` shows a checkmark within the button

The UI requests maybe should just be their own issue though.

robincornett:
Removing a fee throws this JS error and the total is not recalculated:

```
Uncaught TypeError: Cannot read property 'forEach' of undefined
    at n.remove (edd-admin-orders.js?ver=1619626192:1)
    at p (backbone.min.js?ver=1.4.0:2)
    at f (backbone.min.js?ver=1.4.0:2)
    at c (backbone.min.js?ver=1.4.0:2)
    at n.trigger (backbone.min.js?ver=1.4.0:2)
    at n._onModelEvent (backbone.min.js?ver=1.4.0:2)
    at p (backbone.min.js?ver=1.4.0:2)
    at f (backbone.min.js?ver=1.4.0:2)
    at c (backbone.min.js?ver=1.4.0:2)
    at n.trigger (backbone.min.js?ver=1.4.0:2)
```

assets/css/admin/style.scss
spencerfinnell: Looking again I think I like having these separate and we can adjust some of the repetition when more properly converted to `.scss.`

includes/admin/views/tmpl-order-tax.php
```diff
@@ -0,0 +1,72 @@
...
+					<span class="screen-reader-text"><?php esc_html_e( 'Dismiss this notice.', 'easy-digital-downloads' ); ?></span>
+				</button>
+			</div>
+		</td>
+	</tr>
+	<# } #>
+
+<# } #>
+
+<input type="hidden" value="{{ data.tax }}" name="tax" />
```
robincornett, line 72: @spencerfinnell could we also add a hidden input containing the actual tax rate? I just opened #8591 but I think having the tax rate as a hidden input would help with that. I'm asking about it here because I don't know if it's more than just adding the hidden input--if that's all it is then it can be addressed in that issue once this one is sorted.

spencerfinnell: @robincornett Yep, we'd pass the rate/percentage via a hidden field the same way we are the total, so let's do it separately in #8591.

robincornett (commented):
I ran into a new issue which I don't know if it's truly new to this PR or I just magically happened to encounter it, as it may require just the right set of circumstances and rounding.

I have a tax rate of 9.25% because of a combined state and county tax rate, so it can result in some weird numbers. I created an order with one item, and a taxable fee, which resulted in a total charge of $63.91, because of rounding.

However, when taken apart, the refund modal is coming up with a total refund amount of $63.92, because when the tax is calculated on each item first, the rounding happens earlier, I think, and then the refund validation fails:

image

Looking in the database, for the $50 item, the tax is saved differently:
- `4.625000000` (manual)
- `4.630000000` (storefront)

For the fee:
- `0.786250000` (manual)
- `0.786250000` (storefront)

So the primary difference may just be that the order item tax is being rounded (improperly?) but something is a little sideways somewhere.

robincornett:
It looks like there is a fair amount of rounding happening in the cart details which is where `4.625` became `4.63`:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/feb51a37021ccbd21fd53003ab995e653cf11b23/includes/cart/class-edd-cart.php#L266-L277

assets/js/admin/orders/order-overview/models/state.js
```diff
@@ -154,11 +122,14 @@ export const State = Backbone.Model.extend(
...
+			const { number } = this.get( 'formatters' );
 			const items = this.get( 'items' ).models;
+			const adjustments = this.get( 'adjustments' ).getByType( 'fee' );
 
-			return items.reduce(
+			return [ ...items, ...adjustments ].reduce(
 				( amount, item ) => {
-					return amount += +item.get( 'tax' );
+					// Perform early rounding of tax amounts to match storefront calculations.
+					return amount += +number.format( item.get( 'tax' ) );
```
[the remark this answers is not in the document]
robincornett: Actually, I'm not 100% sure what to do here. I think the rounding does need to be reverted, so that the order data is saved consistently with the storefront, although the final amounts end up being different. However, with my test order discrepancies, this results in the refund process breaking, because the maximum refundable amount is $63.91, as the JS calculated the taxes, but the way the refund amounts add up (by item), it wants to refund $63.92 and there is a validation error.

robincornett (changes requested):
Ran into an issue with applying the tax rate to an existing taxable fee, and based on our conversation earlier, I think it's okay to have the discrepancy between a storefront and manually created order (I think the manual order is calculating correctly). However, then refunds break (see the comment) so I'm a little stuck right now.

spencerfinnell:
@robincornett Pushed changes to make the manual order process match the exact data that gets created via the storefront in your given example.

@ashleyfae I'm not sure how you want to handle the refund issue. The order total is stored as `63.916250000` (cart and manual) but the refund flow JavaScript is using a rounded amount of `63.92` in the validator, creating a larger number than the order total value that is retrieved.

robincornett:
My test case order amounts are now consistent with the cart amounts, but @spencerfinnell is correct, refunds are still not working correctly because:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/70af4dddb2ea5a106998e9c41d092e9ed1d14d2f/includes/orders/classes/class-refund-validator.php#L267

Is comparing a rounded number (`63.92`) with a not rounded number (`63.91625`). Note that this is likely to happen with manually created and storefront orders. My thought is to run both of those values through `edd_format_amount` before using them for a comparison. Spencer suggests we need to do the same for other checks in the refund validator, so I'm going to spend time there. My initial tests suggest that the order creation process is now consistent with the storefront/cart.

ashleyfae:
> @ashleyfae I'm not sure how you want to handle the refund issue. The order total is stored as `63.916250000` (cart and manual) but the refund flow JavaScript is using a rounded amount of `63.92` in the validator, creating a larger number than the order total value that is retrieved.

IMO we should be rounding it when storing it. We should store what we charged. `63.916250000` isn't a real amount that we charged; we should be storing `63.92` (with extra zeros on the end is fine)

robincornett (approved):

LisaCee (approved):
- [ ] Admin order with discount and taxes matches front end
- [ ] Admin order with discount and taxes subtotal matches front end
- [ ] Taxed and non-taxed fees calculated correctly
- [ ] Subtotal only contains products
- [ ] Discounts created with Discount Code Generator taxed correctly
- [ ] Taxed location changes reflected in admin and front end

Note: Discount Pro discounts (negative fees) are included in the subtotal, but that is another issue for another day.  I just wanted to note it here for record.

ashleyfae (approved):

PR #8583, opened 2021-04-22
Update edd_count_file_downloads_of_user/customer functions

### LINKED ISSUE #8570: 3.0 - edd_count_file_downloads_of_customer() always returns 0
### Expected behavior

When I use `edd_count_file_downloads_of_customer()` or `edd_count_file_downloads_of_user()`, I expect it to return an accurate count of the logs.

### Actual behavior

It always returns `0` even if there have been downloads by that customer.

1) Check your `wp_edd_logs_file_downloads` and pick a `customer_id` from the records.
2) Run `edd_count_file_downloads_of_user( $customer_id )`
3) Note the function returns `0`, even though there are records in the database.

This function needs reworking to query our file download database directly.

`edd_count_file_downloads_of_user()` should be updated as well.

PHP Version:

EDD Version (or branch):

WordPress Version:

I'm wondering if we should also update `EDD_Logging::get_log_count()` to try to parse stuff out of the meta query that we KNOW has moved somewhere else. These functions no longer work because they're passing in meta queries like this:

```php
$meta_query = array(
	array(
		'key'   => '_edd_log_customer_id',
		'value' => $customer->id,
	)
);
```

But the customer ID no longer lives in meta, which is why the query ultimately fails. Maybe we should catch out certain pieces of meta and re-map those to other arguments.

robincornett:
Fixes #8570

Proposed Changes:
1. Update `edd_count_file_downloads_of_user` and `edd_count_file_downloads_of_customer` to use the new `edd_count_file_download_logs` function.
2. Add the `user_id` as meta to the file download log if it exists (it looks like this was removed at one point, but we keep this data on migration, so adding it back for now. If we want to remove it, we should likely also remove it from the migration).

To test:
Make sure you have orders on your site. You can run this CLI command:
```
wp edd payments create --number=1000 --date=random --range=365 --generate_users=1
```

That will generate new customers, but not necessarily new users. You will need to link at least some of the customers to users. Then switch to this branch and run:
```
wp edd download_logs create --number=1000
```
Some of the download logs will be linked to a customer only, if the customer does not have a linked user account, but some should be linked to a user account. Check the `wp_edd_logs_file_downloadmeta` table for a user ID with records, and then you can replace the customer and user ID in this code to test:

```php
$customer_id = 101;
$user_id     = 2;
var_dump( edd_count_file_downloads_of_customer( $customer_id ) );
var_dump( edd_count_file_downloads_of_user( $user_id ) );
```

The numbers should be the same for a valid user. You can also pass an email address in to either function. In `release/3.0`, these functions will not work.

It is possible that the download numbers will not be the same. We were migrating the `user_id` for 2.x orders, but until this PR, were not storing the `user_id` for new downloads. If an existing user had downloads on `release/3.0`, the customer download number will be higher than the user download number.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -250,6 +250,7 @@ class Data_Migrator {
 				'_edd_log_price_id',
 				'_edd_log_customer_id',
 				'_edd_log_ip',
+				'_edd_log_user_id',
 			);
 			$meta_to_migrate   = $post_meta;
 			$new_log_id        = edd_add_file_download_log( $log_data );
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -244,7 +244,7 @@ class Data_Migrator {
 				'date_modified' => $data->post_modified_gmt,
 			);
 
-			$meta_to_remove    = array(
+			$meta_to_remove = array(
 				'_edd_log_file_id',
 				'_edd_log_payment_id',
 				'_edd_log_price_id',
@@ -252,6 +252,10 @@ class Data_Migrator {
 				'_edd_log_ip',
 				'_edd_log_user_id',
 			);
+			// If the log doesn't have a customer ID, but does have a user ID, keep the user ID as metadata.
+			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1 ), true ) ) {
+				$meta_to_remove = array_diff( $meta_to_remove, array( '_edd_log_user_id' ) );
+			}
 			$meta_to_migrate   = $post_meta;
 			$new_log_id        = edd_add_file_download_log( $log_data );
 			$add_meta_function = 'edd_add_file_download_log_meta';
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -253,7 +253,7 @@ class Data_Migrator {
 				'_edd_log_user_id',
 			);
 			// If the log doesn't have a customer ID, but does have a user ID, keep the user ID as metadata.
-			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1 ), true ) ) {
+			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1, '-1' ), true ) ) {
 				$meta_to_remove = array_diff( $meta_to_remove, array( '_edd_log_user_id' ) );
 			}
 			$meta_to_migrate   = $post_meta;
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -253,7 +253,7 @@ class Data_Migrator {
 				'_edd_log_user_id',
 			);
 			// If the log doesn't have a customer ID, but does have a user ID, keep the user ID as metadata.
-			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1, '-1' ), true ) ) {
+			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1 ) ) ) {
 				$meta_to_remove = array_diff( $meta_to_remove, array( '_edd_log_user_id' ) );
 			}
 			$meta_to_migrate   = $post_meta;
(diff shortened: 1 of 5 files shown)


LisaCee (approved):
This worked for me, though I always had to use the email address for the `$user_id` to get the two values to match.  Even when the two values didn't match, the `edd_count_file_downloads_of_customer( )` was always > 0.

ashleyfae (changes requested):
Is there actually a use case where a _user_ who is not also a _customer_ would have a file download log?

IMO we should get rid of user ID, and for the `edd_count_file_downloads_of_user()` function, the flow should go like this:

- Given a user ID, fetch the customer record.
- If no customer, bail.
- Then, using the customer ID, call `edd_count_file_downloads_of_customer()`.

That's also better for performance. Doing a meta query with a `value` check is really slow. Querying directly on the `customer_id` column is way better/faster.

robincornett:
@ashleyfae should the user ID be removed from the log metadata as well then?

ashleyfae:
@robincornett Yes, but first to clarify: where are you seeing this kept in the migration?

API logs should have user_id, but my understanding is that they were removed for file download logs via an old migration: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/admin/upgrades/classes/class-file-download-log-migration.php#L64-L88

robincornett:
This is in the `edd_logs_file_downloadmeta` table, migration from 2.x run just this morning:

image

ashleyfae:
Let's strip that out. :+1: That might be the generation before `_edd_log_user_info` meta. :shrug:

robincornett:
Re-requesting reviews as changes have been made. On an existing site, the user ID which matches a customer should return the same number now as the user ID is used to retrieve the customer object.

A migration should be run to test as well, if possible--make sure there is file download log data to migrate. I've run a migration with both the UI and WP-CLI and the user ID is no longer saved to the metadata, but someone else should confirm please. :)

includes/class-edd-cli.php
```diff
@@ -1580,7 +1580,7 @@ public function download_logs( $args, $assoc_args ) {
 
 			$file_id_key = array_rand( $download_ids_with_files[ $product_id ], 1 );
 			$file_key    = $download_ids_with_files[ $product_id ][ $file_id_key ];
-			edd_add_file_download_log( array(
+			$log_id      = edd_add_file_download_log( array(
```
ashleyfae, line 1583:
```suggestion
			edd_add_file_download_log( array(
```

The ID is no longer needed.

includes/download-functions.php
```diff
@@ -772,7 +772,7 @@ function edd_get_download_sales_stats( $download_id = 0 ) {
  *
  * @param int    $download_id Download ID.
  * @param int    $file_id     File ID.
- * @param array  $user_info   User information (Deprecated)
+ * @param array  $user_info   User information.
```
ashleyfae, line 775: Let's add the `(Deprecated)` back in.

includes/download-functions.php
```diff
@@ -789,7 +789,7 @@ function edd_record_download_in_log( $download_id = 0, $file_id = 0, $user_info
 
 	$user_agent = $browser->getBrowser() . ' ' . $browser->getVersion() . '/' . $browser->getPlatform();
 
-	edd_add_file_download_log( array(
+	$download_log = edd_add_file_download_log( array(
```
ashleyfae, line 792:
```suggestion
	edd_add_file_download_log( array(
```

The log ID isn't needed.

robincornett:
@ashleyfae I am running a migration right now and definitely have records with a user ID, but not a customer ID. It looks like maybe the customer ID wasn't saved to entries until sometime in 2014, just looking at the rows in front of me, but even once they start showing in the database, not every row has a customer ID.

I wonder if it might be worth saving the user ID _if_ there is no customer ID? So the log is attached to someone somewhere?

ashleyfae:
> @ashleyfae I am running a migration right now and definitely have records with a user ID, but not a customer ID. It looks like maybe the customer ID wasn't saved to entries until sometime in 2014, just looking at the rows in front of me, but even once they start showing in the database, not every row has a customer ID.
> 
> I wonder if it might be worth saving the user ID _if_ there is no customer ID? So the log is attached to someone somewhere?

For now save it in the migrator under those circumstances, but let's leave all the other code as-is.

robincornett:
Okay, waiting for a current migration to finish so I can test keeping it in--hold, please for another commit.

LisaCee (approved):
This still works for me.  I don't have any customers without a customer ID, but for current customers the numbers match and are > 0.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -244,13 +244,18 @@ public static function logs( $data = null ) {
...
+			$meta_to_remove = array(
 				'_edd_log_file_id',
 				'_edd_log_payment_id',
 				'_edd_log_price_id',
 				'_edd_log_customer_id',
 				'_edd_log_ip',
+				'_edd_log_user_id',
 			);
+			// If the log doesn't have a customer ID, but does have a user ID, keep the user ID as metadata.
+			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1, '-1' ), true ) ) {
```
ashleyfae, line 256:
```suggestion
			if ( empty( $log_data['customer_id'] ) && ! empty( $post_meta['_edd_log_user_id'] ) && ! in_array( $post_meta['_edd_log_user_id'], array( 0, -1 ) ) ) {
```

IMO this is a perfect case for not doing a strict check, because we want to account for both integers and strings.

ashleyfae (approved):

PR #8589, opened 2021-04-27
Use is_ssl() to set cookies to secure

### LINKED ISSUE #8037: Maybe update edd_items_in_cart cookie behavior
The `edd_items_in_cart` cookie attributes may need to be updated.

### Justification or use case
Adding an item to a cart on a customer site generated this in the dev tools console (Firefox):
>Cookie “edd_items_in_cart” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

robincornett:
Fixes #8037

Proposed Changes:
1. Change the `secure` parameter for `setcookie` to check `is_ssl()`.
2. Remove error silencing.

### DIFF
--- a/includes/class-edd-session.php
+++ b/includes/class-edd-session.php
@@ -231,13 +231,11 @@ class EDD_Session {
 	 * @return void
 	 */
 	public function set_cart_cookie( $set = true ) {
-		if( ! headers_sent() ) {
-			if( $set ) {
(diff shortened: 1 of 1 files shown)


ashleyfae (commented):
Code looks good, but I still need to test it.

LisaCee (approved):
I was able to replicate the error message in Firefox and confirmed that this branch eliminated it.

ashleyfae (approved):

PR #8595, opened 2021-05-03
Sanitize discount code & send JSON headers.

ashleyfae:


robincornett (approved):
Add/removing a discount code still works as expected.

JJJ (approved):

PR #8606, opened 2021-05-05
Round all order subtotals, taxes, totals before saving

### LINKED ISSUE #8599: 3.0 - subtotal, tax, totals should be rounded when saved to the database
### Expected behavior
Data in the database should not be rounded:

> > @ashleyfae I'm not sure how you want to handle the refund issue. The order total is stored as `63.916250000` (cart and manual) but the refund flow JavaScript is using a rounded amount of `63.92` in the validator, creating a larger number than the order total value that is retrieved.
> 
> IMO we should be rounding it when storing it. We should store what we charged. `63.916250000` isn't a real amount that we charged; we should be storing `63.92` (with extra zeros on the end is fine)
> 

_Originally posted by @ashleyfae in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8581#issuecomment-831956679_

### Actual behavior
In EDD 3.0, the order item values are rounded when stored, but order and adjustment tax and totals are not.

1. The easiest way to see this in action is to set up a storefront with a tax rate which will generate non "round" (not end with two decimals really) numbers, or set prices which will result in the same with a tax rate applied. My tax rate is 9.25% so it lends itself well to this.
2. Create orders with taxes and taxable fees. Check the database rows.

PHP Version: 7.4.1
WordPress Version: 5.7.1

This is affecting #8578 and breaking refunds, if an order total is stored as `63.916250000`, the refund validation fails when the refund total is `63.92`, because it's technically greater than the order total. #8578 should be finalized and merged before tackling this as its current focus is to get manual orders synced with cart orders.

robincornett:
Fixes #8599

Proposed Changes:
1. Round every subtotal, tax, total, amount before saving to the database. Applies to orders, order items, order adjustments.

### DIFF
--- a/includes/orders/functions/actions.php
+++ b/includes/orders/functions/actions.php
@@ -186,10 +186,10 @@ function edd_add_manual_order( $args = array() ) {
 			'currency'     => edd_get_currency(),
 			'payment_key'  => $data['payment_key'] ? sanitize_text_field( $data['payment_key'] ) : edd_generate_order_payment_key( $email ),
 			'tax_rate_id'  => ! empty( $tax_rate->id ) ? $tax_rate->id : null,
-			'subtotal'     => $order_subtotal,
-			'tax'          => $order_tax,
-			'discount'     => $order_discount,
-			'total'        => $order_total,
+			'subtotal'     => round( $order_subtotal, edd_currency_decimal_filter() ),
+			'tax'          => round( $order_tax, edd_currency_decimal_filter() ),
+			'discount'     => round( $order_discount, edd_currency_decimal_filter() ),
+			'total'        => round( $order_total, edd_currency_decimal_filter() ),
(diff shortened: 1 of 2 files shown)


spencerfinnell (approved):

LisaCee (approved):
I couldn't replicate your issue with refunds exceeding the maximum amount, but all of my database values are nicely rounded.

ashleyfae:
I've created a new issue to refactor how this is handled at a later date. https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8609 Right now the logic is being run separately on each column, but we can actually set it up so we have one general "sanitize" function and make sure each amount column is run through it.

I'll approve this for now since it's done and it works, but at a later date and lower priority we should look at moving all that logic to one central place.

PR #8610, opened 2021-05-07
Re-organize and only display order item notes once.

### LINKED ISSUE #7415: Download Notes duplicated in email receipt
### Expected behavior
When multiple variable priced options are purchased for the same core product, {download_list} should only display Download Notes once per core product.

### Actual behavior
{download_list} duplicates Download Notes for each variable priced option that is purchased.

1) Purchase 2 or more variable options for a single product
2) {download_list} in the email receipt should only display Download Notes once.

PHP Version: 5.4.4

EDD Version (or branch): 2.9.17

WordPress Version: 4.9.11

https://secure.helpscout.net/conversation/978365525/134068?folderId=180499

cklosowski:
Fixes #7415 

Proposed Changes:
1. Collects items on an order that need to display notes on the purchase receipt.
2. Moves the order notes to its own section of the purchase receipt.
3. Only shows a single note for a download ID, instead of duplicating when the same download ID is purchased more than once (with a different price id)


includes/emails/class-edd-email-tags.php
```diff
@@ -486,6 +490,19 @@ function edd_email_tag_download_list( $payment_id ) {
 	}
 	$download_list .= '</ul>';
 
+	// Remove any empty values.
+	$needs_notes = array_filter( $needs_notes );
+	if ( ! empty( $needs_notes ) ) {
+		$download_list .= '<div>';
+		$download_list .= __( 'Additional information about your purchase.', 'easy-digital-downloads' );
```
robincornett, line 497:
```suggestion
		$download_list .= __( 'Additional information about your purchase:', 'easy-digital-downloads' );
```

LisaCee (approved):
This works as expected, though I do like Robin's suggestions.

includes/emails/class-edd-email-tags.php
```diff
@@ -486,6 +490,19 @@ function edd_email_tag_download_list( $payment_id ) {
 	}
 	$download_list .= '</ul>';
 
+	// Remove any empty values.
+	$needs_notes = array_filter( $needs_notes );
+	if ( ! empty( $needs_notes ) ) {
+		$download_list .= '<div>';
```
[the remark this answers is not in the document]
cklosowski: I can see how a UL might be good, but not as a wrapper for the entire section.

includes/emails/class-edd-email-tags.php
```diff
@@ -474,8 +475,11 @@ function edd_email_tag_download_list( $payment_id ) {
 			}
 
 
-			if ( '' != edd_get_product_notes( $item['id'] ) ) {
-				$download_list .= ' &mdash; <small>' . edd_get_product_notes( $item['id'] ) . '</small>';
+			if ( ! array_key_exists( $item['id'], $needs_notes ) ) {
```
robincornett, line 478: The output is totally fine and this is working, but I guess I am looking at this wanting to assign `edd_get_product_notes( $item['id'] )` to a variable and use that as part of the conditional check (in addition to what's already here), instead of building the array for everything in the cart.

ashleyfae (approved):

PR #8614, opened 2021-05-10
Do not migrate amount, user ID for payment meta

### LINKED ISSUE #8603: 3.0 - unneeded payment_meta is still migrated
### Expected behavior
When I migrate my site data from 2.10 to 3.0, I expect to only have `payment_meta` saved to the new orders if there is something unique (for example, from an extension) in the metadata which needs to be preserved.

### Actual behavior
When I run the migration, some [core metadata](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L739-L750) is retained. This may be because I'm looking at a data set for a fairly old site (another migration doesn't have all of these), but I'm seeing a lot of entries with:
- `amount`
- `user_id`

1. Run a migration from EDD 2.10 to 3.0. I'm going to guess the data needs to be from a fairly old site.
2. Check the `edd_ordermeta` table for `payment_meta` key/values with these two items.

PHP Version: 7.3.5
WordPress Version: 5.7.1

robincornett:
Fixes #8603

Proposed Changes:
1. Adds `user_id` and `amount` to the core meta keys which don't need to be migrated.

To Test:
1. Run yet another migration from 2.x order data to 3.0. It may require having an older dataset.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -752,6 +752,8 @@ class Data_Migrator {
 			'discount',
 			'subtotal',
 			'tax',
+			'amount',
+			'user_id',
 		);
 
 		// Remove core keys from `user_info`.


ashleyfae:
@robincornett Is this ready for review?

robincornett:
Sorry, yes, I forgot to put in reviewers!

ashleyfae (approved):

PR #8621, opened 2021-05-11
Create order recovery methods

### LINKED ISSUE #8550: 3.0 - should the order class have recovery functions?
In working on #8539, I'm seeing that when the `EDD_Payment` class is instantiated, that there are two "recovery" functions which are not available elsewhere:
- [`$payment->is_recoverable()`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L2645)
- [`$payment->get_recovery_url()`](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/payments/class-edd-payment.php#L2664)

It might be nice for this logic to be available to an individual order.

### Justification or use case
On the purchase history, for example, having these functions available to an order would be one more step towards being able to not use the `EDD_Payment` class.

robincornett:
Fixes #8550; branched from #8549

Proposed Changes:
1. Create `edd_recoverable_order_statuses` to return an array of recoverable order statuses.
2. Create order recovery methods `$order->is_recoverable()` and `$order->get_recovery_url()`.
3. Use the new order methods in the purchase history template.

### DIFF
--- a/includes/orders/classes/class-order.php
+++ b/includes/orders/classes/class-order.php
@@ -642,7 +642,7 @@ class Order extends Rows\Order {
 		$recovery_url = add_query_arg(
 			array(
 				'edd_action' => 'recover_payment',
-				'payment_id' => $this->id,
+				'payment_id' => urlencode( $this->id ),
 			),
 			edd_get_checkout_uri()
 		);
@@ -660,7 +660,8 @@ class Order extends Rows\Order {
 		 * The order recovery URL.
 		 *
 		 * @since 3.0
-		 * @param \EDD\Orders\Order $this The order object.
+		 * @param string            $recovery_url The order recovery URL.
+		 * @param \EDD\Orders\Order $this         The order object.
 		 */
 		return apply_filters( 'edd_order_recovery_url', $recovery_url, $this );
 	}
--- a/includes/orders/classes/class-order.php
+++ b/includes/orders/classes/class-order.php
@@ -610,4 +610,58 @@ class Order extends Rows\Order {
 	public function is_complete() {
 		return ( 'complete' === $this->status );
 	}
+
+	/**
+	 * Determines if this order is able to be resumed by the user.
+	 *
+	 * @since 3.0
+	 *
+	 * @return bool
+	 */
+	public function is_recoverable() {
+		$recoverable_statuses = edd_recoverable_order_statuses();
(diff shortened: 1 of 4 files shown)


includes/orders/classes/class-order.php
```diff
@@ -610,4 +610,58 @@ public function get_notes() {
...
+	 */
+	public function get_recovery_url() {
+		if ( ! $this->is_recoverable() ) {
+			return false;
+		}
+
+		$recovery_url = add_query_arg(
+			array(
+				'edd_action' => 'recover_payment',
+				'payment_id' => $this->id,
```
ashleyfae, line 645:
```suggestion
				'payment_id' => urlencode( $this->id ),
```

includes/orders/classes/class-order.php
```diff
@@ -610,4 +610,58 @@ public function get_notes() {
...
+		 */
+		if ( has_filter( 'edd_payment_recovery_url' ) ) {
+			$recovery_url = apply_filters( 'edd_payment_recovery_url', $recovery_url, edd_get_payment( $this->id ) );
+		}
+
+		/**
+		 * The order recovery URL.
+		 *
+		 * @since 3.0
+		 * @param \EDD\Orders\Order $this The order object.
```
ashleyfae, line 663: Can you document the `$recovery_url` parameter too?

includes/orders/functions/orders.php
```diff
@@ -1378,6 +1378,26 @@ function edd_get_net_order_statuses() {
...
+	$statuses = apply_filters( 'edd_recoverable_payment_statuses', $statuses );
+
+	/**
+	 * Order statuses which are considered recoverable.
+	 *
+	 * @param $statuses {
+	 *        An array of order status array keys.
+	 * }
+	 */
+	return apply_filters( 'edd_recoverable_order_statuses', $statuses );
```
ashleyfae, line 1398: I'm inclined to say we just keep the old `edd_recoverable_payment_statuses` filter for now and not introduce a new one. I know the naming isn't perfect with our new terminology, but in the event that all the parameters are still the same, I think keeping the old ones is simpler.

ashleyfae (changes requested):
Can you update the methods in the `EDD_Payment` object to use the new order methods instead?

So, `EDD_Payment::is_recoverable()` would be updated to be:

```php
return $this->order->is_recoverable();
```

and same with `get_recovery_url()`.

ashleyfae (approved):

PR #8622, opened 2021-05-11
Filter required fields for modifying on checkout

### LINKED ISSUE #8572: Unexpected checkout fields may become required when a discount is applied
### Expected behavior
I expect not-required checkout fields to not change unexpectedly to be required.

### Actual behavior
As reported in Barn2's EU VAT changelog:
> Fix: Fixed a issue introduced in EDD 2.10.1 which made the VAT number a required field on checkout when a discount code was applied.

I am actually unable to fully replicate because I went ahead and ran the update to 1.3.5 of the VAT extension. Assume we could replicate with 1.3.4. My guess would be that the extension used the `.edd-input` class for the VAT field--adding that class does reproduce the behavior and removing it fixes it.

PHP Version: 7.4.1
EDD Version (or branch): 2.10
WordPress Version: 5.7

My guess is that the fix we applied for #7487 is causing this behavior. Reporting here because although they've resolved it on their end, if we have unintentionally broken similar behavior more universally, it would be great to fix it.

robincornett:
Fixes #8572

Proposed Changes:
1. Update the checkout JS to set an array of required fields when applying a discount, and modifying them after the AJAX call, instead of relying on CSS classes.


ashleyfae:
## Testing

I'm testing with the un-patched version of the EU VAT plugin.

- [x] Reproduce issue on `master` branch: on checkout, verify that VAT Number field is not required; add 25% off discount code; verify that VAT Number field is now required.
- [ ] Verify fix: on checkout, verify that VAT Number field is not required; add 25% off discount code; verify that VAT Number field is still not required.
- [ ] On checkout, verify that billing address fields are required. Add 25% off discount code. Verify that billing address fields are still required.
- [ ] On checkout, verify that billing address fields are required. Add 100% off discount code. Verify that billing address fields are not required.

ashleyfae (approved):

PR #8624, opened 2021-05-13
Issue/8623

### LINKED ISSUE #8623: 3.0 - EDD_Payments_Query failures when using meta queries
### Expected behavior

I expect this query that PDF Invoices uses to work to find relevant payments:

```php
$meta_query = array(
	'relation'  => 'AND',
	array(
		'key'   => '_edd_payment_purchase_key',
		'value' => $key
	),
	array(
		'key'   => '_edd_payment_user_email',
		'value' => $email
	)
);

$args = array(
	'meta_query' => $meta_query,
);

$payments_query = new EDD_Payments_Query( $args );
$payments = $payments_query->get_payments();
```

### Actual behavior

It's not returning any results.

There are a couple problems here:

1. `_edd_payment_purchase_key` is missing from our meta query re-mapping.
2. After re-mapping something that was in meta, we should be unsetting it from the meta query. Otherwise it ends up checking in the orders table **and** in meta.

1) Install PDF invoices.
2) Go to your order history on the front-end and click `Download Invoice` for an order.
3) You'll get an error about the invoice not being found.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8623

Proposed Changes:
1. Add `_edd_payment_purchase_key` to remapping. That will look for it in the orders table instead of meta.
2. After remapping a meta query clause to the orders table, unset that meta query so it doesn't look in meta anymore.
3. BerlinDB: check if `where` is empty before using. See https://github.com/berlindb/core/issues/101

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -746,24 +746,33 @@ class EDD_Payments_Query extends EDD_Stats {
 			$arguments['status'] = edd_get_payment_status_keys();
 		}
 
-		if ( isset( $this->args['meta_query'] ) && is_array( $this->args['meta_query'] ) ) {
-			foreach ( $this->args['meta_query'] as $meta ) {
+		if ( isset( $arguments['meta_query'] ) && is_array( $arguments['meta_query'] ) ) {
+			foreach ( $arguments['meta_query'] as $meta_index => $meta ) {
 				if ( ! empty( $meta['key'] ) ) {
 					switch ( $meta['key'] ) {
 						case '_edd_payment_customer_id':
 							$arguments['customer_id'] = absint( $meta['value'] );
+							unset( $arguments['meta_query'][ $meta_index ] );
 							break;
 
 						case '_edd_payment_user_id':
 							$arguments['user_id'] = absint( $meta['value'] );
+							unset( $arguments['meta_query'][ $meta_index ] );
 							break;
 
 						case '_edd_payment_user_email':
 							$arguments['email'] = sanitize_email( $meta['value'] );
+							unset( $arguments['meta_query'][ $meta_index ] );
 							break;
 
 						case '_edd_payment_gateway':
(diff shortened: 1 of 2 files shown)


robincornett (approved):

includes/payments/class-payments-query.php
```diff
@@ -746,24 +746,33 @@ private function remap_args() {
...
 
-		if ( isset( $this->args['meta_query'] ) && is_array( $this->args['meta_query'] ) ) {
-			foreach ( $this->args['meta_query'] as $meta ) {
+		if ( isset( $arguments['meta_query'] ) && is_array( $arguments['meta_query'] ) ) {
+			foreach ( $arguments['meta_query'] as $meta_index => $meta ) {
 				if ( ! empty( $meta['key'] ) ) {
 					switch ( $meta['key'] ) {
 						case '_edd_payment_customer_id':
 							$arguments['customer_id'] = absint( $meta['value'] );
+							unset( $arguments['meta_query'][ $meta_index ] );
```
JJJ, line 755: Small piece of general feedback. This code will work, but I would caution against it.

`$arguments` is being iterated through inside of a `foreach` loop. In the middle of that loop, its current index is being unset. If in the future, this argument index needs to be referred to lower (outside the `switch` but still inside the `foreach` loop) it will be gone and trigger a notice. Also, if PHP changes the way Iterators work (clone vs. byref) this may trigger a warning/error, unshifting the original array, or something else weird.

Instead of unsetting mid-iteration, it may be safer to do the opposite, and only add the indexes you want to a new array, and then replace `$arguments` with that new array after the `foreach` has completed.

LisaCee (approved):

includes/payments/class-payments-query.php
ashleyfae: Thank you @JJJ . In the interest of getting beta-2 out, I'm going to merge this for now but will be opening another issue to address refactoring this.

PR #8626, opened 2021-05-14
Move amount sanitization to BerlinDB schema settings #8609

### LINKED ISSUE #8609: 3.0 - Move amount sanitization to BerlinDB schema settings

We should move all the sanitization for amounts to the BerlinDB's supported `validate` feature. This would include:

- Casting to float
- Rounding, according to decimal settings

Each relevant column in the schema would just need `'validate' => 'edd_some_function_name'` added. For example, here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/schemas/class-orders.php#L192

### Justification or use case

Avoid code duplication and needing to run this logic manually on each amount.

This would essentially be refactoring https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8606

ashleyfae:
Fixes #8609 

Proposed Changes:
1. Moves amount sanitization logic to BerlinDB, which means it will _always_ run prior to database insertion. This ensures payments created outside the normal checkout flow also get the same sanitization.

I chose `edd_sanitize_amount` amount for this, which handles rounding, but it also does a bunch of other stuff to strip thousand separators and whatnot. We should test this with non-standard decimal separators to make sure nothing breaks. Example:

- Set decimal separator to be a comma instead of a period.
- Make a purchase for $10.50.
- Make sure it shows up for the customer on the receipt, etc. as `10,50`, but it should be stored in the database as `10.50` (with more zeros at the end).

Other tests to run:

- The original issue: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8599
- Also, purchase a subscription for a non-standard amount like that. Then [trigger a Stripe renewal](https://github.com/easydigitaldownloads/edd-recurring/wiki/Stripe---Manually-Triggering-an-Automatic-Renewal) and make sure the amounts on the renewal payment are rounded.

### DIFF
--- a/includes/database/schemas/class-order-adjustments.php
+++ b/includes/database/schemas/class-order-adjustments.php
@@ -118,7 +118,8 @@ class Order_Adjustments extends Schema {
 			'type'       => 'decimal',
 			'length'     => '18,9',
 			'default'    => '0',
-			'sortable'   => true
+			'sortable'   => true,
+			'validate'   => 'edd_sanitize_amount'
 		),
 
 		// tax
@@ -127,7 +128,8 @@ class Order_Adjustments extends Schema {
 			'type'       => 'decimal',
 			'length'     => '18,9',
 			'default'    => '0',
-			'sortable'   => true
+			'sortable'   => true,
+			'validate'   => 'edd_sanitize_amount'
 		),
 
 		// total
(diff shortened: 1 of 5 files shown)


robincornett (commented):
Should the rounding we added for manual orders be removed as well? (See [edd_add_manual_order](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8609/includes/orders/functions/actions.php#L25))

ashleyfae:
Good call @robincornett . Yes, I'll remove it.

robincornett (approved):

PR #8636, opened 2021-05-19
Update refund validation to work more precisely

### LINKED ISSUE #8634: 3.0 - Refund validation can fail when it shouldn't
### Expected behavior
When I do a full refund, I expect it to work.

### Actual behavior
I have an order with a product price of $6.99 and a 9.25% tax rate. Clicking in the modal to select all products for the refund ends up with a validation error and I don't know why:

image

1. Enable taxes. I have a tax rate of 9.25%.
2. Create a product and purchase it with tax applied. My product price is $6.99; the order total after tax is $7.64.
3. Attempt to do a full refund. Mine fails validation.

PHP Version: 7.3.5
WordPress Version: 5.7.2

I'm digging and logging in the refund validation class and cannot sort out why this is happening. I wondered if it was possibly that the `$attempted_amount` and `$maximum_refundable_amounts[ $column_name ]` might be two different types of things, and it turns out they are... `$attempted_amount` is a `string` and `$maximum_refundable_amounts[ $column_name ]` is a `double`. I attempted using `floatval` on both prior to evaluating, and that made them identical, both the same value and `double`, but the validation is still failing. Logging the type and value of both certainly looks like they are equal, but the validator is failing (this is logged _after_ the `>` comparison fails):

```
// $attempted_amount
[19-May-2021 16:11:55 UTC] double
[19-May-2021 16:11:55 UTC] 7.64
// $maxiumum_amount (result of floatval( $maximum_refundable_amounts[ $column_name ] ))
[19-May-2021 16:11:55 UTC] double
[19-May-2021 16:11:55 UTC] 7.64
```

If I purchase the same product without taxes applied, the refund validation is fine.

The behavior is the same for both manual and cart created orders. The database values all appear to be correct.

robincornett:
Fixes #8634

Proposed Changes:
1. Use `bccomp` to compare the attempted refund amounts with the maximum refund amounts, because computers think in binary. See the related issue for more details.

### DIFF
--- a/includes/orders/classes/class-refund-validator.php
+++ b/includes/orders/classes/class-refund-validator.php
@@ -264,7 +264,7 @@ class Refund_Validator {
 
 		// Overall refund total cannot be over total refundable amount.
 		$order_total = edd_get_order_total( $this->order->id );
-		if ( 1 === bccomp( $this->total, $order_total, 2 ) ) {
+		if ( $this->is_over_refund_amount( $this->total, $order_total ) ) {
 			throw new Exception( sprintf(
 				/* Translators: %s - maximum refund amount as formatted currency */
 				__( 'The maximum refund amount is %s.', 'easy-digital-downloads' ),
@@ -345,7 +345,7 @@ class Refund_Validator {
 			}
 
 			// This is our fallback.
-			$attempted_amount = isset( $original_item->{$column_name} ) ? $original_item->{$column_name} : 0.00;
+			$attempted_amount = isset( $original_item->{$column_name} ) ? floatval( $original_item->{$column_name} ) : 0.00;
 			$maximum_amount   = floatval( $maximum_refundable_amounts[ $column_name ] );
 
 			// Only order items are included in the subtotal.
@@ -355,10 +355,10 @@ class Refund_Validator {
 
 			// But grab from specified amounts if available. It should always be available.
 			if ( isset( $amounts_to_refund[ $column_name ] ) ) {
-				$attempted_amount = $amounts_to_refund[ $column_name ];
+				$attempted_amount = floatval( $amounts_to_refund[ $column_name ] );
 			}
 
-			if ( 1 === bccomp( $attempted_amount, $maximum_amount, 2 ) ) {
+			if ( $this->is_over_refund_amount( $attempted_amount, $maximum_amount ) ) {
 				if ( $original_item instanceof Order_Item ) {
 					$error_message = sprintf(
 						/*
@@ -489,5 +489,28 @@ class Refund_Validator {
 		return $new_args;
 	}
 
+	/**
+	 * Checks if the attempted refund amount is over the maximum allowed refund amount.
+	 * The calculation has to be done using a machine epsilon because when rational numbers
+	 * are converted to base 2, there is a small loss of precision.
+	 *
+	 * @see https://www.php.net/manual/en/language.types.float.php
+	 *
+	 * @since 3.0
+	 * @param float $attempted_amount The amount to refund.
+	 * @param float $maximum_amount   The maximum amount which can be refunded.
+	 * @return boolean
+	 */
+	private function is_over_refund_amount( $attempted_amount, $maximum_amount ) {
+		/**
+		 * The machine epsilon, or unit roundoff--the smallest acceptable difference in calculations.
+		 * See https://www.php.net/manual/en/language.types.float.php
+		 *
+		 * @var float
+		 */
+		$epsilon = 0.00000001;
+
+		return ( ( $attempted_amount - $maximum_amount ) / $maximum_amount ) > $epsilon;
+	}
 }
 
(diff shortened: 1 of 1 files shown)


includes/orders/classes/class-refund-validator.php
```diff
@@ -264,7 +264,7 @@ public function validate_and_calculate_totals() {
 
 		// Overall refund total cannot be over total refundable amount.
 		$order_total = edd_get_order_total( $this->order->id );
-		if ( $this->total > $order_total ) {
+		if ( 1 === bccomp( $this->total, $order_total, 2 ) ) {
```
cklosowski, line 267: @robincornett Does this need to use the order's currency decimal filter in order to work here? The last parameter is defined as:
`The optional scale parameter is used to set the number of digits after the decimal place which will be used in the comparison.`

If this is a cryptocurrency comparison, then the precision is 8, or even if there is a zero decimal currency, will this still hold true?

robincornett: Well, generally what EDD has at that point is something with two decimal places at the most.

However, it looks like `bccomp` isn't the right answer for us here--although it's not a new PHP function, all of the unit tests are failing, so I'm guessing that means that it's not necessarily available on every server. Thinking on this again.

cklosowski: I don't think any of those failures are due to this code, looks like an HTTP request was failing from some of the test dependencies. I re-ran them and they are failing elsewhere now, that is code related.

EDD can't always get two decimals at that point if you are running a cryptocurrency, as they always have 8 decimal points, even if they are all zeros. This makes me think there is something up the chain that might not be multi-decimal currency compatible. Worth a look I think.

robincornett: Thanks--I do not have unit testing working on the new machine yet, so missed those. This gives me a direction.

cklosowski: @robincornett I'm running Windows on my new machine and I have it rocking away, so maybe we can get on a call and figure it out on yours too.

I ended up finding a pretty good solution for local dev that didn't stray far from my Linux setup, actually.

ashleyfae:
What about this instead:

When we do comparisons, run the values through `edd_sanitize_amount()`. That will round them to two decimal places, which should fix the problem where we end up comparing `7.640000000000001` to `7.64` (because the former would be rounded to 2 decimal places). I personally prefer this approach over the epsilon approach just because that feels super hacky to me.

This would just require two minor changes on these lines:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-refund-validator.php#L267
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/orders/classes/class-refund-validator.php#L360

From this:

```php
if ( $this->total > $order_total ) {
```

to this:

```php
if ( edd_sanitize_amount( $this->total ) > edd_sanitize_amount( $order_total ) ) {
```

includes/orders/classes/class-refund-validator.php
```diff
@@ -345,7 +345,8 @@ private function validate_item_and_add_totals( $original_item, $amounts_to_refun
 			}
 
 			// This is our fallback.
-			$attempted_amount = isset( $original_item->{$column_name} ) ? $original_item->{$column_name} : 0.00;
+			$attempted_amount = isset( $original_item->{$column_name} ) ? floatval( $original_item->{$column_name} ) : 0.00;
+			$maximum_amount   = floatval( $maximum_refundable_amounts[ $column_name ] );
```
ashleyfae, line 349: Can you revert this? Just because `edd_sanitize_amount()` will already do it and I don't want to confuse ourselves over which one we actually need.

includes/orders/classes/class-refund-validator.php
```diff
@@ -354,10 +355,10 @@ private function validate_item_and_add_totals( $original_item, $amounts_to_refun
 
 			// But grab from specified amounts if available. It should always be available.
 			if ( isset( $amounts_to_refund[ $column_name ] ) ) {
-				$attempted_amount = $amounts_to_refund[ $column_name ];
+				$attempted_amount = floatval( $amounts_to_refund[ $column_name ] );
```
ashleyfae, line 358: Can you revert this? Just because `edd_sanitize_amount()` will already do it and I don't want to confuse ourselves over which one we actually need.

robincornett:
@ashleyfae for the moment I left the new method intact--it felt cleaner to me to have that instead of the function calls in the conditional, but I can change it completely to your suggestion if you prefer. I'm still wondering why the logging I was doing was not revealing that one of the numbers was not rounded at all, but this is currently working on the problem order. Thanks for the pointer!

ashleyfae:
I'm fine with keeping the new method. :+1: Just left a couple comments about `floatval()`. Then I think we're good.

The logging may have just been rounding the number for you too. :shrug:

PR #8638, opened 2021-05-19
#8575 Disable button while processing refund

### LINKED ISSUE #8575: 3.0 - Disable refund button during AJAX request
### Expected behavior
If I click two times on the refund button, I expect to only have the first click register (or be acted on)
### Actual behavior
While testing issue/8564 (now merged into core) with a wonky mouse, I kept seeing an error saying "Order not refundable".  At first I thought it was a bug, but it turned out that my mouse was double clicking and causing two requests for a refund.  
1) Refund an order with your network speed throttled (slow 3G worked for me)
2) Click the refund button twice
3) You should see a confirmation followed by the error.

PHP Version: 7.4

EDD Version (or branch): 3.0

WordPress Version:5.7.1

LisaCee:
Fixes #8575

Proposed Changes:
1. Change button setting to disabled true while refund is processing


robincornett (commented):
This is working as expected (thank you for the instructions about throttling!). With the button being disabled, I'm wondering if we should skip adding the `updating-message` class, as everything ends up being whited out on a disabled button? It generally goes by quickly enough that you may not really notice, but it seems like disabling the button is the desired effect, and the class may just be an unnecessary distraction.

robincornett (approved):

assets/js/admin/orders/order-overview/_refund.js
```diff
@@ -221,7 +221,7 @@ function recalculateRefundTotal() {
 $(document.body).on( 'click', '#edd-submit-refund-submit', function(e) {
 	e.preventDefault();
 	$('.edd-submit-refund-message').removeClass('success').removeClass('fail');
-	$( this ).attr( 'disabled', false ).addClass( 'updating-message' );
+	$( this ).attr( 'disabled', true );
```
robincornett, line 224: I know I approved this but I'm watching a migration in another screen and I like the `updating-message` icon behavior. I'm wondering what you would think of adding that back in, but removing the `button-primary` class from `$( this )` at the same time? That would get the spinner in the button, but the color would be the red spinner, which does work with the disabled button.

LisaCee: Oh, I like that much more. Good suggestion.

assets/js/admin/orders/order-overview/_refund.js
```diff
@@ -221,7 +221,8 @@ function recalculateRefundTotal() {
 $(document.body).on( 'click', '#edd-submit-refund-submit', function(e) {
 	e.preventDefault();
 	$('.edd-submit-refund-message').removeClass('success').removeClass('fail');
-	$( this ).attr( 'disabled', true );
+	$( this ).removeClass( 'button-primary' );
+	$( this ).attr( 'disabled', true ).addClass( 'updating-message' );
```
robincornett, line 225: YES thank you I like this so much!

I _think_ the two lines could be combined if we want to be super finicky (not tested):

```js
$( this ).removeClass( 'button-primary' ).attr( 'disabled', true ).addClass( 'updating-message' );
```
but I love this.

LisaCee: I had meant to test chaining those statements, but then I got too excited by the new button behavior. 🤷‍♀️

robincornett (approved):

assets/js/admin/orders/order-overview/_refund.js
```diff
@@ -253,7 +253,7 @@ $(document.body).on( 'click', '#edd-submit-refund-submit', function(e) {
 				url_target.hide();
 
 				$('#edd-submit-refund-status').show();
-				$( '#edd-submit-refund-submit' ).attr( 'disabled', false ).removeClass( 'updating-message' );
+				$( '#edd-submit-refund-submit' ).attr( 'disabled', false );
```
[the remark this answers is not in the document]
ashleyfae: `button-primary` also needs to be re-added on line 269.

ashleyfae (changes requested):
I like what we've decided here. :raised_hands: Just a couple minor things.

robincornett (approved):

PR #8642, opened 2021-05-21
Move admin notice dismissal back to user meta

### LINKED ISSUE #8633: 3.0 - All dismissed admin notices come back, due to moving dismiss state from user meta to options
### Expected behavior

When I dismiss an admin notice, I expect it will not come back.

### Actual behavior

When moving from EDD 2.x to release/3.0, all dismissed notices come back again.

1) Start on EDD 2.10. Fresh install might be best so you can see the nginx notice.
2) Go to an EDD page (like Downloads > Settings) and you'll see the nginx notice (provided you're on nginx).
3) Dismiss the notice. Refresh the page to make sure it's gone.
4) Switch to `release/3.0` branch.
5) Note that you might need to comment out this return statement: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/class-edd-notices.php#L246 (because it's picking up `edd_is_dev_environment()` for me, at least).
6) Refresh the page.
7) Note how the admin notice has returned.

PHP Version:

WordPress Version:

In EDD 2.x, dismissals were stored in user meta: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/admin/class-edd-notices.php#L321

In release/3.0, they've been moved to options: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/class-edd-notices.php#L173

I couldn't find a corresponding issue for this, so maybe it was just an oversight. But for now we should revert back to meta, and if there's a reason to move to options, that should be done with more careful planning and possibly a migration.

This also affected a [Software Licensing PR](https://github.com/easydigitaldownloads/EDD-Software-Licensing/pull/1777) that was using EDD core dismissal logic. Right now it doesn't work in both EDD 2.x and 3.0 because of the different dismissal storage.

robincornett:
Fixes #8633

Proposed Changes:
1. Changes dismissal notice back to user meta instead of a site option to be consistent with EDD 2.x.

### DIFF
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -215,7 +215,7 @@ class EDD_Notices {
 
 		// Checkout page is missing
 		$purchase_page = edd_get_option( 'purchase_page', '' );
-		if ( empty( $purchase_page ) || ( 'trash' === get_post_status( $purchase_page ) ) && ! edd_get_option( 'dismissed_notice_checkout' ) ) {
+		if ( empty( $purchase_page ) || ( 'trash' === get_post_status( $purchase_page ) ) ) {
 			$this->add_notice( array(
 				'id'             => 'edd-no-purchase-page',
 				'message'        => sprintf( __( 'No checkout page is configured. Set one in <a href="%s">Settings</a>.', 'easy-digital-downloads' ), admin_url( 'edit.php?post_type=download&page=edd-settings&tab=general&section=pages' ) ),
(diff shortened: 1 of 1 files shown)


ashleyfae (commented):
All the notices themselves also need updating when they check the dismissal status. Example: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/d70502bdee8d18a5a02a07fc26cad60715b26879/includes/admin/class-edd-notices.php#L218

robincornett:
I ended up removing the checks that I found (for checkout and uploads) because both of those notices have `is_dismissible` set to `false`. Only found these two, though.

ashleyfae (approved):
I didn't realize we'd actually removed the ability to dismiss the nginx notice. :facepalm:

PR #8647, opened 2021-05-24
Update order migration cart details, discount calculation

### LINKED ISSUE #8643: 3.0 - some [older] discount information may be lost during migration
### Expected behavior
When I migrate orders from 2.x to 3.0, I expect the data to be consistent.

### Actual behavior
Some orders with discounts applied end up not properly showing/applying the discount after migration:

image

The discount applied here is 15% off and the order total should be $41.65, which is what it is in 2.10.

I think to reproduce the behavior, you have to have a dataset with orders (with discounts) dating prior to sometime in early December, 2013. In the database I'm looking at, this appears to be approximately when data stops being inaccurate.

PHP Version: 7.3.5
WordPress Version: 5.7.2

Here's the `cart_details` from the payment meta of an affected order:

```
[cart_details] => a:1:{i:0;a:5:{s:4:"name";s:29:"PayPal Pro and PayPal Express";s:2:"id";s:3:"162";s:11:"item_number";a:2:{s:2:"id";s:3:"162";s:7:"options";a:0:{}}s:5:"price";s:5:"49.00";s:8:"quantity";i:1;}}
```

In 3.0, this order appears this way in the database:

In `edd_orders` (limited to relevant columns):
|id |status |subtotal |discount |tax |total |
--- | --- | --- | --- | --- | ---
|107132631 |complete |0.000000000 |0.000000000 |0.000000000 |49.000000000 |

In `edd_order_items`:
|id |order_id |product_id |product_name |price_id |cart_index |quantity |amount |subtotal |discount |tax |total |
--- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | ---
|835 |107132631 |162 |PayPal Pro and PayPal Express — Single Site |0 |0 |1 |49.000000000 |49.000000000 |0.000000000 |0.000000000 |49.000000000 |

And in `edd_order_adjustments`:
|id |object_id |object_type |type_id |type |description |subtotal |tax |total |
--- | --- | --- | --- | --- | --- | --- | --- | ---
|60 |107132631 |order |21916 |discount |TURKEY2012 |0.000000000 |0.000000000 |0.000000000 |

The order item row is nearly correct, although it's not showing the discount. But the amount and subtotal are properly populated, I think because this code accounts for the array key `price` as the `item_price` fallback:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/677d4f2187938bea4a1680422a84dfd7ff908420/includes/admin/upgrades/v3/class-data-migrator.php#L816-L819

However, the order subtotal/discount are calculated from `$cart_item['subtotal']`, which is present in newer orders, but not in the ones I'm looking at:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/677d4f2187938bea4a1680422a84dfd7ff908420/includes/admin/upgrades/v3/class-data-migrator.php#L468-L471

This causes the discount to be incorrectly calculated for the order because the `$subtotal` is 0:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/677d4f2187938bea4a1680422a84dfd7ff908420/includes/admin/upgrades/v3/class-data-migrator.php#L1086-L1087

As far as I can tell, I think we need to tweak the following:
- copy the order item logic (with `price` fallback) back to the order creation calculations, which I think will fix both the subtotal and the discounts on the order, and on the order adjustment.
- account for `$cart_item['discount']` possibly not being present in the cart details array, although if it's not, I'm not sure that it's possible to accurately account for the discount being applied to the individual item, because it could be an excluded item and the discount just applied to something else in the cart.

I _believe_ this is an issue with older orders, so refunds are not a likely scenario, but with this example order, it would be possible to attempt to use the new refund API to issue a refund for more than what the original order actually was, since the customer paid $41.65 but the refund would validate $49 (the gateway would likely balk here).

This also will affect customer/store statistics, I think.

robincornett:
Fixes #8643

Proposed Changes:
1. Create a new `maybe_update_cart_details` method in the order migration process to update the cart items for use in calculating the order amounts as well as the order items.
2. Move the discount calculation array to earlier in the migration, so that discounts can be calculated before the order is inserted. If the order total doesn't include a discount, but a discount has been calculated, update the order discount/total amounts before adding to the database. The discount/adjustment is still added to the database in its original location.

To Test:
Migrate orders with discounts applied, but not in the expected place in the cart details (`_edd_payment_meta`). Likely only older orders. If you are working with a database that's already been migrated once, empty (don't drop) all of the `edd_order*` tables and run this CLI command:

```
wp edd migrate_payments --force
```

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -440,6 +440,9 @@ class Data_Migrator {
 		// Some old EDD data has the cart details serialized, but starting with something other than a: so it can't be unserialized
 		$cart_details = self::fix_possible_serialization( $cart_details );
 
+		// Update cart details for consistency, defaults.
+		$cart_details = self::maybe_update_cart_details( $cart_details );
+
 		// Account for possible double serialization of the cart_details
 		$cart_downloads = isset( $payment_meta['downloads'] ) ? maybe_unserialize( $payment_meta['downloads'] ) : array();
 
@@ -465,10 +468,10 @@ class Data_Migrator {
 
 			// Loop through the items in the purchase to build the totals.
 			foreach ( $cart_details as $cart_item ) {
-				$subtotal += isset( $cart_item['subtotal'] ) ? (float) $cart_item['subtotal'] : 0;
-				$tax      += isset( $cart_item['tax'] )      ? (float) $cart_item['tax']      : 0;
-				$discount += isset( $cart_item['discount'] ) ? (float) $cart_item['discount'] : 0;
-				$total    += isset( $cart_item['price'] )    ? (float) $cart_item['price']    : 0;
+				$subtotal += $cart_item['subtotal'];
+				$tax      += $cart_item['tax'];
+				$discount += $cart_item['discount'];
+				$total    += $cart_item['price'];
 			}
 
 		} else {
@@ -812,31 +815,6 @@ class Data_Migrator {
 						$product_name .= ' — ' . $option_name;
 					}
 				}
-
-				// Get item price.
-				$cart_item['item_price'] = isset( $cart_item['item_price'] )
-					? (float) $cart_item['item_price']
-					: (float) $cart_item['price'];
-
-				// Get quantity.
-				$cart_item['quantity'] = isset( $cart_item['quantity'] )
-					? $cart_item['quantity']
-					: 1;
-
-				// Get subtotal.
-				$cart_item['subtotal'] = isset( $cart_item['subtotal'] )
-					? (float) $cart_item['subtotal']
-					: (float) $cart_item['quantity'] * $cart_item['item_price'];
-
-				// Get discount.
-				$cart_item['discount'] = isset( $cart_item['discount'] )
-					? (float) $cart_item['discount']
-					: 0.00;
-
-				// Get tax.
-				$cart_item['tax'] = isset( $cart_item['tax'] )
-					? (float) $cart_item['tax']
-					: 0.00;
 
 				$order_item_args = array(
 					'order_id'      => $order_id,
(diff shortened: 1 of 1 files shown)


ashleyfae (commented):
Still doing some final testing, but so far this is looking well. In release/3.0, my database had 66 order adjustments with a subtotal of `0`. With this branch, there's only 8, and they're all discounts used with entirely free products, so it makes sense.

ashleyfae:
Hm so I'm looking at one of my own payment records affected by this and I think the problem might actually be slightly different.

Here's the original record in the live EDD store:

Screenshot from 2021-05-28 12-39-31

Note there's a discount code that was used. This is a 50% off discount. And note the total price is $3.75. Right now the price of that product is $9.99, but I can believe it used to be $7.50 back in 2013. It definitely wasn't $3.75 though.

Here's a screenshot of the live PayPal payment, to confirm that the customer was in fact charged $3.75:

Screenshot from 2021-05-28 12-41-40

Here's the migrated record in a local copy of that database in EDD 3.0. Note that the discount code name is different just because I anonymized this database to change names. But all amounts remain the same.

Screenshot from 2021-05-28 12-41-12

Here's the same order in release/3.0:

Screenshot from 2021-05-28 13-06-59

My EDD 2.10 cart details array looks like this:

```
    [cart_details] => Array
        (
            [0] => Array
                (
                    [name] => UBB Reading Challenge Add-On
                    [id] => 11493
                    [item_number] => Array
                        (
                            [id] => 11493
                            [options] => Array
                                (
                                )

                            [quantity] => 1
                        )

                    [price] => 3.75
                    [quantity] => 1
                    [tax] => 0
                )

        )
```

---

So with my dataset, it looks like `price` in EDD 2.10 is the post-discount **total**.
But in your dataset, @robincornett, it sounds like your `price` in EDD 2.10 was the pre-discount **subtotal**?

So do we have two different problems here?

robincornett:
In the specific orders I'm looking at, the `cart_details` array has a `price` for each item, but it looks to be the full price before discounts. The discount is reflected in the `_edd_payment_meta['amount']` and `_edd_payment_total` (as those two values show the purchase amount after the discount is applied, anyway).

The release branch doesn't, I think, account for orders with incomplete metadata (in the original order I'm looking at, for example, the release branch shows a total for the order, but no subtotal or discount amount). Plus it's currently handling the cart items two different ways (once for getting the order information and then more detailed for the order item information), but if your `price` is an after discount amount and mine is not, and that's all the data we have, I'm not really sure how to account for that off the top of my head.

robincornett:
Converting this PR to a draft in favor of #8675, at least for now--some of the work in here has been moved to that PR and this one _may_ be resolved.

robincornett:
Closing because the other PR does fix what this one was addressing.

PR #8649, opened 2021-05-24
Set currency validation to strtoupper

### LINKED ISSUE #8646: 3.0 - migrated currency can be lowercase
### Expected behavior
When I view an order, I expect the currency to show as a symbol.

### Actual behavior
When I view a migrated order where the `_edd_payment_meta['currency']` is stored as lowercase (`usd` instead of `USD`), the currency is not converted/displayed as `$`.

image

1. Really you could make this happen by manually changing the currency for an order from `USD` to `usd`, but I'm encountering it as part of a site migration.

PHP Version: 7.3.5
WordPress Version: 5.7.2

I think there are a couple of places where we could tackle this, both for migrated orders and for future proofing:
- when migrating orders, I think we could `strtoupper` the currency value so that old orders have the currency properly stored.
- we could also `strtoupper` the `$currency` variable in `edd_currency_symbol()` to catch any orders with currency incorrectly stored (this fixes the display issue for an order with currency stored as `usd`).

Note that the order displays the same way in EDD 2.10, so this is not a regression; I think it's just incorrectly saved metadata.

robincornett:
Fixes #8646

Proposed Changes:
1. Add `validation` to order currencies in BerlinDB to cast currencies to uppercase.

### DIFF
--- a/includes/database/schemas/class-orders.php
+++ b/includes/database/schemas/class-orders.php
@@ -135,7 +135,8 @@ class Orders extends Schema {
 		array(
 			'name'       => 'currency',
 			'type'       => 'varchar',
-			'length'     => '20'
+			'length'     => '20',
+			'validate'   => 'strtoupper',
 		),
 
 		// payment_key


ashleyfae (approved):

ashleyfae:
Manually created an order via:

```php
$id = edd_add_order([
	'user_id' => 1,
	'customer_id' => 1,
	'email' => 'test@email.com',
	'mode' => 'test',
	'currency' => 'usd',
]);
```

Confirmed that the currency was stored in the DB as `USD`.

PR #8656, opened 2021-05-26
Add support for old get_x() methods #8655

### LINKED ISSUE #8655: 3.0 - EDD_Discount class missing `get_name()` method (backwards compat)
### Expected behavior

I expect all methods that exist in the `EDD_Discount` class in 2.10 to still exist/work in 3.0.

### Actual behavior

`$discount->get_name()` no longer works.

Run this code:

```php
$discount = edd_get_discount( $my_discount_id ); // replace ID
var_dump( $discount->get_name() );
```

It will work on EDD 2.10 but not on 3.0.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8655 

Proposed Changes:
1. Strip `get_` on incoming method calls. So `get_name()` becomes `name`. Then it falls into case 2 (below).
2. If a method doesn't exist but a property does, call the method. I updated the existing call for this logic. Previously it was passing `$args` to the property, which... doesn't work. It was treating the property like a method instead of a property. My fix actually treats it like a property.

To test:

See original issue. Test both `get_ID()` and `get_name()` methods.

### DIFF
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -432,12 +432,6 @@ class EDD_Discount extends Adjustment {
 	public function __call( $method, $args ) {
 		$property = str_replace( array( 'setup_', 'get_' ), '', $method );
 		if ( ! method_exists( $this, $method ) ) {
-			// If a function exists for the determined property, use that.
-			if ( method_exists( $this, $property ) ) {
-				return call_user_func( array( $this, $property ), $args );
-			}
-
-			// Otherwise, use the property directly.
 			return $this->{$property};
 		}
 	}
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -430,7 +430,7 @@ class EDD_Discount extends Adjustment {
 	 * @return mixed
 	 */
 	public function __call( $method, $args ) {
-		$property = str_replace( array( 'setup_', 'get_' ), '', $method );
+		$property = strtolower( str_replace( array( 'setup_', 'get_' ), '', $method ) );
 		if ( ! method_exists( $this, $method ) && property_exists( $this, $property ) ) {
 			return $this->{$property};
 		}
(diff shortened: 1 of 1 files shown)


LisaCee (changes requested):
The `get_name` function works for me, but I'm getting `NULL` for the `get_ID` function call.

This is the code I'm using for testing (thanks Debug console). It works for `get_name`.
$discount = edd_get_discount(1);
var_dump( $discount->get_ID());

ashleyfae:
Huh that's really weird. It's working for me. Sorry if this is a silly question, but do you definitely have a discount with the ID `1` ?

LisaCee:
> Huh that's really weird. It's working for me. Sorry if this is a silly question, but do you definitely have a discount with the ID `1` ?

Not a silly question, but yes. I tried it with a different discount in 3.0 and dumped everything.
```<?php
$discount = edd_get_discount(524);

var_dump($discount);
var_dump($discount->get_name());
var_dump($discount->get_ID());
```

resulted in
```
object(EDD_Discount)#2194 (21) { ["id":protected]=> int(524) ["name":protected]=> string(6) "10poff" ["code":protected]=> string(6) "10poff" ["status":protected]=> string(6) "active" ["amount_type":protected]=> string(7) "percent" ["amount":protected]=> string(12) "10.000000000" ["product_reqs":protected]=> array(0) { } ["scope":protected]=> string(6) "global" ["excluded_products":protected]=> array(0) { } ["product_condition":protected]=> string(3) "all" ["date_created":protected]=> string(19) "2021-05-04 18:09:29" ["date_modified":protected]=> string(19) "2021-05-04 19:47:01" ["start_date":protected]=> bool(false) ["end_date":protected]=> bool(false) ["use_count":protected]=> string(1) "2" ["max_uses":protected]=> string(1) "0" ["min_charge_amount":protected]=> string(11) "0.000000000" ["once_per_customer":protected]=> string(1) "0" ["db_global":protected]=> string(4) "wpdb" ["prefix":protected]=> string(3) "edd" ["last_error":protected]=> bool(false) } 

string(6) "10poff" 
NULL
```

LisaCee (approved):

robincornett (approved):
Tested:
- [x] 2.10
- [x] 3.0

PR #8660, opened 2021-05-26
Add Product filter to orders table & remove Sales Log #8658

### LINKED ISSUE #8658: 3.0 - Add "Download" (product) filter to orders table & remove Sales log

We should add a Download filter to the orders table.

### Justification or use case

Then we could completely remove the old Sales log (Downloads > Tools > Logs > change dropdown to "Sales"). This sales log has a lot of code/logic duplication, and instead of spreading that information across two different UIs, we could just merge it all into one. I think the only thing that's missing from the main Orders table is a Download filter. Then it could completely replace that log.

ashleyfae:
Fixes #8658

Proposed Changes:
1. Updates old "Sales" links to point to orders page with a product filter.
2. `class-payments-table.php`
    - Removed unused variables.
    - Add support for product ID filter.
    - Add new "Product" filter in advanced filters.
3. Redirects page requests from the old sales log to the orders page (supporting 3 existing query args in the redirect).
4. BerlinDB (`database/engine/class-query.php`) - Set the second `parse_groupby()` parameter to `true`, which just prefixes the column name with the table alias. This fixes an ambiguous column error when doing a join and groupby at the same time.
5. `database/queries/class-order.php`
    - Adds new `query_by_product()` parameter, and adds support for two new query args: `product_id` and `product_price_id`. If either of these are set, we join on the order items table to get that information.
    - Reworked the `query_by_country()` logic to stop merging the WHERE clause onto the JOIN. This was a bad idea and doesn't work when you have multiple joins (querying by product + country at the same time).

## Testing

- Test new product filter.
    - Test a regular product.
    - Test a variable priced product.
    - Test combining a product filter with a country filter.
- Attempt to visit old sales log manually and check that you are redirected to orders. (Might be easiest to visit it on `release/3.0`, set it up how you want, then check out this branch and refresh.)

### DIFF
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -1318,11 +1318,17 @@ function edd_render_stats_meta_box() {
 	}
 
 	$earnings = edd_get_download_earnings_stats( $post_id );
-	$sales    = edd_get_download_sales_stats( $post_id ); ?>
+	$sales    = edd_get_download_sales_stats( $post_id );
+
+	$sales_url = add_query_arg( array(
+		'page'       => 'edd-payment-history',
+		'product-id' => urlencode( $post_id )
+	), edd_get_admin_base_url() );
+	?>
 
 	<p class="product-sales-stats">
 		<span class="label"><?php _e( 'Sales:', 'easy-digital-downloads' ); ?></span>
-		<span><a href="<?php echo admin_url( '/edit.php?page=edd-tools&view=sales&post_type=download&tab=logs&download=' . $post_id ); ?>"><?php echo $sales; ?></a></span>
+		<span><a href="<?php echo esc_url( $sales_url ); ?>"><?php echo $sales; ?></a></span>
 	</p>
 
 	<p class="product-earnings-stats">
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -1328,7 +1328,7 @@ function edd_render_stats_meta_box() {
 
 	<p class="product-sales-stats">
 		<span class="label"><?php _e( 'Sales:', 'easy-digital-downloads' ); ?></span>
-		<span><a href="<?php echo esc_url( $sales_url ); ?>"><?php echo $sales; ?></a></span>
+		<span><a href="<?php echo esc_url( $sales_url ); ?>"><?php echo esc_html( $sales ); ?></a></span>
 	</p>
 
 	<p class="product-earnings-stats">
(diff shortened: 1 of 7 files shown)


includes/admin/downloads/metabox.php
```diff
@@ -1318,11 +1318,17 @@ function edd_render_stats_meta_box() {
...
+	$sales_url = add_query_arg( array(
+		'page'       => 'edd-payment-history',
+		'product-id' => urlencode( $post_id )
+	), edd_get_admin_base_url() );
+	?>
 
 	<p class="product-sales-stats">
 		<span class="label"><?php _e( 'Sales:', 'easy-digital-downloads' ); ?></span>
-		<span><a href="<?php echo admin_url( '/edit.php?page=edd-tools&view=sales&post_type=download&tab=logs&download=' . $post_id ); ?>"><?php echo $sales; ?></a></span>
+		<span><a href="<?php echo esc_url( $sales_url ); ?>"><?php echo $sales; ?></a></span>
```
robincornett, line 1331: Should `$sales` be escaped here?

includes/admin/tools.php
```diff
@@ -1614,6 +1614,35 @@ function edd_tools_sysinfo_download() {
...
+			'end-date'   => 'end-date'
+		);
+
+		foreach( $args_to_remap as $old_arg => $new_arg ) {
+			if ( ! empty( $_GET[ $old_arg ] ) ) {
+				$query_args[ $new_arg ] = urlencode( $_GET[ $old_arg ] );
+			}
+		}
+
+		wp_safe_redirect( esc_url_raw( add_query_arg( $query_args ), edd_get_admin_base_url() ) );
```
robincornett, line 1640: This redirect is working but the URL ends up having both the old and new parameters in it:

```
http://robin.local/wp-admin/edit.php?post_type=download&page=edd-payment-history&tab=logs&view=sales&download=12357&product-id=12357
```
Wondering if that's just WP being wonky (feel like I was wrestling with this in the Slack extension as well) or if it's fixable?

robincornett (commented):
Just a couple comments but this is working well for me.

Thinking that I just did a PR where I had to loop through order items and then orders to get orders for a specific product/download (maybe SL?) and this could really change that process for sure.

includes/admin/downloads/metabox.php
ashleyfae: Yes, I didn't do it because it wasn't new code, but I'll do it.

includes/admin/tools.php
ashleyfae: No -- it's me. I messed up my commas.

robincornett (approved):

PR #8675, opened 2021-06-01
issue/8671 - 3.0 - Migration - Use Tax and Total Metadata Values in Order Migration

### LINKED ISSUE #8671: 3.0 - Migration - Migrated order totals don't include fees
### Expected behavior

I expect the order total to be all order items + fees.

### Actual behavior

Fees are not being considered.

Screenshot from 2021-05-29 10-09-34

The total should be $25.

1) Start in EDD 2.10 (master).
2) Have an order that contains a fee. In the above example it was an order fee, but we should also test order item fees.
3) Checkout release/3.0 and migrate.
4) Note the order total.

PHP Version:

WordPress Version:

spencerfinnell:
Fixes #8671

Proposed Changes:
1. Use tax and total metadata values (if available) in order migration.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -1083,28 +1083,51 @@ class Data_Migrator {
 		}
 
 		if ( ! empty( $discounts ) && ( 'none' !== $discounts[0] ) ) {
-			foreach ( $discounts as $discount ) {
+			if ( 1 === count( $discounts ) ) {
+				$discount_code = reset( $discounts );
 
-				/** @var \EDD_Discount $discount */
-				$discount = edd_get_discount_by( 'code', $discount );
+				/** @var \EDD_Discount $discount_object */
+				$discount_object = edd_get_discount_by( 'code', $discount_code );
 
-				if ( false === $discount ) {
-					continue;
+				if ( $discount_object instanceof \EDD_Discount ) {
+					edd_add_order_adjustment(
+						array(
+							'object_id'     => $order_id,
+							'object_type'   => 'order',
+							'type_id'       => $discount_object->id,
+							'type'          => 'discount',
+							'description'   => $discount_object->code,
+							'subtotal'      => $order_discount,
+							'total'         => $order_discount,
+							'date_created'  => $date_created_gmt,
+							'date_modified' => $data->post_modified_gmt,
+						)
+					);
 				}
+			} else {
+				foreach ( $discounts as $discount_code ) {
 
-				edd_add_order_adjustment(
-					array(
-						'object_id'     => $order_id,
-						'object_type'   => 'order',
-						'type_id'       => $discount->id,
-						'type'          => 'discount',
-						'description'   => $discount->code,
-						'subtotal'      => $subtotal - $discount->get_discounted_amount( $subtotal ),
-						'total'         => $subtotal - $discount->get_discounted_amount( $subtotal ),
-						'date_created'  => $date_created_gmt,
-						'date_modified' => $data->post_modified_gmt,
-					)
-				);
+					/** @var \EDD_Discount $discount_object */
+					$discount_object = edd_get_discount_by( 'code', $discount_code );
+
+					if ( false === $discount_object ) {
+						continue;
+					}
+
+					edd_add_order_adjustment(
+						array(
+							'object_id'     => $order_id,
+							'object_type'   => 'order',
+							'type_id'       => $discount_object->id,
+							'type'          => 'discount',
+							'description'   => $discount_object->code,
+							'subtotal'      => $order_subtotal - $discount_object->get_discounted_amount( $order_subtotal ),
+							'total'         => $order_subtotal - $discount_object->get_discounted_amount( $order_subtotal ),
+							'date_created'  => $date_created_gmt,
+							'date_modified' => $data->post_modified_gmt,
+						)
+					);
+				}
 			}
 		}
 
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -455,10 +455,17 @@ class Data_Migrator {
 			return;
 		}
 
-		$subtotal = 0;
-		$tax      = 0;
-		$discount = 0;
-		$total    = 0;
+		$subtotal    = 0;
+		$order_tax   = 0;
+		$discount    = 0;
+		$order_total = 0;
+
+		// Track the total value of added fees in case the Order was initially migrated
+		// without _edd_payment_total or _edd_payment_tax and manual calculation was needed.
+		$order_fees_tax         = 0;
+		$order_fees_total       = 0;
+		$order_items_fees_tax   = 0;
+		$order_items_fees_total = 0;
 
 		// Retrieve the tax amount from metadata if available.
 		$meta_tax = isset( $meta['_edd_payment_tax'] )
@@ -466,13 +473,8 @@ class Data_Migrator {
 			: false;
 
 		if ( false !== $meta_tax ) {
-			$meta_tax = maybe_unserialize( $meta_tax );
-
-			if ( is_array( $meta_tax ) ) {
-				$tax = (float) $meta_tax[0];
-			} else {
-				$tax = (float) $meta_tax;
-			}
+			$meta_tax  = maybe_unserialize( $meta_tax );
+			$order_tax = (float) $meta_tax[0];
 		}
 
 		// Retrieve the total amount from metadata if available.
@@ -481,13 +483,8 @@ class Data_Migrator {
 			: false;
 
 		if ( false !== $meta_total ) {
-			$meta_total = maybe_unserialize( $meta_total );
-
-			if ( is_array( $meta_total ) ) {
-				$total = (float) $meta_total[0];
-			} else {
-				$total = (float) $meta_total;
-			}
+			$meta_total  = maybe_unserialize( $meta_total );
+			$order_total = (float) $meta_total[0];
 		}
 
 		// In some cases (very few) there is no cart details...so we have to just avoid this part.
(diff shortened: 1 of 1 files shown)


robincornett (commented):
This is working to fix the order totals for me on the orders I've found that were not getting the correct values from the cart details.

I see that the `$subtotal` is still being calculated from the cart details, or as the `$total - $tax` only if there are not cart details--should this be changed to run if the `$subtotal` is not calculated properly from the cart details? That's an issue I'm still seeing.

spencerfinnell:
@robincornett I moved this back to draft as I found some other inconsistencies when manual calculation needs to be done with fees.

spencerfinnell:
@robincornett This is ready for another look.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -923,11 +955,11 @@ public static function orders( $data = null ) {
 						// Reset any conditional IDs to be safe.
 						$refund_adjustment_id = 0;
 
-						$tax_rate = isset( $meta['_edd_payment_tax_rate'][0] )
-							? (float) $meta['_edd_payment_tax_rate'][0]
-							: 0.00;
```
spencerfinnell, line 928: This was causing incorrect tax calculations for Order Item fees. No other use of `get_calculated_tax()` referenced a context-specific tax rate. Removing it falls back to the previously determined rate information.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -455,35 +455,67 @@ public static function orders( $data = null ) {
...
-				$tax = (float) $meta['_edd_payment_tax'][0];
-			}
+			// Cart details cannot be retrieved.
+		} else {
 
-			$subtotal = $total - $tax;
+			// We lose the discount amount at this point... it remains 0.
 
+			// Calculate the subtotal if we have a total and tax amount.
+			if ( false !== $meta_tax && false !== $meta_total ) {
```
robincornett, line 516: I'm not sure this should be in the `else` (which would mean no `else` is required. In the orders I'm checking for a similar issue with discounts, this PR is correctly setting the order total, but not the subtotal, which is `0`. There are cart details but they're incomplete, so the order(s) are not getting to this point.

spencerfinnell: @robincornett So the cart details exist, but don't contain a `subtotal` entry for each item? I believe `release/3.0` would still result in a 0 subtotal?

Maybe we could check if subtotal is still `0` after the cart loop and subtract any found tax and discounts from the total and use that?

robincornett: I've logged it in issue #8643 ... it's somewhat related to what you're doing here, and that PR does fix the subtotal issue (mainly by making sure that the cart details are normalized before using, as they're used twice, but potentially adjusted before using in `release/3.0`).

Here's the cart details from the affected order (I do have more):

```
[cart_details] => a:1:{i:0;a:5:{s:4:"name";s:29:"PayPal Pro and PayPal Express";s:2:"id";s:3:"162";s:11:"item_number";a:2:{s:2:"id";s:3:"162";s:7:"options";a:0:{}}s:5:"price";s:5:"49.00";s:8:"quantity";i:1;}}
```

I think we need to fix both issues separately, but your PR should be finalized first and merged and then I can update mine, because I also was tinkering with order totals and your approach is better.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -1214,6 +1234,55 @@ public static function tax_rates( $data = null ) {
 		edd_add_adjustment( $adjustment_data );
 	}
 
+	/**
+	 * Normalizes and backfills legacy payment cart data.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @param array $cart_details Cart details.
+	 * @return array
```
robincornett, line 1243: Technically this could also return a string--I think that is the other possible type of `$cart_details` that could be passed to the function (same is true for the parameter I guess).

robincornett (commented):
With the cart details being normalized, the wonky orders I'm keeping an eye on appear to be fixed (discounts), so this may make the PR for #8643 unneeded. I'm curious to see how this PR affects the order @ashleyfae mentioned being broken in that PR.

Note that for those orders, the order item totals do _not_ include the discounts, so the refund modal shows a potential number to refund which is too high. However, the refund validator prevents that amount from being refunded, so I think it's something I could live with.

ashleyfae:
> I'm curious to see how this PR affects the order @ashleyfae mentioned being broken in that PR.

The order total is correct, but the discount amount is not.

See my notes here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8647#issuecomment-850374685

Here's how it looks in this PR:

Screenshot from 2021-06-03 14-57-42

I've ended up with:

Order:
- Subtotal: 3.75
- Discount: 0
- Tax: 0
- Total: 3.75

Order adjustment:
- Total: 1.88

Screenshot from 2021-06-03 15-11-12

Screenshot from 2021-06-03 15-12-09

Screenshot from 2021-06-03 15-12-38

I'm pretty sure what it's meant to be is:

Order:
- Subtotal: 7.50
- Discount: 3.75
- Tax: 0
- Total: 3.75

Order adjustment:
- Total: 3.75

It's a 50% off discount.

However.. the payment meta has no subtotal information or discount information. Here it is in full:

```
array(6) {
  ["currency"]=>
  string(3) "USD"
  ["downloads"]=>
  array(1) {
    [0]=>
    array(3) {
      ["id"]=>
      string(5) "11493"
      ["options"]=>
      array(0) {
      }
      ["quantity"]=>
      int(1)
    }
  }
  ["user_info"]=>
  array(6) {
    ["id"]=>
    int(0)
    ["email"]=>
    string(36) "redacted@easydigitaldownloads.com"
    ["first_name"]=>
    string(4) "Mrs."
    ["last_name"]=>
    string(7) "Kirsten"
    ["discount"]=>
    string(8) "goodwin4"
    ["address"]=>
    bool(false)
  }
  ["cart_details"]=>
  array(1) {
    [0]=>
    array(6) {
      ["name"]=>
      string(28) "UBB Reading Challenge Add-On"
      ["id"]=>
      string(5) "11493"
      ["item_number"]=>
      array(3) {
        ["id"]=>
        string(5) "11493"
        ["options"]=>
        array(0) {
        }
        ["quantity"]=>
        int(1)
      }
      ["price"]=>
      float(3.75)
      ["quantity"]=>
      int(1)
      ["tax"]=>
      float(0)
    }
  }
  ["tax"]=>
  int(0)
}
```

However, since the subtotal/discount data literally doesn't exist, we can't expect to get it perfect. Maybe this is just as good as it's going to be? :shrug:

robincornett:
@ashleyfae sorry, just curious--since that order is missing data, I'm curious as to how accurate it was after migration in `release/3.0` ... do you happen to know? Wondering if this PR makes it better or worse, as it's made my wonky orders better.

ashleyfae:
Here's how it looks in release/3.0:

Screenshot from 2021-06-03 15-47-58

Screenshot from 2021-06-03 15-49-13

TBH I kind of like having `0.00` when we don't actually know what the value is supposed to be.

LisaCee:
Lurking, but the code so far works to resolve the original issue for me.

robincornett (commented):
Running through another migration with this again and thinking about the discounts/total issues that @ashleyfae is seeing, which I think are the result of two things:
1. The `$cart_details` are being normalized now before they are ever used, which I think is important and needs to stay intact. The migration loops through the `$cart_details` twice, and it seems logical to not change the details between loops 1 and 2.
2. The discounts are being calculated fresh during the migration (lines 1102-1103) and as I think about it, this concerns me. It seems like the discount information should be pulled from the order metadata instead, and maybe not calculated at all if it doesn't exist. One scenario I can see the current setup causing problems with is if a discount wss created, say `DISCOUNT10` was 10% off the order when it was originally created, but someone changed it at some point to be 20% off, and just edited the existing discount instead of creating a new one. (Or if they created a discount and inadvertently typoed to have the wrong amount and fixed it after realizing the error because some orders came in.) In that case, the discount was applied at two different rates, but the migrator doesn't know that, and is just calculating it off of what's currently in the database. The orders with the old/incorrect discount amount will be recalculated and show an incorrect amount. Won't they?

So my feeling, as I understand how the migrator is working, anyway, is that the discount loop needs to be updated, similar to what this PR is doing for the tax/total values.

ashleyfae:
@robincornett Yes I have the same concern as you RE: discounts. So what are you thinking we do if the discount amount isn't recorded in 2.x?

robincornett:
Pushed an update to tweak how the discount migration is handled and I'm cautiously optimistic. Based on a conversation with @ashleyfae, we decided to use the order data for the discount amounts, instead of recalculating them, as much as possible--basically, if there is only one discount on the order, we're using the order information we have. If there are multiple discounts on the order, we do have to run the calculations.

This does result in orders with missing discount information showing a `$0.00` value for the discount, but the order **total** is correct, which is more important. It's actually not different than what is known in EDD 2.x, as orders in 2.x show only the discount code anyway, not the discount amount. The discount amount is implied in the difference between the order item amounts and the total.

So the orders I am spot checking with missing discount information are showing correctly:

image

It may just _look_ off because in 3.0, we expect to have more information than we actually do.

spencerfinnell:
Is it worth adding some metadata to the adjustment if we know we can't determine the true value? Then we could show something in the UI explaining why it shows the $0.00 amount.

spencerfinnell:
@robincornett Yep, any sort of indicator that can be utilized if we need to. We wouldn't even need to implement it in the UI right away but it would give us something to use if there ends up being too much confusion down the road.

robincornett:
I decided to do the order ID as the `meta_value` instead of just a boolean, only because it helps me looking at the database table. Happy to change if that's annoying as I understand it's not technically necessary.

ashleyfae:
My order with the new changes:

Screenshot from 2021-06-15 15-25-15

And confirmed I got the adjustment meta record.

Still checking a few other orders, but I like the direction we've chosen. :+1:

ashleyfae (approved):

PR #8676, opened 2021-06-02
Multi currency enhancements

ashleyfae:
- Payments export
    - Remove currency symbol from header. Orders may be in multiple currencies.
    - Format order amount according to order currency.
- Reports
    - Added new "Currency" filter. This only appears if you have orders with more than 1 currency.
    - Reports themselves are updated to support converting all amounts to base shop currency.
- 3.0 Migration
    - Added filters.
- New currency classes: `EDD\Currency\Currency` and `EDD\Currency\MoneyFormatter`. This is an attempt to make it easier for different currencies to have different settings at the same time (different thousands separator, decimals separator, etc.).
- Existing currency functions have been moved to `includes/currency/functions.php`.
- Database
    - A new `rate` column has been added to order adjustments, order items, order transactions, and orders. This is the exchange rate, relative to the base shop currency. This is the most "useless" feature for EDD core to have, but having this in core makes everything work so much better than trying to hack it into the add-on itself.
- Templates
    - Templates have been updated to ensure we always pass the order currency to `edd_currency_filter()` instead of falling back to the default (store currency).


ashleyfae:
@LisaCee and @robincornett You're welcome to start testing this. For now just look at reports on this branch vs `release/3.0`. All the numbers should hopefully be identical. The one exception being Downloads > Earnings by Taxonomy > Monthly Sales Average. I noticed it was formatted for currency, but it's actually meant to be a sales count, so I turned it into an integer.

Screenshot from 2021-06-30 16-55-35

vs

Screenshot from 2021-06-30 16-56-27

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -405,6 +405,20 @@ public static function orders( $data = null ) {
 			$user_info = array();
 		}
 
+		/**
+		 * Last change to filter payment meta before we use it!
```
robincornett, line 409:
```suggestion
		 * Last chance to filter payment meta before we use it!
```

includes/currency/functions.php
```diff
@@ -0,0 +1,120 @@
...
+ * @return string The currency code
+ */
+function edd_get_currency() {
+	$currency = edd_get_option( 'currency', 'USD' );
+	return apply_filters( 'edd_currency', $currency );
+}
+
+/**
+ * Given a currency determine the symbol to use. If no currency given, site default is used.
+ * If no symbol is determine, the currency string is returned.
```
robincornett, line 87:
```suggestion
 * If no symbol is determined, the currency string is returned.
```

includes/reports/data/customers/class-top-five-customers-list-table.php
```diff
@@ -125,6 +134,32 @@ public function get_columns() {
...
+
+		$currency          = '';
+		$selected_currency = Reports\get_filter_value( 'currencies' );
+		if ( ! empty( $selected_currency ) && 'convert' !== $selected_currency ) {
+			$currency = $selected_currency;
+		}
+
+		$value = edd_currency_filter( edd_format_amount( $item[ $column_name ] ), $currency );
+
+		return apply_filters( 'edd_customers_column_' . $column_name, $value, $item['id'] );
```
robincornett, line 160: If the `$column_name` will be `spent` here (as I read the conditional at line 148 it seems that the rest of the code is running only if it is `spent`, can the return line instead be:

```suggestion
		return apply_filters( 'edd_customers_column_spent', $value, $item['id'] );
```

I like dynamic filters but it feels like this really isn't one and it would be easier to search for it.

includes/reports/data/payment-gateways/class-gateway-stats-list-table.php
```diff
@@ -97,35 +98,36 @@ public function get_data() {
...
 
 			$refunded_count = $stats->get_gateway_sales( array(
-				'range'   => $filter['range'],
-				'gateway' => $gateway_id,
-				'status'  => array( 'complete' ),
-				'type'    => array( 'refund' ),
+				'range'    => $filter['range'],
+				'gateway'  => $gateway_id,
+				'status'   => array( 'complete' ),
+				'type'     => array( 'refund' ),
```
robincornett, line 120: Commenting here because I don't know where the heading is defined--but the heading on the table is "Refunded Sales" and I feel like that is not technically correct. I have two Euro payments, and one of them is partially refunded, but it's not being included in the `$complete_count` (which is maybe a different question), but this count is technically the number of _refunds issued_, not the number of sales which have been refunded. So if I have one order in a gateway, and then issue two separate partial refunds on the order, this column will show `2`, completed orders will show `0`, and the total sales will be `1`. It is a little sideways anyway but I'm basically wondering if we can change the heading to say just "Refunds Processed" or something like that?

robincornett (commented):
One error on initial load, which could be normal, and some questions, but so far, it seems to be working pretty seamlessly. Still testing.

includes/reports/data/payment-gateways/class-gateway-stats-list-table.php
ashleyfae: I think that's probably new issue territory. All I did in this class is add the currency logic. But what it's actually querying ('complete' refunds) hasn't changed. So I think we should open a new issue to look into that.

Don't milestone it for 3.0 though, as it's not a blocker to launch.

includes/reports/data/customers/class-top-five-customers-list-table.php
ashleyfae: Although you're right, I'm kind of inclined to leave it as is. The reason being that this filter is copied from the base class - `EDD_Customer_Reports_Table` which has this same dynamic filter, which is applied to all columns. I like keeping it this way to better show that they are actually the same filter, if that makes sense.

includes/admin/reporting/export/class-batch-export-payments.php
```diff
@@ -52,8 +52,8 @@ public function csv_cols() {
 			'products'     => __( 'Products (Verbose)', 'easy-digital-downloads' ),
 			'products_raw' => __( 'Products (Raw)', 'easy-digital-downloads' ),
 			'skus'         => __( 'SKUs', 'easy-digital-downloads' ),
-			'amount'       => __( 'Amount', 'easy-digital-downloads' ) . ' (' . html_entity_decode( edd_currency_filter( '' ) ) . ')',
-			'tax'          => __( 'Tax', 'easy-digital-downloads' ) . ' (' . html_entity_decode( edd_currency_filter( '' ) ) . ')',
+			'amount'       => __( 'Amount', 'easy-digital-downloads' ),
+			'tax'          => __( 'Tax', 'easy-digital-downloads' ),
```
robincornett, line 56: On this export, the currency column already exists, but it's separate from the amount/tax columns. Could it be moved to the same location?

ashleyfae: Done via https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8676/commits/c510dfaba2bcd9a11672e1001b479fb1d5296049

includes/admin/reporting/reports.php
```diff
@@ -1038,10 +1077,11 @@ function edd_register_refunds_report( $reports ) {
 			'label' => __( 'Average Time to Refund', 'easy-digital-downloads' ),
 			'views' => array(
 				'tile' => array(
-					'data_callback' => function () use ( $dates ) {
+					'data_callback' => function () use ( $dates, $currency ) {
 						$stats = new EDD\Stats();
 						return $stats->get_average_refund_time( array(
-							'range' => $dates['range'],
+							'range'    => $dates['range'],
+							'currency' => $currency
```
robincornett, line 1084: I'm not sure that the `get_average_refund_time` method is taking the currency into account. I've got orders in two different currencies and have issued some refunds in just one of them. When I filter by the other one, all of the tiles are showing 0/-- except for this one, which is still showing the overall average refund time (unchanged from the other views).

robincornett:
Noting that locally, `test_get_filters_should_return_records_for_all_official_filters` is failing for me. It looks like `currencies` maybe should be added to the `$expected` array?

ashleyfae:
> Noting that locally, `test_get_filters_should_return_records_for_all_official_filters` is failing for me. It looks like `currencies` maybe should be added to the `$expected` array?

Good catch. https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8676/commits/abb578ea44619668c99a65d50fe7c34a03c54d3d

includes/admin/reporting/reports.php
ashleyfae: I'll investigate this tomorrow.

ashleyfae: Fixed via https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8676/commits/6ea2552f8da0845f5a13809a4a997167f602b5bb

robincornett (approved):
I _think_ this is solid (and I'm totally impressed). At this point, I think I can approve this. I wonder about maybe updating the docblock for `edd_currency_filter` to maybe explain that when that function is used to display _existing_ orders, that the order currency should always be passed as a parameter. Because in places like the ajax cart function, the currency isn't passed, but it doesn't need to be, because it should use the store setting.

I think I would also consider adding as a feature request (new issue would be fine) to add a currency filter to the advanced filters on the orders list table, if the store has orders in more than one currency:

image

ashleyfae:
Good feedback. :+1: 

> I think I would also consider adding as a feature request (new issue would be fine) to add a currency filter to the advanced filters on the orders list table, if the store has orders in more than one currency:

That might be something to go in the add-on instead (this PR is designed to go alongside the add-on), but I'll keep this in mind. Doesn't have to be done immediately. We'll eventually put it wherever is easiest.

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8676/commits/4ea85d20aeb718c6082aa4a8b84ae3c31f6c0680 updates docblocks.

PR #8678, opened 2021-06-03
Add filter to pagination function

davidsherlock:
Fixes #

Added filter to edd_pagination args.

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -1128,6 +1128,8 @@ function edd_pagination( $args = array() ) {
 	);
 
 	$args = wp_parse_args( $args, $defaults );
+	
+	$args = apply_filters( 'edd_pagination_args', $args );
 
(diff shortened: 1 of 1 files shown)


includes/template-functions.php
```diff
@@ -1128,6 +1128,8 @@ function edd_pagination( $args = array() ) {
 	);
 
 	$args = wp_parse_args( $args, $defaults );
+	
+	$args = apply_filters( 'edd_pagination_args', $args );
```
ashleyfae, line 1132: Can you add some PHPDocs for this please? Here's an example of how they're formatted: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/privacy-functions.php#L881-L889 Just document the parameter and `@since 3.0`

robincornett (approved):

ashleyfae (approved):

PR #8682, opened 2021-06-03
issue/7911 - 3.0 - Provide the Ability to Link Directly to Reports

### LINKED ISSUE #7911: 3.0 - Improve report direct linking capability.
## Enhancement
In order to allow direct linking to reporting screens not only within the UI but between users, we should support direct links to filtered reports. Currently this data is handled in some session data, which is not shareable, or repeatable without someone entering in all the filter data again.

# Original Report
### Expected behavior

When I go to Downloads > Downloads and click on the `Earnings` link for that download, I expect to be taken to the report page that shows the overview for that download. It should be the equivalent of doing this manually:

- Go to Downloads > Reports.
- Clicking on "Downloads" tab.
- Changing the "All Downloads" dropdown to the specific download in question.
- Clicking "Filter".

### Actual behavior

I do get taken to the Downloads report page, but my filter settings are from whatever I'd last saved. So if I'd previously had this filter set to `All Downloads`, then clicking the earnings link just takes me to the report showing `All Downloads`. It seems like the `download-id=36386` query arg is just ignored.

1) First set your report settings. Go to Downloads > Reports > Downloads and set the download filter to `All Downloads`. Click `Filter` to save.
2) Go to Downloads > Downloads and click on the `Earnings` link for a specific download.
3) You'll be taken to the reports page, but `All Downloads` is still selected. It should have filtered to the specific download you clicked on.

PHP Version:

WordPress Version:

spencerfinnell:
Fixes #7911 

Proposed Changes:
1. Provide the ability to link directly to reports.


includes/reports/reports-functions.php
```diff
@@ -384,73 +384,86 @@ function get_filter_value( $filter ) {
...
+
+			/**
+			 * Filters the value of a report filter.
+			 *
+			 * @since 3.0
+			 *
+			 * @param string $value Report filter value.
+			 * @param string $filter Report filter.
+			 */
+			$value = apply_filters( 'edd_reports_get_filter_value', $value, $filter );
```
robincornett, line 436: Is this `apply_filters` within the `default` case? It seems like it should be at line 439 so that it applies to all?

includes/reports/reports-functions.php
```diff
@@ -384,73 +384,86 @@ function get_filter_value( $filter ) {
...
-	return "reports:filter-{$filter}:site-{$site}:user-{$user}";
-}
+	/**
+	 * Filters registered report filters that should be persisted across views.
+	 *
+	 * @since 3.0
+	 *
+	 * @param array $filters List of registered filters to persist.
+	 */
+	$filters = apply_filters( 'edd_reports_get_persisted_filters', $filters );
```
robincornett, line 464: I would vote to `return apply_filters( ...` but that's a very minor thing.

robincornett (commented):
This all seems to be working for me, although I'm noticing some wonkiness (like the tax thing you mentioned) which would be outside of this PR.

One thing I'm seeing with some reports registered by extensions is that they are not picking up the persistent filters (date). Here's a URL I end up on but the date is defaulting to the last 30 days:

```
https://easydigitaldownloads.local/wp-admin/edit.php?post_type=download&page=edd-reports&range=last_30_days&filter_from=2021-05-04&filter_to=2021-06-04&view=software_licensing_upgrades
```

If I manually reset the date filter, the URL becomes:

```
https://easydigitaldownloads.local/wp-admin/edit.php?post_type=download&page=edd-reports&view=software_licensing_upgrades&filter_from=2017-01-01&filter_to=2017-12-31&range=other
```

... the parameters are in a different order but I don't see anything that would cause the date filter not to work, unless it's something in SL?

I'd also like to clarify how "persistent" the persistent filters are--can you confirm that they only persist as long as you are within the reports screen? I clicked a link to view a download or customer and when I went back to the reports, the date reverted to the last 30 days.

includes/reports/reports-functions.php
spencerfinnell: @robincornett My thinking here was an effort to reduce the potential for extension developers to accidentally break our built in filters. It can always be moved out at a later date but this gives us a bit of a buffer initially.

includes/reports/reports-functions.php
spencerfinnell: I switched to this style to reduce the size of future diffs if we need to do anything with the filtered value.

spencerfinnell:
> I'd also like to clarify how "persistent" the persistent filters are--can you confirm that they only persist as long as you are within the reports screen? I clicked a link to view a download or customer and when I went back to the reports, the date reverted to the last 30 days.

@robincornett Correct. They persist within the current "session" of viewing reports. Once you leave the report area and revisit the default filters will apply again (just the date at this point).

spencerfinnell:
Pushed a fix to address @robincornett's original error. It was reproducible via:

1. Set a non-default date range (anything but "Last 30 days")
2. Visit a report that does not include a date filter (such as All Access) and submit the filters
3. View another report that does include date filters

The recent commit outputs the values of the required URL querystring arguments for a persisted filter if they are not registered for the current report.

robincornett:
I'm experiencing odd behavior with the latest commit and the only way I can think to explain is that the date filters are now overly persistent? I am unable to change the date range without modifying the URL directly.

If I start on the main reports screen, and then enter a custom date range, when the page loads to show that, the date drop down shows Custom, but the date inputs are hidden; the URL shows different dates (maybe just today?) and there is no range set, although the query parameter is there. If I try to change to a predefined range, it does not work--the URL dates adjust slightly (oddly the to date appears to increase by one with each load?), but the range is still empty and the dates are wrong.

spencerfinnell:
@robincornett Reverting that last commit and drafting this until I'm able to focus on this more. Anyone is welcome to pick up from here though.

cklosowski:
Just updated this with the latest from `release/3.0`. Let's take this for another spin @robincornett

cklosowski:
I think I'm good with this, even with the small quirks for reports that don't do date filters resetting things. Merging.

cklosowski (approved):

PR #8688, opened 2021-06-08
Set single price mode to true only if download has variable prices

### LINKED ISSUE #8640: Multi purchase mode can be enabled even without variable pricing
### Expected behavior
Enabling multi purchase mode BUT variable prices disabled to be treated as single purchase mode.

### Actual behavior
Multi purchase mode is enabled.

1) https://share.getcloudapp.com/2NuwGyjA
2) Check Variable prices box then check the Multiple purchase box.
3) Uncheck the Variable prices box then save the download.
This is affecting the Custom Prices extension as if you follow the steps above for a product with Custom price it will not show the custom input box and the download will be free.

PHP Version: 7.4

EDD Version (or branch): latest

WordPress Version: latest

I have a download with Enable multi-option purchase mode unchecked and when I run edd_single_price_option_mode( 398 ) is returning false cause the meta value is not in the post_meta and when you have it checked it returns true but its incorrect (based on the method name) as single mode is not enabled since I just checked the multi price mode

robincornett:
Fixes #8640

Proposed Changes:
1. Adds a `has_variable_prices` check to `is_single_price_mode`.

### DIFF
--- a/includes/class-edd-download.php
+++ b/includes/class-edd-download.php
@@ -339,7 +339,7 @@ class EDD_Download {
 	 */
 	public function is_single_price_mode() {
 
-		$ret = get_post_meta( $this->ID, '_edd_price_options_mode', true );
+		$ret = $this->has_variable_prices() && get_post_meta( $this->ID, '_edd_price_options_mode', true );
 
 		/**
 		 * Override the price mode for a download when checking if is in single price mode.


ashleyfae (approved):
- [x] Reproduced original issue.
- [x] Confirmed fix.
- [x] Confirmed that multi option purchase mode still works when enabled.

mihaijoldis (commented):
Good here

PR #8703, opened 2021-06-15
Show fees as negative if the raw subtotal is less than zero

### LINKED ISSUE #8695: 3.0 - negative fees are showing as positive on orders screen
### Expected behavior
When I view an order in the admin with a negative fee applied, I expect to be able to tell that it's negative.

### Actual behavior
When I view an order like this, the fee doesn't show that it's a discount:

image

1. Create or find an order with a negative fee applied.
2. View it in the admin.

PHP Version: 7.3.5
WordPress Version: 5.7.2

The fee does show correctly on the order receipt.

The fee is negative when the data is localized in `edd_order_details_overview`. It is not negative when it's sent to the template, so I'm guessing that somewhere in the JS formatting that it got cast to a positive number.

robincornett:
Fixes #8695

Proposed Changes:
1. Add a second check in the order adjustment template to show negative fees as negative based on the original, non-formatted subtotal value.

### DIFF
--- a/includes/admin/views/tmpl-order-adjustment.php
+++ b/includes/admin/views/tmpl-order-adjustment.php
@@ -53,7 +53,7 @@ $view_url = edd_get_admin_url(
 </td>
 
 <td class="column-right" data-colname="<?php esc_html_e( 'Amount', 'easy-digital-downloads' ); ?>">
-	<# if ( 'fee' !== data.type ) { #>&ndash;<# } #>{{ data.subtotalCurrency }}
+	<# if ( 'fee' !== data.type || data.subtotal < 0 ) { #>&ndash;<# } #>{{ data.subtotalCurrency }}
 </td>
 
 <input type="hidden" value="{{ data.objectId }}" name="adjustments[{{ data.id }}][object_id]" />


LisaCee (approved):
This works for existing and new orders. 

Can you explain what the `#` is doing here? I haven't noticed that syntax before.

robincornett:
I _think_ the `<#` and `#>` are comparable to `<?php` and `?>` in straight HTML > PHP, and this syntax is because of how the templates are loaded (JS)? @spencerfinnell would be the real one to ask--I'm just inferring based on what I'm seeing, honestly.

spencerfinnell:
@LisaCee @robincornett That is special syntax for the templating engine that is used by the JavaScript. It is powered by [Underscore's](https://underscorejs.org/#template) template utility however the standard "evaluate" delimiter `<%` is a very old school ASP-style PHP opening tag that can mess up on some WordPress installs with older PHP configurations.

WordPress adjust the delimiters slightly to the syntax you see in the `tmpl` files: https://github.com/WordPress/WordPress/blob/master/wp-includes/js/wp-util.js#L24-L36

https://core.trac.wordpress.org/ticket/22344

tl;dr it is syntax specific to the JavaScript templating engine used by Backbone and Underscore.

ashleyfae (approved):

PR #8710, opened 2021-06-21
Add a filter on core payment meta keys

### LINKED ISSUE #8673: 3.0 - add a filter on migrated core payment metadata
However, it might be nice to add a filter on the core payment meta array, the way we did on the user info, so that extensions could append to that to save a step in the migration process (when I worked on the user history extension ([PR](https://github.com/easydigitaldownloads/edd-user-history/pull/26)), I added steps to update/delete the payment meta once the user history was migrated, but with a filter in place I would not have to do that).

_Originally posted by @robincornett in https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8603#issuecomment-847978830_

robincornett:
Fixes #8673

Proposed Changes:
1. Create the `edd_30_core_payment_meta` filter, similar to the user info filter immediately following.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -772,14 +772,14 @@ class Data_Migrator {
 
 		/**
 		 * By default, this is what is stored in payment meta. These array keys are part of the core payment meta in 2.x
-		 * but are not needed as part of the order meta.
+		 * but are not needed as part of the order meta and will not be migrated.
 		 * Extensions can add their keys to this filter if they use the payment meta array to store data and have
 		 * established a migration process to keep the data intact with the new order tables.
 		 *
 		 * @since 3.0
 		 * @param array The array of payment meta keys.
 		 */
-		$core_meta_keys = apply_filters( 'edd_30_core_payment_meta', array(
+		$core_meta_keys = apply_filters( 'edd_30_payment_meta_keys_not_migrated', array(
 			'fees',
 			'key',
 			'email',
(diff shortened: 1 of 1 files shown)


includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -770,8 +770,16 @@ public static function orders( $data = null ) {
...
+		/**
+		 * By default, this is what is stored in payment meta. These array keys are part of the core payment meta in 2.x
+		 * but are not needed as part of the order meta.
+		 * Extensions can add their keys to this filter if they use the payment meta array to store data and have
+		 * established a migration process to keep the data intact with the new order tables.
+		 *
+		 * @since 3.0
+		 * @param array The array of payment meta keys.
+		 */
+		$core_meta_keys = apply_filters( 'edd_30_core_payment_meta', array(
```
LisaCee, line 782: I think you may be missing the closing `)` from `apply_filters()`.

robincornett: Ack! You are completely correct! Thanks for catching that!

LisaCee (approved):

includes/admin/upgrades/v3/class-data-migrator.php
ashleyfae, line 782: Can we rename the filter? Although it makes sense with how _we're_ using it, the point of the filter is to support third parties, in which case it's no long really "core" payment meta. Maybe something like:

`edd_30_payment_meta_to_exclude` or `edd_30_payment_meta_keys_not_migrated`

:shrug: Something like that. Just to better convey what actually happens to those keys (or _doesn't_ happen).

ashleyfae (approved):

PR #8713, opened 2021-06-21
Update discounts API query to use discount objects

### LINKED ISSUE #8708: 3.0 - Discount API endpoint not 3.0-ready
### Expected behavior

I expect the discount API to work in EDD 3.0.

### Actual behavior

I have not actually tested to confirm, but looking at [the code for it](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/api/class-edd-api.php#L1623-L1638) it does not look 3.0-ready. It runs `get_post()` on a discount ID and uses functions like `get_post_field()`.

I have not actually tested this yet - I just saw the code. But I suspect this is how you'd do it:

1) Set up a download that you know exists.
2) Query the API for that download specifically: `curl "https://yoursite.com/edd-api/discounts/?discount=DISCOUNT_ID&key=YOUR KEY&token=YOUR TOKEN`
3) Results may not be there at all, or may not be correct/complete.

PHP Version:

WordPress Version:

robincornett:
Fixes #8708

Proposed Changes:
1. Update the `get_discounts` method for the API to use the new discount objects.
2. Update the query parameters to correctly handle pagination.

To test:
To retrieve an array of discounts:

```
https://3dd.local/edd-api/discounts/?key=API_KEY&token=API_TOKEN
```

Add `&page=2` to get a second page of discounts (you should get new discounts; with 3.0, I did not).

To retrieve one discount:

Add `&discount=DISCOUNT_ID` to the URL and comment out the line indicated in the issue.

Noting that I opted to use the discount object properties directly instead of the helper functions because the general query was noticeably faster (1.8s to 0.03s).


includes/api/class-edd-api.php
```diff
@@ -1589,8 +1589,8 @@ public function get_discounts( $discount = null ) {
 			$paged     = $this->get_paged();
 			$per_page  = $this->per_page();
 			$discounts = edd_get_discounts( array(
-				'posts_per_page' => $per_page,
-				'paged'          => $paged,
+				'number' => $per_page,
+				'offset' => $paged * $per_page,
```
ashleyfae, line 1593:
```suggestion
				'offset' => ( absint( $paged ) - 1 ) * $per_page,
```

Otherwise it's setting an offset on page 1.

ashleyfae (changes requested):

ashleyfae (approved):

PR #8715, opened 2021-06-22
Introduce Tokenizer to replace some nonce verifications #8714

### LINKED ISSUE #8714: edd_ajax_add_to_cart() - Add alternative to nonce verification for better caching support

Ticket: https://secure.helpscout.net/conversation/1542727636/187014/

Our ajax add to cart buttons use a nonce ([background](https://easydigitaldownloads.com/development/2018/07/09/important-update-to-ajax-requests-in-easy-digital-downloads-2-9-4/#comment-887)). However, nonces don't work well with full page caching. After a certain period of time, they will expire, and the buttons will stop working until the cache expires and a new nonce gets generated.

A customer has recommended that we implement a custom token verification instead of nonces, which will still have the security benefits, but without the downside of causing problems with full page caching.

### Justification or use case

Better compatibility with sites using full page caching.

ashleyfae:
Fixes #8714 

Proposed Changes:
1. Introduces a new `Tokenizer` class for hashing a piece of data.
2. A timestamp and a token are now added to the "add to cart" buttons and sent via AJAX.
3. `edd_ajax_add_to_cart()` A few changes have been made here:
    - Exit early if `download_id` is not provided. (This is instead of nesting everything under an `if ( isset() )` check.)
    - If the timestamp + token are provided and the token is valid, consider the request verified.
    - Otherwise, fall back to still using the nonce. This ensures that we don't break anything if the add to cart links have been filtered and the filtered versions don't contain our timestamp + token.
    - If the token isn't valid and the nonce isn't valid, bail.

Security-wise, the one downside this method has over the caching method is that the tokens never actually expire. That's kind of the point, to make them work with caching. However, if a site does have problems with this endpoint being abused, the signing key could be regenerated, which would invalidate existing tokens.

To test:
1. Make sure add to cart links still work.
2. Manually delete `data-timestamp` and `data-token` from the cart buttons ([here](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/63fa0794453e91ad64578085d6f78b43ab95dc53/includes/template-functions.php#L187)) and make sure add to cart links still work.
3. Manually delete `data-nonce`, `data-timestamp`, and `data-token` from the cart buttons, and make sure add to cart links **do not** work.

### DIFF
--- a/assets/js/edd-ajax.js
+++ b/assets/js/edd-ajax.js
@@ -174,6 +174,8 @@ jQuery(document).ready(function ($) {
 			price_ids : item_price_ids,
 			post_data: $(form).serialize(),
 			nonce: nonce,
+			timestamp: $this.data( 'timestamp' ),
+			token: $this.data( 'token' )
 		};
 
 		$.ajax({
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -319,6 +319,7 @@ final class Easy_Digital_Downloads {
 		require_once EDD_PLUGIN_DIR . 'includes/shortcodes.php';
 		require_once EDD_PLUGIN_DIR . 'includes/admin/tracking.php'; // Must be loaded on frontend to ensure cron runs
 		require_once EDD_PLUGIN_DIR . 'includes/privacy-functions.php';
+		require_once EDD_PLUGIN_DIR . 'includes/utils/class-tokenizer.php';
 
 		if ( is_admin() || ( defined( 'WP_CLI' ) && WP_CLI ) ) {
 			require_once EDD_PLUGIN_DIR . 'includes/admin/add-ons.php';
--- a/includes/ajax-functions.php
+++ b/includes/ajax-functions.php
@@ -186,76 +186,78 @@ add_action( 'wp_ajax_nopriv_edd_remove_from_cart', 'edd_ajax_remove_from_cart' )
  * @return void
  */
 function edd_ajax_add_to_cart() {
-	if ( ! isset( $_POST['nonce'] ) ) {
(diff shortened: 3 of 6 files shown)


robincornett (approved):
- [x] Make sure add to cart links still work.
- [x] Manually delete data-timestamp and data-token from the cart buttons (here) and make sure add to cart links still work.
- [x] Manually delete data-nonce, data-timestamp, and data-token from the cart buttons, and make sure add to cart links do not work.

Note on the last that when the nonce check fails, the "add to cart" button just spins indefinitely on the front end. The log message shows in the debug log, but it seems off to give nothing to the customer attempting to make a purchase? However, that's the behavior in `master`, so it's not changed in this PR.

ashleyfae:
> Note on the last that when the nonce check fails, the "add to cart" button just spins indefinitely on the front end. The log message shows in the debug log, but it seems off to give nothing to the customer attempting to make a purchase? However, that's the behavior in `master`, so it's not changed in this PR.

Yeah I noticed that too. It's weird for sure. But out of scope for this PR.

LisaCee:
- [x] Make sure add to cart links still work.
- [ ] Manually delete data-timestamp and data-token from the cart buttons  and make sure add to cart links still work.
- [ ] Manually delete data-nonce, data-timestamp, and data-token from the cart buttons, and make sure add to cart links do not work.

LisaCee (approved):
- [x] Make sure add to cart links still work.
- [x] Manually delete data-timestamp and data-token from the cart buttons  and make sure add to cart links still work.
- [x] Manually delete data-nonce, data-timestamp, and data-token from the cart buttons, and make sure add to cart links do not work.

PR #8720, opened 2021-06-23
Add edd_post_insert_discount hook to edd_add_discount

### LINKED ISSUE #8718: 3.0 - No hook is fired after discount is added from admin
### Expected behavior
When I  create a new discount in the admin, I expect that custom fields are saved too.

### Actual behavior
`edd_add_discount()` function, that is used to create a discount, does not have any hooks. This is different behaviour from function `edd_update_discount()` that is used when discount is edited in admin, which has hooks so submitted value can be saved.

1) Add custom field to Add Discount screen.
2) Save discount.
3) Value of custom field is not saved.

EDD branch:
`release/3.0`

robincornett:
Fixes #8718

Proposed Changes:
1. Sets the original `$data` to a variable for use in the `edd_post_insert_discount` hook and adds it to `edd_add_discount`. (Pulled from `edd_store_discount`.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -29,6 +29,7 @@ function edd_add_discount( $data = array() ) {
 	$product_requirements = isset( $data['product_reqs'] )      ? $data['product_reqs']      : null;
 	$excluded_products    = isset( $data['excluded_products'] ) ? $data['excluded_products'] : null;
 	$product_condition    = isset( $data['product_condition'] ) ? $data['product_condition'] : null;
+	$pre_convert_args     = $data;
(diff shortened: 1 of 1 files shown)


LisaCee (approved):
I was able to replicate the original issue. This issue works to save the pushover notification settings to the `wp_postmeta` table with the correct `meta_value` of `on` (default) or `off`.

ashleyfae (commented):
I think we also need to remove the hook from `edd_store_discount()`.

`edd_store_discount()` calls `edd_add_discount()` so as it is, I believe the hook would now be running twice if someone calls the store function.

ashleyfae (approved):

PR #8726, opened 2021-06-29
Do not replace the cart contents if the ajax response is empty

### LINKED ISSUE #8719: edd_ajax_recalculate_taxes() deletes cart HTML when making Wallet deposit
## Bug Report

See https://github.com/easydigitaldownloads/edd-wallet/issues/76 for a description of the issue.

The cart disappears because the `edd_ajax_recalculate_taxes()` function just returns false if `! edd_get_cart_contents()`. However, in the JavaScript, we don't actually check the return value at all. As long as the ajax call returns 200 (which a `false` response technically does), we do this:

```js
jQuery( '#edd_checkout_cart_form' ).replaceWith( tax_response.html );
```

So we end up replacing the cart HTML with... nothing.

We have two options here:

1. Check that `tax_response.html` actually contains something before using.
2. Remove the `! edd_get_cart_contents()` check from `edd_ajax_recalculate_taxes()` and just let it process anyway (this shouldn't cause any harm).

PHP Version:

EDD Version (or branch): 2.10.5

WordPress Version:

robincornett:
Fixes #8719

Proposed Changes:
1. Update the JS to only replace the cart contents if the ajax response is not empty.

I did not update the corresponding ajax function as this is working, but I can.

To Test:
1. Enable two gateways (eg Test and Stripe).
2. Enable Wallet.
3. Enable taxes.
3. Make a deposit to the wallet.
4. On the checkout screen, switch to the second gateway.

In `master`, the cart will disappear. In this branch, it should not.


ashleyfae (approved):

LisaCee (approved):
Tested 2.x and 3.0

PR #8728, opened 2021-06-30
Apply count_where() to ensure date range is respected & query sales only #8727

### LINKED ISSUE #8727: 3.0 - EDD_Payments_Stats methods are returning unexpected results
### Expected behavior
When I use the `EDD_Payments_Stats` methods, I expect it to return correct data.

### Actual behavior
When I use the `get_stats` and `get_earnings` methods, the data is incorrect.

1. In EDD 3.0, enable the EDD Email Reports plugin.
2. Preview an email report email.

PHP Version: 7.3.5
WordPress Version: 5.7.2

Two of the functions that plugin has use the `EDD_Payments_Stats` class like this:

```php
/**
 * Returns the total earnings for today.
 *
 * @return string
 */
function edd_email_reports_daily_total() {
	$stats = new EDD_Payment_Stats();

	return edd_format_amount( $stats->get_earnings( 0, 'today', false ) );
}

/**
 * Returns the number of transactions for today.
 *
 * @return int
 */
function edd_email_reports_daily_transactions() {
	$stats = new EDD_Payment_Stats();

	return $stats->get_sales( false, 'today' );
}
```

There are others, but they're similar.

The second one is returning the total number of completed payments for the entire storefront; the first is returning `$0` (same for each total function in the plugin), even though it should not be.

ashleyfae:
Fixes #8727

Proposed Changes:
1. Re-adds the filter on `edd_count_payments_where`, which should make the date range apply correctly.
2. Updates an array of statuses to use `edd_get_gross_order_statuses()`.
3. Query sales only (no refunds).

### DIFF
--- a/includes/payments/class-payment-stats.php
+++ b/includes/payments/class-payment-stats.php
@@ -55,6 +55,9 @@ class EDD_Payment_Stats extends EDD_Stats {
 		}
 
 		if ( empty( $download_id ) ) {
+			// Global sale stats
+			add_filter( 'edd_count_payments_where', array( $this, 'count_where' ) );
+
 			if ( is_array( $status ) ) {
 				$count = 0;
 				foreach ( $status as $payment_status ) {
@@ -63,6 +66,8 @@ class EDD_Payment_Stats extends EDD_Stats {
 			} else {
 				$count = edd_count_payments()->$status;
 			}
+
+			remove_filter( 'edd_count_payments_where', array( $this, 'count_where' ) );
 		} else {
 			$this->timestamp = false;
 
@@ -146,8 +151,6 @@ class EDD_Payment_Stats extends EDD_Stats {
 		}
 
 		if ( empty( $download_id ) ) {
-			$statuses = array( 'complete', 'publish', 'revoked', 'refunded', 'partially_refunded' );
-
 			/**
 			 * Filters Order statuses that should be included when calculating stats.
 			 *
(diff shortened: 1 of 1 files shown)


robincornett (commented):
This looks to have fixed the `get_sales` method, but `get_earnings` is still returning $0.

ashleyfae:
> This looks to have fixed the `get_sales` method, but `get_earnings` is still returning $0.

Are your values cached by any chance? There is a transient on `get_earnings()`.

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/blob/b47f4fffdd549f9e0e020e49d8068e95f76e3782/includes/payments/class-payment-stats.php#L178

robincornett:
Yes, I cleared that transient to double check, and just did so again. Still showing an incorrect $0 earnings value.

ashleyfae:
Hm it seems to be working for me. Exactly which one isn't right for you? (or all of them?)

earnings

robincornett:
All of the earnings ones are broken:
image

but the orders are present:
image

ashleyfae:
@robincornett Can you do a little more debugging on your end to figure out what's going on? It's working fine for me so I'm not really sure where to start.

robincornett:
Found it! It's Recurring running this filter on the `edd_stats_earnings_args`:

https://github.com/easydigitaldownloads/edd-recurring/blob/27d3d12cc2bfe97f79e78363db13c13a4458887c/edd-recurring.php#L1497-L1508

It's completely overwriting the `post_status` value and that's borking 3.0. If I comment out the `apply_filters` line in the stats method, then I get numbers like this:

image

So I think I can approve this with a good will and then make a new issue in Recurring, as I assume we need to keep this filter in place.

robincornett (approved):

PR #8737, opened 2021-07-04
Introduce new edd_add_discount_form_after_use_once action

davidsherlock:
Fixes #

Added new action to add discount screen.

### DIFF
--- a/includes/admin/discounts/add-discount.php
+++ b/includes/admin/discounts/add-discount.php
@@ -210,6 +210,9 @@ defined( 'ABSPATH' ) || exit;
 						<span class="description"><?php esc_html_e( 'Prevent customers from using this discount more than once.', 'easy-digital-downloads' ); ?></span>
 					</td>
 				</tr>
+				
+				<?php do_action( 'edd_add_discount_form_after_use_once' ); ?>
+				
 			</tbody>
 		</table>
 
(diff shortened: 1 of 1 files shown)


includes/admin/discounts/add-discount.php
```diff
@@ -210,6 +210,9 @@
 						<span class="description"><?php esc_html_e( 'Prevent customers from using this discount more than once.', 'easy-digital-downloads' ); ?></span>
 					</td>
 				</tr>
+				
+				<?php do_action( 'edd_add_discount_form_after_use_once' ); ?>
```
ashleyfae, line 214: I know the others in this file don't have them, but can you please add a docblock with `@since 3.0` ?

includes/admin/discounts/add-discount.php
```diff
@@ -210,6 +210,14 @@
 						<span class="description"><?php esc_html_e( 'Prevent customers from using this discount more than once.', 'easy-digital-downloads' ); ?></span>
 					</td>
 				</tr>
+				
+				/**
```
ashleyfae, line 214: Looks good, but needs PHP tags. :stuck_out_tongue:

robincornett (commented):
Should this hook also be added to the `edit-discount.php` file?

davidsherlock (commented):
> Should this hook also be added to the `edit-discount.php` file?

One already exists that is similar before and below the status field, so no need for an additional one!

ashleyfae (approved):

PR #8745, opened 2021-07-12
Use a zip for the nginx rules test #8744

### LINKED ISSUE #8744: 3.0 - edd_is_uploads_url_protected() needs to check a zip file -- not a jpeg
### Expected behavior

I expect to not get this admin notice when my files are protected:

Screenshot from 2021-07-12 11-43-08

### Actual behavior

I've protected my files according to the docs, which by default only protect `zip` files. However, I'm getting the above admin notice because the test code in `edd_is_uploads_url_protected()` is checking a `.jpg` extension.

1) Make sure you have your nginx rules in place to protect downloads.
2) Manually run this code:
```php
delete_transient('edd_is_uploads_url_protected');
var_dump(edd_is_uploads_url_protected());
```
3) It will return `false`, meaning your files are not protected. But they are, according to the recommendation in our docs.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8744 

Proposed Changes:
1. Changes the extension to `.zip` to line up with our default nginx rules as per the docs.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1018,7 +1018,7 @@ function edd_is_uploads_url_protected() {
 		if ( wp_is_writable( $upload_path ) ) {
 
 			// Get the file path
-			$file_name = wp_unique_filename( $upload_path, 'edd-temp.jpg' );
+			$file_name = wp_unique_filename( $upload_path, 'edd-temp.zip' );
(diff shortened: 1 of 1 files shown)


rwebster85 (approved):
Okay.

robincornett (approved):

Issue #8764, opened 2021-07-19
3.0 - Third parameter passed to `edd_log_file_download_download_files` filter is in different format than in pre-3.0 versions


dimadin:
Before EDD 3.0, third parameter passed to `edd_log_file_download_download_files` filter contained post meta values of `file_download` log post. Those meta values [were retrieved](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/5657a2b4355874b30a2a08d0b79162a739f609ef/includes/admin/reporting/class-file-downloads-logs-list-table.php#L359-L374) by using `get_post_custom()` function. That function returns all meta values of a post. Which means parameter was array in format `'meta_key' => array( 'meta_value_1', 'meta_value_2'...)`. In EDD 3.0, that array was [changed](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/8128b5383cd3610778da33e27b8aa35f8b797701/includes/admin/reporting/class-file-downloads-logs-list-table.php#L162-L169) to format `'meta_key' => 'meta_value'`. Code that referenced meta value in a format `$meta['meta_key'][0]` doesn't work anymore, and it can create a bug.

PR #8765, opened 2021-07-19
issue/8764 - 3.0 - Convert meta values to arrays before passing to the filter

dimadin:
This is proposed fix for #8764.

### DIFF
--- a/includes/admin/reporting/class-file-downloads-logs-list-table.php
+++ b/includes/admin/reporting/class-file-downloads-logs-list-table.php
@@ -168,6 +168,11 @@ class EDD_File_Downloads_Log_Table extends EDD_Base_Log_List_Table {
 					'_edd_log_price_id'   => $log->price_id,
 				);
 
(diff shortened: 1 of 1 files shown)


ashleyfae (commented):
This will work, but also noting that compared to EDD 2.x, what we have here doesn't actually include _all_ the meta. So if someone has added custom meta, then it won't show up here.

I'm thinking we should start with this:

```php
$meta = edd_get_file_download_log_meta( $log->id );
```

That will return an array of all meta, Then merge in the values we already have here (`_edd_log_user_info`, `_edd_log_user_id`, etc.) as those are no longer stored in meta.

dimadin:
True, I wasn't paying attention. Now that I see, `_edd_log_customer_id` meta key is also missing in 3.0 but was there pre-3.0. It also needs to be manually added.

ashleyfae:
Closing this in favour of https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8808

PR #8775, opened 2021-07-22
issue/6923 - Include Plugin URI in the System Info report

### LINKED ISSUE #6923: Enhancement: System Info page to show plugin author or URI
### Expected behavior
On the System info page for plugins to show the author and/or the URL

### Actual behavior
At the moment it only shows the plugin name and sometimes its hard or takes a while to locate the developers of a specific plugin

1) Generate system report

PHP Version: All

EDD Version (or branch): Master

WordPress Version: latest

Would make it easier for Support staff to find out quickly who is the developer of plugin X and point the customer in their direction for issues regarding a 3`rd party plugin.

Something as simple as this could work https://www.screencast.com/t/VyEYy99FkeQA

mihaijoldis:
Fixes #6923

Proposed Changes:

  1. Add the Plugin URI or Author URI in the plugins report as it makes it easier to tell who is the developer for a plugin in support

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -1382,7 +1382,7 @@ function edd_tools_sysinfo_get() {
 			$plugin_url = $plugin['PluginURI'];
 		} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
 			$plugin_url = $plugin['AuthorURI'];
-		} else {
+		} elseif {
 			$plugin_url = $plugin['Author'];
 		}
 		if ( $plugin_url ) {
@@ -1406,7 +1406,7 @@ function edd_tools_sysinfo_get() {
 			$plugin_url = $plugin['PluginURI'];
 		} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
 			$plugin_url = $plugin['AuthorURI'];
-		} else {
+		} elseif {
 			$plugin_url = $plugin['Author'];
 		}
 		if ( $plugin_url ) {
@@ -1437,7 +1437,7 @@ function edd_tools_sysinfo_get() {
 				$plugin_url = $plugin['PluginURI'];
 			} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
 				$plugin_url = $plugin['AuthorURI'];
-			} else {
+			} elseif {
 				$plugin_url = $plugin['Author'];
 			}
 			if ( $plugin_url ) {
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -1382,6 +1382,8 @@ function edd_tools_sysinfo_get() {
 			$plugin_url = $plugin['PluginURI'];
 		} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
 			$plugin_url = $plugin['AuthorURI'];
+		} elseif ( ! empty( $plugin['Author'] ) ) {
+			$plugin_url = $plugin['Author'];
 		}
 		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . ' - ' . $plugin_url . "\n";
 	}
@@ -1401,6 +1403,8 @@ function edd_tools_sysinfo_get() {
 			$plugin_url = $plugin['PluginURI'];
 		} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
 			$plugin_url = $plugin['AuthorURI'];
+		} elseif ( ! empty( $plugin['Author'] ) ) {
+			$plugin_url = $plugin['Author'];
 		}
 		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . ' - ' . $plugin_url . "\n";
 	}
@@ -1427,6 +1431,8 @@ function edd_tools_sysinfo_get() {
 				$plugin_url = $plugin['PluginURI'];
 			} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
 				$plugin_url = $plugin['AuthorURI'];
+			} elseif ( ! empty( $plugin['Author'] ) ) {
+				$plugin_url = $plugin['Author'];
 			}
 			$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . ' - ' . $plugin_url . "\n";
 		}
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -1377,7 +1377,12 @@ function edd_tools_sysinfo_get() {
 			continue;
 
 		$update = ( array_key_exists( $plugin_path, $updates ) ) ? ' (needs update - ' . $updates[$plugin_path]->update->new_version . ')' : '';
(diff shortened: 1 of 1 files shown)


includes/admin/tools.php
```diff
@@ -1377,7 +1377,8 @@ function edd_tools_sysinfo_get() {
 			continue;
 
 		$update = ( array_key_exists( $plugin_path, $updates ) ) ? ' (needs update - ' . $updates[$plugin_path]->update->new_version . ')' : '';
-		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . "\n";
+		$plugin_url = isset( $plugin['PluginURI'] ) && ! empty( $plugin['PluginURI'] ) ? $plugin['PluginURI'] : $plugin['AuthorURI'];
```
robincornett, line 1380: This is not happening for me, at least not with the plugins I have active, but what if a plugin header is missing both the `PluginURI` and the `AuthorURI`? I'm wondering about doing something like this instead:

```php
		$plugin_url = '';
		if ( ! empty( $plugin['PluginURI'] ) ) {
			$plugin_url = $plugin['PluginURI'];
		} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
			$plugin_url = $plugin['AuthorURI'];
		}
```

Note that `empty()` will also satisfy `isset()` so you don't need both.

includes/admin/tools.php
```diff
@@ -1377,7 +1377,8 @@ function edd_tools_sysinfo_get() {
...
+		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . ' - ' . $plugin_url . "\n";
```
robincornett, line 1381: Just a question for you--as you look at the list, if a plugin needs an update, is it more helpful to have them in the order you do already, or would you want the `$update` bit at the end? I'm seeing this:

```
Advanced Custom Fields PRO: 5.9.5 (needs update - 5.9.9) - https://www.advancedcustomfields.com
AffiliateWP: 2.7-beta1 (needs update - 2.8-alpha2) - https://affiliatewp.com
```
(I am just asking--either way is fine with me.)

includes/admin/tools.php
```diff
@@ -1390,7 +1391,8 @@ function edd_tools_sysinfo_get() {
 			continue;
 
 		$update = ( array_key_exists( $plugin_path, $updates ) ) ? ' (needs update - ' . $updates[$plugin_path]->update->new_version . ')' : '';
-		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . "\n";
+		$plugin_url = isset( $plugin['PluginURI'] ) && ! empty( $plugin['PluginURI'] ) ? $plugin['PluginURI'] : $plugin['AuthorURI'];
```
robincornett, line 1394: I would do the same logic here.

includes/admin/tools.php
```diff
@@ -1410,7 +1412,8 @@ function edd_tools_sysinfo_get() {
 
 			$update = ( array_key_exists( $plugin_path, $updates ) ) ? ' (needs update - ' . $updates[$plugin_path]->update->new_version . ')' : '';
 			$plugin  = get_plugin_data( $plugin_path );
-			$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . "\n";
+			$plugin_url = isset( $plugin['PluginURI'] ) && ! empty( $plugin['PluginURI'] ) ? $plugin['PluginURI'] : $plugin['AuthorURI'];
```
robincornett, line 1415: And here.

robincornett (commented):
This looks good, @mihaijoldis, thanks! Just made a suggestion (x3) which might make the logic a little bit safer, maybe.

includes/admin/tools.php
mihaijoldis: I think its easier and more "in front" of you to see the Needs update message right at the start than having to perhaps scroll to the right if its a long URI

includes/admin/tools.php
mihaijoldis: Updated!

includes/admin/tools.php
mihaijoldis: Updated!

includes/admin/tools.php
mihaijoldis: Updated!

mihaijoldis:
@robincornett Do you think its too much to also add a final else check and put the Author in there just in case the URI and Author URI are not set ?
```
			$plugin_url = '';
			if ( ! empty( $plugin['PluginURI'] ) ) {
				$plugin_url = $plugin['PluginURI'];
			} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
				$plugin_url = $plugin['AuthorURI'];
			} else {
				$plugin_url = $plugin['Author'];
			}
```

I noticed on my local site that I have one of our snippets from hte library that has neither of the PluginURI/AuthorURI set but it does the Author so mybe even if it just shows this 
`EDD All Access - Do Not Count Re-Downloads: 1.0 - Pippin Williamson`

Its still helpful?

robincornett:
Sure, I would say just make it a final elseif rather than an else, so you're never trying to use something that doesn't exist.

mihaijoldis:
@robaggior updated to check for `Author` as last step

includes/admin/tools.php
robincornett: @mihaijoldis this is looking great, but I discovered a plugin in my site which meets none of these parameters, so the output is:

```
Plugin Name: 1.5 - 
```

What would you think of updating the logic a little to drop that if the `$plugin_url` can't be retrieved? Something like:

```php
		if ( $plugin_url ) {
			$plugin_url = ' &mdash; ' . $plugin_url;
		}
		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . $plugin_url . "\n";
```
(I threw in the `&mdash;` as our extension names have a hyphen in them already and thought that might help differentiate a bit.)

mihaijoldis:
@robincornett PR updated with latest changes.

includes/admin/tools.php
```diff
@@ -1377,7 +1377,18 @@ function edd_tools_sysinfo_get() {
 			continue;
 
 		$update = ( array_key_exists( $plugin_path, $updates ) ) ? ' (needs update - ' . $updates[$plugin_path]->update->new_version . ')' : '';
-		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . "\n";
+		$plugin_url = '';
+		if ( ! empty( $plugin['PluginURI'] ) ) {
+			$plugin_url = $plugin['PluginURI'];
+		} elseif ( ! empty( $plugin['AuthorURI'] ) ) {
+			$plugin_url = $plugin['AuthorURI'];
+		} elseif {
```
robincornett, line 1385:
```suggestion
		} elseif ( ! empty( $plugin['Author'] ) ) {
```
The `elseif` has to have something for the `if` please--on all three of these.

robincornett (approved):

ashleyfae (approved):

PR #8776, opened 2021-07-23
BUGFIX: Get price ID of product with variable prices included in Bundle

sergeykomlev:
For product(s) with variable prices included in Bundle(s) we need to get price ID in a different way. Without this change, customers see all downloadable files for variable pricing products included in the Bundle.

### DIFF
--- a/templates/history-downloads.php
+++ b/templates/history-downloads.php
@@ -36,6 +36,10 @@ if ( $purchases ) :
 					<tr class="edd_download_history_row">
 						<?php
 						$price_id       = edd_get_cart_item_price_id( $download );
+						// Get price ID of product with variable prices included in Bundle
+						if ( $download['in_bundle'] && edd_has_variable_prices( $download['id'] ) ) {
+							$price_id = edd_get_bundle_item_price_id( $download['id'] );
+						}
 						$download_files = edd_get_download_files( $download['id'], $price_id );
 						$name           = $download['name'];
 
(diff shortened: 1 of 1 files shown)


templates/history-downloads.php
```diff
@@ -36,6 +36,10 @@
 					<tr class="edd_download_history_row">
 						<?php
 						$price_id       = edd_get_cart_item_price_id( $download );
+						// Get price ID of product with variable prices included in Bundle
+						if ( $download['in_bundle'] && edd_has_variable_prices( $download['id'] ) ) {
```
robincornett, line 40:
```suggestion
						if ( ! empty( $download['in_bundle'] ) && edd_has_variable_prices( $download['id'] ) ) {
```
I think we need to add the `empty` check as the `in_bundle` may not be set at all--I got an undefined index notice.

Once I made that change, the PR works as expected and does prevent all bundled items from showing for the download if they should not be.

robincornett (changes requested):

templates/history-downloads.php
sergeykomlev: @robincornett, Thanks! Great suggestion.

sergeykomlev (commented):
Great suggestion!

robincornett (approved):
Thanks for the PR, @sergeykomlev!

sergeykomlev:
Hi @ashleyfae, will this change will be applied in the next releases?

cklosowski (approved):

PR #8781, opened 2021-07-26
Update order migration for orders missing payment total

### LINKED ISSUE #8780: 3.0 - migrated orders lacking _edd_payment_total may have incorrect amounts
### Expected behavior
When I run a migration from EDD 2.11 to 3.0, I expect order totals to match up (even if the discounts or cart items are off).

### Actual behavior
I was spot checking a site with old orders and found one where the order total should have been $7.50 (after a 50% discount), but it is showing in 3.0 as $15. I narrowed it down to there being no `_edd_payment_total` meta key/value on the payment.

I wanted to know how widespread this issue is and ran this query on the site in question:

```php
		$args  = array(
			'post_type'   => 'edd_payment',
			'post_status' => 'publish',
			'meta_query'  => array(
				'relation' => 'AND',
				array(
					'key'     => '_edd_payment_meta',
					'compare' => 'EXISTS',
				),
				array(
					'key'     => '_edd_payment_total',
					'compare' => 'NOT EXISTS',
				),
			),
		);
		$query = new WP_Query( $args );
		var_dump( ( $query->found_posts ) );
```
There are 242 affected orders (I think this query reflects what I'm looking for). I _believe_ the date on the last affected order is 2012-09-15 (I'm testing a migration adjustment and logging these orders), so it may have been just a short term issue.

1. Create some orders.
2. Make sure at least some of the orders have discount codes applied.
3. Delete any `_edd_payment_total` meta key rows. That should effectively replicate what I'm seeing here.
4. Migrate the orders.
5. The new order totals are likely to be incorrect.

PHP Version: 7.4.1
WordPress Version: 5.8

What I'm currently testing locally is changing [these lines](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/admin/upgrades/v3/class-data-migrator.php#L482-L490) to this:

```php		$meta_total = false;
		// Retrieve the total amount from metadata if available.
		if ( isset( $meta['_edd_payment_total'] ) ) {
			$meta_total  = maybe_unserialize( $meta['_edd_payment_total'] );
			$order_total = (float) $meta_total[0];
		} elseif ( isset( $payment_meta['amount'] ) ) {
			$meta_total  = maybe_unserialize( $payment_meta['amount'] );
			$order_total = (float) $meta_total[0];
			error_log( print_r( $data->ID . ' ' . $order_total, true ) );
		}
```

It appears to be working correctly, and I have a list of order IDs which were affected, so I can spot check a little more accurately.

robincornett:
Fixes #8780

Proposed Changes:
1. Add an additional payment metadata check for looking for the order total during migration.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -479,13 +479,13 @@ class Data_Migrator {
 			$order_tax = (float) $meta_tax[0];
 		}
 
+		$meta_total = false;
 		// Retrieve the total amount from metadata if available.
-		$meta_total = isset( $meta['_edd_payment_total'] )
-			? $meta['_edd_payment_total']
-			: false;
-
-		if ( false !== $meta_total ) {
-			$meta_total  = maybe_unserialize( $meta_total );
+		if ( isset( $meta['_edd_payment_total'] ) ) {
+			$meta_total  = maybe_unserialize( $meta['_edd_payment_total'] );
+			$order_total = (float) $meta_total[0];
+		} elseif ( isset( $payment_meta['amount'] ) ) {
+			$meta_total  = maybe_unserialize( $payment_meta['amount'] );
 			$order_total = (float) $meta_total[0];
 		}
 
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -486,7 +486,7 @@ class Data_Migrator {
 			$order_total = (float) $meta_total[0];
 		} elseif ( isset( $payment_meta['amount'] ) ) {
 			$meta_total  = maybe_unserialize( $payment_meta['amount'] );
-			$order_total = (float) $meta_total[0];
+			$order_total = (float) $meta_total;
 		}
 
 		// In some cases (very few) there is no cart details...so we have to just avoid this part.


ashleyfae (commented):
I'm getting some weird results on a payment that I deleted the total for. The total is coming out as `8.00` instead of `80.00`. Going to dig into it.

Out of curiosity, with your orders that naturally didn't have `_edd_payment_total`, what was the total displaying as in the 2.x UI?

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -479,13 +479,13 @@ public static function orders( $data = null ) {
...
-			: false;
-
-		if ( false !== $meta_total ) {
-			$meta_total  = maybe_unserialize( $meta_total );
+		if ( isset( $meta['_edd_payment_total'] ) ) {
+			$meta_total  = maybe_unserialize( $meta['_edd_payment_total'] );
+			$order_total = (float) $meta_total[0];
+		} elseif ( isset( $payment_meta['amount'] ) ) {
+			$meta_total  = maybe_unserialize( $payment_meta['amount'] );
 			$order_total = (float) $meta_total[0];
```
ashleyfae, line 489:
```suggestion
			$order_total = (float) $meta_total;
```

Data retrieved from `$meta` is an array, but `$payment_meta` is not.

ashleyfae (changes requested):

includes/admin/upgrades/v3/class-data-migrator.php
ashleyfae: Patched that change in locally and it's now working well for me. So I'll be happy to merge this after that change.

robincornett: Confirmed here--my $7.50 order was showing in 3.0 as $7, but doing this fixed it.

ashleyfae (approved):

PR #8784, opened 2021-07-27
Add offset to remap_args for payments queries

### LINKED ISSUE #8783: 3.0 - `offset` has no effect when passed to `edd_get_payments()`
### Expected behavior
When I pass `offset` to `edd_get_payments()`, I expect that payments are offset for that value. 

### Actual behavior
When I pass `offset` to `edd_get_payments()`, it has no effect. It's not mapped in `EDD_Payments_Query::remap_args()`, it's always [calculated](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/aebe33952b1330bce68ec95ecf4439724868ddd8/includes/payments/class-payments-query.php#L718-L720).

EDD Version (or branch):
`release/3.0`

robincornett:
Fixes #8783 

Proposed changes:
1. Adds a check for `$this->args['offset']` to `remap_args` and use it if possible, preferring that to `paged`/`posts_per_page`.

### DIFF
--- a/includes/payments/class-payments-query.php
+++ b/includes/payments/class-payments-query.php
@@ -715,7 +715,9 @@ class EDD_Payments_Query extends EDD_Stats {
 			$this->args['parent'] = $this->args['post_parent'];
 		}
 
-		if ( isset( $this->args['paged'] ) && isset( $this->args['posts_per_page'] ) ) {
+		if ( ! empty( $this->args['offset'] ) ) {
+			$arguments['offset'] = $this->args['offset'];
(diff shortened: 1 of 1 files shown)


dimadin (approved):
Confirmed that this works. It also takes precedence over 'paged' [in `WP_Query`](https://github.com/WordPress/wordpress-develop/blob/e6ed1b2be9f7bf769ccb41184c88b7e5c0752971/src/wp-includes/class-wp-query.php#L2625-L2631) as well.

ashleyfae (approved):

PR #8796, opened 2021-08-02
BugFix: set custom discount expire time

AkramiPro:
add support for set custom discount expire time . (but edd force to 23:59:59)


ashleyfae:
This feature is actually available in our `release/3.0` branch if you want to check that out. (On a test/staging site only.)

PR #8797, opened 2021-08-04
Set default order gateway to manual

### LINKED ISSUE #8792: 3.0 - set manual as default order gateway for CLI generated orders
### Expected behavior
When I generate a bunch of dummy orders using CLI, I'm generally doing so to test something like reports.

### Actual behavior
I generated a bunch of dummy orders and none showed in the Gateway Stats table in Reports because none of the orders had a gateway assigned at all. It seems like `manual` should be the default for CLI orders (if there could be another option, I'm not even sure).

1. In EDD 3.0, generate some orders using WP-CLI (something like `wp edd payments create --number=10 --date=random --range=30`)
2. Check the `edd_orders` table. None of those orders will have an assigned gateway.

PHP Version: 7.4.1
WordPress Version: 5.8

robincornett:
Fixes #8792

Proposed Changes:
1. Set the default order gateway to `manual`.
2. Upgrade the orders table to fill any empty gateways.

### DIFF
--- a/includes/database/schemas/class-orders.php
+++ b/includes/database/schemas/class-orders.php
@@ -122,6 +122,7 @@ class Orders extends Schema {
 			'type'       => 'varchar',
 			'length'     => '100',
 			'sortable'   => true,
+			'default'    => 'manual',
 		),
 
 		// mode
--- a/includes/database/tables/class-orders.php
+++ b/includes/database/tables/class-orders.php
@@ -38,7 +38,7 @@ final class Orders extends Table {
 	 * @since  3.0
 	 * @var    int
 	 */
-	protected $version = 202103261;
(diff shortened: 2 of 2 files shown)


ashleyfae (changes requested):
This is good so far, and I want to keep everything you've done, but I want to set the default in the actual database schema here: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/25a45e173202027cb07dd2d6188dc45528ffdcf1/includes/database/tables/class-orders.php#L76 Right now it's `default ''` but I want that to be `default 'manual'`

ashleyfae (approved):
- [x] Tested update on existing install.
- [x] Tested fresh install.

PR #8799, opened 2021-08-09
Drop payment_ids column when legacy data is removed #8798

### LINKED ISSUE #8798: 3.0 - Don't drop old payment_ids column until removing legacy data

As soon as you upgrade to 3.0, we drop the `payment_ids` column from the customers table: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/tables/class-customers.php#L105

This is a destructive action and should not be done until legacy data is chosen to be removed. It doesn't cause any harm or stop 3.0 from working to keep that column available.

### Justification or use case

Make it easier to roll back to 2.10 from 3.0.

ashleyfae:
Fixes #8798

Proposed Changes:
1. Removes the column drop from the database upgrade routine. (This was run immediately upon upgrading to 3.0.)
2. Add the column drop to both the CLI legacy removal and the UI legacy removal.

## Testing

1. Upgrade to release/3.0 & migrate.
2. Ensure the `wp_edd_customers` table still has the `payment_ids` column.
3. Maybe make a new order with a brand new account, which will create a new customer, just to make sure that all still works okay.
4. Remove legacy data via the UI. Ensure the `payment_ids` column is gone.

Repeat steps 1 and 2 (skip 3), but this time remove legacy data via CLI. Ensure the `payment_ids` column is gone.

### DIFF
--- a/includes/admin/upgrades/v3/class-remove-legacy-data.php
+++ b/includes/admin/upgrades/v3/class-remove-legacy-data.php
@@ -50,8 +50,15 @@ class Remove_Legacy_Data extends Base {
 	 * @return bool True if data was migrated, false otherwise.
 	 */
 	public function get_data() {
-		// Delete meta on first step.
+		// Perform some database operations on the first step.
 		if ( 1 === $this->step ) {
+			// Drop custoemr `payment_ids` column. It's no longer needed.
+			$customer_table = edd_get_component_interface( 'customer', 'table' );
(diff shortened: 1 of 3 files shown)


robincornett (commented):
Ran a UI and CLI migration. Database update did not happen until I removed the legacy data, no errors. Creating a new customer in between migration and removing the legacy data worked as expected, no errors with the payment IDs column.

Just a couple typo fixes in the comments is all I've got.

robincornett (approved):

PR #8802, opened 2021-08-10
Set order parent from original data->post_parent in migration

### LINKED ISSUE #8801: 3.0 - Order parents not migrated
### Expected behavior

When migrating, I expect the order parent ID (`wp_posts.post_parent`) to be copied over to the order (`wp_edd_orders.parent`)

### Actual behavior

The parent is always `0`

The migration runs this code:

```php
// Find the parent payment, if there is one.
$parent = 0;
if ( ! empty( $data->post_parent ) ) {
	$parent = $wpdb->get_var( $wpdb->prepare( "SELECT edd_order_id FROM {$wpdb->edd_ordermeta} WHERE meta_key = %s AND meta_value = %d", esc_sql( 'legacy_order_id' ), $data->ID ) );
}
```

I assume that's code left over from before we were retaining post IDs. That just needs to be removed and replaced with `$data->post_parent`.

1) Have a renewal payment in 2.10 that has a parent ID.
2) Migrate to release/3.0
3) Note that the `parent` is `0` in the database

PHP Version:

EDD Version (or branch):

WordPress Version:

robincornett:
Fixes #8801

Proposed Changes:
1. Use the original `post_parent` as the new order `parent` if it is not empty.

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -577,7 +577,7 @@ class Data_Migrator {
 		// Find the parent payment, if there is one.
 		$parent = 0;
 		if ( ! empty( $data->post_parent ) ) {
-			$parent = $wpdb->get_var( $wpdb->prepare( "SELECT edd_order_id FROM {$wpdb->edd_ordermeta} WHERE meta_key = %s AND meta_value = %d", esc_sql( 'legacy_order_id' ), $data->ID ) );
+			$parent = $data->post_parent;
 		}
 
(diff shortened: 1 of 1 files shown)


ashleyfae (commented):
This is absolutely doing what we want, but honestly I think we can just simplify it.

Since we're copying the value over directly, there's really no need for a `$parent` variable or an empty check. I reckon we can just delete this whole block and on line 667, assign `$data->post_parent` to `'parent'` directly.

ashleyfae (commented):
Code looks great, just need to test a migration to confirm.

robincornett:
The migration I ran looks good. 😇

ashleyfae (approved):
Yep looks good. :raised_hands:

PR #8807, opened 2021-08-10
Add EDD 3.0 rollback handling #8805

### LINKED ISSUE #8805: Add EDD 3.0 rollback handling

In EDD 2.11 we should add 3.0 rollback support.

- Detect if the installation was previously running EDD 3.0.
- If so, ask the user if they want to roll back the database tables.
- If so, we'll perform some operation to undo the changes that EDD 3.0 made.

Most notable are the changes done to the customer meta table: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/tables/class-customer-meta.php#L89-L91

The main destructive action we do is changing the `customer_id` column to `edd_customer_id` (and change the index accordingly). So our rollback will handle reverting that change.

There are also a few changes made to the customers table, but they're not as big of a deal, as the changes are backwards compatible: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/database/tables/class-customers.php#L97-L102

For example, the `email` column length is increased, and `purchase_value` is changed from `mediumtext` to `decimal`.

### Justification or use case

Make it easy to roll back from 3.0.

ashleyfae:
Fixes #8805

Proposed Changes:
1. Detects a downgrade from EDD 3.0. This is done by checking if the _database_ version number is >= 3.0, but the _constant_ within the plugin is < 3.0.
2. If a downgrade has been detected, show the user a notice asking if that's correct and if they want to proceed.
3. Completing a downgrade rolls back the database changes.
4. Dismissing the notice just removes the downgrade flag with no other action.

## Testing

- Make sure a site that has only ever been on EDD 2.x doesn't see any new notices.
- Do a fresh EDD 3.0 install. Make sure you don't see any new notices.
- Migrate from EDD 2.x to 3.0. Then downgrade. Make sure notice appears.
    - Before clicking Upgrade, confirm that your `wp_edd_customermeta` table has a column called `edd_customer_id`.
    - Click Upgrade.
    - Confirm that your `wp_edd_customermeta` table now has a column called `customer_id` instead. (And should have the same values as before, just different column name.)
    - Do a couple test purchases, make sure new customers get created okay and customer meta gets added okay. (A good way to test the meta is to checkout via Stripe, which should add a piece of Stripe meta, I think).
    - Then upgrade to 3.0 once again. Make sure that meta column gets changed from `customer_id` to `edd_customer_id`.
    - Do a few test purchases, make sure new customers + meta get created okay.
- Start on a new 2.x test site. Migrate to 3.0. Then downgrade. Make sure notice appears.
    - Click the dismiss button and make sure the notice goes away.

Feedback on the notice content is welcome!

### DIFF
--- a/includes/admin/upgrades/downgrades.php
+++ b/includes/admin/upgrades/downgrades.php
@@ -13,8 +13,9 @@
  * and updates the version number.
  *
  * @since 2.11
+ * @return bool Whether or not a downgrade was performed.
  */
-add_action( 'admin_init', function() {
+function edd_do_downgrade() {
 	$did_downgrade   = false;
 	$edd_version     = preg_replace( '/[^0-9.].*/', '', get_option( 'edd_version' ) );
 	$downgraded_from = get_option( 'edd_version_downgraded_from' );
@@ -40,7 +41,10 @@ add_action( 'admin_init', function() {
 		update_option( 'edd_version', preg_replace( '/[^0-9.].*/', '', EDD_VERSION ) );
 		delete_option( 'edd_version_downgraded_from' );
 	}
-} );
+
+	return $did_downgrade;
+}
+add_action( 'admin_init', 'edd_do_downgrade' );
 
 /**
  * Display downgrade notices.
--- a/includes/admin/upgrades/downgrades.php
+++ b/includes/admin/upgrades/downgrades.php
@@ -17,7 +17,7 @@
 add_action( 'admin_init', function() {
 	$did_downgrade   = false;
 	$edd_version     = preg_replace( '/[^0-9.].*/', '', get_option( 'edd_version' ) );
-	$downgraded_from = get_option( 'edd_version_upgraded_from' );
+	$downgraded_from = get_option( 'edd_version_downgraded_from' );
 
 	/**
 	 * Check for downgrade from 3.0 to 2.11.
@@ -32,15 +32,13 @@ add_action( 'admin_init', function() {
 			 * @see edd_show_downgrade_notices()
 			 */
 			update_option( 'edd_v3_downgrade', time() );
+			$did_downgrade = true;
 		}
-	}
-
-	if ( version_compare( $downgraded_from, $edd_version, '>' ) ) {
-		$did_downgrade = true;
 	}
 
 	if ( $did_downgrade ) {
 		update_option( 'edd_version', preg_replace( '/[^0-9.].*/', '', EDD_VERSION ) );
+		delete_option( 'edd_version_downgraded_from' );
 	}
 } );
 
--- a/includes/admin/upgrades/downgrades.php
+++ b/includes/admin/upgrades/downgrades.php
@@ -107,7 +107,7 @@ add_action( 'edd_downgrade_v3', function () {
 
 	global $wpdb;
 	$customer_meta_table = EDD()->customer_meta->table_name;
-	$wpdb->query( "ALTER TABLE {$customer_meta_table} CHANGE `edd_customer_id` `edd_customer` bigint(20) unsigned NOT NULL default '0'" );
+	$wpdb->query( "ALTER TABLE {$customer_meta_table} CHANGE `edd_customer_id` `customer_id` bigint(20) unsigned NOT NULL default '0'" );
 	$wpdb->query( "ALTER TABLE {$customer_meta_table} DROP INDEX edd_customer_id" );
 	$wpdb->query( "ALTER TABLE {$customer_meta_table} ADD INDEX customer_id (customer_id)" );
 	EDD()->customer_meta->create_table(); // This re-adds the version number for us.
--- a/includes/admin/upgrades/downgrades.php
+++ b/includes/admin/upgrades/downgrades.php
@@ -80,6 +80,8 @@ add_action( 'edd_downgrade_v3', function () {
 
 	EDD()->customers->create_table();
 
+	delete_option( 'edd_v3_downgrade' );
+
 	wp_safe_redirect( remove_query_arg( array( '_wpnonce', 'edd_action' ) ) );
 	exit;
 } );
--- a/includes/admin/upgrades/downgrades.php
+++ b/includes/admin/upgrades/downgrades.php
@@ -9,6 +9,42 @@
  */
 
 /**
+ * Checks if the current site has downgraded, and if so, sets any necessary flags
+ * and updates the version number.
+ *
+ * @since 2.11
+ */
+add_action( 'admin_init', function() {
+	$did_downgrade   = false;
+	$edd_version     = preg_replace( '/[^0-9.].*/', '', get_option( 'edd_version' ) );
+	$downgraded_from = get_option( 'edd_version_upgraded_from' );
+
+	/**
+	 * Check for downgrade from 3.0 to 2.11.
+	 */
+	if ( version_compare( EDD_VERSION, '3.0-beta1', '<' ) ) {
+		if (
+			version_compare( $edd_version, '3.0-beta1', '>=' ) ||
+			( $downgraded_from && version_compare( $downgraded_from, '3.0-beta1', '>=' ) )
+		) {
+			/*
+			 * This site probably just downgraded from EDD 3.0. Let's store a flag so we can give them the option to properly roll back.
+			 * @see edd_show_downgrade_notices()
+			 */
+			update_option( 'edd_v3_downgrade', time() );
+		}
+	}
+
+	if ( version_compare( $downgraded_from, $edd_version, '>' ) ) {
+		$did_downgrade = true;
+	}
+
+	if ( $did_downgrade ) {
+		update_option( 'edd_version', preg_replace( '/[^0-9.].*/', '', EDD_VERSION ) );
+	}
+} );
+
+/**
  * Display downgrade notices.
  *
  * @since 2.11
--- a/includes/admin/upgrades/downgrades.php
+++ b/includes/admin/upgrades/downgrades.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * Downgrades
+ *
+ * @package   easy-digital-downloads
+ * @copyright Copyright (c) 2021, Sandhills Development, LLC
+ * @license   GPL2+
+ * @since     2.11
+ */
+
+/**
+ * Display downgrade notices.
+ *
+ * @since 2.11
+ */
+function edd_show_downgrade_notices() {
+	if ( ! current_user_can( 'manage_options' ) ) {
+		return;
+	}
+
+	/**
+	 * EDD 3.0 downgrade.
+	 */
+	if ( get_option( 'edd_v3_downgrade' ) ) {
+		$downgrade_url = wp_nonce_url( add_query_arg( 'edd_action', 'downgrade_v3' ), 'edd_downgrade_v3' );
+		$dismiss_url   = wp_nonce_url( add_query_arg( 'edd_action', 'dismiss_v3_downgrade' ), 'edd_dismiss_v3_downgrade' );
+		?>
+		<div class="notice notice-warning">
+			<h2><?php esc_html_e( 'Did you downgrade from Easy Digital Downloads 3.0?', 'easy-digital-downloads' ); ?></h2>
+			<p>
+				<?php _e( 'We\'ve detected that your site may have just downgraded from Easy Digital Downloads version 3.0. If that is correct, please click the "Complete Downgrade" button below to complete this process.', 'easy-digital-downloads' ); ?>
+			</p>
+			<p>
+				<?php _e( 'If you believe this message to be in error, please click "Dismiss Notice" instead, which will remove this notice with no further action.', 'easy-digital-downloads' ); ?>
+			</p>
+			<p>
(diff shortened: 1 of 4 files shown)


includes/admin/upgrades/upgrade-functions.php
```diff
@@ -21,13 +21,16 @@
...
+	$did_upgrade = $did_downgrade = false;
 	$edd_version = preg_replace( '/[^0-9.].*/', '', get_option( 'edd_version' ) );
 
-	if( version_compare( $edd_version, '2.6', '<' ) ) {
-
+	if ( version_compare( $edd_version, '2.6', '<' ) ) {
 		edd_v26_upgrades();
-
+	}
+	if ( version_compare( $edd_version, '3.0-beta1', '>=' ) && version_compare( EDD_VERSION, '3.0-beta1', '<' ) ) {
```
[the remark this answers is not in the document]
robincornett: Although this could be because of how I'm switching between versions (I have both installed and am not replacing the directory/files?)?

robincornett (commented):
I'm having issues with the notice not showing--if I force it, I can run the downgrade and everything appears correct. Going back to 3.0 changes the database tables again. If I force the notice to show, I can't dismiss it, but that could be because of how I'm forcing it.

LisaCee:
I can confirm that I am not seeing the downgrade database either.

includes/admin/upgrades/upgrade-functions.php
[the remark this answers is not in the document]
LisaCee: How did you force the notice? I've upgraded and would like to downgrade for further testing.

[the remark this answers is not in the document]
ashleyfae: @robincornett So are you deactivating v3 and activating v2? I was just switching between branches and that was working, but maybe it doesn't work if actually reactivating the plugin entirely.

ashleyfae:
Updated with some changes. I've confirmed this still works for my method (switching branches). @robincornett and @LisaCee can you test with what you were doing?

robincornett (commented):
The notice is now working properly for me.

However, I'm ending up with both an `edd_customer_id` column and a `customer_id` column. The latter has all `0` values.

Would it be possible to add an extra check to not show the downgrade notice if the legacy data has been removed? It seems like we should not allow a downgrade in that situation.

ashleyfae:
> > However, I'm ending up with both an `edd_customer_id` column and a `customer_id` column. The latter has all `0` values.
> 
> Is that after downgrading? (Or after downgrading and then upgrading again?)

Oh ugh. I think I see the problem.

robincornett:
I have the two columns after downgrading, without trying to upgrade again. (I noticed this earlier and trashed the database so I could do the migration again and start from scratch with this process.)

ashleyfae:
Yeah the problem is specifically that with your method, `edd_run_install()` runs, and that function runs this:

```php
@EDD()->customers->create_table();
@EDD()->customer_meta->create_table();
```

Which it cannot do before the downgrade logic runs.

I'll look into it..

includes/admin/upgrades/downgrades.php
```diff
@@ -0,0 +1,151 @@
...
+	 * EDD 3.0 downgrade.
+	 */
+	if ( get_option( 'edd_v3_downgrade' ) ) {
+		$downgrade_url = wp_nonce_url( add_query_arg( 'edd_action', 'downgrade_v3' ), 'edd_downgrade_v3' );
+		$dismiss_url   = wp_nonce_url( add_query_arg( 'edd_action', 'dismiss_v3_downgrade' ), 'edd_dismiss_v3_downgrade' );
+		?>
+		<div class="notice notice-warning">
+			<h2><?php esc_html_e( 'Did you downgrade from Easy Digital Downloads 3.0?', 'easy-digital-downloads' ); ?></h2>
+			<p>
+				<?php _e( 'We\'ve detected that your site may have just downgraded from Easy Digital Downloads version 3.0. If that is correct, please click the "Complete Downgrade" button below to complete this process.', 'easy-digital-downloads' ); ?>
```
robincornett, line 69:
```suggestion
				<?php esc_html_e( 'We\'ve detected that your site may have just downgraded from Easy Digital Downloads version 3.0. If that is correct, please click the "Complete Downgrade" button below to complete this process. This will change your database, but is necessary to use this version of Easy Digital Downloads.', 'easy-digital-downloads' ); ?>
```
Maybe? That way they know at least a little bit amount what it's doing, and indicates that they can't use their store without doing it?

includes/admin/upgrades/downgrades.php
```diff
@@ -0,0 +1,151 @@
...
+			</p>
+			<p>
+				<?php _e( 'If you believe this message to be in error, please click "Dismiss Notice" instead, which will remove this notice with no further action.', 'easy-digital-downloads' ); ?>
```
robincornett, line 72:
```suggestion
				<?php esc_html_e( 'If you believe this message to be in error, please click "Dismiss Notice" instead, which will remove this notice with no further action.', 'easy-digital-downloads' ); ?>
```
I feel like this is a little stilted, but don't have a better suggestion, either.

robincornett (commented):
Just a wording comment.

1. I migrated a database to EDD 3.0.
2. Downgraded. Clicked the button. The database updates correctly.
3. I can place a new order and customer metadata is added correctly.
4. I switch back to EDD 3.0. The database updates correctly.
5. Note that the order I placed during the downgrade is gone.
6. I run `wp edd migrate_payments --force` .... the missing order is included in the "Old Records" count, but is not migrated...so it is lost. I wonder if we need to offer a way to "reset" the v3 migration, to empty the newly created tables? Or something else? Just because if someone does downgrade, but doesn't do so, and has orders come in during the interim, and then tries to update again, they'll have problems. Or at that point we need to tell them they need more support (in terms of knowing what tables to clear out before forcing a new migration?).

LisaCee:
- [x] Make sure a site that has only ever been on EDD 2.x doesn't see any new notices.

- [x] Do a fresh EDD 3.0 install. Make sure you don't see any new notices.

 - [x] Migrate from EDD 2.x to 3.0. Then downgrade. Make sure notice appears.
      - [ ] Before clicking Upgrade, confirm that your wp_edd_customermeta table has a column called edd_customer_id.
      - [ ] Confirm that your wp_edd_customermeta table now has a column called customer_id instead. (And should have the same values as before, just different column name.)
      - [ ] Do a couple test purchases, make sure new customers get created okay and customer meta gets added okay. (A good way to test the meta is to checkout via Stripe, which should add a piece of Stripe meta, I think).
      - [ ] Then upgrade to 3.0 once again. Make sure that meta column gets changed from customer_id to edd_customer_id.
             _Here the `wp_edd_customermeta` metadata does get updated, but as Robin mentioned the orders placed in 2.11 don't 
             get moved into the `wp_edd_orders` table. They are still in `wp_postmeta`._
      - [ ] Do a few test purchases, make sure new customers + meta get created okay.
- [ ] Start on a new 2.x test site. Migrate to 3.0. Then downgrade. Make sure notice appears.
     - [ ] Click the dismiss button and make sure the notice goes away.

ashleyfae:
> 6\. I run `wp edd migrate_payments --force` .... the missing order is included in the "Old Records" count, but is not migrated...so it is lost. I wonder if we need to offer a way to "reset" the v3 migration, to empty the newly created tables? Or something else? Just because if someone does downgrade, but doesn't do so, and has orders come in during the interim, and then tries to update again, they'll have problems. Or at that point we need to tell them they need more support (in terms of knowing what tables to clear out before forcing a new migration?).

Hmm that's weird. The new order should have been migrated.

I think though for the sake of this PR we just focus on getting EDD 2.x back into a usable state -- not doing fancy stuff with resetting the entire 3.0 migration.

ashleyfae:
> * Start on a new 2.x test site. Migrate to 3.0. Then downgrade. Make sure notice appears.
>   _I keep running into an issue here. I upgrade and get confirmation that the migration is complete and then I downgrade. On two different sites, after I change branches to downgrade, I get an notice that EDD needs to complete a previously started upgrade. I click it and it puts me in a loop that goes to the upgrade screen for just a second and then back to the dashboard with the same notice. I am not seeing the downgrade notice, probably due to this. Has anyone else successfully tested this step? I'll test again tomorrow to confirm it wasn't just some random hiccup in my system._

Can you post a screenshot of that notice you're getting? Curious which upgrade it's talking about.

includes/admin/upgrades/downgrades.php
ashleyfae: I didn't want people to be afraid to click it. :sweat_smile: I'm in two minds about it. I want to treat it like no big deal, because it really is no big deal. Does `change your database` sound scary?

Under other circumstances this is the kind of update we'd do behind the scenes without even telling the user about it. (Which is what we do on upgrade to 3.0. We make this change before you even run the migration.)

robincornett: Yes, "change your database" does sound a bit scary, and I understand it's not a destructive change. I can live without it, but I do think the message needs to communicate that their store will not operate correctly in 2.x if they don't click the button.

> Under other circumstances this is the kind of update we'd do behind the scenes without even telling the user about it.

Is that something to consider here instead maybe?

ashleyfae: > Is that something to consider here instead maybe?

I think originally we didn't because we were kind of guessing if it was a downgrade, and it was less certain than an upgrade. But now that it's done... I think we can be pretty certain. I'll think on it for an hour. :stuck_out_tongue:

LisaCee:
I make a new site this morning and now I'm getting a hybrid of notices. I am seeing the downgrade notice, but I'm still getting the complete the database notice. GitHub isn't letting me add an image here, but [here's](https://drive.google.com/file/d/16zo7HCd-xllUoxoFQ-YxcqI3uCYfYaB2/view?usp=drivesdk) a screenshot of the messages.

ashleyfae:
@LisaCee  Weird. So are these steps correct?

1. Totally fresh install of EDD master(?) branch (or `issue/8805` branch).
2. Upgrade to release/3.0. No "incomplete upgrade" notice here.
3. Downgrade to `issue/8805` branch. This is where "incomplete upgrade" notice appears?

LisaCee:
@ashleyfae, I guess this was a false alarm. I tried a few more times over the last few days and I couldn't replicate my own issue. I added my testing details below, in case there's anything of use.  

1. Brand new site with EDD `issue/8805`. I add some products so that I have something to migrate. (Themedd demo content, no attachments)
2. Upgrade to 3.0. It's weird because I'm not getting the migration notice anymore. The database just magically is updated.

So, trying again...

1. Brand new site with EDD `issue/8805`. I add some products and make an order so that I have something to migrate. (Themedd demo content, no attachments)
2. Upgrade to 3.0. Upgrade database, remove legacy data
3. Downgrade to `issue/8805`, complete downgrade - no notice.

ashleyfae:
> 2\. It's weird because I'm not getting the migration notice anymore. The database just magically is updated.

You won't get the EDD 3.0 notice if you've upgraded and migrated once before. This PR doesn't bring the "Migrate to 3.0" notice back after upgrading a second time; it just allows you to downgrade and have a functioning 2.x site. So that part is normal.

Weird you're not getting the other database notice again. :shrug: 

So I guess we can call this good then?

includes/admin/upgrades/downgrades.php
robincornett: If we are leaving the message in so they have to trigger the downgrade manually, should the notice say something like, "You must complete the downgrade to successfully process orders in version 2.11." ?

I guess my concern is just that someone might downgrade and then ignore the notice. Functionally, the PR is working for me and I can approve if we don't want to tweak this part.

ashleyfae: I'll rework it to happen automatically.

ashleyfae:
@LisaCee and @robincornett I've updated the PR so that the upgrade now happens automatically without an admin notice. On downgrade, it does the same version check as before, but it checks to see if the `edd_customer_id` column exists. If so, then it just performs the downgrade without asking.

robincornett (approved):
This is working well and invisibly for me at this point. I guess my last question would be if the `edd_do_downgrade` function should include a `function_exists` check for an EDD 3.0 function and return early if it exists, since this will be merged into 3.0 when the next master -> 3.0 merge occurs ... or maybe it won't be added to the 3.0 codebase at all, as I think about it.

ashleyfae:
> This is working well and invisibly for me at this point. I guess my last question would be if the `edd_do_downgrade` function should include a `function_exists` check for an EDD 3.0 function and return early if it exists, since this will be merged into 3.0 when the next master -> 3.0 merge occurs ... or maybe it won't be added to the 3.0 codebase at all, as I think about it.

We do check to make sure that the defined `EDD_VERSION` (constant -- not DB value) is less than `3.0-beta1` and the code only runs then: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8807/files#diff-82fb95df29660acfacc2a483e9fb158d9db977895cfeb68ef18fa01125eb1d3bR25

So that should be handling it just fine.

But once this logic does get merge into `release/3.0` then we can just remove that entire thing.

LisaCee (approved):
This works seamlessly for me. I like having the downgrade done behind the scenes.

PR #8808, opened 2021-08-10
Issue/8764

### LINKED ISSUE #8764: 3.0 - Third parameter passed to `edd_log_file_download_download_files` filter is in different format than in pre-3.0 versions

ashleyfae:
Fixes #8764

Proposed Changes:
1. Moves log file retrieval logic into a new method `get_log_files()` to clean things up a bit.
2. Retrieve all meta via `$meta = edd_get_file_download_log_meta( $log->id );` as a starting point.
3. When adding backwards compat meta, ensure values are an array.


includes/admin/reporting/class-file-downloads-logs-list-table.php
```diff
@@ -210,6 +168,76 @@ function get_logs( $log_query = array() ) {
...
+			: array();
+
+		/*
+		 * Build up an array of meta.
+		 */
+		$meta = edd_get_file_download_log_meta( $log->id );
+
+		// These meta keys no longer exist, but we need to create them for use in the filter.
+		$backwards_compat_meta = array(
+			'_edd_log_user_info'  => $user_info,
```
robincornett, line 219: Comparing a var_dump  from 2.11 with this PR, I think there also needs to be this line:

```php
'_edd_log_customer_id' => $customer->id,
```

Noting that my 2.11 log data does not include the `user_info` array:
```
array (size=6)
  '_edd_log_customer_id' => 
    array (size=1)
      0 => string '228' (length=3)
  '_edd_log_user_id' => 
    array (size=1)
      0 => string '1' (length=1)
  '_edd_log_file_id' => 
    array (size=1)
      0 => string '1' (length=1)
  '_edd_log_ip' => 
    array (size=1)
      0 => string '127.0.0.1' (length=9)
  '_edd_log_payment_id' => 
    array (size=1)
      0 => string '3463' (length=4)
  '_edd_log_price_id' => 
    array (size=1)
      0 => string '0' (length=1)
```

But if it's supposed to be there and I just don't happen to have it, that's fine, too. Other than the `customer_id`, this PR makes the `$meta` match the values in 2.x.

ashleyfae: I'm inclined to say we keep it anyway because I don't know why it was there and it won't do any harm to keep it. :shrug:

robincornett (approved):

Issue #8659, opened 2021-05-26
3.0 - Download stats re-count tool is counting refunds


ashleyfae:
### Expected behavior

Download stats should be of completed _sales_. I wouldn't see any refund information.

### Actual behavior

Refund records are skewing the results.

### Steps to reproduce the behavior
1) For easier testing, create a brand new product.
2) Purchase the product once.
3) Then refund that purchase.
4) Now the sale & earnings counts. For me:
![Screenshot from 2021-05-26 13-44-20](https://user-images.githubusercontent.com/6324272/119661740-812b4d00-be28-11eb-94f7-eb70ea7feb74.png)
5) Then go to Downloads > Tools and in "Recount Stats", re-count for the download you just created.
6) Re-check the numbers.
![Screenshot from 2021-05-26 13-45-09](https://user-images.githubusercontent.com/6324272/119663226-12e78a00-be2a-11eb-9a20-8d665eafe4fa.png)

What's actually happening here is that the recount tool does a query like this:

```php
$order_items = edd_get_order_items( array(
    'status__in' => array( 'complete', 'revoked' ),
    'product_id' => $your_product_id
) );
```

Well the original order record is `refunded`, so that won't get picked up at all. Instead it picks up the _refund_ record (which is order_type = refund; status = complete). It adds that amount to the earnings, which is actually a negative number (`-10.93` for me). But on display, the earnings isn't allowed to be below `0`, so it just displays `0` instead.

tl;dr: This whole thing is weird.

Copying from Slack discussion: I like @spencerfinnell's line of thinking which is that:

1. Sales number should always be number of orders. Maybe orders of any status that would have been complete at one time. (`edd_get_gross_order_statuses()`)
2. `Earnings` column should be re-named to `Revenue`.
3. Earnings is the same as 1, but for earnings instead of sales count.

With that in mind... Perhaps the easiest solution here is to add a new argument to `edd_get_order_items()` that would allow you to pass in `'order_type' => 'sale'`. Then we could make sure we only actually pick up sales and not refunds.

### Information (if a specific version is affected):
PHP Version:

EDD Version (or branch): release/3.0

WordPress Version:

Any other relevant information:

ashleyfae:
More from Slack:

> If we choose for it to include refunds, it needs to at least be gross or have both, if wechoose to not have refunds in there then it needs to have a prefix of net

tl;dr: We should rename the column to `Gross Revenue`

PR #8809, opened 2021-08-10
Move `edd_update_payment_status` to a transition hook

### LINKED ISSUE #8804: 3.0 - "edd_update_payment_status" hook not triggering for orders
### Expected behavior

When an order's status changes, I expect the `edd_update_payment_status` hook to fire.

### Actual behavior

It doesn't fire unless you're running custom code and working with an `EDD_Payment` object.

See https://github.com/easydigitaldownloads/EDD-Software-Licensing/issues/1870 This issue was discovered with Software Licensing, which uses that hook to revoke license keys after a refund.

1. Be on EDD release/3.0
2. Purchase a new SL product.
3. Refund the payment.
4. Note that the license key is not disabled.

PHP Version:

WordPress Version:

Here's kind of what I'm thinking.

We should _remove_ this from `EDD_Payment::update_status()`:

```php
do_action( 'edd_update_payment_status', $this->ID, $status, $old_status );
```

and move that to trigger when the order status changes. So something like:

 ```php
add_action( 'edd_transition_order_status', function( $old_status, $new_status, $order_id ) {
	do_action( 'edd_update_payment_status', $order_id, $new_status, $old_status );
}, 10, 3 );
```

That will ensure `edd_update_payment_status` runs for orders. Removing it from `EDD_Payment` ensures it doesn't run twice. (Ultimately `EDD_Payment::update_status()` calls an update on the order, which should then trigger the hook.)

This change will require a lot of testing and potentially updates for add-ons. We want to make sure that we don't have any logic running twice.

A quick check revealed the following add-ons using this hook:

- AffiliateWP: `add_action( 'edd_update_payment_status', array( $this, 'revoke_referral_on_refund' ), 10, 3 );`
- All Access: `add_action( 'edd_update_payment_status', 'edd_all_access_check_updated_payment_on_status_change', 101, 3 );`
- Commissions: `add_action( 'edd_update_payment_status', 'eddc_record_commission', 10, 3 );`
- Conditional Emails: `add_action( 'edd_update_payment_status', 'edd_conditional_emails_status_change_email', 100, 3 );`
- Cross Sell / Upsell: `add_action( 'edd_update_payment_status', 'edd_csau_undo_purchase_on_refund', 100, 3 );`
- Fraud Monitor: `add_action( 'edd_update_payment_status', array( $this, 'remove_fraud_flag_on_update' ), 10, 3 );`
- Free Downloads: `add_action( 'edd_update_payment_status', 'edd_free_downloads_remove_optin', -10 );`
- Mailchimp: `add_action( 'edd_update_payment_status', array( $this, 'cancel_order' ), 10, 3 );`
- Slack: `add_action( 'edd_update_payment_status', array( $this, 'edd_failed_purchase' ), 10, 3 );`
- Software Licensing: `add_action( 'edd_update_payment_status', array( $this, 'revoke_license' ), 0, 3 );`
- Stripe: `add_action( 'edd_update_payment_status', 'edd_stripe_process_refund', 200, 3 );`
- Wallet: `add_action( 'edd_update_payment_status', array( $this, 'process_refund' ), 200, 3 );`
- Zapier: `add_action( 'edd_update_payment_status', 'edd_zapier_update_payment_status', 10, 3 );`

### PR DESCRIPTION
**Pull request by robincornett**
Fixes #8804

Proposed Changes:
1. Move the `edd_update_payment_status` hook to an anonymous function hooked into `edd_transition_order_status`.
2. Decrease download earnings/sales as part of `edd_refund_order`. I believe this is consistent with how the download earnings are handled when adding a new order (both via cart and admin). These stats were not being updated when the hook was moved.
3. Oddly, with these changes, unit tests are experiencing errors in `edd_get_sale_notification_body_content` and `edd_get_email_names`. Although not technically related to this issue, I updated the logic in both of these to start with the fallback values, and only use the user data if it's available. Not sure why it would magically start failing with the  changes I made, but the extra checks seem reasonable to me.

At this point, unit tests related to refunds are failing. As far as I can tell, logging the data, what's happening (for example) in `test_refund_payment` is that the download earnings are being updated twice in the refund process. I think we want to remove the [decrease functions](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8804/includes/payments/functions.php#L306-L318) from `edd_undo_purchase`, maybe, but at this point I'm just putting this up as a draft for further discussion.

### DIFF
--- a/includes/orders/functions/refunds.php
+++ b/includes/orders/functions/refunds.php
@@ -342,8 +342,6 @@ function edd_refund_order( $order_id, $order_items = 'all', $adjustments = 'all'
 		if ( ! empty( $order_item['parent'] ) && ! empty( $order_item['original_item_status'] ) ) {
 			edd_update_order_item( $order_item['parent'], array( 'status' => $order_item['original_item_status'] ) );
 		}
-		edd_decrease_earnings( $order_item['product_id'], $order_item['amount'] );
-		edd_decrease_purchase_count( $order_item['product_id'], $order_item['quantity'] );
 	}
 
 	/** Insert order adjustments **********************************************/
--- a/includes/orders/functions/refunds.php
+++ b/includes/orders/functions/refunds.php
@@ -342,6 +342,8 @@ function edd_refund_order( $order_id, $order_items = 'all', $adjustments = 'all'
 		if ( ! empty( $order_item['parent'] ) && ! empty( $order_item['original_item_status'] ) ) {
 			edd_update_order_item( $order_item['parent'], array( 'status' => $order_item['original_item_status'] ) );
 		}
+		edd_decrease_earnings( $order_item['product_id'], $order_item['amount'] );
+		edd_decrease_purchase_count( $order_item['product_id'], $order_item['quantity'] );
 	}
 
 	/** Insert order adjustments **********************************************/
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -1987,14 +1987,6 @@ class EDD_Payment {
 					$this->process_pending();
 					break;
 			}
-
-			do_action( 'edd_update_payment_status', $this->ID, $status, $old_status );
-
-			if ( 'complete' === $old_status ) {
-				// Trigger the action again to account for add-ons listening for status changes from "publish".
-
-				do_action( 'edd_update_payment_status', $this->ID, $status, 'publish' );
-			}
 		}
 
 		return $updated;
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -1979,6 +1979,7 @@ class EDD_Payment {
 			switch ( $status ) {
 				case 'refunded':
 					$this->process_refund();
+					do_action( 'edd_update_payment_status', $this->ID, $status, $old_status );
 					break;
 				case 'failed':
 					$this->process_failure();
--- a/includes/emails/functions.php
+++ b/includes/emails/functions.php
@@ -226,12 +226,16 @@ function edd_get_email_names( $user_info, $payment = false ) {
 
 	if ( $payment instanceof EDD_Payment ) {
 
+		$email_names['name']     = $payment->email;
+		$email_names['username'] = $payment->email;
 		if ( $payment->user_id > 0 ) {
 
-			$user_data = get_userdata( $payment->user_id );
-			$email_names['name']      = $payment->first_name;
-			$email_names['fullname']  = trim( $payment->first_name . ' ' . $payment->last_name );
-			$email_names['username']  = $user_data->user_login;
+			$user_data               = get_userdata( $payment->user_id );
+			$email_names['name']     = $payment->first_name;
+			$email_names['fullname'] = trim( $payment->first_name . ' ' . $payment->last_name );
+			if ( ! empty( $user_data->user_login ) ) {
+				$email_names['username'] = $user_data->user_login;
+			}
 
 		} elseif ( ! empty( $payment->first_name ) ) {
 
@@ -239,13 +243,7 @@ function edd_get_email_names( $user_info, $payment = false ) {
 			$email_names['fullname'] = trim( $payment->first_name . ' ' . $payment->last_name );
 			$email_names['username'] = $payment->first_name;
 
-		} else {
-
-			$email_names['name']     = $payment->email;
-			$email_names['username'] = $payment->email;
-
 		}
-
 	} else {
 
 		if ( is_serialized( $user_info ) ) {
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by ashleyfae** [comment]
Just letting you know I'm looking into this and will probably have some info shortly.

File: includes/orders/functions/refunds.php [hunk 1]
```diff
@@ -342,6 +342,8 @@ function edd_refund_order( $order_id, $order_items = 'all', $adjustments = 'all'
 		if ( ! empty( $order_item['parent'] ) && ! empty( $order_item['original_item_status'] ) ) {
 			edd_update_order_item( $order_item['parent'], array( 'status' => $order_item['original_item_status'] ) );
 		}
+		edd_decrease_earnings( $order_item['product_id'], $order_item['amount'] );
+		edd_decrease_purchase_count( $order_item['product_id'], $order_item['quantity'] );
```
ashleyfae, line 346: I think we want to remove these. The download earnings specifically says "Gross Revenue", so it's not net.

Some discussion here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8659

ashleyfae: Although actually I see there is logic to decrease earnings in `edd_undo_purchase()`. So maybe the better move is to leave this addition in and open an entirely new issue to address this.

includes/payments/class-edd-payment.php
```diff
@@ -1987,14 +1987,6 @@ public function update_status( $status = '' ) {
 					$this->process_pending();
 					break;
 			}
-
-			do_action( 'edd_update_payment_status', $this->ID, $status, $old_status );
```
ashleyfae, line 1991: I _think_ we want to bring this back _but_ wrap it in:

```php
if ( 'refunded' === $status ) {
    // code here
}
```

Reason being:

- The order only gets updated if the status is NOT `refunded`. 
- That means, if changing to `refunded`, the new order transition hook will never run.
- But previously this hook _did_ run for all statuses, including `refunded`.
- So in short: we need this back _just_ to handle refunds, then all other statuses are covered by the order update.

includes/orders/functions/refunds.php
robincornett: If I take these out, though, and add the payment status hook for refunds back into that class, the download earnings numbers and counts start passing again.

However, the overall store earnings start to fail. In `test_refund_payment`, with those changes, I log this kind of information:
- `$store_earnings`: $360
- `$this->payment->total`: 120
- `edd_get_total_earnings()`: 120

The two failures specifically are:

> There were 2 failures:
> 
> 1) EDD\Orders\EDD_Payment_Tests::test_refund_payment
> Failed asserting that 120.0 matches expected 240.0.
> 
> /tmp/RWNMriU0H05xvvWnjeXJYHmrTcPWdM3b/tests/orders/tests-edd-payment.php:501
> 
> 2) EDD\Orders\EDD_Payment_Tests::test_refund_affecting_stats
> Failed asserting that 157.0 matches expected 277.0.

So....removing these lines and adding the extra hook in the payment class is some kind of progress.

includes/payments/class-edd-payment.php
robincornett: I did this but put it inside the `refunded` case in the switch statement above.

ashleyfae:
Got a solution for this, will follow up shortly.

ashleyfae:
So `edd_get_total_earnings()` has caching in it. The unit tests pass if you add this:

```php
delete_option('edd_earnings_total');
```

before each test does this:

```php
$store_earnings = edd_get_total_earnings();
$store_sales    = edd_get_total_sales();
```

That clears the cache.

I noticed this because the payment in question we're testing with is for `120`. But in the tests, `$store_earnings = edd_get_total_earnings();` was coming back as `240`, which... doesn't make sense.

robincornett:
Tested so far:
- Software Licensing: license key is disabled on refund
- Wallet: money is put back in the wallet on refund
- Stripe: refund processed (although this actually uses the `edd_refund_order` hook as of 2.8.7)
- All Access: using Stripe with preapproval enabled, the pass is invalid. When I mark the payment complete, the pass is activated.

ashleyfae:
## Tested

- [x] AffiliateWP (revoke unpaid referral on refund)

ashleyfae (approved):

cklosowski:
@ashleyfae Do we really need to cache `edd_get_total_earnings` and `edd_get_total_sales` still if it is a SUM query to get this data? Basically, do we need to keep the value for the ledger balance for those values?

We did this because we were having to run expensive queries to get these numbers previously, but with the new structure a SUM query should be able to achieve this pretty easily. Since we could rely on WP's query cache, maybe those option based cache values are not needed. They do cause some level of support tickets when they don't re-count.

PR #8812, opened 2021-08-11
Show error message if we think the modal didn't open #8811

### LINKED ISSUE #8811: 2.11 - PayPal Commerce - Try to detect if PayPal connect wasn't opened in the modal

And if not, show an error message.

### Justification or use case

Providing more useful feedback to the end user / UX.

ashleyfae:
Fixes #8811

Proposed Changes:
1. Moved some duplicated `json_decode( get_option( 'edd_paypal_commerce_connect_details_' . $mode ) );` code into a new function: `get_partner_details()`.
2. If we don't have REST API credentials, but we do have partner connect information, then we assume that the modal didn't work and display an error message:

Screenshot from 2021-08-11 16-04-50

@mihaijoldis I'd love for you to test this since you had popups blocked and were able to trigger this behaviour reliably. Basically just attempt to connect with PayPal when popups are blocked, make sure the PayPal modal DOESN'T open, continue through the process and click back to your site. Hopefully you should see the above error message.

@LisaCee and @robincornett you're welcome to check the above too, but also do a normal connection with the modal and make sure you DON'T see this error message.

### DIFF
--- a/includes/gateways/paypal/admin/connect.php
+++ b/includes/gateways/paypal/admin/connect.php
@@ -46,7 +46,14 @@ function connect_settings_field() {
 			?>
 			<div class="notice notice-error inline">
 				<p>
-					<?php esc_html_e( 'Connection failure. Please ensure browser popups are enabled then click the button below to connect again. If you continue to experience this error, please contact support.', 'easy-digital-downloads' ); ?>
+					<?php
+					echo wp_kses( sprintf(
+						/* Translators: %1$s opening <strong> tag; %2$s closing </strong> tag */
+						__( '%1$sConnection failure:%2$s Please ensure browser popups are enabled then click the button below to connect again. If you continue to experience this error, please contact support.', 'easy-digital-downloads' ),
+						'<strong>',
+						'</strong>'
+					), array( 'strong' => array() ) );
+					?>
 				</p>
 			</div>
 			<?php
--- a/includes/gateways/paypal/admin/connect.php
+++ b/includes/gateways/paypal/admin/connect.php
@@ -37,6 +37,20 @@ function connect_settings_field() {
 		/**
 		 * Show Connect
 		 */
+
+		/*
+		 * If we have Partner details but no REST credentials then that most likely means
+		 * PayPal wasn't opened in the modal. We'll show an error message about popups.
+		 */
+		if ( get_partner_details() ) {
+			?>
+			<div class="notice notice-error inline">
+				<p>
+					<?php esc_html_e( 'Connection failure. Please ensure browser popups are enabled then click the button below to connect again. If you continue to experience this error, please contact support.', 'easy-digital-downloads' ); ?>
+				</p>
+			</div>
+			<?php
+		}
 		?>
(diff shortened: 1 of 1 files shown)


ashleyfae:
Also input on the wording is welcome.

LisaCee (approved):
I love this. Successful login works perfectly. I also get the notice if I close the modal without filling it out and then refresh the page.

includes/gateways/paypal/admin/connect.php
```diff
@@ -37,6 +37,20 @@ function connect_settings_field() {
...
+
+		/*
+		 * If we have Partner details but no REST credentials then that most likely means
+		 * PayPal wasn't opened in the modal. We'll show an error message about popups.
+		 */
+		if ( get_partner_details() ) {
+			?>
+			<div class="notice notice-error inline">
+				<p>
+					<?php esc_html_e( 'Connection failure. Please ensure browser popups are enabled then click the button below to connect again. If you continue to experience this error, please contact support.', 'easy-digital-downloads' ); ?>
```
robincornett, line 49:
```suggestion
					<?php wp_kses( printf( __( '%sConnection failure:%s Please ensure browser popups are enabled, then click the button below to connect again. If you continue to experience this error, please contact support.', 'easy-digital-downloads' ), '<strong>', '</strong>' ), array( 'strong' => array() ) ); ?>
```
That's sloppy and doesn't include translator comments, but what would you think of adjusting to match the general notices pattern (strong: notice)?

Note: if my site is connected in test mode, and I enable Live mode, I do get this error. My only concern about that process is that you only see this notice on the PayPal settings page. If I switch from Test to Live without visiting that screen, I won't know there is an issue. I think we wrapped an inline notice with the Test Mode checkbox in Stripe to account for this kind of scenario--wonder if we should here as well?

robincornett (commented):
This is working for me (I can trigger reliably by switching to Live mode and I don't have those credentials on my local site). Just a thought about the error message format, and question about switching from test to live mode.

mihaijoldis:
With popups blocked I see the message that @ashleyfae mentioned so that is good.

includes/gateways/paypal/admin/connect.php
ashleyfae: Updated the message here: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8812/commits/8e584adf9396cdf617ebf89a97ee58ce287e4369

Will address your other question separately.

ashleyfae:
> Note: if my site is connected in test mode, and I enable Live mode, I do get this error. My only concern about that process is that you only see this notice on the PayPal settings page. If I switch from Test to Live without visiting that screen, I won't know there is an issue. I think we wrapped an inline notice with the Test Mode checkbox in Stripe to account for this kind of scenario--wonder if we should here as well?

@robincornett I think you might have had some stray data left over from somewhere. Had you previously started (but not finished) the live connection?

I started everything fresh and did not get this notice.

- I was connected in both test and live mode, so I disconnected from both, which ensures all PayPal-related data got wiped.
- I then connected in test mode. Did not get the notice.
- I turned on live mode.
- Checked and I did not have the notice.

If this is related to your question in Slack the other day, it does sound like you had half of the live information saved from a connection at one point, so that's why you would have seen the notice.

robincornett (approved):

mihaijoldis (approved):

PR #8814, opened 2021-08-12
Only start session when adding an item to the cart #8813

### LINKED ISSUE #8813: Only start session when adding an item to the cart

Right now we start a session all the time. We'd make a lot of people happy if we only started a session when first adding an item to your cart. From there we'd maintain the session until the cart is completely emptied.

### Justification or use case

This would allow sites to actually utilize caching --- at least until an item is added to the cart.

ashleyfae:
Fixes #8813

Proposed Changes:
1. `EDD_Session::should_start_session()`
    - The default value used to be `true`, but now it defaults to whether or not there's an item in the cart. To figure that out, we use the `edd_items_in_cart` cookie. If that is set, we assume the session should start. If that's not set, we assume it shouldn't be.
    - Added a `$force` parameter. When set to `true`, we bypass the `should_start_session()` check.
    - Removed PHP `< 5.4` logic
2. Introduced `EDD_Session::force_start_session()`. When called, this sets the `edd_items_in_cart` cookie and forcibly starts the session.
3. Introduced `EDD_Session::close_session()`. When called, this unsets the `edd_items_in_cart` cookie. There's an opt in filter to also completely terminate the session. Off by default because maybe other plugins are using sessions and we don't want to mess with that. :shrug: 
4. `EDD_Session::set()` now always calls `force_start_session()` if the value being set isn't empty. This is how we start the session when an item is being added to the cart. (Adding to cart calls `set()`, which then calls `force_start_session()`.
5. `EDD_Cart` - Emptying the cart completely triggers the `EDD_Session::close_session()`.

---

Testing this is a low priority for now. (In other words: don't yet. :stuck_out_tongue: ) Just felt like experimenting with it to get a basic proof of concept.

### DIFF
--- a/includes/cart/class-edd-cart.php
+++ b/includes/cart/class-edd-cart.php
@@ -512,6 +512,10 @@ class EDD_Cart {
 		$this->contents = $cart;
 		$this->update_cart();
 
+		if ( $this->is_empty() && ! $this->has_discounts() ) {
+			EDD()->session->close_session();
+		}
+
 		do_action( 'edd_post_remove_from_cart', $key, $item_id );
 
 		edd_clear_errors();
@@ -584,6 +588,9 @@ class EDD_Cart {
 		// Remove any active discounts
 		$this->remove_all_discounts();
 		$this->contents = array();
+
+		// Close session.
+		EDD()->session->close_session();
 
 		do_action( 'edd_empty_cart' );
 	}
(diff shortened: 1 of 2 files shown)


cklosowski:
This is working great when the `EDD_USE_PHP_SESSOINS` is true, but as soon as it is set to `false` we get empty carts. Going to see if I can account for that.

PR #8817, opened 2021-08-16
Only calculate diff if dates are not equal & catch exceptions #8816

### LINKED ISSUE #8816: 3.0 - Migration - PHP Fatal error:  Uncaught Exception: DateInterval::__construct(): Unknown or bad format (PT)
### Expected behavior

I expect to not get any errors during the migration to 3.0.

### Actual behavior

I got the following fatal error:

```
PHP Fatal error:  Uncaught Exception: DateInterval::__construct(): Unknown or bad format (PT) in easy-digital-downloads/includes/admin/upgrades/v3/class-data-migrator.php:570
```

The code block in question:

```php
// Account for a situation where the post_date_gmt is set to 0000-00-00 00:00:00
$date_created_gmt = $data->post_date_gmt;
if ( '0000-00-00 00:00:00' === $date_created_gmt ) {

	$date_created_gmt  = new \DateTime( $data->post_date );
	$modified_time     = new \DateTime( $data->post_modified );
	$modified_time_gmt = new \DateTime( $data->post_modified_gmt );

	$diff = $modified_time_gmt->diff( $modified_time );

	$time_diff = 'PT';

	// Add hours to the offset string.
	if ( ! empty( $diff->h ) ) {
		$time_diff .= $diff->h . 'H';
	}

	// Add minutes to the offset string.
	if ( ! empty( $diff->i ) ) {
		$time_diff .= $diff->i . 'M';
	}

	// Account for -/+ GMT offsets.
	if ( 1 === $diff->invert ) {
		$date_created_gmt->add( new \DateInterval( $time_diff ) );
	} else {
		$date_created_gmt->sub( new \DateInterval( $time_diff ) );
	}

	$date_created_gmt = $date_created_gmt->format('Y-m-d H:i:s');
}
```

The payment that caused the error looks like this:

```
select ID,post_date,post_date_gmt,post_modified,post_modified_gmt from wp_posts where post_type = 'edd_payment';
+----+---------------------+---------------------+---------------------+---------------------+
| ID | post_date           | post_date_gmt       | post_modified       | post_modified_gmt   |
+----+---------------------+---------------------+---------------------+---------------------+
| 17 | 2021-08-12 12:18:49 | 0000-00-00 00:00:00 | 2021-08-12 12:18:49 | 2021-08-12 12:18:49 |
```

Note how `post_date_gmt` is `0000-00-00 00:00:00` -- that's required to get in this `if` block. And the `post_modified` and `post_modified_gmt` dates are the same.

I don't actually know how I got this data in the first place. The payment in question is `pending` but I actually created this payment last week and don't remember what I was doing. But I reckon you could spoof the error with these steps:

1. Have a payment record in EDD `master` branch.
2. Manually edit the DB to set the dates to what's shown above in the table. (Set all of them.)
3. Checkout `release/3.0` and run the upgrade to 3.0.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8816

Proposed Changes:
1. Only calculate the time difference if `$modified_time` is different from `$modified_time_gmt`
2. Catch exceptions from `DateInterval` operations. Worst case scenario here we end up with `$date_created_gmt` unmodified, which is no biggie.

See issue for steps to reproduce this problem. If you use my data, then post-migration you should see the date like this:

Screenshot from 2021-08-16 13-28-22

(Note my store time zone is UTC+0)

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -534,25 +534,31 @@ class Data_Migrator {
 			$modified_time     = new \DateTime( $data->post_modified );
 			$modified_time_gmt = new \DateTime( $data->post_modified_gmt );
 
-			$diff = $modified_time_gmt->diff( $modified_time );
+			if ( $modified_time != $modified_time_gmt ) {
+				$diff = $modified_time_gmt->diff( $modified_time );
 
-			$time_diff = 'PT';
+				$time_diff = 'PT';
 
-			// Add hours to the offset string.
-			if ( ! empty( $diff->h ) ) {
-				$time_diff .= $diff->h . 'H';
-			}
+				// Add hours to the offset string.
+				if ( ! empty( $diff->h ) ) {
+					$time_diff .= $diff->h . 'H';
+				}
 
-			// Add minutes to the offset string.
-			if ( ! empty( $diff->i ) ) {
-				$time_diff .= $diff->i . 'M';
-			}
+				// Add minutes to the offset string.
+				if ( ! empty( $diff->i ) ) {
+					$time_diff .= $diff->i . 'M';
+				}
 
-			// Account for -/+ GMT offsets.
-			if ( 1 === $diff->invert ) {
-				$date_created_gmt->add( new \DateInterval( $time_diff ) );
-			} else {
-				$date_created_gmt->sub( new \DateInterval( $time_diff ) );
+				// Account for -/+ GMT offsets.
+				try {
+					if ( 1 === $diff->invert ) {
+						$date_created_gmt->add( new \DateInterval( $time_diff ) );
+					} else {
+						$date_created_gmt->sub( new \DateInterval( $time_diff ) );
+					}
+				} catch ( \Exception $e ) {
+
+				}
 			}
 
 			$date_created_gmt = $date_created_gmt->format('Y-m-d H:i:s');


includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -534,25 +534,31 @@ public static function orders( $data = null ) {
 			$modified_time     = new \DateTime( $data->post_modified );
 			$modified_time_gmt = new \DateTime( $data->post_modified_gmt );
 
-			$diff = $modified_time_gmt->diff( $modified_time );
+			if ( $modified_time != $modified_time_gmt ) {
```
ashleyfae, line 537: @robincornett because I know you'll notice: this is intentionally not a strict check. We're comparing two `DateTime` objects, and a strict check won't actually work. But lose one does.

robincornett:
I am trying to edit the `post_date_gmt` two different ways to the `0000` value and both ways, I can't. I get these errors in both programs I'm using:

> Incorrect datetime value: '0000-00-00 00:00:00' for column 'post_date_gmt' at row 1 12:49:24 SQL command
> Incorrect datetime value: '' for column 'post_date_gmt' at row 1 12:46:19 SQL command

But reading the logic of the code, I am willing to believe it works. Is there a way I can override the database protecting me from myself?

ashleyfae:
> I am trying to edit the `post_date_gmt` two different ways to the `0000` value and both ways, I can't. I get these errors in both programs I'm using:
> 
> > Incorrect datetime value: '0000-00-00 00:00:00' for column 'post_date_gmt' at row 1 12:49:24 SQL command
> > Incorrect datetime value: '' for column 'post_date_gmt' at row 1 12:46:19 SQL command
> 
> But reading the logic of the code, I am willing to believe it works. Is there a way I can override the database protecting me from myself?

I believe that's a default setting in the MySQL version you're using. You'll have to disable `NO_ZERO_DATE`.

You can try this: https://stackoverflow.com/questions/60186508/how-to-remove-no-zero-date-from-sql-mode-in-mysql

robincornett (approved):
Thanks for the instructions on the database. I was able to break that data column. The order migrated successfully with the correct date.

PR #8824, opened 2021-08-23
Remove locale from PayPal SDK parameters

### LINKED ISSUE #8823: PayPal Commerce: get_locale can provide an incorrect string to PayPal
### Expected behavior
When using a different locale, I expect PayPal Commerce to load.

### Actual behavior
If the `get_locale` function returns anything other than a 5 character locale string, PayPal Commerce throws a 400 error. Basically, there are some locales that WordPress returns a 2 or 3 character locale for like `ja` instead of a locale string like `ja_JP`. This does not line up with a local that PayPal Supports. We need to find a way to map any 2/3 character locales to their fully formatted locale_LANG string.

Report: https://wordpress.org/support/topic/2-11-paypal-locale-error/

1) Set your store to use a locale of `ja` for Japan: `日本語` in the dropdown.
2) Attempt to load the checkout with Paypal Commerce
3) Notice that the dev tools shows a 400 and the `locale` parameter is `ja` not `ja_JP`

PHP Version: All

EDD Version (or branch): 2.11

WordPress Version: 5.8 (likely all)

robincornett:
Fixes #8823

Proposed Changes:
1. Based on the PayPal documentation (see notes in the issue), I believe we are best off removing the `locale` parameter from this entirely.

### DIFF
--- a/includes/gateways/paypal/scripts.php
+++ b/includes/gateways/paypal/scripts.php
@@ -79,7 +79,6 @@ function register_js( $force_load = false ) {
 	 */
 	$sdk_query_args = apply_filters( 'edd_paypal_js_sdk_query_args', array(
 		'client-id'       => urlencode( $api->client_id ),
-		'locale'          => urlencode( get_locale() ),
 		'currency'        => urlencode( strtoupper( edd_get_currency() ) ),
 		'intent'          => 'capture',
 		'disable-funding' => 'card,credit,bancontact,blik,eps,giropay,ideal,mercadopago,mybank,p24,sepa,sofort,venmo'


LisaCee (approved):

cklosowski:
I'm waiting on feedback from our PayPal contacts to verify that this is ideal. The docs we have for the SDK mention 'Checkout' not the 'Commerce' API, so I want to just make sure we don't run into any issues.

cklosowski:
Still waiting here, our technical contact is checking with the Commerce product team to be sure:
```
Hi Chris, Yes our JS SDK documentation specifically says not to use the locale: “The locale used to localize any components. PayPal recommends not setting this parameter, as the buyer's locale is automatically set by PayPal.“

But we’ll confirm once with our Product team just to be sure and get back.
```

PR #8835, opened 2021-08-26
Add order items to the PayPal API data

### LINKED ISSUE #8833: PayPal Commerce: Product name is missing inside PayPal
### Expected behavior
PayPal transactions should include the product name in the Order details.

CleanShot 2021-08-25 at 15 06 44@2x

### Actual behavior
Product name is missing from the PayPal Order details. 

CleanShot 2021-08-25 at 15 03 17@2x

1) Make a purchase with PayPal Commerce enabled
2) From EDD Order Details, click Transaction ID to open in PayPal. Confirm product is listed  under the order details.

PHP Version:

EDD Version (or branch): 2.11

WordPress Version:

Reported by customer: https://secure.helpscout.net/conversation/1612030000/192611?folderId=180499
I replicated this issue on my site.

robincornett:
Fixes #8833

Proposed Changes:
1. Create an array of `$items` from the cart details and send that to PayPal.

What the modal shows:

image

PayPal transaction:

image

This needs to be tested with discounts, taxes, SKU, probably fees...the order total is still coming from the original setup, so it's not calculated from this new array of data. The individual item amount calculation is copied verbatim from PayPal standard.


includes/gateways/paypal/checkout-actions.php
```diff
@@ -128,12 +128,13 @@ function create_order( $purchase_data ) {
 
 	try {
 		// Create pending payment in EDD.
+		$currency     = edd_get_currency();
```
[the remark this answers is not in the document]
robincornett: Tested with Aelia, store currency USD but order completed in EUR.

Order data sent to PayPal:
```
[30-Aug-2021 12:11:28 UTC] $order_data
[30-Aug-2021 12:11:28 UTC] Array
(
    [intent] => CAPTURE
    [purchase_units] => Array
        (
            [0] => Array
                (
                    [reference_id] => 0dbf90f43f57a483e5a9d0f38f842c16
                    [amount] => Array
                        (
                            [currency_code] => EUR
                            [value] => 17.16
                            [breakdown] => Array
                                (
                                    [item_total] => Array
                                        (
                                            [currency_code] => EUR
                                            [value] => 15.7
                                        )

                                    [tax_total] => Array
                                        (
                                            [currency_code] => EUR
                                            [value] => 1.46
                                        )

                                )

                        )

                    [custom_id] => 3748
                    [items] => Array
                        (
                            [0] => Array
                                (
                                    [name] => Download with a SKU
                                    [quantity] => 1
                                    [unit_amount] => Array
                                        (
                                            [currency_code] => EUR
                                            [value] => 14.88
                                        )

                                    [sku] => ANB324
                                )

                            [1] => Array
                                (
                                    [name] => PayPal Commerce Fee
                                    [unit_amount] => Array
                                        (
                                            [currency_code] => EUR
                                            [value] => 0.82
                                        )

                                    [quantity] => 1
                                )

                        )

                )

        )

    [application_context] => Array
        (
            [shipping_preference] => NO_SHIPPING
            [user_action] => PAY_NOW
            [return_url] => https://edd.local/checkout/
            [cancel_url] => https://edd.local/checkout/transaction-failed/?payment-id=3748
        )

    [payment_instructions] => Array
        (
            [disbursement_mode] => INSTANT
        )

    [payer] => Array
        (
            [email_address] => dev-email@flywheel.local
            [name] => Array
                (
                    [given_name] => Robin
                    [surname] => Cornett
                )

        )

)
```

PayPal response:
```
[30-Aug-2021 12:11:28 UTC] $response
[30-Aug-2021 12:11:28 UTC] stdClass Object
(
    [id] => 90352532UExxxxxx
    [status] => CREATED
    [links] => Array
        (
            [0] => stdClass Object
                (
                    [href] => https://api.sandbox.paypal.com/v2/checkout/orders/90352532UExxxxxx
                    [rel] => self
                    [method] => GET
                )

            [1] => stdClass Object
                (
                    [href] => https://www.sandbox.paypal.com/checkoutnow?token=90352532UExxxxxx
                    [rel] => approve
                    [method] => GET
                )

            [2] => stdClass Object
                (
                    [href] => https://api.sandbox.paypal.com/v2/checkout/orders/90352532UExxxxxx
                    [rel] => update
                    [method] => PATCH
                )

            [3] => stdClass Object
                (
                    [href] => https://api.sandbox.paypal.com/v2/checkout/orders/90352532UExxxxxx/capture
                    [rel] => capture
                    [method] => POST
                )

        )

)
```

Transaction view in PayPal:
image

includes/gateways/paypal/checkout-actions.php
```diff
@@ -154,20 +155,84 @@ function create_order( $purchase_data ) {
...
+						'currency_code' => $currency,
+						'value'         => edd_sanitize_amount( $item_amount ),
+					),
+					'quantity'    => $item['quantity'],
+				);
+				if ( (float) $item['discount'] > 0 ) {
+					$order_subtotal -= ( $item['discount'] * $item['quantity'] );
+				}
+				if ( edd_use_skus() ) {
+					$items[ $i ]['sku'] = edd_get_download_sku( $item['id'] );
```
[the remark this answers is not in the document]
robincornett: I've ended up checking if the sku function returns empty (possible if the filter is used) or `-` (default if no sku is set).

includes/gateways/paypal/checkout-actions.php
```diff
@@ -154,20 +155,87 @@ function create_order( $purchase_data ) {
 			);
 		}
 
+		$order_subtotal = $purchase_data['subtotal'];
+		// Create an array of items for the order.
+		$items = array();
+		$i     = 0;
+		if ( is_array( $purchase_data['cart_details'] ) && ! empty( $purchase_data['cart_details'] ) ) {
+			foreach ( $purchase_data['cart_details'] as $item ) {
+				$item_amount = round( ( $item['subtotal'] / $item['quantity'] ) - ( $item['discount'] / $item['quantity'] ), 2 );
```
ashleyfae, line 164: I'm inclined to say we remove this `round()`. On line 174 it gets run through `edd_sanitize_amount()`, which already has logic for rounding and with a filter in place:

```php
/**
 * Filter number of decimals to use for prices
 *
 * @since unknown
 *
 * @param int $number Number of decimals
 * @param int|string $amount Price
 */
$decimals = apply_filters( 'edd_sanitize_amount_decimals', 2, $amount );
$amount   = number_format( (double) $amount, $decimals, '.', '' );
```

includes/gateways/paypal/checkout-actions.php
```diff
@@ -178,8 +246,9 @@ function create_order( $purchase_data ) {
 					// @todo We could put the breakdown here (tax, discount, etc.)
```
ashleyfae, line 246: This `@todo` can be removed now.

ashleyfae (changes requested):
Can you split off all the new code (everything new that's used to build `$items`) into a new function? That way this `create_order()` function will be a bit simpler and we could also even write unit tests for the separate items output if we want.

dgoldak:
Re-tested and passed:
- 1 time purchase, 1 item
- 1 time purchase, 2 items
- recurring with fee
- recurring with fee and taxes
- taxes (inclusive & exclusive)
- taxes (inclusive & exclusive) with discount code
- taxes (inclusive & exclusive) with Discounts Pro
- SKU

includes/gateways/paypal/checkout-actions.php
```diff
@@ -154,32 +155,79 @@ function create_order( $purchase_data ) {
...
+		$tax          = (float) $purchase_data['tax'] > 0 ? $purchase_data['tax'] : 0;
 		$order_amount = array(
-			'currency_code' => edd_get_currency(),
-			'value'         => (string) $purchase_data['price']
-		);
-		if ( (float) $purchase_data['tax'] > 0 ) {
-			$order_amount['breakdown'] = array(
+			'currency_code' => $currency,
+			'value'         => (string) ( $order_subtotal + $tax - $discount ),
+			'breakdown'     => array(
```
ashleyfae, line 200: I'm feeling semi uncomfortable about the way these breakdowns work. We're essentially recalculating the subtotal, discount, tax, and order total, instead of using the values provided in `$purchase_data`. My concern being ending up in another situation like this: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6837

If we ever change order calculations somehow, we now have to update that logic in two places (standard purchase data place, and PayPal-specific overview).

One of the things I do want to do at some point is improve the information that's currently available in `$purchase_data`, because it could definitely be more useful, but I guess my point still stands that I'm still feeling concerned about how much we're rewriting here.

Is there any reason we can't just use the values already available in `$purchase_data` ?

robincornett: I think maybe we can for the overall order value. Looking back through my commits, it looks like I changed that because the item_total value was giving me fits and I was mentally tying them together. I think the item_total one does need to be handled differently because of how taxes, discounts, and fees end up resolving, but some tentative tests suggest that this can be reverted to `$purchase_data['price']`.

ashleyfae (commented):
This is working well but I'd still like to move some more of the logic out of this function and into the new `get_order_items()` function (which might need a rename after my proposal).

I think everything you've left in this `create_order()` function is there because it affects the `$order_subtotal` variable so it needs to be available here?

Could we instead move everything that goes into `purchase_units` into the `get_order_items()` function? (Which we could now called `get_order_purchase_units()` ? Or you could have two functions overall: `get_order_purchase_units()` which then also includes the existing `get_order_items()` (actually I kind of like this).

So that would include all of `$order_amount`, all of the fees logic, all the discount logic, etc.  Then we'd just be left with:

```php
$order_data = array(
	'intent'               => 'CAPTURE',
	'purchase_units'       => get_order_units( $purchase_data ),
	// ...
);
```

I just want to keep `create_order()` smaller.

includes/gateways/paypal/checkout-actions.php
```diff
@@ -154,32 +155,79 @@ function create_order( $purchase_data ) {
...
+				if ( ! empty( $fee['download_id'] ) ) {
+					continue;
+				}
+				// Positive fees.
+				if ( floatval( $fee['amount'] ) > 0 ) {
+					$items[]         = array(
+						'name'        => stripslashes_deep( html_entity_decode( wp_strip_all_tags( $fee['label'] ), ENT_COMPAT, 'UTF-8' ) ),
+						'unit_amount' => array(
+							'currency_code' => $currency,
+							'value'         => edd_sanitize_amount( $fee['amount'] ),
```
ashleyfae, line 184: Just in case, cast this to a string please. That's how PayPal wants all amount values.

```suggestion
							'value'         => (string) edd_sanitize_amount( $fee['amount'] ),
```

includes/gateways/paypal/functions.php
```diff
@@ -138,3 +138,44 @@ function get_button_styles() {
...
+
+		if ( $item_amount <= 0 ) {
+			$item_amount = 0;
+		}
+		$items[ $i ] = array(
+			'name'        => stripslashes_deep( html_entity_decode( edd_get_cart_item_name( $item ), ENT_COMPAT, 'UTF-8' ) ),
+			'quantity'    => $item['quantity'],
+			'unit_amount' => array(
+				'currency_code' => edd_get_currency(),
+				'value'         => edd_sanitize_amount( $item_amount ),
```
ashleyfae, line 167:
```suggestion
				'value'         => (string) edd_sanitize_amount( $item_amount ),
```

ashleyfae:
Getting an error:

```
{"name":"UNPROCESSABLE_ENTITY","details":[{"field":"/purchase_units/@reference_id=='6ce4009808ecbdbafeec6955a40faa13'/amount/breakdown/item_total/value","value":"0","issue":"ITEM_TOTAL_MISMATCH","description":"Should equal sum of (unit_amount * quantity) across all items for a given purchase_unit"}]
```

My set up:

Screenshot from 2021-09-14 11-00-38

- Product costs $10. I have two in my cart.
- I'm using a 50% off discount code.
- Taxes enabled, at 20%.

Purchase units sent to PayPal:

```
  "purchase_units": [
    {
      "reference_id": "0faaf0b2f7442a0b7915020457b8b265",
      "amount": {
        "currency_code": "USD",
        "value": "11",
        "breakdown": {
          "item_total": {
            "currency_code": "USD",
            "value": "0"
          },
          "tax_total": {
            "currency_code": "USD",
            "value": "1.00"
          }
        }
      },
      "custom_id": 1710,
      "items": [
        {
          "name": "Buy Now Button",
          "quantity": 2,
          "unit_amount": {
            "currency_code": "USD",
            "value": "5.00"
          },
          "discount": 10
        }
      ]
    }
  ]
```

I believe the problem is here:

https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8835/files#diff-592139a8fe2bc5db27cc0632e1effe26e8d432588c57471193aeb8ef66b6e2a2R160

`$item['discount']` is already the **total** for all items. Multiplying it by the quantity doubles it, when it shouldn't be doubled.

robincornett:
Fixing that, and I think that possibly the [initial item amount](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/issue/8833/includes/gateways/paypal/functions.php#L244) needs to be changed as well (thinking to do subtotal - quantity and divide result by quantity?).

However, running into an issue if the calculated item total doesn't match the order total and funny numbers/quantities are involved. This is the order data. You can see that the item_total doesn't match the calculated value of the quantities/discounted items:

```
    [purchase_units] => Array
        (
            [0] => Array
                (
                    [reference_id] => 48ca5c74420ec7ba80bde25552cd4b1f
                    [amount] => Array
                        (
                            [currency_code] => USD
                            [value] => 45.05
                            [breakdown] => Array
                                (
                                    [item_total] => Array
                                        (
                                            [currency_code] => USD
                                            [value] => 40.95
                                        )

                                    [tax_total] => Array
                                        (
                                            [currency_code] => USD
                                            [value] => 4.10
                                        )

                                )

                        )

                    [custom_id] => 161
                    [items] => Array
                        (
                            [0] => Array
                                (
                                    [name] => Ship This
                                    [quantity] => 2
                                    [unit_amount] => Array
                                        (
                                            [currency_code] => USD
                                            [value] => 20.48
                                        )

                                )

                        )
```

What's happening here is that really one product value is 20.48 and one is 20.47 (or both are 20.475, but PayPal won't accept that), at least the way the math works out in the cart, I think? I'm trying to buy two of a simple shipping product so there is a fee involved (plus a 10% discount, but the discount being there or not doesn't change the problem). EDD adds the item fee into the item price, so the math is mismatched with the quantity:

image

Wondering if we need to reconsider the item calculation if item specific fees are involved, and add them separately as their own items, the way the global fees are currently. The item `subtotal` and `price` include the fees, but if we go down to the `item_price` then the fee is not included (also the discount is not.

ashleyfae:
Thinking about what I want to do here...

I just hate how fragile this whole thing is, and the fact that the whole API request fails if the amounts don't add up.

ashleyfae:
I think at minimum we definitely need to start adding unit tests for some of the trickier scenarios.

ashleyfae:
> Thinking about what I want to do here...
> 
> I just hate how fragile this whole thing is, and the fact that the whole API request fails if the amounts don't add up.

To be clear, @robincornett , when I said this I wasn't criticizing your work. That's great. Just a general comment on how frustrating it is how much we have to rebuild all this to work with PayPal.

Do you mind if I work up a sample unit test here? I'll do the one case I found above then we can work on a few other edge case scenarios and if you want you can take over adding more after I've set up an example.

robincornett:
Oh no worries, I didn't think you were talking about my work--I agree this does feel fragile and unfriendly. Some scenarios that chuck a wobbly that I know of are:

- positive "discounts" (apparently they exist) where a discount code is set to -50%
- quantities, evidently
- fees can be negative or positive and each are handled differently, plus if they are global or download specific
- I actually feel like taxes have been relatively well behaved here but probably inclusive taxes may be gnarly

ashleyfae:
I don't know if this is possible within PayPal's flow here... but it would also be really cool if we could detect the price mismatch error, and if it happens, make a new order without the line items.

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8835/commits/64dd347b0712ca44032547765a2b9e136ad3cb59 adds the above mentioned retry behaviour along with an initial unit test.

I tested the retry by undoing your fix on the discounts. Expected behaviour is that the modal pops up with the correct price but no item breakdown. Upon fixing the discount again, modal should pop up with correct price **and with** an item breakdown.

robincornett:
Noting that right now, this PR (with the items breakdown) breaks the `add_shipping_to_paypal_commerce` method in Simple Shipping. This is the response/error log information:

> Response Code: 422; Message: Unexpected response when creating order:
>
> Log data:
>
> {"name":"UNPROCESSABLE_ENTITY","details":[{"field":"/purchase_units/@reference_id=='597d9e595f3a7a42c9be86aeefd0b5b5'/shipping/address","issue":"MISSING_SHIPPING_ADDRESS","description":"The shipping address is required when `shipping_preference=SET_PROVIDED_ADDRESS`."}],"message":"The requested action could not be performed, semantically incorrect, or failed business validation.","debug_id":"521ff508f7e4","links":[{"href":"https://developer.paypal.com/docs/api/orders/v2/#error-MISSING_SHIPPING_ADDRESS","rel":"information_link","method":"GET"}]}

ashleyfae:
https://github.com/easydigitaldownloads/easy-digital-downloads/commit/ecf69557fe4897a19acfb476b96a6e358579e2ae re-applies the filter for the second attempt, which should fix that issue.

robincornett:
Okay, now working with Simple Shipping again!

Also--inclusive taxes!
image

ashleyfae:
I want to get some more unit tests written but so far I'm feeling a lot better about this. It makes me way more comfortable knowing we have a backup!

robincornett:
Inclusive taxes with a `-50%` discount applied (sorry for repeated screenshots but it is SO EXCITING):
image

includes/gateways/paypal/checkout-actions.php
```diff
@@ -219,6 +195,18 @@ function create_order( $purchase_data ) {
 			$api      = new API();
 			$response = $api->make_request( 'v2/checkout/orders', $order_data );
 
+			if ( ! isset( $response->id ) && _is_item_total_mismatch( $response ) ) {
+				// Try again without the item breakdown. That way if we have an error in our totals the whole API request won't fail.
```
robincornett, line 199: What would you think about adding an `edd_debug_log` here indicating that there was an item total mismatch and then logging the `$order_data`? That way we could see exactly what we were sending and maybe get some clues about where it could be fixed.

cklosowski: I think that is a good idea @robincornett, these types of things are always hard to track down.

ashleyfae: I like the idea but might opt for a gateway payment error instead, just because it's a bit more persistent than the debug log.

Do you want to do it @robincornett or do you want me to?

robincornett: Sure, I can do it, since I can consistently get the error to happen.

includes/gateways/paypal/checkout-actions.php
```diff
@@ -196,6 +196,17 @@ function create_order( $purchase_data ) {
 			$response = $api->make_request( 'v2/checkout/orders', $order_data );
 
 			if ( ! isset( $response->id ) && _is_item_total_mismatch( $response ) ) {
+
+				edd_record_gateway_error(
+					__( 'PayPal Gateway Error', 'easy-digital-downloads' ),
```
ashleyfae, line 201:
```suggestion
					__( 'PayPal Gateway Warning', 'easy-digital-downloads' ),
```

Just to clarify that the whole process did not fail.

ashleyfae:
@robincornett Do you want to add unit tests for:

1. Positive fee
2. Negative fee
3. Taxes

You can use this example as a boilerplate: https://github.com/easydigitaldownloads/easy-digital-downloads/blob/22ebc47737aa53f8934f2bbd5a742153fe76456a/tests/gateways/tests-paypal.php#L432-L483

I pretty much just made a purchase using the config I wanted to test, recorded [the purchase data that was present at the time](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/22ebc47737aa53f8934f2bbd5a742153fe76456a/tests/gateways/tests-paypal.php#L433-L449) then used that in the unit test.

ashleyfae (approved):
Awesome work on this @robincornett :raised_hands:

PR #8837, opened 2021-08-30
Fix hour/minute inputs on order details screen

### LINKED ISSUE #8803: 3.0 - time inputs for order attributes can be missized
### Expected behavior
When I view an individual order, I expect the order time inputs to look somewhat like a clock. Here's what it looks like in Chrome, which is visually what I expect:

image

### Actual behavior
However, in Firefox, the order time looks like this:

image

1. View an order in EDD 3.0/Firefox.

PHP Version: 7.4.1
WordPress Version: 5.8

I think these inputs just need a class of `small-text` to get the correct sizing.

robincornett:
Fixes #8803

Proposed Changes:
1. Add `class="edd-form-group__input small-text"` to the input fields for hour/minute on the order details screen.


LisaCee (approved):
Looks good in Chrome and FireFox

ashleyfae (approved):

PR #8842, opened 2021-09-01
Add support for required attribute to select/text inputs

### LINKED ISSUE #8433: Allow EDD_HTML_Elements fields to be required
It would be great if the input fields built with the `EDD_HTML_Elements` class allowed fields to be required so that the `required` attribute could be added via a parameter.

### Justification or use case
I've worked with a couple different extensions recently which use the elements class to build custom forms in the admin, and each have required fields which currently need extra validation to enforce it:

- EDD Message, for emailing individual customers. Subject/content/recipient are required.
- EDD Recurring, specifically for manually adding a subscription. Expiration date, product ID, initial amount, and recurring amount are all required.

robincornett:
Fixes #8433

Proposed Changes:
1. Adds support for `required` HTML5 attribute to select/text inputs, set to `false` by default.
2. Allow methods which support custom parameters to set `required`: `product_dropdown`, `customer_dropdown`, `user_dropdown`, `discount_dropdown`, `country_select`, `region_select`.

### DIFF
--- a/tests/tests-html-elements.php
+++ b/tests/tests-html-elements.php
@@ -198,6 +198,9 @@ class Test_HTML_Elements extends EDD_UnitTestCase {
 		$this->assertContains( '<option value="12"', $out );
 	}
 
+	/**
+	 * @covers EDD_HTML_Elements::select
+	 */
 	public function test_select_is_required() {
 		$select = EDD()->html->select(
 			array(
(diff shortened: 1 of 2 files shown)


ashleyfae (commented):
This looks good, but do you want to add a unit test? (see `tests-html-elements.php`).

tests/tests-html-elements.php
```diff
@@ -197,4 +197,41 @@ public function test_month_dropdown() {
 		$this->assertContains( '<option value="11"', $out );
 		$this->assertContains( '<option value="12"', $out );
 	}
+
+	public function test_select_is_required() {
```
ashleyfae, line 201: These are great, but for each one can you add a docblock like this:

```php
/**
 * @covers EDD_HTML_Elements::select
 */
```

For the one that covers text, you'd change `select` to `text` (the `::select` references the method name in the HTML class).

ashleyfae (approved):

PR #8844, opened 2021-09-01
Make sure an order object exists before using it in payment actions

### LINKED ISSUE #8843: 3.0 - fatal error creating a payment when there is no customer
### Expected behavior
When I create a new subscription (using EDD Recurring), I expect it to work without breaking things.

### Actual behavior
Unfortunately, I get a fatal error:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_address() on bool in \wp-content\plugins\easy-digital-downloads.30\includes\payments\class-edd-payment.php:2271
```

1. Activate Recurring on 3.0.
2. Set up a new subscription. I created a subscription which was not attached to a customer (I think that should not be possible but that's a different issue), so don't select a customer. Selecting a customer with no address does not trigger the error.
3. Click the button to create the subscription.

PHP Version: 7.4.1
WordPress Version: 5.8

It looks like the error is being triggered because the order is not created (because no customer). But the condition includes a call to the method to retrieve the address and the ID at the same time, but the order wasn't created. So I think maybe the update_meta function in the payments class needs to return early if the `ID` property isn't an order?

robincornett:
Fixes #8843

Proposed Changes:
1. If `edd_get_order` does not return an order in `edd_complete_purchase`, this prevents the rest of the function from running.
2. If there is not an order object/property in `check_status` in the payment class, this does an early return.
3. And the `update_meta` method now returns early if there is not a payment ID/property.

Basically I started with trying to create a subscription improperly in 3.0 and kept fixing errors until they stopped occurring. Note that in 3.0, creating a subscription without a customer attached fails to create an order at all. However, the same steps in 2.11.1 do create a new payment, although there is no customer attached. I'm not sure which behavior is preferable, to be honest, so putting the PR up as it is for now.

### DIFF
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -369,12 +369,12 @@ class EDD_Payment {
 
 		if ( $by_txn ) {
 			$payment_id = edd_get_order_id_from_transaction_id( $payment_or_txn_id );
+
+			if ( empty( $payment_id ) ) {
+				return false;
+			}
 		} else {
 			$payment_id = absint( $payment_or_txn_id );
-		}
-
-		if ( empty( $payment_id ) ) {
-			return false;
 		}
 
 		$this->setup_payment( $payment_id );
--- a/includes/payments/class-edd-payment.php
+++ b/includes/payments/class-edd-payment.php
@@ -369,12 +369,12 @@ class EDD_Payment {
 
 		if ( $by_txn ) {
 			$payment_id = edd_get_order_id_from_transaction_id( $payment_or_txn_id );
-
-			if ( empty( $payment_id ) ) {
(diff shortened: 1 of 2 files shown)


includes/payments/class-edd-payment.php
```diff
@@ -1898,6 +1898,10 @@ public function refund() {
 	 */
 	public function update_status( $status = '' ) {
 
+		if ( ! $this->order ) {
+			return;
```
ashleyfae, line 1902:
```suggestion
			return false;
```

Looks like this method is meant to return a boolean.

ashleyfae (changes requested):

ashleyfae:
Feel free to merge when tests pass.

PR #8846, opened 2021-09-07
Merge 2.11 & 2.11.1 releases into release/3.0

ashleyfae:
We should re-test PayPal connect and a front-end PayPal purchase as it's possible I totally broke the JavaScript on those (due to new 3.0 webpack config).


ashleyfae:
Failing unit tests are actually due to an undiscovered bug in 3.0! Will be making a separate issue for that. We can hold off on this until that's fixed.

ashleyfae:
This is ready for review.

Note that PayPal Commerce does need updating to use the 3.0 refunds API but I am creating a new issue for that.

includes/admin/upgrades/downgrades.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
```
robincornett, line 2: Wondering if this needs to be merged into 3.0, or do we need to make a 3.0 issue to remove it (either the entire file, or the pieces that call on it)?

includes/install.php
```diff
@@ -196,6 +196,11 @@ function edd_run_install( $site_id = false ) {
 
 	// Maybe save the previous version, only if different than current
 	if ( ! empty( $current_version ) && ( edd_format_db_version( EDD_VERSION ) !== $current_version ) ) {
+		if ( version_compare( $current_version, edd_format_db_version( EDD_VERSION ), '>' ) ) {
```
robincornett, line 199: eg. this could be removed because the downgrading won't apply.

robincornett (approved):
I'm going to go ahead and approve--everything looks correct in here and new PayPal works in 3.0.

I did note that we maybe could remove the downgrading file/functions since it was to handle going back down to 2.x from 3.0, but that's maybe worth just opening a separate issue over.

ashleyfae:
> I'm going to go ahead and approve--everything looks correct in here and new PayPal works in 3.0.
> 
> I did note that we maybe could remove the downgrading file/functions since it was to handle going back down to 2.x from 3.0, but that's maybe worth just opening a separate issue over.

Yeah let's handle that separately. It ultimately doesn't cause any harm to keep it and I might want to keep at least some of the pieces in place for future considerations anyway. (You know... when we do EDD 4.0? :stuck_out_tongue: )

PR #8847, opened 2021-09-07
Create tag label from tag if label is empty

### LINKED ISSUE #8829: 3.0 - Email tags without labels have empty markup added to the inserter
### Expected behavior
When I click the "Insert Marker" button to add an email tag to the editor, I expect the email tags to look consistent and not have empty markup.

### Actual behavior
Some extensions register email tags without labels, because the label parameter is new to EDD 3.0. For those extensions, the tag list looks like this:

image

The markup is:

```html
<button class="edd-email-tags-list-button" data-to_insert="{edd_invoice}">
	<strong></strong><code>{edd_invoice}</code>
	<span>Creates a link to a printable invoice</span>
</button>
```

It's a minor thing, but it would be preferable for the missing label to be bypassed, instead of added ([ref](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/release/3.0/includes/emails/tags-inserter.php#L131)).

1. Activate an extension which registers an email tag. Software Licensing, Reviews, PDF Invoices, Invoices (release/1.3) all do.
2. Visit `/wp-admin/edit.php?post_type=download&page=edd-settings&tab=emails&section=purchase_receipts`.
3. Click "Insert Marker" and check out the email tag registered by the extension.

PHP Version: 7.4.1
WordPress Version: 5.8

robincornett:
Fixes #8829

Proposed Changes:
1. If the email `$tag['label']` is empty, then the label will be created from the `$tag['tag']` instead.

### DIFF
--- a/includes/emails/class-edd-email-tags.php
+++ b/includes/emails/class-edd-email-tags.php
@@ -57,7 +57,7 @@ class EDD_Email_Template_Tags {
 		if ( is_callable( $func ) ) {
 			$this->tags[ $tag ] = array(
 				'tag'         => $tag,
-				'label'       => $label,
+				'label'       => ! empty( $label ) ? $label : ucwords( str_replace( '_', ' ', $tag ) ),
 				'description' => $description,
 				'func'        => $func,
 			);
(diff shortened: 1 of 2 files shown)


ashleyfae (changes requested):
Can we move this logic to the `EDD_Email_Template_Tags` class instead? I'm thinking inside the `add()` method. Just check if `$label` is `null`, and if so, then add this label fallback logic.

That way we have a label as soon as the tag is saved and it's available always, instead of just when we render.

ashleyfae (approved):

PR #8850, opened 2021-09-07
Don't require the purchase button to have a specific ID attribute #8834

### LINKED ISSUE #8834: Custom checkout button markup can cause a JS error when clicked.
### Expected behavior
When I customize my checkout button, I expect it to work still when clicked.

### Actual behavior
Due to these changes in the `edd-ajax.js` file https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/assets/js/edd-ajax.js#L402-L405, a customized checkout button will not have the attribute, and therefore possibly fail. 

1) Customize your checkout button markup.
2) Try and checkout.

PHP Version: All

EDD Version (or branch): 2.11

WordPress Version: All

cklosowski:
Fixes #8834 

Proposed Changes:
1. Removes the assignment of the purchase button to a variable based on the ID
2. Uses the instance of `$(this)` to address the purchase button


cklosowski:
Steps to replicate are here: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8834#issuecomment-914360440

robincornett (approved):
This is working, but noting that PayPal Commerce Pro removes all of the checkout button filters with:

```php
remove_all_filters( 'edd_checkout_button_purchase' );
```

so the sample code in the issue does not work for PayPal Commerce Pro.

ashleyfae (approved):

PR #8852, opened 2021-09-07
Issue/8851

### LINKED ISSUE #8851: Improved `eddfile` token validation, to be more compatible with caching.
Right now the file download token validation relies heavily on the query string parameters being in a specific order. This was fine back when it was built in #603, but since then, many caching solutions like Varnish and Cloudflare have taken to re-ordering the query string parameters before sending them to PHP, which causes our validation to fail.

We've setup a doc on this for store owners: https://docs.easydigitaldownloads.com/article/1351-invalid-download-links-due-to-query-string-order-modification

### Justification or use case
While we do have the doc available, we can likely add support for this and make the validation less dependent on the order of the parameters in order to provide a more stable experience, and avoid the support tickets from ever being created.

cklosowski:
Fixes #8851 

Proposed Changes:
1. Enforces allowed list of query string values instead of removing the not-allowed list.
2. Some small PHPCS fixes.

### DIFF
--- a/includes/download-functions.php
+++ b/includes/download-functions.php
@@ -1254,8 +1254,8 @@ function edd_validate_url_token( $url = '' ) {
 			'edd_url_token_allowed_params',
 			array(
 				'eddfile',
-				'file',
 				'ttl',
+				'file',
 				'token',
 			)
 		);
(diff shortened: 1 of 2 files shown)


robincornett (commented):
This is working to reformat/correct the URL. As I'm reading the function, though, the `ttl` condition confused me because I think it's interrupting the flow of reformatting and then using the `$url` variable. Would it be possible to consider moving that up to be the first check after the `$query_args` have been parsed? That seems like it's all that's needed to perform that check.

cklosowski:
Yeah we can probably move that up. I didn’t change the order of the operations here. We’ve always done the URL construction, TTL check, and then URL validation in that order, perilously it was just a restrictive check instead of an inclusion check. I’ll move the TTL check up and see if that messes with anything. I want to get some unit tests in here too btw.

robincornett (approved):

cklosowski:
@ashleyfae being that this is a critical function of the offering of EDD, I'd love your input on the approach as well before merging.

ashleyfae (approved):
- [x] Confirmed reordered parameters in `master` doesn't let me download a file.
- [x] Confirmed same reordered parameters works in this branch.
- [x] Confirmed invalid token doesn't let me download.

PR #8858, opened 2021-09-13
Update SendWP account link

### LINKED ISSUE #8857: Update SendWP link
Just heard from SendWP that the URL we are using in our "Access your SendWP account" link in our settings is no longer the preferred one. We should use https://app.sendwp.com/dashboard instead of linking to the marketing site going forward.

Nothing is broken; that is just the better destination for users.

robincornett:
Fixes #8857

Proposed changes:
1. Update the SendWP account link.
2. Reorganize the `sprintf` so that the string for translation does not include the link.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -2173,7 +2173,10 @@ function edd_sendwp_callback($args) {
 
 	// Connection status partial label based on the state of the SendWP email sending setting (Tools -> SendWP)
 	$connected    = sprintf(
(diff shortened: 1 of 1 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -2173,7 +2173,10 @@ function edd_sendwp_callback($args) {
 
 	// Connection status partial label based on the state of the SendWP email sending setting (Tools -> SendWP)
 	$connected    = sprintf(
```
ashleyfae, line 2175: Seems like we could just ditch the sprintf entirely? And just concatenate the different strings.

ashleyfae (approved):

PR #8860, opened 2021-09-14
Fix price_id parameter for CLI payments

### LINKED ISSUE #8859: WP-CLI: price_id parameter does not work correctly
### Expected behavior
When I generate a payment using CLI, I expect valid parameters to be properly parsed.

### Actual behavior
I'm trying to create a payment for a specific download/price ID combination and the price ID is not being honored, just randomly assigned.

1. Create a product with multiple prices.
2. Run this CLI command: `wp edd payments create --number=1 --id=35 --price_id=1` where `id` matches your download ID and `price_id` is a valid price ID.
3. Run that command more than once just to verify, but it will not always use price ID 1 for the new order.

PHP Version: 7.4.1
WordPress Version: 5.8

It looks like the issue is here:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/2e55064124cfdfe319cf409ac7290ab104449255/includes/class-edd-cli.php#L541

It's _checking_ for the `price_id` parameter in the array but actually _using_ the `id`.

Note that the issue is present in EDD 3.0 as well.

robincornett:
Fixes #8859

Proposed Changes:
1. Changes the `$price_id` variable to be properly assigned from `$assoc_args['price_id']`.

Submitting a PR now because I'm actively using this.

### DIFF
--- a/includes/class-edd-cli.php
+++ b/includes/class-edd-cli.php
@@ -538,7 +538,7 @@ class EDD_CLI extends WP_CLI_Command {
 		if( count( $assoc_args ) > 0 ) {
 			$number     = ( array_key_exists( 'number', $assoc_args ) )   ? absint( $assoc_args['number'] )             : $number;
 			$id         = ( array_key_exists( 'id', $assoc_args ) )       ? absint( $assoc_args['id'] )                 : $id;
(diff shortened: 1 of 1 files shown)


ashleyfae (approved):
Feel free to merge when tests pass.

PR #8881, opened 2021-09-22
Sanitize all amounts to ensure rounding & formatting #8870

### LINKED ISSUE #8870: PayPal Commerce: make sure all values are rounded
### Expected behavior
When I send an order to PayPal Commerce, I expect it to be formatted correctly so it will process.

### Actual behavior
It's possible to send a non-rounded amount to PayPal, which returns this error:

> {"name":"UNPROCESSABLE_ENTITY","details":[{"field":"/purchase_units/@reference_id=='de368136c10a0c51a4091d853545a926'/amount/value","value":"6.7999999999999","issue":"DECIMAL_PRECISION","description":"If the currency supports decimals, only two decimal place precision is supported."}],"message":"The requested action could not be performed, semantically incorrect, or failed business validation.","debug_id":"1d4bfae0e4b18","links":[{"href":"https://developer.paypal.com/docs/api/orders/v2/#error-DECIMAL_PRECISION","rel":"information_link","method":"GET"}]}

Still needs to be replicated, but here's what the customer setup was:
1. Currency is `NOK` (Norwegian Krone).
2. Item price is `680 NOK`. (Possibly this is with taxes included as the cart shows a line item of `MVA: NOK 1.36` and the item price shows there as `NOK 544.00`)
3. 99% discount was applied.
4. Total shows in cart as `NOK 6.80`.

PHP Version:
EDD Version (or branch): 2.11.1
WordPress Version: 5.8

In working on #8833 we are using `edd_sanitize_amount` to make sure that item and fee values are rounded, but we should review anything sent to PayPal as a `value` and make sure that all are sanitized. I suggest this be handled after #8833 is finalized/merged.

ashleyfae:
Fixes #8870

Proposed Changes:
1. Run all amounts through `edd_sanitize_amount()`, which handles rounding and formatting.

### DIFF
--- a/includes/gateways/paypal/functions.php
+++ b/includes/gateways/paypal/functions.php
@@ -227,7 +227,7 @@ function get_order_purchase_units( $payment_id, $purchase_data, $payment_args ) 
 
 	$order_amount = array(
 		'currency_code' => $currency,
-		'value'         => (string) $purchase_data['price'],
+		'value'         => (string) edd_sanitize_amount( $purchase_data['price'] ),
 		'breakdown'     => array(
 			'item_total' => array(
 				'currency_code' => $currency,
--- a/includes/gateways/paypal/functions.php
+++ b/includes/gateways/paypal/functions.php
@@ -287,7 +287,7 @@ function get_order_items( $purchase_data ) {
 				'currency_code' => edd_get_currency(),
 				'value'         => (string) edd_sanitize_amount( $item_amount ),
 			),
-			'discount'    => (string) edd_sanitize_amount( $item['discount'] ),
+			'discount'    => $item['discount'], // This is unset later and never sent to PayPal.
 		);
 		if ( edd_use_skus() ) {
 			$sku = edd_get_download_sku( $item['id'] );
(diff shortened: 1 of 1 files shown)


includes/gateways/paypal/functions.php
```diff
@@ -287,7 +287,7 @@ function get_order_items( $purchase_data ) {
 				'currency_code' => edd_get_currency(),
 				'value'         => (string) edd_sanitize_amount( $item_amount ),
 			),
-			'discount'    => $item['discount'],
+			'discount'    => (string) edd_sanitize_amount( $item['discount'] ),
```
robincornett, line 290: I think we should revert this one--this is not passed through to PayPal; it's just added to the `$items` array for calculating and is `unset` in line 200.

ashleyfae: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8881/commits/c23324cef4d62b752215377a523136f9e6b2d586

robincornett (changes requested):
One more spot, I think--line 230, we need to run it on the `$purchase_data['price']`.

robincornett (approved):

PR #8886, opened 2021-09-23
Update discount metadata checks to empty instead of isset

### LINKED ISSUE #8885: 3.0 - during migration, empty discount metadata is being saved to the database
### Expected behavior
When I run a migration from 2.11 to 3.0, I would love to not have empty metadata saved to the database.

### Actual behavior
Discounts are all being migrated with empty `product_condition`, `product_requirement`, and/or `product_requirement` rows.

1. Start with a 2.x EDD dataset.
2. Make sure there are discounts which are pretty basic and do not have any product conditions, requirements, and exclusions.
3. Migrate to EDD 3.0.
4. Check the edd_adjustmentmeta database table and see the empty rows.

PHP Version: 7.4
WordPress Version: 5.8

The migration function calls `edd_store_discount` to migrate the discounts. That function, in turn, calls `edd_add_discount`. I think this is where the issue is:

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/32749e96000b542bdff8b2e7b8f3a25038d5fdaf/includes/discount-functions.php#L29-L31

Each of these variables is just checking if the array key `isset` and not whether it has any data in it. So apparently these are all set, but empty, so they're being added to the new database table. I expect that changing each to a `! empty()` check should resolve the issue (it's working locally for me so far).

robincornett:
Fixes #8885

Proposed Changes:
1. This changes the checks for the product condition, product requirements, and excluded products metadata to be an `empty` check instead of `isset`, to prevent empty metadata rows being saved to the new database tables during migration.

Additionally this will impact any use of `edd_store_discount` and `edd_add_discount`.

In the database I'm migrating, the existing discount posts had this data saved as empty post meta rows.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -52,7 +52,7 @@ function edd_add_discount( $data = array() ) {
 	if ( ! empty( $discount_id ) ) {
 
 		// Product requirements.
-		if ( ! is_null( $product_requirements ) ) {
+		if ( ! empty( $product_requirements ) ) {
 			if ( is_string( $product_requirements ) ) {
 				$product_requirements = maybe_unserialize( $product_requirements );
 			}
@@ -65,7 +65,7 @@ function edd_add_discount( $data = array() ) {
 		}
 
 		// Excluded products.
-		if ( ! is_null( $excluded_products ) ) {
+		if ( ! empty( $excluded_products ) ) {
 			if ( is_string( $excluded_products ) ) {
 				$excluded_products = maybe_unserialize( $excluded_products );
 			}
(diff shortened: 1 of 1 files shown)


ashleyfae (commented):
I'm wondering if it's better to change where the data is added.

So keep an `isset()` here, but when we actually add it, change the `! is_null()` check to be `! empty()`

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/32749e96000b542bdff8b2e7b8f3a25038d5fdaf/includes/discount-functions.php#L55

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/32749e96000b542bdff8b2e7b8f3a25038d5fdaf/includes/discount-functions.php#L68

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/32749e96000b542bdff8b2e7b8f3a25038d5fdaf/includes/discount-functions.php#L80

ashleyfae (approved):

PR #8890, opened 2021-09-27
Update admin CSS

### LINKED ISSUE #8111: 3.0 - Review admin CSS relating to forms, inputs
We've been updating markup and CSS in various admin components to be more consistent, manageable, and accessible ([current recommended markup](https://github.com/easydigitaldownloads/easy-digital-downloads/wiki)). Some significant corresponding CSS has been added already, but this should be reviewed for full coverage and specific use cases, especially with extensions.

### Justification or use case
As I have been working on some extensions, updating their markup, I've run into some styling issues/conflicts which are not present with 2.9.x, but are related to CSS added for EDD 3.0. Ideally, extensions will rarely need to override EDD styling (I'm working on some custom form groups in the Git Download Updater which would be an exception). My suggestion is that we would note conflicts, or internal (to EDD) CSS issues here and try to manage them together.

Specific issues I've encountered:
- working on an [input update for Software Licensing](https://github.com/easydigitaldownloads/EDD-Software-Licensing/pull/1662), I copied minimal EDD 3.0 CSS to the plugin. However, to be consistent with SL's other form fields, I had to add `<strong>` tags to the new `<label>` elements, as EDD 3.0 doesn't modify the label weights.
- for inputs such as radio/checkboxes, we are overriding the EDD input/label default styles, so that the label is inline with the input. I am wondering if the styles need to be tweaked so that this override is not necessary?
- Working on an issue for Git Download Updater, I needed to add a `max-width: 100%;` to `.edd-form-group__control` ([note](https://github.com/easydigitaldownloads/edd-git-download-updater/pull/88#discussion_r490905122)) to prevent a `<select>` overflowing the parent metabox; this may need to be added to EDD 3.0 directly (this affects selects with really long labels).

robincornett:
Fixes #8111

Proposed Changes:
1. pending


robincornett:
Download Files in the block editor with release/3.0:

image

robincornett:
@cklosowski 3b02a1c somewhat addresses the chart height issue flagged for #9147--just makes sure they're not full width on larger screens.

cklosowski:
@robincornett Overall I think this is much better. I think I would prefer if we do the following (if possible)

On a larger screen, maybe a viewport of 1800px+ show both graphs on a single line, but anything under 1800px, maybe try and go to a single graph per line?

cklosowski:
I think this new `$break-huge` is better. I would like to use something like `1680px`. While it isn't in the `most-common` list it is in the list of media queries that WP Core is using in places. So let's go with that.

```
/**
 * Breakpoints & Media Queries
 */

// Most used breakpoints
$break-huge: 1440px;
$break-wide: 1280px;
$break-xlarge: 1080px;
$break-large: 960px;	// admin sidebar auto folds
$break-medium: 782px;	// adminbar goes big
$break-small: 600px;
$break-mobile: 480px;
$break-zoomed-in: 280px;

// All media queries currently in WordPress:
//
// min-width: 2000px
// min-width: 1680px
// min-width: 1250px
// max-width: 1120px *
// max-width: 1000px
// min-width: 769px and max-width: 1000px
// max-width: 960px *
// max-width: 900px
// max-width: 850px
// min-width: 800px and max-width: 1499px
// max-width: 800px
// max-width: 799px
// max-width: 782px *
// max-width: 768px
// max-width: 640px *
// max-width: 600px *
// max-width: 520px
// max-width: 500px
// max-width: 480px *
// max-width: 400px *
// max-width: 380px
// max-width: 320px *
//
// Those marked * seem to be more commonly used than the others.
// Let's try and use as few of these as possible, and be mindful about adding new ones, so we don't make the situation worse
```
Since `1440px` (or `$break-huge`) is the maximum size that WP core is adopting as a break point that's probably good enough for now.

cklosowski (approved):

PR #8899, opened 2021-09-30
Add our submenu item to the Dashboard menu list #8898

### LINKED ISSUE #8898: Add reports menu entry in top level Dashboard menu.
Add a link to the EDD Reports view to the main Dashboard menu items.

### Justification or use case
To help us encourage store owners to see the performance of their store, we can expose them to the reports menu easier.

cklosowski:
Fixes #8898 

Proposed Changes:
1. Adds a 'Reports' menu item under the main Dashboard widget, which links to the EDD Reports view.

### DIFF
--- a/includes/admin/admin-pages.php
+++ b/includes/admin/admin-pages.php
@@ -28,7 +28,7 @@ if ( ! defined( 'ABSPATH' ) ) exit;
  * @return void
  */
 function edd_add_options_link() {
-	global $edd_discounts_page, $edd_payments_page, $edd_settings_page, $edd_reports_page, $edd_add_ons_page, $edd_settings_export, $edd_upgrades_screen, $edd_tools_page, $edd_customers_page;
+	global $submenu, $edd_discounts_page, $edd_payments_page, $edd_settings_page, $edd_reports_page, $edd_add_ons_page, $edd_settings_export, $edd_upgrades_screen, $edd_tools_page, $edd_customers_page;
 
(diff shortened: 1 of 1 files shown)


cklosowski:
I attempted to do this with `add_dashboard_page`, however that generates a URL that is for `index.php` which our reports are not registered for. Our reports are registered for `edit.php`.

cklosowski:
image

ashleyfae (commented):
I'm personally not a fan of this change. It feels like we're starting to put our stuff where it doesn't quite belong.

Cod works fine though.

At the very least, should we change the `Reports` label? Because we're infringing on space that isn't ours, it's not clear what these reports are for. People could have reports for all kinds of stuff: site traffic, affiliates, etc. Do we need an `EDD` prefix?

cklosowski:
@ashleyfae we could do something like `Store Reports` or `Shop Reports`?

`EDD Reports` is also an option, I think a more generic naming convention there is better (so long as it isn't too generic, as you've pointed out).

ashleyfae:
EDD 3.0 uses `Store` in various places, so maybe we should stick with that. I imagine the only conflict there would be if someone is using more than one shop plugin (like EDD + Woo). Which, I'm sure happens, but is minimal.

ashleyfae (approved):

robincornett (approved):
I agree that the label needs to be more specific than just "Reports"; adding "Store" works for me.

PR #8900, opened 2021-09-30
Switch from edd_die() to die() on API responses #8250

### LINKED ISSUE #8250: API - Extra output possible in API requests
### Expected behavior
When I query an API endpoint, I want just the data to be returned.

### Actual behavior
When I query an API endpoint and have a plugin such as Query Monitor active, there is extra output at the end of the data, which appears to be the QM stylesheet and a QM div.

1) Activate Query Monitor.
2) Query an EDD API endpoint (eg `http://example.com/edd-api/products/`)

PHP Version: 7.4.1
EDD Version (or branch): 2.9.x, release/3.0
WordPress Version: 5.5.3

We dealt with a similar issue with exporting CSV files in EDD3 (#7883), where `edd_die()` needed to be replaced with `die()`.

cklosowski:
Fixes #8250 

Proposed Changes:
1. Switch `edd_die` to `die` after outputting API Data.

This will avoid the `edd_die` from calling `wp_die` which is where the output from Query Monitor comes in.

### DIFF
--- a/includes/api/class-edd-api.php
+++ b/includes/api/class-edd-api.php
@@ -1917,7 +1917,7 @@ class EDD_API {
 
 		do_action( 'edd_api_output_after', $this->data, $this, $format );
 
-		edd_die();
+		die();
 	}
 
 	/**


ashleyfae (approved):
I'm actually unable to reproduce the original error anymore, but confirmed this doesn't cause any new problems.

robincornett (approved):
I am still having the original issue and this fixes it.

PR #8901, opened 2021-09-30
Only output a backtrace in deprecation methods if it isn't empty #7755

### LINKED ISSUE #7755: _edd_deprecated_function() - only trigger error with backtrace if there is a backtrace provided
### Expected behavior

When a function calls `_edd_deprecated_function()` I expect to have an error triggered, and if available, I expect to see a backtrace. However, the backtrace is passed as an argument with the default value `null`, so if this value was never provided (and is still `null`) then I expect to only see the triggered error message and no backtrace log.

### Actual behavior

If a function calls `_edd_deprecated_function()` and doesn't provide a backtrace, then I get the normal error:

```
[20-Apr-2020 11:24:28 UTC] PHP Notice:  edd_tools_banned_emails_display is <strong>deprecated</strong> since Easy Digital Downloads version 3.0 with no alternative available. in /var/www/edd3/wp-content/plugins/easy-digital-downloads/includes/misc-functions.php on line 644
```

That's fine.

Then I also get an empty notice, from the empty backtrace:

```
[20-Apr-2020 11:24:28 UTC] PHP Notice:   in /var/www/edd3/wp-content/plugins/easy-digital-downloads/includes/misc-functions.php on line 645
```

This is a useless log because it's logging nothing.

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/misc-functions.php#L553

We should just wrap this in an `if` statement that checks if `$backtrace` isn't empty.

1) Manually run a deprecated function that doesn't include a backtrace. If you check out `release/3.0` then one example is `edd_agree_to_terms_js()`.
2) Check your error logs.
3) You'll see two log entries: one with the standard error message, and another blank error.

PHP Version:

EDD Version (or branch): master and release/3.0

WordPress Version:

cklosowski:
Fixes #7755

Proposed Changes:
1. Wrap all backtrace outputs in an empty check.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -605,7 +605,6 @@ function _edd_deprected_argument( $argument, $function, $version, $replacement =
 
 			if ( ! empty( $backtrace ) ) {
 				trigger_error( print_r( $backtrace, 1 ) );// Limited to previous 1028 characters, but since we only need to move back 1 in stack that should be fine.
-
 			}
 			// Alternatively we could dump this to a file.
 		}
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -545,11 +545,19 @@ function _edd_deprecated_function( $function, $version, $replacement = null, $ba
 	if ( WP_DEBUG && apply_filters( 'edd_deprecated_function_trigger_error', $show_errors ) ) {
 		if ( ! is_null( $replacement ) ) {
 			trigger_error( sprintf( __( '%1$s is <strong>deprecated</strong> since Easy Digital Downloads version %2$s! Use %3$s instead.', 'easy-digital-downloads' ), $function, $version, $replacement ) );
-			trigger_error(  print_r( $backtrace, 1 ) ); // Limited to previous 1028 characters, but since we only need to move back 1 in stack that should be fine.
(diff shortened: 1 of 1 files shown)


includes/misc-functions.php
```diff
@@ -586,11 +594,19 @@ function _edd_deprected_argument( $argument, $function, $version, $replacement =
...
+			}
+
 			// Alternatively we could dump this to a file.
 		} else {
 			trigger_error( sprintf( __( 'The %1$s argument of %2$s is <strong>deprecated</strong> since Easy Digital Downloads version %3$s with no alternative available.', 'easy-digital-downloads' ), $argument, $function, $version ) );
-			trigger_error( print_r( $backtrace, 1 ) );// Limited to previous 1028 characters, but since we only need to move back 1 in stack that should be fine.
+
+			if ( ! empty( $backtrace ) ) {
+				trigger_error( print_r( $backtrace, 1 ) );// Limited to previous 1028 characters, but since we only need to move back 1 in stack that should be fine.
+
```
ashleyfae, line 608:
```suggestion
```

(Removes the empty line.)

robincornett (approved):

PR #8904, opened 2021-10-01
Replace the gateway selectors when a PayPal order is cancelled

### LINKED ISSUE #8883: Changing Payment Gateway on Checkout Page doesn't work after trying to use PayPal Gateway
### Expected behavior
I am filling in my data to proceed with my purchase by using the PayPal payment gateway on checkout page. When I am done and everything is properly filled out, I try to use the PayPal purchase button. It opens a new window that loads the PayPal website. But now I close this window for whatever reason - e.g. I just remember that I can't use PayPal for some reason. So I try now to choose another payment gateway.

When I try to change the payment gateway on checkout page, I expect it to work so I can see the new payment gateway form to proceed with my purchase.

### Actual behavior
But, when I try to change the payment gateway on checkout page, it doesn't display the new payment gateway form. Instead, it doesn't load anything. It is just blank. Also, if I try to go back to the first payment gateway, I get the same problem: no form is displayed.

That can be really frustrating for many customers and they might leave the website without making any purchase as the process doesn't work well.

1) Use your website without a logged in user.
2) Try to add some product to your cart.
3) Go to the checkout page.
4) Choose the PayPal payment gateway in case it is not your default.
5) Fill in all the required data correctly.
6) Try to use the PayPal purchase button.
7) Wait for the new window to open and load.
8) Close the PayPal window.
9) Try to change the payment gateway.
10) See your form fade away without any other form being displayed.

PHP Version: 8.0.9

EDD Version (or branch): master / 2.11.1

WordPress Version: 5.8

If I am correct, my debug shows that the nonce verification fails - edds_prb_load_gateway function located at easy-digital-downloads/includes/gateways/stripe/includes/payment-methods/payment-request/checkout.php.

It seems that, if the data collected on checkout page is correct and the new PayPal window is open and loaded, a new user is created for the customer - because currently you are not logged in. I am also using Easy Digital Downloads - Auto Register.

There are nonces on checkout page for each payment gateway, generated on page load - at this point you don't have a user.

Later on it creates a new user for you and now you are logged in, but the page is not refreshed - it means that we might have nonces that will not work correctly anymore. It seems that subsequent requests to change the payment gateway fails because of these nonces, that are not maching anymore (client/server).

This happens because nonces takes into account several pieces of data, including the current user - which has changed, right?

So I guess that is the problem right now.

Refreshing the checkout page, after closing the newly opened PayPal window... would that be the solution for this problem?

Thank you.

robincornett:
Fixes #8883

Proposed Changes:
1. When a PayPal transaction is cancelled, run an action via ajax to generate new payment select options with updated nonce values.

Marking as draft because this is currently incomplete--no validation happening in the new ajax function and I expect there should be, for one, but need to discuss if this is a practical approach (it is working to solve the original problem for me).


includes/gateways/paypal/checkout-actions.php
```diff
@@ -424,3 +424,33 @@ function capture_order() {
...
+		$output .= '<label for="edd-gateway-' . esc_attr( $gateway_id ) . '" class="edd-gateway-option' . $checked_class . '" id="edd-gateway-option-' . esc_attr( $gateway_id ) . '">';
+		$output .= '<input type="radio" name="payment-mode" class="edd-gateway" id="edd-gateway-' . esc_attr( $gateway_id ) . '" value="' . esc_attr( $gateway_id ) . '"' . $checked . $nonce . '>' . esc_html( $label );
+		$output .= '</label>';
+	}
+
+	wp_send_json_success(
+		array(
+			'select' => $output,
+		)
+	);
```
ashleyfae, line 453: I think I'd prefer to _just_ return / replace the nonces instead of updating the whole markup.

ashleyfae (commented):
Code is looking good. Still need to test though.

ashleyfae (commented):
So this is behaving weirdly for me. For some reason, it's working when I switch to the test (store) gateway, but not others (Stripe). However, it does work if I change the code that gets the nonce to vanilla JavaScript instead of jQuery.

Replace this: https://github.com/awesomemotive/easy-digital-downloads/blob/issue/8883/assets/js/edd-ajax.js#L518

with this:

```js
var nonce = document.getElementById( 'edd-gateway-' + payment_mode ).getAttribute( 'data-' + payment_mode + '-nonce' );
```

includes/gateways/paypal/checkout-actions.php
```diff
@@ -424,3 +424,25 @@ function capture_order() {
 
 add_action( 'wp_ajax_nopriv_edd_capture_paypal_order', __NAMESPACE__ . '\capture_order' );
 add_action( 'wp_ajax_edd_capture_paypal_order', __NAMESPACE__ . '\capture_order' );
+
+/**
+ * Gets a fresh set of gateway options when a PayPal order is cancelled.
+ *
```
ashleyfae, line 430: Can you link to https://github.com/awesomemotive/easy-digital-downloads/issues/8883 so we can remember why we did this? :grin:

PR #8905, opened 2021-10-01
Merge 2.11.2 into 3.0

ashleyfae:
Merges the 2.11.2 release into release/3.0.


robincornett:
The original file works with `SCRIPT_DEBUG` but if I disable that, then the minified file is a 404:

`https://3dd.local/wp-content/plugins/easy-digital-downloads.30/assets/js/paypal-checkout.min.js?ver=3.0-rc1`

It looks like in 3.0, `paypal-checkout.js` is the minified file?

ashleyfae:
Yeah we no longer have separate minified versions. Think the script file just needs an update.

robincornett (approved):

ashleyfae:
Updated some of the unit tests that were failing for me locally: https://github.com/easydigitaldownloads/easy-digital-downloads/pull/8905/commits/4ce8ecf6f4793f76efbc6f64faf22498e226770e

Now they all pass.

robincornett (approved):

PR #8906, opened 2021-10-01
Check if notices have been dismissed before displaying #8876

### LINKED ISSUE #8876: 3.0 - Unable to dismiss nginx file protection notice
## Bug Report

Screenshot from 2021-09-20 10-58-13

### Expected behavior

When I click the link to dismiss the nginx notice, I expect it to go away.

### Actual behavior

It does not go away.

I don't have this problem on my main local test site. This report is coming from a totally different 3.0 install from what I normally use. I have not yet dug into what's going on, but just making this issue so I remember to investigate.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8876

Proposed Changes:
1. Changes the capability check to `manage_shop_settings`, as that's what's in 2.x currently.
2. Check if notice is dismissed before displaying. (If it has been, then don't display.)
3. Customize the dismissal URL for nginx vs apache notices (as the dismissal keys are different for each).

### DIFF
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -247,7 +247,7 @@ class EDD_Notices {
 		}
 
 		// Bail if user cannot manage options
-		if ( ! current_user_can( 'manage_options' ) ) {
+		if ( ! current_user_can( 'manage_shop_settings' ) ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


robincornett (approved):
Noting that `edd_is_dev_environment()` was returning true for me and I had to disable that early return in order to test this, but the dismiss notice is working now.

We may want to update that particular function to check `wp_get_environment_type()`, but that would be a new issue.

ashleyfae:
Made a new issue for that: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/8913

PR #8916, opened 2021-10-06
Adjust system info output to make plugins list more readable

### LINKED ISSUE #8914: Enhancement: Improve formatting of plugins list for easier viewing
Since the addition of URI (https://github.com/awesomemotive/easy-digital-downloads/issues/6923) added to the system info's plugin list, it's made the list harder to review and find important information. It would be good if we could improve the formatting to make it more readable.

#### Plugins list in 2.11:
Markup on 2021-10-06 at 13:32:17

#### List prior to 2.11:
Markup%202021-10-06%20at%2013 20 07

A couple suggestions that might help:
1. Double space the plugins list
2. Remove the URL for plugins we own. That would make reviewing our plugins a lot easier and this is likely one of the main uses in support of the system info.
3. Move the URL to the line below the Plugin Name.
4. Remove `https://www.` (Only from the plugins list, this needs to exist in the Site Info.)

robincornett:
Fixes #8914

Proposed Changes:
1. In the system info output, if a plugin URI (or equivalent) is retrieved for a plugin, add it to the output as a new line.
2. Add an empty line between each plugin in the output.

Milestoning for 2.11.3 as I think this is a lot easier to read and it sounds like our support team would all find it helpful.

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -1386,9 +1386,9 @@ function edd_tools_sysinfo_get() {
 			$plugin_url = $plugin['Author'];
 		}
 		if ( $plugin_url ) {
-			$plugin_url = ' &mdash; ' . $plugin_url;
+			$plugin_url = "\n" . $plugin_url;
 		}
-		$return .= $plugin['Name'] . ': ' . $plugin['Version'] . $update . $plugin_url . "\n";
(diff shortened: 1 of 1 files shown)


shaynesanderson (commented):
Looks great!

PR #8920, opened 2021-10-07
Update edd_is_dev_environment to use wp_get_environment_type

### LINKED ISSUE #8913: 3.0 - Use wp_get_environment_type() in edd_is_dev_environment()

In 3.0 we have a new function `edd_is_dev_environment()`. We should add some checks for `wp_get_environment_type()`.

### Justification or use case

`wp_get_environment_type()` supports a user-defined environment, which could be more accurate than our guesswork.

robincornett:
Fixes #8913

Proposed Changes:
1. If `wp_get_environment_type` exists, use that to check if the current site is a development site. This function returns one of `'local'`, `'development'`, `'staging'`, or`'production'`. If the environment is `local` or `development`, this is considered to be true and returns early, instead of sorting through the URL options.
2. Updates the doc block return value, as this function actually returns either false or a string, matching whatever was determined to be the determining condition.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -80,20 +80,19 @@ function edd_is_debug_mode() {
  *
  * @since 3.0
  *
- * @return bool|string $retval False if not a development environment, otherwise string matching environment, port, or domain.
+ * @return bool $is_dev_environment True if development environment; otherwise false.
  */
 function edd_is_dev_environment() {
 
 	// wp_get_environment_type was added in WordPress 5.5.
 	if ( function_exists( 'wp_get_environment_type' ) ) {
 		$environment = wp_get_environment_type();
-		if ( in_array( $environment, array( 'local', 'development' ), true ) ) {
-			return apply_filters( 'edd_is_dev_environment', $environment );
-		}
+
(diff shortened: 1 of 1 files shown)


ashleyfae:
Not directly related to your PR since you didn't change this behaviour, but maybe relevant here..

Looking at the name of the function and the docblocks, I feel like returning a string is actually an error. :/ Kind of irks me. What are your thoughts on changing that?

robincornett:
Based on the function name and how the return value was originally described, I would agree with you. Seems like it would make more sense to return a boolean. Since this is new to 3.0 it seems reasonable to think we could still change it to always return a boolean.

ashleyfae:
The Stripe gateway does have a polyfill for this function, but based on its one usage, I don't think changing the function would cause any problems.

```php
// Current site is a development environment, Apple Pay won't be able to be used, do nothing.
if ( false !== edd_is_dev_environment() ) {
	return;
}
```

Let's go ahead and rework this function then (you can keep your current logic to use `wp_get_environment_type`, just change the return value).

ashleyfae (approved):

PR #8923, opened 2021-10-08
Ensure file URL creation and validation use the same arary of parameters

### LINKED ISSUE #8917: Download links unexpectedly producing "You do not have permission to download this file" error
Since updating to 2.11.2, at least some of our download links are producing a "You do not have permission to download this file" error.

I tracked this down to `edd_validate_url_token()` returning false when it should return true. I'm not sure why that's the case though, and don't know how to debug further ☹️ 

Here's an example URL:

```
https://www.wptasty.com/index.php?eddfile=167710%3A10%3A1%3A1&ttl=1634051586&beta=1&file=1&token=4b5cffaee74944c9903a20a7cab2d38737fac9bc2f04dbd4c0318dffafef1034
```

Notably, we're adding `beta=1` to the URL with the `edd_get_download_file_url_args` filter:

```php
/**
 * Filters the arguments used in a download link.
 *
 * @param array   $args       Existing query args.
 * @param integer $payment_id ID for the payment.
 * @param array   $params     Original parameters.
 * @return array
 */
public static function filter_edd_get_download_file_url_args( $args, $payment_id, $params ) {
	if ( ! empty( $params['download_id'] )
		&& WPT_EDD_TASTY_RECIPES_ID === (int) $params['download_id'] ) {
		$args['beta'] = 1;
	}
	return $args;
}
```

I also previously added `beta` as an allowed token:

```php
/**
 * Parameters allowed in signed URLs.
 *
 * @param array $args Existing parameters.
 * @return array
 */
public static function filter_edd_url_token_allowed_params( $args ) {
	$args[] = 'beta';
	return $args;
}
```

Maybe related to #8852? My hacky temporary fix is `add_filter( 'edd_validate_url_token', '__return_true' );`.

robincornett:
Fixes #8917

Based on a discussion with @cklosowski I'm putting together this draft PR to see how it does. The issue, I think, is that we have the download URL being built off of one array, and validated off of another array, and the order now matters. However, since two different filters/functions/filters are involved, the problem is that the two arrays can be disordered/out of sync with each other.

Proposed Changes:
1. Creates a new function, `edd_get_url_token_parameters`, which returns a one dimensional array of the URL parameters which are allowed for URL validation.
2. That array is replaced 1:1 in `edd_validate_url_token`.
3. The same array is now used as the base array in `edd_get_download_file_url` and populated with empty values, so that the initial array order is set to match the validation array, and then the rest of the function fills in array values as previously done.

What this does not fix: links which would have been sent out via email in which the parameters are in a different order. One thought I had was to add back the original code (prior to 2.11.2), to run that at the end of the validation if `$ret` ended up still being false. So ideally, everything would be in order, but if it's not, possibly it could still validate.

Tested locally and with the fix, the dynamically generated file URL is:
```
https://edd.local/index.php?eddfile=614%3A35%3A1%3A1&ttl=1633794013&file=1&token=262cfc100cea02545ff705a7f447898d3fd60ca79e142f52bd5d188aae68ad09&beta=1
```

The link generated by EDD 2.11.2 does not validate with this fix or in 2.11.2:
```
https://edd.local/index.php?eddfile=614%3A35%3A1%3A1&ttl=1633793963&beta=1&file=1&token=8a38bbf5aed7deff5cf7e0a93b1a5c38a2b1306f544383b799ffbffbda6dde6d
```

The filters I used, based on the information in the linked issue:
```php
add_filter( 'edd_get_download_file_url_args', 'filter_edd_get_download_file_url_args', 10, 3 );
function filter_edd_get_download_file_url_args( $args, $payment_id, $params ) {
	if ( ! empty( $params['download_id'] )
		&& 35 === (int) $params['download_id'] ) { // 35 is my download ID
		$args['beta'] = 1;
	}
	return $args;
}

add_filter( 'edd_url_token_allowed_params', 'filter_edd_url_token_allowed_params' );
function filter_edd_url_token_allowed_params( $args ) {
	$args[] = 'beta';

	return $args;
}
```

Marking as draft for review/discussion.

### DIFF
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -993,3 +993,34 @@ function maybe_add_jilt_notice_to_abandoned_payment( $payment_id ) {
 		<?php
 	}
 }
+
+/**
+ * Reverts to the original download URL validation.
+ *
+ * @since 2.11.3
+ * @todo  Remove this function in 3.0.
+ *
+ * @param bool   $ret
+ * @param string $url
+ * @param array  $query_args
+ * @param string $original_url
+ */
+add_filter( 'edd_validate_url_token', function( $ret, $url, $query_args, $original_url ) {
+	// If the URL is already validated, we don't need to validate it again.
+	if ( $ret ) {
+		return $ret;
+	}
+	$allowed = edd_get_url_token_parameters();
+	$remove  = array();
+	foreach ( $query_args as $key => $value ) {
+		if ( ! in_array( $key, $allowed, true ) ) {
+			$remove[] = $key;
+		}
+	}
+
+	if ( ! empty( $remove ) ) {
+		$original_url = remove_query_arg( $remove, $original_url );
+	}
+
+	return isset( $query_args['token'] ) && hash_equals( $query_args['token'], edd_get_download_token( $original_url ) );
+}, 10, 4 );
--- a/includes/deprecated-functions.php
+++ b/includes/deprecated-functions.php
@@ -997,7 +997,7 @@ function maybe_add_jilt_notice_to_abandoned_payment( $payment_id ) {
 /**
  * Reverts to the original download URL validation.
  *
- * @since 2.11.3
+ * @since 2.11.4
  * @todo  Remove this function in 3.0.
  *
  * @param bool   $ret
--- a/includes/download-functions.php
+++ b/includes/download-functions.php
@@ -1260,8 +1260,10 @@ function edd_get_download_token( $url = '' ) {
  */
 function edd_validate_url_token( $url = '' ) {
 
-	$ret   = false;
-	$parts = parse_url( $url );
+	$ret          = false;
+	$parts        = parse_url( $url );
+	$query_args   = array();
+	$original_url = $url;
 
 	if ( isset( $parts['query'] ) ) {
 
@@ -1285,9 +1287,6 @@ function edd_validate_url_token( $url = '' ) {
 				$validated_query_args[ $key ] = $query_args[ $key ];
 			}
 		}
-
-		// Save the original URL for the filter.
-		$original_url = $url;
 
 		// strtok allows a quick clearing of existing query string parameters, so we can re-add the allowed ones.
 		$url = add_query_arg( $validated_query_args, strtok( $url, '?' ) );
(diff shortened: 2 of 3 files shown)


robincornett:
Update: I added the original URL as a final parameter to the `edd_validate_url_token` filter. That way, a user who had custom parameters (and therefore existing links which might cause problems) could add the filter to their site and the original URLs would be validated (this is the original validation code, basically, which was replaced in 2.11.2):

```php
/**
 * Reverts to the original download URL validation.
 * 
 * @since 2.11.3
 * @param bool   $ret
 * @param string $url
 * @param array  $query_args
 * @param string $original_url
 */
add_filter( 'edd_validate_url_token', function( $ret, $url, $query_args, $original_url ) {
	// If the URL is already validated, we don't need to validate it again.
	if ( $ret ) {
		return $ret;
	}
	$allowed = edd_get_url_token_parameters();
	$remove  = array();
	foreach ( $query_args as $key => $value ) {
		if ( ! in_array( $key, $allowed, true ) ) {
			$remove[] = $key;
		}
	}

	if ( ! empty( $remove ) ) {
		$original_url = remove_query_arg( $remove, $original_url );
	}

	return isset( $query_args['token'] ) && hash_equals( $query_args['token'], edd_get_download_token( $original_url ) );
}, 10, 4 );
```

Optionally, we could add this to EDD as well, with the intention to remove it eventually, as I believe email links all have expiration/ttl values, so with the new parameter order in place for long enough, the issue would become moot, I think?

ashleyfae:
I'm fine adding it but let's add a `@todo` to remove it. Possibly even in 3.0.

cklosowski (approved):

includes/deprecated-functions.php
```diff
@@ -993,3 +993,34 @@ function maybe_add_jilt_notice_to_abandoned_payment( $payment_id ) {
 		<?php
 	}
 }
+
+/**
+ * Reverts to the original download URL validation.
+ *
+ * @since 2.11.3
```
ashleyfae, line 1000:
```suggestion
 * @since 2.11.4
```

includes/download-functions.php
```diff
@@ -1048,12 +1049,29 @@ function edd_get_download_file_url( $key, $email, $filekey, $download_id = 0, $p
...
-	$download_url = add_query_arg( $args, site_url( 'index.php' ) );
+	return add_query_arg( array_filter( $args ), site_url( 'index.php' ) );
+}
 
-	return $download_url;
+/**
+ * Gets the array of parameters to be used for the URL token generation and validation.
+ * Used by `edd_get_download_file_url` and `edd_validate_url_token` so that their parameters are ordered the same.
+ *
+ * @since 2.11.3
```
ashleyfae, line 1062:
```suggestion
 * @since 2.11.4
```

ashleyfae (approved):

PR #8927, opened 2021-10-14
Update DB schema to set default price_id to null #8926

### LINKED ISSUE #8926: 3.0 - edd_order_items "price_id" column defaults to "0", which is an actual price ID & thus confusing
## Bug Report

See https://github.com/awesomemotive/easy-digital-downloads/issues/8924

> Related: I'm concerned that our default database value for this column is 0, because 0 is (or, was, but still supported) is a legitimate price ID. This means we're mixing a legitimate price ID value with "no price ID", which I find very confusing.

We should change the default to be `null`, where `null` means "no price ID", then we can allow `0` to be set for actual price ID `0`, which can be a legitimate value.

This provides better clarity when purchasing a product that isn't variably priced.

PHP Version:

WordPress Version:

ashleyfae:
Fixes #8926

Proposed Changes:
1. Updates the `wp_edd_order_items` table to set the `price_id` default to `null`.
2. Update all areas that use `edd_add_order_item()` to ensure default price ID is `null`.
3. Add unit tests demonstrating that you can query by price ID = `0` and price ID = `null`
4. Updates the 3.0 migration for order items so that:
    - If a product does not currently have variable prices enabled, the price ID is _always_ set to `null`.
    - For variable prices, we check that the cart item's price ID currently exists, and if so, use that.
    - If it doesn't exist, we set it to the default price ID instead.

## Testing

- Purchase a product that does NOT have variable prices. Your order item should be added with `price_id` = `null`
- Purchase a variably priced product. Your order item should have the correct `price_id`.
- Set up a variable priced product that has a price ID `0`. Purchase it. Make sure it's added to the database with `price_id` = `0`

## Existing data

I'm not quite sure how to handle existing data. Keeping the values at `0` doesn't actually cause any harm but it isn't ideal..  Could write a CLI command to fix it..


ashleyfae:
Ugh! That's annoying.

Well saving it as `0` doesn't really cause any real world _problems_, it's just confusing. So although it does suck, the best option might be to keep the old data as `0` if that's what it is in 2.x. But we can handle it with nulls moving forward.

tests/orders/tests-orders.php
```diff
@@ -305,6 +305,52 @@ public function test_get_orders_with_product_price_id_should_return_1() {
...
+
+		// Now set one to `1`.
+		$items = edd_get_order_items( array( 'order_id' => self::$orders[0] ) );
+		foreach( $items as $item ) {
+			edd_update_order_item( $item->id, array(
+				'price_id' => 1
+			) );
+		}
+
+		// There should be 4 records at `null`, and 1 at `1`.
```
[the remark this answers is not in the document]
ashleyfae: https://github.com/awesomemotive/easy-digital-downloads/pull/8927/commits/c503296e39b275329173709d1588f91a16ae0fac

ashleyfae:
Okay @robincornett here's what I'm thinking:

- We get this working with `null` for all **new** orders that do not have a price ID.
- Legacy data just uses `0`.
- We don't worry about doing any kind of migration for existing 3.0 orders (since legacy data from 2.x will use `0` anyway).

This is annoying, and will lead to a little data inconsistency, but I think this is the better design in the long term. It shouldn't actually pose a problem to have a mixture of `0` and `null`, because everywhere we use price IDs we first check:

```
if ( has_variable_prices() ) {
    do_something_with_price_id();
}
```

So this PR doesn't actually impact functionality, it just clarifies the data a little for future orders. It makes it nicer to scan the order items table and be able to distinguish between a variable priced product and a normal one without having to look it up. I like that.

What do you think?

robincornett:
I think we should do it, because it gives an immutable reference for the price ID--relying on `has_variable_prices()` seems significantly less reliable, because a product can change from a single price to variable pricing, or vice versa, so using that on existing orders is less than ideal, I think.

robincornett:
Interestingly, running unit tests with this branch results in one failure which seems unexpected:

> 1) EDD\Discounts\Tests_Discounts::test_discount_status_update_inactive
> Failed asserting that 1 is true.

It looks like that test was updated in 2.11.3 and may be giving different results in 3.0?

cklosowski:
> The download itself has an empty array saved for `edd_variable_prices`. We could theoretically check that during migration, but if a product was variable at one time and is not anymore, or if a variation was sold and then deleted, that would be fairly problematic.

Quite honestly, if they've removed a price ID after selling it or went from a variable priced product to a single price product after it has had sales, then there are far more problematic issues they'll have in 2.x already. I would honestly say this is our opportunity to correct this going forward. Things like Software Licensing activation counts and recurring already fail if someone purchases a variable priced product and then the store owner deletes the price ID or alters the variable price status of a product. Since it already breaks down in 2.x then we might as well go ahead and just make it uniform in 3.0 is my opinion.

ashleyfae:
> Interestingly, running unit tests with this branch results in one failure which seems unexpected:
> 
> > 1. EDD\Discounts\Tests_Discounts::test_discount_status_update_inactive
> >    Failed asserting that 1 is true.
> 
> It looks like that test was updated in 2.11.3 and may be giving different results in 3.0?

I don't think that's specific to this issue; I think that's the whole `release/3.0` branch right now.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -1306,6 +1304,40 @@ public static function orders( $data = null ) {
...
+		}
+
+		// Get the price ID that's set to the cart item right now.
+		$price_id = isset( $cart_item['item_number']['options']['price_id'] ) && is_numeric( $cart_item['item_number']['options']['price_id'] )
+			? absint( $cart_item['item_number']['options']['price_id'] )
+			: null;
+
+		// Now let's confirm it's actually a valid price ID.
+		$variable_prices = edd_get_variable_prices( $cart_item['id'] );
+		if ( ! is_array( $variable_prices ) || empty( $variable_prices ) ) {
```
robincornett, line 1332: Tested on a migration of my own site data and it worked beautifully. Going to test EDD as well, but what if this early return was moved above the `$price_id` definition since that variable isn't actually used until the very end?

robincornett: Looks like this is working nicely on EDD data as well, even with retired products, which do have a `0` price ID involved. I can approve with or without moving the early return but will wait to see what you think.

ashleyfae: Good idea. :+1: https://github.com/awesomemotive/easy-digital-downloads/pull/8927/commits/3b190bbf4f17e0d83333236387bdaa088c83ceef

robincornett (changes requested):
This is working for migrations, new orders from the front end, and CLI generated orders. But I just discovered that manually added orders (through the admin) are getting a price ID set to `0` when variable pricing is not enabled. It looks like we need to update some JS lines (I changed these and got a good order item):
- https://github.com/awesomemotive/easy-digital-downloads/blob/issue/8926/assets/js/admin/orders/order-overview/models/order-item.js#L27
- https://github.com/awesomemotive/easy-digital-downloads/blob/issue/8926/assets/js/admin/orders/order-overview/views/form-add-order-item.js#L147
- https://github.com/awesomemotive/easy-digital-downloads/blob/issue/8926/assets/js/admin/orders/order-overview/views/form-add-order-item.js#L189

robincornett (approved):
Seeing all of the `null` values in the database is immensely satisfying.

PR #8935, opened 2021-10-19
Add filter to discount row title

davidsherlock:
Fixes #

Proposed Changes:
1.
2.
3.

### DIFF
--- a/includes/admin/discounts/class-discount-codes-table.php
+++ b/includes/admin/discounts/class-discount-codes-table.php
@@ -238,6 +238,9 @@ class EDD_Discount_Codes_Table extends List_Table {
 			'edd-action' => 'edit_discount',
 			'discount'   => $discount->id,
 		), $base ) . '">' . stripslashes( $discount->name ) . '</a>' . esc_html( $state ) . '</strong>';
+		
+		// Filter the discount row title
+		$discount_title = apply_filters( 'edd_discount_row_title', $discount_title, $discount, $base, $state );
 
 		// Return discount title & row actions
 		return $discount_title . $this->row_actions( $row_actions );
--- a/includes/admin/discounts/class-discount-codes-table.php
+++ b/includes/admin/discounts/class-discount-codes-table.php
@@ -239,11 +239,11 @@ class EDD_Discount_Codes_Table extends List_Table {
 			'discount'   => $discount->id,
 		), $base ) . '">' . stripslashes( $discount->name ) . '</a>' . esc_html( $state ) . '</strong>';
 		
-		// Filter the discount row title
-		$discount_title = apply_filters( 'edd_discount_row_title', $discount_title, $discount, $base, $state );
+		// Filter to additional content after the discount title
+		$additional_content = apply_filters( 'edd_discount_row_after_title', '', $discount, $base, $state );
 
 		// Return discount title & row actions
-		return $discount_title . $this->row_actions( $row_actions );
+		return $discount_title . $additional_content . $this->row_actions( $row_actions );
 	}
 
 	/**
--- a/includes/admin/discounts/class-discount-codes-table.php
+++ b/includes/admin/discounts/class-discount-codes-table.php
@@ -246,10 +246,10 @@ class EDD_Discount_Codes_Table extends List_Table {
 		 *
 		 * @param EDD_Discount $discount Discount object.
 		 * @param string $base The base URL for the discount list table.
-		 * @param string $state The discount status label.
+		 * @param string $status The discount status.
 		 * @return string Additional data shown in the Name column
 		 */
-		$additional_content = apply_filters( 'edd_discount_row_after_title', '', $discount, $base, $state );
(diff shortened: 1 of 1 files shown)


davidsherlock:
@robincornett where I added the filter essentially makes it possible to change the markup, since everything passed in from the `$discount_title.` I need to add a custom icon next to the title, because I have built an extension that allows automatic discounts (with a lot of other features). But I need to be able to show which ones are automatic or not is important, thus this suggestion.

davidsherlock:
@spencerfinnell I wanted to use that approach, but it strips slashes so I can't close the span I need added, hence this filter request.

davidsherlock:
For more content - I am building an extension for EDD that uses the discounts, but I want to show (visually) which ones are linked like this:

Screenshot 2022-03-13 at 22 35 26

This is because discounts can be linked to upsells, which I want to show like this :)

cklosowski:
I think I would prefer a filter that does not directly affect the title itself, as it would be possible to accidently remove core's functionality

I think I'd like to see an 'additional' content filter. Something like:
```
		// Return discount title & row actions
		$additional_content = apply_filters( 'edd_discount_row_after_title', '', $discount, $base, $state );
		return $discount_title . $additional_content . $this->row_actions( $row_actions );
```

This would allow adding content to the title I believe without the possibility of breaking the core markup. Thoughts @davidsherlock ?

davidsherlock:
Completely fine with that @cklosowski. I will edit my original PR so it conforms to this structure.

cklosowski (changes requested):
Just one more small change for some inline documentation for the filter.

includes/admin/discounts/class-discount-codes-table.php
```diff
@@ -238,9 +238,21 @@ public function column_name( $discount ) {
...
+		 * Filter to allow additional content to be appended to the discount title.
+		 *
+		 * @since 3.0
+		 *
+		 * @param EDD_Discount $discount Discount object.
+		 * @param string $base The base URL for the discount list table.
+		 * @param string $state The discount status label.
+		 * @return string Additional data shown in the Name column
+		 */
+		$additional_content = apply_filters( 'edd_discount_row_after_title', '', $discount, $base, $state );
```
robincornett, line 252: I know `$state` came from @cklosowski but I'm wondering if this should be the `$status` variable instead ... the `$state` is less easy to evaluate, I think, because it's been modified.  If that's solid, though, then I'm good--just wanting to check first.

cklosowski: @robincornett While `$state` could be modified, `$status` may not be the actual `$discount->status`, as you can see we have `$this->get_status()` which is a method of our list table and not the actual `status` of the `EDD_Discount` class, and therefore `$status` is purely used for the column displays.

`$state` is primarily defined as something more accurate than `$status`. Or at least `$discount->status`

robincornett: My concern is that `$state` could be extremely different since it's a "user facing" string, with the em dash and then pulling a discount label (which could be translated, capitalized, etc.).

davidsherlock: I switched it back to status, but I am fine with either

cklosowski: So thinking about this, if we are passing in `$discount` as the `EDD_Discount`, do we even need the `display` version of `$status` that we are getting from `$this->get_status()`? Seems like it would be confusing and people could end up using the wrong `status` indicator. We should revert to a single use of the values when possible and avoid confusion.

davidsherlock: That seems the best to me for my use case, yeah!

robincornett (approved):

includes/admin/discounts/class-discount-codes-table.php
robincornett: @cklosowski the PR currently includes the parameters listed above and I've approved it. If it looks all right to you as well, I can merge once you've approved. Thanks!

cklosowski: Since `$status` is not the status of the discount, but the requested status filter, the docblock needs to be updated as it currently reads:
`* @param string $status The discount status.`

Since it is _not_ the discount status, but instead the status requested/filtered by, we need to clarify that.

includes/admin/discounts/class-discount-codes-table.php
```diff
@@ -238,9 +238,21 @@ public function column_name( $discount ) {
...
 		), $base ) . '">' . stripslashes( $discount->name ) . '</a>' . esc_html( $state ) . '</strong>';
+		
+		/**
+		 * Filter to allow additional content to be appended to the discount title.
+		 *
+		 * @since 3.0
+		 *
+		 * @param EDD_Discount $discount Discount object.
+		 * @param string $base The base URL for the discount list table.
+		 * @param string $status The discount status.
```
robincornett, line 249:
```suggestion
		 * @param string $status The queried discount status.
```

PR #8940, opened 2021-10-21
Feature/review request

robincornett:
Proposed Changes:
1. Revises the dashboard widget to add a static content hook--anything added here will not be replaced by the ajax call.
2. Adds an `edd_activation` option which is a timestamp. Added during install, or the first time the review request attempts to display (for existing users).
3. Creates two new promo notices:
    - Dashboard Widget Review
    - Settings Screen Review
4. The content of the two promo notices are identical, but the hooks and conditions are slightly different.
5. The promo handler has been updated to allow for a new constant to be pulled from the notice class(es), so that promo notices can accept different capability levels, rather than just `manage_options`.
6. The new notices will show to store managers and admins:
   - 30 days after activation
   - AND when the store has at least 50 sales
7. The content/behavior of the notice is a Q&A sequence, allowing a user who is not ready to leave a review an opportunity to reach out for feedback.


robincornett:
@cklosowski do you have anything specific in mind for styling? The example I'm working off of doesn't have any. One [simple] thought I had was to incorporate Edd, at least in the dashboard widget--he would be a bit eye catching there. Or were you thinking to use buttons/links that look like buttons? Or ??

robincornett:
@ashleyfae @cklosowski this is now updated to remove the testing condition, add version numbers for 2.11.4, and include the UTM parameters for the plugin feedback link. Think we are ready for final reviews.

includes/admin/promos/notices/class-five-star-review-dashboard.php
```diff
@@ -0,0 +1,193 @@
...
+				return false;
+			}
+		}
+		// @todo Change this to edd_count_orders in 3.0
+		$payments = edd_count_payments();
+		if ( $payments && 15 > $payments->publish ) {
+			return false;
+		}
+
+		return true;
```
ashleyfae, line 170:
```suggestion
		return isset( $payments->publish ) && $payments->publish >= 15;
```

I went through several other suggestions before ending up on this one lol.

Originally I was going to change the `if` you have now to:

```
if ( isset( $payments->publish ) && 15 > $payments->publish ) {
```

But then I realized it probably makes more sense to return `false` if `$payments->publish` isn't set for some reason... Then ultimately I realized it feels a bit better to me visually for the `if` check to be looking for a successful return, which led to:

```
if ( isset( $payments->publish ) && $payments->publish >= 15 )
```

Then simplified it. :stuck_out_tongue:

includes/admin/promos/notices/class-five-star-review-dashboard.php
```diff
@@ -0,0 +1,190 @@
...
+	 * @return bool
+	 */
+	protected function _should_display() {
+
+		$activated = edd_get_activation_date();
+		if ( $activated ) {
+			if ( ( $activated + ( DAY_IN_SECONDS * 30 ) ) > time() ) {
+				return false;
+			}
+		}
```
ashleyfae, line 163:
```suggestion
		// Do not show if EDD was activated less than 30 days ago.
		if ( ! is_numeric( $activated ) || ( $activated + ( DAY_IN_SECONDS * 30 ) ) > time() ) {
			return false;
		}
```

Main reason for the change is I don't love the double ifs. I think the comment is helpful too.

includes/admin/dashboard-widgets.php
```diff
@@ -37,8 +37,9 @@ function edd_register_dashboard_widgets() {
  * @since 1.2.2
  * @return void
  */
-function edd_dashboard_sales_widget( ) {
-	echo '<p><img src=" ' . esc_attr( set_url_scheme( EDD_PLUGIN_URL . 'assets/images/loading.gif', 'relative' ) ) . '"/></p>';
+function edd_dashboard_sales_widget() {
+	do_action( 'edd_dashboard_sales_widget' );
```
ashleyfae, line 41: Can we add a docblock with a `@since` version?

cklosowski:
On larger screens (my window here is 1518px wide) in the plugin settings prompt, I feel like Edd is too far detached to make it look like 'he' is asking the question:
image

Where as on the widget, the proximity of the Edd image is close enough to the prompt that the 'I notice' makes sense.
image

If that is an easy problem to solve, it'd be great, but that is my only feedback here. I like this a lot.

cklosowski (approved):

ashleyfae (approved):

PR #8945, opened 2021-10-26
Refactor settings experience/delight

robincornett:
Proposed Changes (so far):
- Create a Marketing tab in EDD settings.
- Move Reviews section to Marketing.
- Move Recurring section to Payments.
- Move Free Downloads section to Marketing.
- Deprecate SendWP functions.
- Create an Extension Manager class to handle installing/activating extensions:
    - Adds two new ajax callbacks--one to activate; one to install and activate.
    - Registers a new/separate JS script for handling the manager.
    - Adds three files to `/libraries/` for helping with the installation/activation.
    - Includes a button creation method for other code to use.
- Add a new WP Mail SMPT "setting" if that plugin is not installed/active to automatically install and/or activate.


robincornett:
@ashleyfae if you look at this before I come in on Monday, just know that I know that it is a hot mess right now, but it is working for an extension. With a license.

I need to check with you about a fix I had to add to the Pass Manager and how I want to revise it--you will probably see exactly what I'm going for but I want to make sure it's okay by you.

ashleyfae:
What about refactoring it slightly like this:

1. Turn the `get_highest_license_key()` method into `set_highest_pass_data()` or something. Point being: it will now be responsible for setting both properties (highest license key & highest pass ID) and will no longer have a return value.
2. Remove the `get_highest_pass_id()` method.
3. The constructor will now just call `$this->set_highest_pass_data()`.
4. Change the properties  (`public $highest_pass_id` and the other) to give them a default value of `null`. That way things like this: https://github.com/awesomemotive/easy-digital-downloads/blob/735cdb78f9f14f57c82eecacafc084c9aaa736dd/includes/admin/class-pass-manager.php#L99 can just be turned into `return;`

I think that makes a bit more sense because with what that method is doing, both values are available at that point in time, so I think it makes sense for that method to just set both values itself.

Does that make sense?

robincornett:
Yes, thanks, that does make sense--I considered something like that but it seemed to contradict how the class was set up to start with. But if it's fine to do that then I shall.

ashleyfae:
FYI I think there are some unit tests that reference `get_highest_pass_id` in the docblock (something like `@covers Pass_Manager::get_highest_pass_id`). Can you update those to the new method name when you make this change?

includes/admin/class-pass-manager.php
```diff
@@ -83,28 +90,30 @@ public function __construct() {
...
+	private function set_highest_pass_data() {
 
 		if ( ! $this->has_pass_data || ! is_array( $this->pass_data ) ) {
-			return $highest_pass_id;
+			return;
 		}
 
-		foreach ( $this->pass_data as $pass_data ) {
+		$highest_license_key = false;
+		$highest_pass_id     = false;
```
ashleyfae, line 114:
```suggestion
		$highest_license_key = null;
		$highest_pass_id     = null;
```

I'd prefer null defaults for these.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,524 @@
...
+	 */
+	public function register_scripts() {
+		$minify = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+		wp_register_script( 'edd-extension-manager', EDD_PLUGIN_URL . "assets/js/extension-manager{$minify}.js", array( 'jquery' ), EDD_VERSION, true );
+		wp_localize_script(
+			'edd-extension-manager',
+			'EDDExtensionManager',
+			array(
+				'activating'              => __( 'Activating', 'easy-digital-downloads' ),
+				'installing'              => __( 'Installing', 'easy-digital-downloads' ),
```
robincornett, line 60: Note to self: probably need a string for updating the button text on failure.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,524 @@
...
+			'msg'          => $generic_error,
+			'is_activated' => false,
+			'basename'     => $plugin_basename,
+		);
+
+		// Check for permissions.
+		if ( ! current_user_can( 'activate_plugins' ) ) {
+			$result['msg'] = 'plugin' === $type ? esc_html__( 'Plugin installed.', 'easy-digital-downloads' ) : esc_html__( 'Extension installed.', 'easy-digital-downloads' );
+
+			wp_send_json_success( $result );
```
robincornett, line 396: Need to account for this result--if a user can install, but not activate a plugin, what needs to happen?

includes/admin/extensions/class-extensions-download-url.php
```diff
@@ -0,0 +1,107 @@
...
+			$this->get_api_url(),
+			array(
+				'timeout'   => 15,
+				'sslverify' => true,
+				'body'      => array(
+					'edd_action'  => 'get_version',
+					'license'     => $license,
+					'item_id'     => $item_id,
+					'php_version' => phpversion(),
+					'wp_version'  => get_bloginfo( 'version' ),
```
robincornett, line 48: Should add EDD version as well.

includes/admin/extensions/class-extensions-download-url.php
```diff
@@ -0,0 +1,107 @@
...
+				),
+			)
+		);
+
+		// If there was an API error, set transient for only 10 minutes.
```
robincornett, line 53: Need to implement this.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,209 @@
+<?php
+
+namespace EDD\Admin\Extensions;
+
+use \EDD\Admin\Pass_Manager;
+use \EDD\Admin\Extensions\ExtensionsAPI;
```
ashleyfae, line 6:
```suggestion
```
Not needed, as it never gets used. (`ExtensionsAPI` is referenced, but they're in the same namespace so it doesn't need to be imported.)

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,521 @@
...
+			$data['title'] = $product->info->title;
+		}
+		if ( ! empty( $config['image'] ) ) {
+			$data['image'] = $config['image'];
+		} elseif ( ! empty( $product->info->thumbnail ) ) {
+			$data['image'] = $product->info->thumbnail;
+		}
+		if ( ! empty( $config['desription'] ) ) {
+			$data['desription'] = $config['desription'];
+		} elseif ( ! empty( $$product->info->excerpt ) ) {
```
ashleyfae, line 141:
```suggestion
		} elseif ( ! empty( $product->info->excerpt ) ) {
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,521 @@
...
+		} elseif ( ! empty( $product->info->title ) ) {
+			$data['title'] = $product->info->title;
+		}
+		if ( ! empty( $config['image'] ) ) {
+			$data['image'] = $config['image'];
+		} elseif ( ! empty( $product->info->thumbnail ) ) {
+			$data['image'] = $product->info->thumbnail;
+		}
+		if ( ! empty( $config['desription'] ) ) {
+			$data['desription'] = $config['desription'];
```
ashleyfae, line 140:
```suggestion
		if ( ! empty( $config['description'] ) ) {
			$data['description'] = $config['description'];
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,521 @@
...
+	 * @since 2.11.x
+	 * @param object $product The extension data from the Products API.
+	 * @param array  $config  The configuration array for the specific extension.
+	 * @return array
+	 */
+	private function get_card_data( $product, $config ) {
+		$data = array(
+			'title'      => '',
+			'image'      => '',
+			'desription' => '',
```
ashleyfae, line 127:
```suggestion
			'description' => '',
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,209 @@
...
+	private function get_type( array $array ) {
+		if ( function_exists( 'array_key_first' ) ) {
+			return array_key_first( $array );
+		}
+		foreach ( $array as $key => $unused ) {
+			return $key;
+		}
+
+		return 'product';
+	}
```
ashleyfae, line 138: I might have a concern about where this is getting the type information, but first, can you help me understand what the other options might be? I understand `$array` is coming from `get_api_body()`. What other options other than `product` would we have?

My potential concerns is that this method is requiring that the type---whatever that is---always be the first item in the `get_api_body()` array and I'm seeing that as a potential pitfall. I'd rather not have specific ordering requirements.

includes/admin/settings/class-email-marketing.php
```diff
@@ -0,0 +1,106 @@
...
+
+	public function __construct() {
+		add_filter( 'edd_settings_sections_marketing', array( $this, 'add_section' ) );
+		add_action( 'edd_settings_tab_top_marketing_email_marketing', array( $this, 'field' ) );
+
+		parent::__construct();
+	}
+
+	/**
+	 * Gets the configuration for the newsletter extensions.
```
ashleyfae, line 27: Can you expand the docblock to:

1. Document the `$item_id` parameter.
2. Note that it could be returning the configuration for all extensions or for a specific one.

includes/admin/settings/class-email-marketing.php
```diff
@@ -0,0 +1,106 @@
...
+	 * Whether any email marketing extension is active.
+	 *
+	 * @since 2.11.x
+	 *
+	 * @return bool True if any email marketing extension is active.
+	 */
+	protected function is_activated() {
+		$config = $this->get_configuration();
+		foreach ( $config as $extension ) {
+			if ( is_plugin_active( $extension['basename'] ) ) {
```
ashleyfae, line 97: I'm putting this here but I want it to apply to all `is_activated()` methods that extend the `Extension` class.

I have some concerns about relying on `is_plugin_active()`. It will work in most cases, but not if someone has renamed the plugin directory. Software Licensing is a great example. (Not in use in this PR, but a good example.)

My SL directory is `EDD-Software-Licensing`, as that's the repo name. But for most of our customers, it will be `edd-software-licensing`. Point being: a check for `is_plugin_active( 'edd-software-licensing/edd-software-licenses.php' )` will _fail_ for me.

Perhaps we can improve this by using multiple checks. If you merge in `release/2.11.4` then I think the `ExtensionRegistry` could help with this. We could add a new method called `isExtensionActive( $extensionId )` (takes the EDD product ID). This could just return `$this->offsetExists( $extensionId )`.

Obviously this will only work for new extensions, but it's a good way forward, and would just be one of several.

Then the general flow for `is_activated()` could be this:

1. Check `is_plugin_active()`. If that returns `true`, return `true`. Otherwise continue to next checks.
1. Check extension registry `isExtensionActive()`.
2. We could potentially use the `$edd_licensed_products` global. That only adds products that have a license key entered, so it's not great, but could be a second step.
3. I noticed that the Recurring one does a `class_exists()` check. We could add something to the configs for a certain function to check.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,209 @@
...
+		$type    = $this->get_type( $body );
+		$id      = $body[ $type ];
+		$button  = array(
+			'type'    => $type,
+			'id'      => $id,
+			'product' => $item_id,
+		);
+		// If the extension is not installed, the button will prompt to install and activate it.
+		if ( ! $this->manager->is_plugin_installed( $config['basename'] ) ) {
+			$download_url = $this->manager->get_download_url( $item_id, 'extension' );
```
ashleyfae, line 160: We need to rework this. The new "Email Marketing" tab is an example. It's doing a separate API request for every single extension that's not already installed. This is going to 1) impact our site's performance; and 2) impact the performance of actually loading that page.

I think a better method is to actually fetch the download URL as part of the ajax callback (so only after the button is clicked). So when you click an "Install & Activate" button, the first part of that ajax callback can be to get the download URL, then show an error if it cannot be retrieved for some reason.

includes/admin/extensions/class-extensions-download-url.php
```diff
@@ -0,0 +1,107 @@
...
+		if ( ! $request || $request instanceof WP_Error ) {
```
ashleyfae, line 54:
```suggestion
		if ( ! $request || $request instanceof \WP_Error ) {
```

Because `WP_Error` isn't imported at the top of the file, it needs the backslash.

ashleyfae:
I don't have an idea yet, but something I'm thinking of...

It would be really nice if we didn't have to maintain a config for each extension. I'm just thinking: I believe a goal we have (not in this PR specifically, but in the not-so-distant future) is to allow downloading/activating of _any_ extension via the dashboard. So basically any extension listed on our site. It would be tedious to have to put together a config for every single extension.

I guess I'm wondering if that kind of data is actually something we could move to our products API endpoint? So that by just querying the API we have all the information we need and don't have to maintain any hard-coded configs.

Anyway just thinking out loud. I'll ponder this a bit more.

includes/admin/extensions/abstract-extension.php
robincornett: Oh, that's a great idea--and yes I was concerned about that. I can work on that.

includes/admin/extensions/abstract-extension.php
robincornett: So originally type was going to be `plugin` or `extension` and then as I started working with the products API, I felt like I could get more use out of it than just whether it was a plugin or something EDD specific--so the default type is `plugin` (outside of this class, I mean--but for managing the installation/activation process), and if it's anything else, the language at the end should change over to `extension`. But as I was working, I realized that I wanted to be able to set the product API result to an option, but I also want to delete it when the installation is done, because I don't want to leave an unnecessary option hanging in the database (that was one reason I liked a transient). So I'm using the first array item to determine the option name that I set when I query the products API, and then I delete that option on successful activation. So it could be `product_{$item_id}` or `tag_{$tag_id}` (that's what I have so far, as I was querying the entire email marketing tag, but only the recurring ID).

I don't think ordering should be an issue--the array is used this way twice (to extract the type and ID here and then to name the option when querying the API), but it's not filterable so the data should be the same for both places. But I'm open to suggestions too!

robincornett:
I wondered about adding more data to the products API, because I was thinking that would be a much better source of truth than a hard coded configuration array. That was part of the reason I was condensing the configuration down as much as I could, so that the product name, description/excerpt, image would all come from the API. But if we need to also get the plugin basename and a function/class from the API, that starts to feel a little messy also? Some of this could be mitigated with the changes we are implementing on our site--a different API. I'll need to read more about how that all will work.

ashleyfae:
I think the basename (or an array of possible basenames?) could work in the API. Plus it would be nice to be able to hotfix that on our server rather than having to deploy a new EDD update to push out changes to it.

But I agree that the function/class is weird there.

ashleyfae:
@robincornett When you have a chance can you merge in the release branch? Kind of want to see this stuff next to the new licensing stuff we added to see if we should move any directories around, etc.

robincornett:
Marking this as ready for review.

Additional Proposed Changes:

- Add a new Email Marketing section to "Marketing". This pulls the products in the Newsletter tag from our site and outputs them as extension cards (vertical) to get more information (if minimum pass level is not met), or to automatically download and install a selected email marketing extension. Right now, the initial array is hard coded, but it would be nice to flip this so that if an extension is added/removed, the cards would sync automatically. A blocker for this is that the plugin basename is needed, plus the settings tabs/sections.
- Email marketing extension settings have been moved, as possible, to the Marketing tab.
- This section will remain active, even if a listed extension is active, so that a user can review other options if they want.
- A Recurring section is added, if not installed or active. If the user uses the automatic installer, they're redirected to the Recurring settings section (now under Payments) and this section is no longer available.
- An Invoices section/installer has been implemented, consistent with Recurring.
- Recurring extension card content is all defined by the Products API. The Invoices content is customized, so that you can see the different options.
- The extension cards are styled to be vertical by default (as on the email marketing screen), but optionally can be horizontal (Recurring, Invoices) so they are not so narrow.
- I've [documented all possible user facing responses/messages](https://docs.google.com/spreadsheets/d/1mlkzxAzOHEuOyyThTHTIO34AMkPPxzDVUST94UgsB2g/edit?usp=sharing). Open to suggestions on rewording.
- Whether the plugin to activate is from the WP repo (WPSMTP) or from our site, the same installation/activation methods are used. The responses vary slightly ("extension" v. "plugin" language).
- If the pass level is not met, the "install" button is a link to the individual product page on our site (I expect this to change, and it will need UTM parameters added).

Screenshots:

Email Marketing: one failed to install, the other succeeded:
image

WPSMTP installed successfully:
image

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,248 @@
...
+	 * @return string
+	 */
+	private function get_upgrade_url( $config, $item_id ) {
+		if ( ! empty( $config['upgrade_url'] ) ) {
+			return $config['upgrade_url'];
+		}
+
+		return add_query_arg(
+			array(
+				'p' => $item_id,
```
robincornett, line 207: This needs UTM parameters.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,545 @@
...
+	 *
+	 * @param string $plugin_basename
+	 * @return void
+	 */
+	public function delete_options( $plugin_basename ) {
+		// Delete the option generated by the request to the Products API.
+		$type = filter_input( INPUT_POST, 'type', FILTER_SANITIZE_STRING );
+		$id   = filter_input( INPUT_POST, 'id', FILTER_SANITIZE_NUMBER_INT );
+		if ( $type && $id ) {
+			// delete_option( "edd_extension_{$type}_{$id}_data" );
```
robincornett, line 447: I do want to delete these options, but leaving them commented out for now.

includes/admin/settings/class-recurring.php
```diff
@@ -0,0 +1,107 @@
...
+		return $sections;
+	}
+
+	/**
+	 * Registers the setting/hook to display the extension card.
+	 *
+	 * @param array $settings
+	 * @return array
+	 */
+	public function setting( $settings ) {
```
robincornett, line 75: Not currently used because of hooking the output into `edd_settings_tab_top_gateways_recurring` but this could be reverted.

assets/js/extension-manager.js
```diff
@@ -0,0 +1,75 @@
+/* global EDDExtensionManager */
```
ashleyfae, line 1:
```suggestion
/* global EDDExtensionManager, ajaxurl */
```

assets/js/extension-manager.js
```diff
@@ -0,0 +1,75 @@
...
+						nextStep.fadeIn();
+					}
+				} else {
+					thisStep.fadeOut();
+					var message = res.data.message;
+					/**
+					 * The install class returns an array of error messages, and res.data.message will be undefined.
+					 * In that case, we'll use the standard failure messages.
+					 */
+					if ( !message ) {
```
ashleyfae, line 64:
```suggestion
					if ( ! message ) {
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,545 @@
...
+	 *
+	 * @since 2.11.x
+	 * @param object $item_id           The extension information retrieved from the Products API.
+	 * @param array  $config            The array of custom extension information.
+	 * @param array  $button_parameters The array of information needed to build the button.
+	 * @param array  $link_parameters   The array of information needed to build the link.
+	 * @return void
+	 */
+	public function do_extension_card( $product, $config, $button_parameters, $link_parameters ) {
+		if ( ! $product ) {
```
ashleyfae, line 84: Is it actually possible for this to be empty?

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,545 @@
...
+				'extension_manager_nonce'  => wp_create_nonce( 'edd_extensionmanager' ),
+			)
+		);
+	}
+
+	/**
+	 * Outputs a standard extension card.
+	 *
+	 * @since 2.11.x
+	 * @param object $item_id           The extension information retrieved from the Products API.
```
ashleyfae, line 77:
```suggestion
	 * @param object $product           The extension information retrieved from the Products API.
```

ashleyfae:
There's one thing I'm pondering organizationally: the `EmailMarketing` class. It breaks a pattern that's set up and I'm not sure how I feel about that.

There's the base `Extension` class, which feels very much like "I am one single extension." And the `Invoices` and `Recurring` etc. classes re-enforce that. But then `EmailMarketing` is suddenly a _collection_ of extensions, but still extending the same class. You've made it work, but it makes the whole `Extension` class feel a little inconsistent and I'm wondering if there's another way to approach it. I get why you wouldn't want to create a new class for every.single marketing extension -- that's tedious.

So I guess it leads me to two potential thoughts:

1. I like your email marketing config array. I wonder if we could refactor `EmailMarketing` so that it's just one single extension, but utilize that config array to dynamically create new instances of `EmailMarketing` for each extension in that array. (That would require being able to pass that config to the constructor to then set the config within the class somewhere).
2. That sort of ties into another question/thought though. I think our goal for sometime in 2022 is to be able to install _any_ extension via EDD? The marketing issue demonstrates how it can be tedious to create a new class for a large number of add-ons. What would this process look like down the line if we eventually wanted to add install buttons to _everything_ in Downloads > Extensions? Even if that's not part of this PR I think it's something to think about, as we'll likely use this infrastructure to make it work. When looking at this in a much larger scale, it would help immensely if we could just use arrays to dynamically create these `Extension` classes. So that would probably mean passing the config into the constructor instead of setting it within the class itself.

---

> One thing I'm not sure how to manage...if the Products API request fails, what should show?

We could go with some marketing copy about how we have extensions available and they can see them here: https://easydigitaldownloads.com/downloads/category/extensions/marketing/
(In other words, don't act like something broke, just fall back to _"Hey check out what we got via this link!"_

robincornett:
@ashleyfae I'm thinking about your comment here:

> I wonder if we could refactor EmailMarketing so that it's just one single extension, but utilize that config array to dynamically create new instances of EmailMarketing for each extension in that array.

I'm not _opposed_ to that but I'm not sure how it would simplify things. Honestly, I feel like setting up the configuration array for a group of products is more work than setting up the new class...setting up Invoices took maybe 15 minutes, and that was mostly me adding in custom configuration data.

Setting this up to replace the current Downloads > Extensions page to toy with it a bit, and it's pretty fast to implement, but building the configuration array is definitely going to be the tedious part. Also, for tags/categories, I don't _love_ that it's not dynamic. This is the point at which I'm thinking it would be nice to add the data I need to the API response and turn this around a bit to be totally dynamic.

PR #8957, opened 2021-11-01
new: Added a filter inside AJAX remove discount function.

rohitp-bsf:
Fixes #
Proposed Changes:
1. Added a filter while removing discount code via ajax call just like the way you did in adding discount code.

### DIFF
--- a/includes/ajax-functions.php
+++ b/includes/ajax-functions.php
@@ -394,6 +394,9 @@ function edd_ajax_remove_discount() {
 			'html'        => edd_get_cart_discounts_html()
 		);
 
+		// Allow for custom remove discount code handling
+		$return = apply_filters( 'edd_ajax_remove_discount_response', $return );
+
 		wp_send_json( $return );
 	}
 	edd_die();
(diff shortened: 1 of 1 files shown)


ashleyfae:
Hi @rohitp-bsf , just curious what you're using this for? :)

rohitp-bsf:
> Hi @rohitp-bsf , just curious what you're using this for? :)

@ashleyfae For internal purposes, we have modified the checkout page and have added a bump order functionality. Which adds and removes download with discount code. So for the calculations and to be consistent need the above filter which is the same like you have added in apply discount functionality.

includes/ajax-functions.php
```diff
@@ -394,6 +394,9 @@ function edd_ajax_remove_discount() {
 			'html'        => edd_get_cart_discounts_html()
 		);
 
+		// Allow for custom remove discount code handling
```
ashleyfae, line 397:
```suggestion
		/**
		 * Allow for custom remove discount code handling.
		 * 
		 * @since 2.11.4
		 */
```

ashleyfae (approved):

PR #8964, opened 2021-11-01
Create edd_get_activation_date()

### LINKED ISSUE #8963: Create an activation date option/function
Let's create an option so we know when EDD was activated, please.

### Justification or use case
It would be helpful to know when an install of EDD was activated (give or take, since we can't see that for existing installs).

robincornett:
Fixes #8963

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1052,6 +1052,23 @@ function edd_is_promo_active() {
 }
 
 /**
+ * Gets the date that this EDD install was activated (for new installs).
+ * For existing installs, this option is added whenever the function is first used.
+ *
+ * @since 2.11.4
+ * @return string The timestamp when EDD was activated.
+ */
+function edd_get_activation_date() {
+	$activation_date = get_option( 'edd_activation_date', '' );
+	if ( ! $activation_date ) {
+		update_option( 'edd_activation_date', time() );
+		$activation_date = get_option( 'edd_activation_date', '' );
+	}
+
+	return $activation_date;
+}
+
+/**
  * Polyfills for is_countable and is_iterable
  *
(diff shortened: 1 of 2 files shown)


includes/misc-functions.php
```diff
@@ -1051,6 +1051,23 @@ function edd_is_promo_active() {
...
+ * For existing installs, this option is added whenever the function is first used.
+ *
+ * @since 2.11.4
+ * @return string The timestamp when EDD was activated.
+ */
+function edd_get_activation_date() {
+	$activation_date = get_option( 'edd_activation_date', '' );
+	if ( ! $activation_date ) {
+		update_option( 'edd_activation_date', time() );
+		$activation_date = get_option( 'edd_activation_date', '' );
```
ashleyfae, line 1065:
```suggestion
		$activation_date = time();
		update_option( 'edd_activation_date', $activation_date );
```

What do you think about this instead of calling `get_option()` ?

robincornett: Yes, thank you, I like that better for sure.

includes/misc-functions.php
```diff
@@ -1051,6 +1051,23 @@ function edd_is_promo_active() {
 	return false;
 }
 
+/**
+ * Gets the date that this EDD install was activated (for new installs).
+ * For existing installs, this option is added whenever the function is first used.
+ *
+ * @since 2.11.4
+ * @return string The timestamp when EDD was activated.
```
ashleyfae, line 1059:
```suggestion
 * @return int The timestamp when EDD was activated.
```

`time()` will be an integer.

ashleyfae (commented):
I think we still need to set this initially on install?

ashleyfae (approved):

PR #8972, opened 2021-11-02
Add extension registry. #8970

### LINKED ISSUE #8970: Add license registry for extensions to initialize EDD_License

This issue is a long term improvement for the https://github.com/awesomemotive/easy-digital-downloads/issues/8969 issue.

I'd like to add two things:

1. A new hook for extensions to use to initialize their `EDD_License` class.
2. A new registry for those class instances to live. The registry would be a long term replacement for the `$edd_licensed_products` global variable.

### Justification or use case

1. The hook allows us to have extensions _consistently_ initialize their licensing. It gives us more control over when it happens and in what context. (This helps fix the problem we have now where some extensions are only doing it if `is_admin()`.)
2. The registry gives us a better way in EDD core to see all the premium EDD extensions that are running. Right now we have to use a pretty hacky global variable, but a registry would give us an easily accessible array of all `EDD_License` instances, which then allows us to better determine if any premium extensions are activated, if so, which ones. This will be a much better solution for the `Pass_Manager` class, which tries to determine if the current site is "free" (no licenses activated) or has à la carte license(s) activated.

ashleyfae:
Fixes #8970

Proposed Changes:
1. Adds a new `ExtensionRegistry` class, which is initialized via the main EDD class. This has three methods:
    1. `addExtension()` - This is what extensions would use to register themselves. The arguments are the only real things we require right now from `EDD_License`. (I omitted the ones that would be the same across all extensions, like author name and API URL.) Right now, all this method does is create a new `EDD_License` object and saves it in the registry.
    2. `getExtensions()` - This returns all registered extensions. For now it's private so we can have flexibility on what the return value is. More on this later.
    3. `countLicensedExtensions()` - This counts the number of licensed extensions. This is what we could ultimately use in the `Pass_Manager` class for `$number_license_keys`.
2. Adds a new action hook `edd_extension_license_init`. This fires after plugins have been loaded, because the extensions need to add their `add_action()` before we run the `do_action()`.

Here's an example of how extensions would use this new method: https://github.com/awesomemotive/EDD-Software-Licensing/pull/1924/files

## Testing

If this PR is installed alone, it doesn't really do much. But you can run this via debug bar:

```php
echo EDD()->extensionRegistry->countLicensedExtensions();
```

It should return `0`.

Then you can checkout the above linked SL PR, make sure you have a license key entered, run the code again, and it should return `1`.

@robincornett If you remember, I had actually talked to you about making this a lot more involved and refactoring a lot of the `EDD_License` class. I did start doing that.... and changed my mind _for now_. Reason being: I found that I was unintentionally duplicating a lot of the work I've done in https://github.com/awesomemotive/edd-sl-sdk/ . The things I wanted were things I'd already built in there. That just felt really silly and weird so I decided it's best to keep this PR very minimal for now, and maybe in the longer (but not too long!) term re-visit that project. That way maybe it's something we could bundle with EDD and other people could use too.


easy-digital-downloads.php
```diff
@@ -152,17 +160,18 @@ public static function instance() {
...
+			self::$instance->api             = new EDD_API();
+			self::$instance->session         = new EDD_Session();
+			self::$instance->html            = new EDD_HTML_Elements();
+			self::$instance->emails          = new EDD_Emails();
+			self::$instance->email_tags      = new EDD_Email_Template_Tags();
+			self::$instance->customers       = new EDD_DB_Customers();
+			self::$instance->customer_meta   = new EDD_DB_Customer_Meta();
+			self::$instance->payment_stats   = new EDD_Payment_Stats();
+			self::$instance->cart            = new EDD_Cart();
+			self::$instance->extensionRegistry = new \EDD\Licensing\ExtensionRegistry();
```
robincornett, line 174: Since the lines above had spaces added to accommodate the original variable name, can we go ahead and add a couple more due to the new variable name?

robincornett (commented):
Initial testing with the SL branch has this working nicely.

One thought--I was thinking about this last night with the work I've been doing to manage extensions/upgrades and was thinking to move some of my new files into an `extensions` directory. Should that be filed in the same place as this new class, and if so, should the directory be `extensions` or `licensing`?

ashleyfae:
I'll rename this `licensing` one to `extensions`. Ultimately the licensing is for extensions, so I think the `extensions` naming makes good sense.

robincornett (approved):
Working well for me!

I was pondering the `plugins_loaded` hook and wondering if it was definitely the correct one. I think it is, but wondering if we should set up a wiki/KB to codify correct hooks/practices for extensions to load/manage themselves (how to use this, for example, and what hook to initialize on, things like that).

ashleyfae:
Right now extensions initialize on `plugins_loaded`. So that means:

- Extension loads its code (require files, set up hooks, etc.) via `plugins_loaded`, with priority of `98` (that's what `ExtensionLoader` uses).
- Extension immediately uses `edd_extension_license_init` hook to initialize updater.
- On `plugins_loaded` with a really really high priority (like, highest you can go), EDD finally triggers the `do_action( 'edd_extension_license_init' )`.

Would be nice to have a more official like `edd_loaded` hook or something that extensions would load into.

robincornett:
> Would be nice to have a more official like edd_loaded hook or something that extensions would load into.

+1000

I mainly am not sure off the top of my head if every extension is using `plugins_loaded` or possibly something later, but that's something to check as we work to implement this.

PR #8976, opened 2021-11-04
Update edd_get_activation_date to optionally set date to first order

### LINKED ISSUE #8975: For existing installs, set activation date to first payment date
It would be shiny to set the activation date to the first payment date if we are adding the `edd_activation_date` to an existing install (not during the install process).

### Justification or use case
It wouldn't be perfect, but if we ever wanted to use that information for something else, it would be nice to know that it is more accurate than if we didn't add in the check. eg if I set my store up in 2014, but the option was just added in 2021, the timestamp from the first payment would be a lot more accurate than when 2.11.4 was released.

robincornett:
Fixes #8975

Proposed Changes:
1. Adds an optional boolean parameter `$use_first_order` to the `edd_get_activation_date` function. Set to false by default.
2. If called outside of the initial installation, set the parameter to `true` to run a quick payments query to get the first order from the database.
3. If an order is retrieved, use its date as the activation date (this is not necessarily a complete order so I'm just using the date, not completed date).
4. So when we run the installer, we would not pass a parameter, but using it elsewhere, which might happen for an existing install, we would.

On my own storefront, this changes the activation date from today (`1636029820`) to `1434612134`, which is June 2015.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1067,7 +1067,7 @@ function edd_get_activation_date( $use_first_order = false ) {
 			// Gets the first order placed in the store (any status).
 			$payments = edd_get_payments(
 				array(
-					'output'  => 'payments',
+					'output'  => 'posts',
 					'number'  => 1,
 					'orderby' => 'ID',
 					'order'   => 'ASC',
@@ -1075,9 +1075,9 @@ function edd_get_activation_date( $use_first_order = false ) {
 			);
 			if ( $payments ) {
 				$first_payment = reset( $payments );
-				// Use just the payment date, rather than completed date (first payment may not be complete).
-				if ( ! empty( $first_payment->date ) ) {
-					$activation_date = strtotime( $first_payment->date );
+				// Use just the post date, rather than looking for the completed date (first payment may not be complete).
+				if ( ! empty( $first_payment->post_date_gmt ) ) {
+					$activation_date = strtotime( $first_payment->post_date_gmt );
 				}
 			}
 		}
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1067,10 +1067,11 @@ function edd_get_activation_date( $use_first_order = false ) {
 			// Gets the first order placed in the store (any status).
 			$payments = edd_get_payments(
 				array(
-					'output'  => 'posts',
-					'number'  => 1,
-					'orderby' => 'ID',
-					'order'   => 'ASC',
+					'output'        => 'posts',
+					'number'        => 1,
+					'orderby'       => 'ID',
+					'order'         => 'ASC',
+					'no_found_rows' => true,
 				)
 			);
 			if ( $payments ) {
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1056,30 +1056,27 @@ function edd_is_promo_active() {
  * For existing installs, this option is added whenever the function is first used.
  *
  * @since 2.11.4
- * @param bool $use_first_order Optional flag to use the first existing order date.
- * @return int The timestamp when EDD was activated.
+ * @return int The timestamp when EDD was marked as activated.
  */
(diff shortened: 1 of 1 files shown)


includes/misc-functions.php
```diff
@@ -1056,12 +1056,31 @@ function edd_is_promo_active() {
...
 	$activation_date = get_option( 'edd_activation_date', '' );
 	if ( ! $activation_date ) {
 		$activation_date = time();
+		if ( $use_first_order ) {
+			// Gets the first order placed in the store (any status).
+			$payments = edd_get_payments(
+				array(
+					'output'  => 'payments',
+					'number'  => 1,
+					'orderby' => 'ID',
```
ashleyfae, line 1072: Nice one. I think I told you to order by date but actually ordering by ID is better here, so well done. :grin:

includes/misc-functions.php
```diff
@@ -1056,12 +1056,31 @@ function edd_is_promo_active() {
...
-function edd_get_activation_date() {
+function edd_get_activation_date( $use_first_order = false ) {
 	$activation_date = get_option( 'edd_activation_date', '' );
 	if ( ! $activation_date ) {
 		$activation_date = time();
+		if ( $use_first_order ) {
+			// Gets the first order placed in the store (any status).
+			$payments = edd_get_payments(
+				array(
+					'output'  => 'payments',
```
ashleyfae, line 1070: Can we actually make this posts? We don't actually need the payment object for what we're doing.

If you change this then we'll also need to adjust the date thing to use `post_date_gmt` (use the GMT one, please).

includes/misc-functions.php
```diff
@@ -1056,12 +1056,31 @@ function edd_is_promo_active() {
...
 	if ( ! $activation_date ) {
 		$activation_date = time();
+		if ( $use_first_order ) {
+			// Gets the first order placed in the store (any status).
+			$payments = edd_get_payments(
+				array(
+					'output'  => 'posts',
+					'number'  => 1,
+					'orderby' => 'ID',
+					'order'   => 'ASC',
```
ashleyfae, line 1073:
```suggestion
					'order'   => 'ASC',
					'no_found_rows' => true,
```

(You may want to fix array formatting.)

includes/misc-functions.php
```diff
@@ -1056,12 +1056,32 @@ function edd_is_promo_active() {
  * For existing installs, this option is added whenever the function is first used.
  *
  * @since 2.11.4
+ * @param bool $use_first_order Optional flag to use the first existing order date.
  * @return int The timestamp when EDD was activated.
  */
-function edd_get_activation_date() {
+function edd_get_activation_date( $use_first_order = false ) {
```
ashleyfae, line 1062: Are there any circumstances you can think of where we WOULDN'T want to use the first order?

ashleyfae:
- [x] Confirmed that my first payment gets logged as activation date.

includes/misc-functions.php
robincornett: Sure, during `edd_run_install`, I don't think we want to use the first order, since there won't be one. The query may not take much juice in that case, I guess, so maybe removing the parameter and check is okay....

Otherwise, if we are calling this function, then maybe we could invert this logic and set it to true here, but `false` during `edd_run_install`.

ashleyfae: Yeah this should be a pretty quick query, and should only ever run one time. So I don't think it will hurt if we run it unnecessarily on activation.

I'm fine with either:

1. Removing the parameter and always running this; or
2. Inverting the logic and defaulting to `true`.

Go with whichever one you prefer and I'll approve it. :grin:

robincornett: I'm good with removing the parameter--simplifies things and there won't be any questions in the future. And if the Queen of Queries is okay with it then I can't disagree. 👑

ashleyfae (approved):
:raised_hands:

PR #8977, opened 2021-11-04
Performance Improvement in batch export earnings report

chriscct7:
The counting of cols is done twice per iteration even though the number of colls is not being manipulated in the loop. Pre-computing the number ahead of the loop will same 2n performance of the operation (O(n) savings of performance).

Fixes #: (none)

Using PR to test a SAST. 

Proposed Changes:
1. Pre-compute count of $cols as $count ahead of loop
2. Replace count( $cols) with $count inside loop


ashleyfae (commented):
This is good, but can you rename the `$count` variable to like `$number_of_cols` or `$column_count` or something? Just to provide more clarity about what it's a count of.

robincornett (approved):

ashleyfae (approved):

PR #8978, opened 2021-11-04
Add unlisted property

chriscct7:
In PHP it is possible to write to properties without declaring them. For example, the following is perfectly valid PHP code:
```
class MyClass { }

$x = new MyClass();
$x->foo = true;
```

Generally, it is a good practice to explictly declare properties to avoid accidental typos and provide IDE auto-completion:
```
class MyClass {
    public $foo;
}

$x = new MyClass();
$x->foo = true;
```

Fixes # (none)

Proposed Changes:
1. Add unlisted payment_stats property

### DIFF
--- a/easy-digital-downloads.php
+++ b/easy-digital-downloads.php
@@ -130,6 +130,14 @@ final class Easy_Digital_Downloads {
 	public $cart;
 
 	/**
+	 * EDD Payment Stats Object
+	 *
+	 * @var object|EDD_Payment_Stats
+	 * @since 1.8
+	 */
+	public $payment_stats;
+
+	/**
 	 * Main Easy_Digital_Downloads Instance.
 	 *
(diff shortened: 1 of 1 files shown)


easy-digital-downloads.php
```diff
@@ -129,6 +129,14 @@ final class Easy_Digital_Downloads {
 	 */
 	public $cart;
 
+	/**
+	 * EDD Payment Stats Object
+	 *
+	 * @var object|EDD_Payment_Stats
```
ashleyfae, line 135:
```suggestion
	 * @var EDD_Payment_Stats
```

It's always going to be `EDD_Payment_Stats` so we don't need the generic `object` in there. (I know the others have it, but they don't need it either.)

ashleyfae (approved):

PR #8941, opened 2021-10-21
Inbox for notifications

ashleyfae:
Proposed Changes:
1. Creates a new database table for `wp_edd_notifications`.
2. Introduces a new `NotificationImporter` class, which handles importing notifications into the local table from an API. This gets executed once per day via cron.
3. The following conditional slugs are supported for determining if a pass should be displayed:
    - free - No license key.
    - ala-carte - Individual product license key, but no pass.
    - pass-personal - Personal pass activated.
    - pass-extended - Extended pass activated.
    - pass-professional - Professional pass activated.
    - pass-all-access - All access pass activated.
    - pass-any - Any type of pass activated.
    - Exact version number using full stops or slugs. Example: `2.11.3` or `2-11-3`
    - Version number with `x` as a wildcard. Example: `2.x` or `2-x`; `3.x` or `3-x`
4. Conditionals are checked at both ingest time and on display time. If any one condition is not met then the notification is not displayed.
3. Adds an icon next to the EDD header image. The number is how many notifications you have.
4. Clicking this icon makes the notification panel slide out.
5. Notifications can then be viewed and/or dismissed.

Screenshot from 2021-10-22 16-30-02

Screenshot from 2021-10-25 16-05-45

---

## Testing

You can check the EDD debug log for notification import errors/notices.

### Generated notification data

If desired you can generate some random notifications via:

1. Clone https://github.com/ashleyfae/edd-generator and inside that directory run `composer install --no-dev`
2. Activate the `EDD Generator` plugin.
3. Via WP CLI run `wp edd generate notifications --number=10`

This is useful to import a bunch of "dummy" notifications just to add a lot to your database, test the UI, and test dismissals.

### Import via official JSON

You can test importing from our final URL via these steps:

1. Make sure you have NOT defined the `EDD_NOTIFICATIONS_API_URL` constant.
2. Use a plugin like [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/) to manually trigger the `edd_daily_scheduled_events` hook.
3. If you do not have any passes or individual license keys activated, you should successfully have 1 (at time of writing) notification imported.
4. If you do have a pass or individual license activated, you should not have gotten any notifications imported.

### Import via local JSON

You can test the notification import with these steps:

1. Create a new file in `wp-content/uploads` called `edd-notifications.json`.
2. Paste in the contents below ("Notification JSON")
[...]

PR #8982, opened 2021-11-10
Use highest pass license key for updating if a specific key is not set

### LINKED ISSUE #8980: If an AA license is active on the site, use that as a fallback license
As an AA pass holder, it would be nice if my stored license key was used as a fallback for getting plugin updates, if I forget to entire the license key for an extension.

### Justification or use case
Sometimes when I install a new extension, I forget that I still need to enter my license key there. If I haven't entered a license key and I have an active pass, checking that key (instead of none) would still allow me to have an update.

Note that this is not the best long term solution but could ease friction temporarily.

robincornett:
Fixes #8980

Proposed Changes:
1. Adds a highest license key property to the `Pass_Manager` class.
2. Instead of just getting the highest pass ID, set the highest pass ID and license key together when evaluating passes.
3. Update related tests docs.
4. When setting up the auto update class in `EDD_License`, if no license key is saved for the extension, but a pass license key is available, use that for the auto updater.

Note that the Pass Manager class changes are pulled directly from #8945. Additionally, I have branched this off of 2.11.4 although the issue is not currently milestoned.

### DIFF
--- a/includes/class-edd-license-handler.php
+++ b/includes/class-edd-license-handler.php
@@ -160,9 +160,8 @@ class EDD_License {
 		// Fall back to the highest license key if one is not saved for this extension.
 		if ( empty( $license ) ) {
 			$pass_manager = new \EDD\Admin\Pass_Manager();
-			$highest_key  = $pass_manager->highest_license_key;
-			if ( $highest_key ) {
-				$license = $highest_key;
+			if ( $pass_manager->highest_license_key ) {
+				$license = $pass_manager->highest_license_key;
 			}
 		}
 		$betas = edd_get_option( 'enabled_betas', array() );
--- a/includes/class-edd-license-handler.php
+++ b/includes/class-edd-license-handler.php
@@ -156,11 +156,20 @@ class EDD_License {
 			return;
 		}
 
+		$license = $this->license;
+		// Fall back to the highest license key if one is not saved for this extension.
+		if ( empty( $license ) ) {
+			$pass_manager = new \EDD\Admin\Pass_Manager();
+			$highest_key  = $pass_manager->highest_license_key;
+			if ( $highest_key ) {
+				$license = $highest_key;
+			}
+		}
 		$betas = edd_get_option( 'enabled_betas', array() );
 
 		$args = array(
 			'version'   => $this->version,
-			'license'   => $this->license,
+			'license'   => $license,
 			'author'    => $this->author,
 			'beta'      => function_exists( 'edd_extension_has_beta_support' ) && edd_extension_has_beta_support( $this->item_shortname ),
 		);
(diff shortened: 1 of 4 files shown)


includes/class-edd-license-handler.php
```diff
@@ -156,11 +156,20 @@ public function auto_updater() {
...
 		}
 
+		$license = $this->license;
+		// Fall back to the highest license key if one is not saved for this extension.
+		if ( empty( $license ) ) {
+			$pass_manager = new \EDD\Admin\Pass_Manager();
+			$highest_key  = $pass_manager->highest_license_key;
+			if ( $highest_key ) {
+				$license = $highest_key;
+			}
```
ashleyfae, line 166:
```suggestion
			if ( $pass_manager->highest_license_key ) {
				$license = $pass_manager->highest_license_key;
			}
```

The new variable isn't needed.

ashleyfae (commented):
Code looks good. Just need to test it.

robincornett:
I checked my http logs after implementing and everything had a pass. One thing we should also look into implementing is combining the updater checks--I know we built in the capability but we haven't implemented here.

robincornett:
Actually this is [sometimes] throwing a fatal error for me:

> PHP Fatal error:  Uncaught Error: Class 'EDD\Admin\Pass_Manager' not found in \plugins\easy-digital-downloads\includes\class-edd-license-handler.php:162

ashleyfae:
> Actually this is [sometimes] throwing a fatal error for me:
> 
> > PHP Fatal error:  Uncaught Error: Class 'EDD\Admin\Pass_Manager' not found in \plugins\easy-digital-downloads\includes\class-edd-license-handler.php:162
> 
> Since the license handler can run outside of the admin (cron), I think that's why this is happening. So I think for this to fully work we might need to load the Pass Manager class all the time, not just when `is_admin` is true.

Yeah I also had problems with that and actually changed it in one of my PRs so it's included all the time. (Forget which one.) You're welcome to make that change here too.

> I checked my http logs after implementing and everything had a pass. One thing we should also look into implementing is combining the updater checks--I know we built in the capability but we haven't implemented here.

Definitely something we want to do and I've talked to Chris about it. But IMO in order to do it **cleanly** we need to make some large structural changes to our updater. So that's the hold up.

robincornett:
It looks like you moved where the file was being required in #8941, but no changes in the pass manager file itself. I'll move it to the same location here.

ashleyfae (approved):
Confirmed I was able to install an update for EDD Reviews when I had an All Access Pass license key entered for another product.

The UI is slightly wonky (forgot to get a screenshot). On the plugin row it invites me to install the update, but also prints out:

> Enter valid license key for automatic updates

But I think this is a good first step. I think we'll need some additional structural changes to make this 100% seamless.

PR #8988, opened 2021-11-11
Set the hidden input value when cloning a row

### LINKED ISSUE #8987: Variable prices are missing the index in the meta value
### Expected behavior
The `edd_variable_prices` meta value should have the `index` value set for variable prices when they are created

### Actual behavior
Only the first price option has the `index` set but if you hit Update for a published Download then it will assign them.

1) Create a new variable product and click Publish
2) Look in the wp_postmeta table for the `edd_variable_prices` key for that product
3) Check the `index` values
```
array (
  1 => 
  array (
    'index' => '1',
    'name' => 'Price A',
    'amount' => '5.00',
    'license_limit' => '0',
    'recurring' => 'no',
  ),
  2 => 
  array (
    'index' => '',
    'name' => 'Price B',
    'amount' => '10.00',
    'license_limit' => '',
  ),
  3 => 
  array (
    'index' => '',
    'name' => 'Price C',
    'amount' => '15.00',
    'license_limit' => '',
  ),
)
```

PHP Version: 7+

Reported by a customer here https://secure.helpscout.net/conversation/1693630434/1636990/ and he mentions this was introduced in 2.13.3 or .1

From @robincornett 
_It looks like when a row is cloned, most of the inputs in the new row are updated, but there is a hidden input in the price ID title which should have its value set as the row/price ID, and that's not happening. The input affected has a class of edd_repeatable_index; I think we need to update our JS to set its value on clone._

robincornett:
Fixes #8987

Proposed Changes:
1. When a repeatable row is cloned, make sure the hidden input with the row title has the value populated with the correct index. This affects (but may not be limited to):
    * variable prices
    * download files
    * upgrade paths (SL)

The variable pricing metadata does use the value of this input, but the others may not.


ashleyfae (approved):

ashleyfae:
@robincornett Can you fix the merge conflict please?

ashleyfae (approved):

PR #8989, opened 2021-11-11
Add failure notice if purchase receipt could not be re-sent

### LINKED ISSUE #8986: "Resend Purchase Receipt" assumes sending the email was successful
### Expected behavior
If I opt to re-send a customer's purchase receipt, and sending the email fails, I expect to see an error message.

### Actual behavior
I see an admin notice that says it was successful.

1) Easiest thing to do is to manually edit `EDD_Emails::send()` so that it returns `false`.
2) Click the link via admin payment page to resend a purchase receipt.
3) Note success message.

PHP Version:

EDD Version (or branch): 2.9.x

WordPress Version:

The `edd_email_purchase_receipt()` function should return the result of `EDD_Emails::send()`.
Then in `edd_resend_purchase_receipt()` we should be checking the result of that and changing the `edd-message` query arg accordingly.

robincornett:
Fixes #8986

Proposed Changes:
1. Updates the emails `send` method to document the `return`.
2. Updates `edd_email_purchase_receipt` to use that boolean and also return it.
3. If that returns `false` when resending the purchase receipt, generate a failure/error notice.

### DIFF
--- a/includes/admin/class-edd-notices.php
+++ b/includes/admin/class-edd-notices.php
@@ -222,7 +222,7 @@ class EDD_Notices {
 						$notices['updated']['edd-payment-sent'] = __( 'The purchase receipt has been resent.', 'easy-digital-downloads' );
 						break;
 					case 'email_send_failed':
-						$notices['error']['edd-payment-sent'] = __( 'The purchase receipt was not successfully sent.', 'easy-digital-downloads' );
+						$notices['error']['edd-payment-sent'] = __( 'Failed to send purchase receipt.', 'easy-digital-downloads' );
 						break;
 					case 'refreshed-reports' :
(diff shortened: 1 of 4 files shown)


includes/admin/class-edd-notices.php
```diff
@@ -221,6 +221,9 @@ public function show_notices() {
 					case 'email_sent' :
 						$notices['updated']['edd-payment-sent'] = __( 'The purchase receipt has been resent.', 'easy-digital-downloads' );
 						break;
+					case 'email_send_failed':
+						$notices['error']['edd-payment-sent'] = __( 'The purchase receipt was not successfully sent.', 'easy-digital-downloads' );
```
ashleyfae, line 225:
```suggestion
						$notices['error']['edd-payment-sent'] = __( 'Failed to send purchase receipt.', 'easy-digital-downloads' );
```

What do you think about this wording instead?

ashleyfae (approved):

PR #8991, opened 2021-11-15
Add support for gateway conditions.

ashleyfae:
Proposed Changes:
1. Add support for payment gateway conditions. This targets installs that have a certain payment gateway enabled. Notifications targeting a gateway should only be imported if you have that gateway activated.

## Testing

Test notifications have been set up. For the gateway specific ones (there are a few others too), here's what we have:

- `Test Notification - Amazon` - should only be imported if you have the "Amazon" gateway active. Notification type should be "warning" (orange exclamation mark icon).
- `Test Notification - PayPal Commerce` - should only be imported if you have the new PayPal Commerce gateway active.
- `Test Notification - Stripe` - should only be imported if you have the Stripe gateway active.
- `Test Notification - PayPal Standard` - should only be imported if you have the PayPal Standard gateway active. This notification doesn't have any buttons associated.

### Steps

- To test, make sure you DO NOT have the `EDD_NOTIFICATIONS_API_URL` constant set.
- You may want to truncate your notifications table, but it's not required.
- Activate a certain payment gateway to test a scenario in the above list. (for example: to test that Amazon imports, activate the Amazon gateway.)
- Manually trigger the `edd_daily_scheduled_events` cron event.
- Check your notifications to see that the notification imported and that irrelevant ones did not.
- Rinse and repeat with different gateway(s) activated to test different notifications.


robincornett (approved):
Tested with each gateway and got the notifications expected.

PR #8993, opened 2021-11-19
Refactor settings for product education

robincornett:
Proposed Changes:
1. Replace SendWP installation with WPSMTP in Emails.
2. Creates a Marketing tab in Downloads > Settings.
3. Moves some extension settings to more logical places:
    - Recurring is now under Payments
    - Invoices is now under Payments
    - Email marketing extensions are under Marketing
    - Free Downloads will be under Marketing
4. Expose certain extensions to users' notice for product education:
    - If their active pass has access to the extension, provides a link to their account on EDD to download so they can install.
    - If their pass doesn't have access, or they don't have a pass, provides a link to the product page.
    - If the extension is installed, but not active, it will offer to activate, then link to the settings page to configure.

To Come:
The extensions API class will be updated to use the new AM product file once that information has been updated.


robincornett:
At this point, to test, you will need to have a local copy of the marketing site with the AM licensing API working, per this issue: https://github.com/awesomemotive/easy-digital-downloads-site/issues/23

Define the `EDD_PRODUCTS_API_URL` constant to be your local site's `products.json` file.

Without the custom post meta added to the site, you will only see the "Log in to download" buttons, if you have an appropriate pass available.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+		<?php
+	}
+
+	/**
+	 * Gets the extension card data.
+	 * Generally, the data from the products API will be used,
+	 * but extending classes can override that in the configuration array.
+	 *
+	 * @since 2.11.x
+	 * @param object $product The extension data from the Products API.
```
ashleyfae, line 129:
```suggestion
	 * @param array $product The extension data from the Products API.
```

It's being treated as an array in the method, so I assume it is one.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+				'extension_manager_nonce'  => wp_create_nonce( 'edd_extensionmanager' ),
+			)
+		);
+	}
+
+	/**
+	 * Outputs a standard extension card.
+	 *
+	 * @since 2.11.x
+	 * @param object $product             The extension information retrieved from the Products API.
```
ashleyfae, line 76:
```suggestion
	 * @param array $product             The extension information retrieved from the Products API.
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+
+	/**
+	 * Gets the download URL.
+	 *
+	 * @param string|int $url_or_item_id  Either the download URL (for a plugin) or the item ID (for an extension).
+	 * @param string     $type            Plugin or extension.
+	 * @return string|false Returns the download URL if possible, or false if not.
+	 */
+	public function get_download_url( $url_or_item_id, $type = 'plugin' ) {
+
```
ashleyfae, line 208:
```suggestion
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+	 * @since 2.11.x
+	 */
+	public function install() {
+
+		// Run a security check.
+		check_ajax_referer( 'edd_extensionmanager', 'nonce', true );
+
+		$generic_error = esc_html__( 'There was an error while performing your request.', 'easy-digital-downloads' );
+		$type          = filter_input( INPUT_POST, 'type', FILTER_SANITIZE_STRING );
+		$required_pass = filter_input( INPUT_POST, 'pass', FILTER_SANITIZE_STRING );
```
ashleyfae, line 250: These need changing. `FILTER_SANITIZE_STRING` is deprecated in PHP 8.1.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+		$result['message'] = 'plugin' === $type
+			? __( 'Could not install the plugin. Please download and install it manually via Plugins > Add New.', 'easy-digital-downloads' )
+			: sprintf(
+				/* translators: 1. opening anchor tag, do not translate; 2. closing anchor tag, do not translate */
+				__( 'Could not install the extension. Please %1$sdownload it from your account%2$s and install it manually.', 'easy-digital-downloads' ),
+				'<a href="https://easydigitaldownloads.com/your-account/" target="_blank" rel="noopener noreferrer">',
+				'</a>'
+			);
+
+		$plugin = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_STRING );
```
ashleyfae, line 273: `FILTER_SANITIZE_STRING` is deprecated in PHP 8.1.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+	}
+
+	/**
+	 * Gets the download URL.
+	 *
+	 * @param string|int $url_or_item_id  Either the download URL (for a plugin) or the item ID (for an extension).
+	 * @param string     $type            Plugin or extension.
+	 * @return string|false Returns the download URL if possible, or false if not.
+	 */
+	public function get_download_url( $url_or_item_id, $type = 'plugin' ) {
```
ashleyfae, line 207: I'm finding this method a little odd, since it just returns exactly what was passed in (provided the `$type === 'plugin'`). Can you explain why you find it useful?

includes/admin/settings/register-settings.php
```diff
@@ -2635,28 +2574,3 @@ function edd_add_setting_tooltip( $html, $args ) {
...
-		 * settings for 'checkout' from 'misc' to 'payments'.
-		 */
-		if ( function_exists( 'edd_recurring_guest_checkout_description' ) && false !== has_filter( 'edd_settings_misc', 'edd_recurring_guest_checkout_description' ) ) {
-			remove_filter( 'edd_settings_misc', 'edd_recurring_guest_checkout_description', 10 );
-			add_filter( 'edd_settings_gateways', 'edd_recurring_guest_checkout_description', 10 );
-		}
-
-	},
-	99
-);
```
ashleyfae, line 2662: Was this definitely removed intentionally?

ashleyfae: Never mind. I see it was moved to a different file.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+
+		// Check for permissions.
+		if ( ! check_ajax_referer( 'edd_extensionmanager', 'nonce', false ) || ! current_user_can( 'activate_plugins' ) ) {
+			$result['message'] = __( 'Plugin activation is not available for you on this site.', 'easy-digital-downloads' );
+			wp_send_json_error( $result );
+		}
+
+		$already_installed = false;
+		if ( empty( $plugin_basename ) ) {
+			$plugin_basename   = filter_input( INPUT_POST, 'plugin', FILTER_SANITIZE_STRING );
```
ashleyfae, line 367: `FILTER_SANITIZE_STRING` is deprecated in PHP 8.1.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+			$already_installed = true;
+		}
+
+		$plugin_basename = sanitize_text_field( wp_unslash( $plugin_basename ) );
+		$type            = filter_input( INPUT_POST, 'type', FILTER_SANITIZE_STRING );
```
ashleyfae, line 372: `FILTER_SANITIZE_STRING` is deprecated in PHP 8.1.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,520 @@
...
+
+	/**
+	 * Upon successful activation, delete the API request options from the database.
+	 *
+	 * @param string $plugin_basename
+	 * @return void
+	 */
+	public function delete_options( $plugin_basename ) {
+		// Delete the option generated by the request to the Products API.
+		$type = filter_input( INPUT_POST, 'type', FILTER_SANITIZE_STRING );
```
ashleyfae, line 424: `FILTER_SANITIZE_STRING` is deprecated in PHP 8.1.

includes/admin/extensions/class-extension-manager.php
robincornett: Yes, it is one--for a while it had been an object, but then I decided I didn't want to store all of the API data and since I wanted to potentially merge it with the custom configuration array, I changed it to an array.

includes/admin/extensions/class-extension-manager.php
robincornett: If the type was not a plugin, the first parameter would have been a product ID and this method called up a class to get the download package from the licensing API. I left the method in as hopefully a future placeholder but I can remove it.

cklosowski:
When there is an error getting the list, we're left with a message (which is good) but also a 'save settings' button, which I think we should try and remove if possible:
image

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,169 @@
...
+	 */
+	public function get_product_data( $body = array(), $item_id = false ) {
+		if ( empty( $body ) ) {
+			if ( empty( $item_id ) ) {
+				return false;
+			}
+			$body = $this->get_api_body( $item_id );
+		}
+		$key         = $this->array_key_first( $body );
+		$option_name = "edd_extension_{$key}_{$body[ $key ]}_data";
```
cklosowski, line 23: Is there any risk here of the `$body` causing an option name that is too long? Should we MD5 this key so that there is less risk of a large option name?

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,169 @@
...
+	 *
+	 * @since 2.11.x
+	 * @return string
+	 */
+	private function get_products_api_url() {
+		if ( defined( 'EDD_PRODUCTS_API_URL' ) ) {
+			return EDD_PRODUCTS_API_URL;
+		}
+
+		return 'https://easydigitaldownloads.com/wp-content/uploads/cache/products/products.json';
```
cklosowski, line 134: Can we adjust this to be the query stringed version?

`https://easydigitaldownloads.com/?edd_action=extension_data`? That way we can adjust the location of the `.json` file if needed or naming, without having to maintain backwards compatibility. They should both return the same data set.

cklosowski (changes requested):

cklosowski:
seeing a small breakpoint issue on the styles in Marketing > Email Marketing:
image

When I have it comfortably in 3 columns, things look fine, but as soon as I just get into 4 columns the images and buttons can extend out of their cards:
image

This happens the same as I just break into 5 columns and 3 columns when we fall just into the breakpoint for the adjustment in the number of columns:
image
image

includes/admin/extensions/class-extensions-api.php
robincornett: At this point, I don't _think_ so, as the values will be:
- `$key`: `product`, `category`, or `tag`
- `$body[ $key ]`: a product or term ID (integer)

image

I'm willing to hash it if that's safest but for right now I confess it's been helpful to have a recognizable option name.

ashleyfae:
@robincornett Can you sort out the merge conflict please?

includes/admin/extensions/class-extensions-api.php
ashleyfae: > I confess it's been helpful to have a recognizable option name.

I like that too.

ashleyfae:
When I try to activate an extension that's already installed, I get this error:

Screenshot from 2021-11-30 13-24-40

Looks like it's not using `plugin` from the ajax request, so `$plugin_basename` is just empty.

includes/admin/extensions/class-extensions-api.php
cklosowski: I do like it as well, just want to make sure we keep that in mind as we use this in the future. :+1: Can we make a comment or something about how the key is created so we don't overrun it by accident?

includes/admin/extensions/class-extension-manager.php
ashleyfae: Based on the context, I feel like right now it's acting more like `can_product_be_downloaded()` (given that if it returns `false`, you then do `wp_send_json_error()`).

I think for clarity it would be better to either remove it or rename it. Once we work on the licensing API stuff we can add it back in as it was if necessary.

ashleyfae:
Activating works now. :raised_hands: 

Screenshot from 2021-12-01 15-27-54

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,269 @@
...
+			$this->get_button_parameters( $product_data, $item_id ),
+			$this->get_link_parameters( $product_data )
+		);
+	}
+
+	/**
+	 * Gets the product data for a specific extension.
+	 *
+	 * @param boolean $item_id
+	 * @return bool|object False if there is no data; API object if there is.
```
ashleyfae, line 81:
```suggestion
	 * @return false|array False if there is no data; API object if there is.
```

Looks like it's actually an array, not an object.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,169 @@
...
+
+class ExtensionsAPI {
+
+	/**
+	 * Gets the product data from the EDD Products API.
+	 *
+	 * @since 2.11.x
+	 * @param array $body    The body for the API request.
+	 * @param int   $item_id The product ID, if querying a single product.
+	 * @return object
```
ashleyfae, line 13:
```suggestion
	 * @return array|false
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		return $type;
+	}
+
+	/**
+	 * Gets the button parameters.
+	 * Classes should not need to replace this method.
+	 *
+	 * @param object $product_data The extension data returned from the Products API.
```
ashleyfae, line 154:
```suggestion
	 * @param array $product_data The extension data returned from the Products API.
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,269 @@
...
+			'utm_medium'   => urlencode( $tab ),
+			'utm_campaign' => 'admin',
+			'utm_term'     => urlencode( $product_data['slug'] ),
+		);
+
+		if ( $has_access ) {
+			$url = 'https://easydigitaldownloads.com/your-account/your-downloads/';
+			unset( $utm_parameters['p'] );
+		} elseif ( ! empty( $product_data['upgrade_url'] ) ) {
+			$url = esc_url( $product_data['upgrade_url'] );
```
ashleyfae, line 225:
```suggestion
			$url = esc_url_raw( $product_data['upgrade_url'] );
```

I think this is more appropriate. `esc_url()` would be good on display (which I think you're already doing). `esc_url_raw()` is more of a sanitize function.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,269 @@
...
+	protected function get_link_parameters( $product_data ) {
+		return array(
+			/* translators: The extension name. */
+			'button_text' => sprintf( __( 'Configure %s', 'easy-digital-downloads' ), $product_data['title'] ),
+			'href'        => add_query_arg(
+				array(
+					'post_type' => 'download',
+					'page'      => 'edd-settings',
+					'tab'       => $product_data['tab'],
+					'section'   => $product_data['section'],
```
ashleyfae, line 251:
```suggestion
					'tab'       => urlencode( $product_data['tab'] ),
					'section'   => urlencode( $product_data['section'] ),
```

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,169 @@
...
+	public function get_product_data( $body = array(), $item_id = false ) {
+		if ( empty( $body ) ) {
+			if ( empty( $item_id ) ) {
+				return false;
+			}
+			$body = $this->get_api_body( $item_id );
+		}
+		$key = $this->array_key_first( $body );
+		// The option name is created from the first key/value pair of the API "body".
+		$option_name = "edd_extension_{$key}_{$body[ $key ]}_data";
```
ashleyfae, line 24:
```suggestion
		$option_name = sanitize_key( "edd_extension_{$key}_{$body[ $key ]}_data" );
```

Feels like a good idea. Will make sure it's appropriate as an option name.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,169 @@
...
+		$option      = $this->get_stored_extension_data( $option_name );
```
ashleyfae, line 25: Not urgent, so this could wait for a future release instead. But I think we can rework this cache a little to be more error-friendly. Right now you could run into this scenario:

- You get product data. It gets cached.
- Cache expires, so it gets deleted in `get_stored_extension_data()`.
- You do a new API request.
- It fails.
- You get no data for at least an hour.

Here's a way we could improve that:

- You get product data. It gets cached.
- In `get_product_data()` you note that the cache has expired, so you need to do an API request, but you don't actually delete the cache yet.
- You do a new API request.
- It fails.
- Because you didn't delete the cache, you can just extend the timeout.

This way we reduce the chance of someone seeing no product information. Instead we just keep serving them the stale cache until the API request is successful again.

includes/admin/settings/class-reviews.php
```diff
@@ -0,0 +1,106 @@
...
+	 * If Reviews is not active, registers a metabox on individual download edit screen.
+	 *
+	 * @since 2.11.x
+	 * @return void
+	 */
+	public function maybe_do_metabox() {
+		if ( $this->is_activated() ) {
+			return;
+		}
+		\add_meta_box(
```
ashleyfae, line 78:
```suggestion
		add_meta_box(
```

Doesn't do any harm, but it's not required.

ashleyfae:
I'm getting the following warning when viewing an individual download (admin side):

```
PHP Warning:  Undefined array key "slug" in easy-digital-downloads/includes/admin/extensions/abstract-extension.php on line 218
```

includes/admin/extensions/abstract-extension.php
robincornett: I will commit this but I want to leave it as a placeholder for half a moment. I need to think about how to handle extensions which may not have settings at all...I don't know if we have any but if we do, I think this needs to be handled differently.

includes/admin/extensions/class-extensions-api.php
robincornett: I agree that sounds like a better way to handle it--I had thought of that and would prefer to show old data rather than nothing.

ashleyfae:
Is this intentional styling for Reviews?

Screenshot from 2021-12-03 09-34-07

Feels a little odd to have the background span the whole page, but the content is in a tighter container.

robincornett:
> Is this intentional styling for Reviews?

I kind of backed out the styling for individual extensions (removed max-width as you can see) because Chris said he wanted a different look--will update styling when I have details on what's wanted.

robincornett:
Update: per the latest directive, I've adjusted the extension card management markup somewhat to allow for different views:
- `stacked`: This is the default CSS class, currently has no specific rules.
- `horizontal`: Currently used for Reviews/Invoices, puts the image on the left of the card and text/button on the right.
- `detailed`: Currently used for Reviews, intended to be used when there is a lot of content.

The content can be anything that's included in a post (it's run through `wp_kses_post` and `wpautop`).

The custom configurations now use default title/description array keys so anything added to the `get_configuration` array which matches anything in the product data array will override the data pulled from our site.

Although the product education content is output mostly on settings screens (the Reviews metabox would be the only difference currently), since they aren't actually settings, I've moved them to a `product-education` directory inside `admin/extensions`.

ashleyfae:
When I view the Email Marketing tab and it does a fresh API request, I get this error:

```
Fatal error: Uncaught Error: array_merge(): Argument #1 must be of type array, int given
in easy-digital-downloads/includes/admin/extensions/abstract-extension.php on line 101
```

To reproduce:

1. Edit `class-extensions-api.php` and at the top of the `get_stored_extension_data()` method, add `return false;` to force a new API request. (Could also delete the option.)
2. View the Email Marketing page.
3. Note fatal error.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,286 @@
...
+	public function get_product_data( $item_id = false ) {
+		require_once EDD_PLUGIN_DIR . 'includes/admin/extensions/class-extensions-api.php';
+		$api          = new ExtensionsAPI();
+		$body         = $this->get_api_body();
+		$product_data = $api->get_product_data( $body, $this->item_id );
+		if ( ! $product_data ) {
+			return false;
+		}
+
+		$config = $this->get_configuration( $item_id );
```
ashleyfae, line 92:
```suggestion
		$config = $this->get_configuration();
```

This method no longer has a parameter.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,286 @@
...
+
+	/**
+	 * Gets the type for the button data-type attribute.
+	 * This is intended to sync with the Products API request.
+	 * Default is product.
+	 *
+	 * Really a shim for array_key_first.
+	 *
+	 * @param array $array
+	 * @return void
```
ashleyfae, line 136:
```suggestion
	 * @return string
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,484 @@
+<?php
+
+namespace EDD\Admin\Extensions;
+
+use \EDD\Admin\Pass_Manager;
+use \EDD\Admin\Extensions\ExtensionsDownloadURL;
```
ashleyfae, line 6:
```suggestion
```

No longer exists? Not used, anyway.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,484 @@
...
+		$card_classes = array(
+			$base_class,
+		);
+		$variation    = 'stacked';
+		if ( ! empty( $product['card'] ) ) {
+			$variation = $product['card'];
+		}
+		$card_classes[] = "{$base_class}--{$variation}";
+		?>
+		<div class="<?php echo esc_attr( implode( ' ', $card_classes ) ); ?>">
```
ashleyfae, line 95:
```suggestion
		<div class="<?php echo esc_attr( implode( ' ', array_map( 'sanitize_html_class', $card_classes ) ) ); ?>">
```

I like the sanitize function here because the source of the class names isn't really clear.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,484 @@
...
+	public function __construct( $required_pass_id = null ) {
+		if ( $required_pass_id ) {
+			$this->required_pass_id = $required_pass_id;
+		}
+		$this->pass_manager = new Pass_Manager();
+
+		add_action( 'wp_ajax_edd_activate_extension', array( $this, 'activate' ) );
+		add_action( 'wp_ajax_edd_install_extension', array( $this, 'install' ) );
+		add_action( 'admin_enqueue_scripts', array( $this, 'register_scripts' ) );
+		// add_action( 'edd_plugin_activated', array( $this, 'delete_options' ) );
```
ashleyfae, line 42: I feel like I saw you address this at some point, but I can't find it. Can you refresh my memory as to why this is here (but also commented out)?

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	/**
+	 * Gets the license key associated with the pass.
+	 *
+	 * @return null|string
+	 */
+	private function get_license_key() {
+		return $this->pass_manager->highest_license_key;
+	}
```
ashleyfae, line 445: This function appears to no longer be used.

includes/admin/extensions/class-extensions-api.php
robincornett: Reworked the data request in aa99898 to return stale data if there is an API error, and only delete the option if it is stale and new data is retrieved.

robincornett:
@ashleyfae @cklosowski update as of today:
- The product API request should now use stale data if it's available and the fresh request fails for any reason.
- The custom configuration accepts the product API data as an optional parameter so that the custom data can be modified along with it (this was necessary for the email marketing headings to say "Get [Extension] Today!")
- I've reworked the markup to optionally allow for an additional two column setting per Chris' most recent email. So now there are four "style" options. The newest is `detailed-2col`.
- Features are now separated from the general product description and can be added to the custom data as an array. The features will automatically be added to the extension card as an unordered list with checkmarks.

Here's the Recurring screen with the `detailed` card style:
image

Here it is with the `detailed-2col` style:
image

This all makes the CSS a little less automatically responsive but I think it's all working.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,284 @@
...
+	 *
+	 * @since 2.11.x
+	 * @return bool
+	 */
+	abstract protected function is_activated();
+
+	/**
+	 * Output the settings field (installation helper).
+	 *
+	 * @param array $args
```
ashleyfae, line 45:
```suggestion
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,284 @@
...
+			return;
+		}
+		$this->do_single_extension_card();
+	}
+
+	/**
+	 * Outputs a single extension card.
+	 *
+	 * @since 2.11.x
+	 * @param bool|int $item_id Optional: the individual extension product ID.
```
ashleyfae, line 59:
```suggestion
	 * @param false|int $item_id Optional: the individual extension product ID.
```

I think this is either `false` or an item ID, but never `true`? If so, I think saying `false` explicitly is a bit clearer.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,284 @@
...
+			$this->get_button_parameters( $product_data, $item_id ),
+			$this->get_link_parameters( $product_data ),
+			$this->get_configuration( $product_data )
+		);
+	}
+
+	/**
+	 * Gets the product data for a specific extension.
+	 *
+	 * @param boolean $item_id
```
ashleyfae, line 81:
```suggestion
	 * @param bool|false $item_id
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,284 @@
...
+
+		return $type;
+	}
+
+	/**
+	 * Gets the button parameters.
+	 * Classes should not need to replace this method.
+	 *
+	 * @param array $product_data The extension data returned from the Products API.
+	 * @param int    $item_id      Optional: the item ID.
```
ashleyfae, line 156:
```suggestion
	 * @param array     $product_data The extension data returned from the Products API.
	 * @param int|false $item_id      Optional: the item ID.
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,284 @@
...
+				array(
+					'post_type' => 'download',
+					'page'      => 'edd-settings',
+					'tab'       => urlencode( $product_data['tab'] ),
+					'section'   => urlencode( $product_data['section'] ),
+				),
+				admin_url( 'edit.php' )
+			),
+		);
+	}
```
ashleyfae, line 271:
```suggestion
	protected function get_link_parameters( $product_data ) {
		if ( empty( $product_data['tab'] ) && empty( $product_data['section'] ) ) {
			return array(
				/* translators: the plural Downloads label. */
				'button_text' => sprintf( __( 'View %s', 'easy-digital-downloads' ), edd_get_label_plural() ),
				'href'        => add_query_arg(
					array(
						'post_type' => 'download',
					),
					admin_url( 'edit.php' )
				),
			);
		}

		return array(
			/* translators: The extension name. */
			'button_text' => sprintf( __( 'Configure %s', 'easy-digital-downloads' ), $product_data['title'] ),
			'href'        => add_query_arg(
				array(
					'post_type' => 'download',
					'page'      => 'edd-settings',
					'tab'       => urlencode( $product_data['tab'] ),
					'section'   => urlencode( $product_data['section'] ),
				),
				admin_url( 'edit.php' )
			),
		);
	}
```

Silly minor formatting thing:

1. Removed empty line on line 245.
2. Added empty line between 257 and 258.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,168 @@
...
+			)
+		);
+
+		// If there was an API error, set timeout for 1 hour and return stale product data if it exists.
+		if ( is_wp_error( $request ) || ( 200 !== wp_remote_retrieve_response_code( $request ) ) ) {
+			$data = array(
+				'timeout' => strtotime( '+1 hour', time() ),
+			);
+			if ( $option && $is_stale ) {
+				$data = array_merge( $option, array( 'timeout' => strtotime( '+1 hour', $data['timeout'] ) ) );
```
ashleyfae, line 60: Just wondering if this timeout needs to be relative to the original `$data['timeout']` ? Could we just make it `strtotime( '+1 hour', time() )` ?

Potential scenario:

1. Cache expired 1 day ago.
2. User views the page to try again.
3. Request fails.
4. New timeout would be set to: {1 day ago} + 1 hour.
5. Which means it would try again immediately, right?

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,168 @@
...
+
+			if ( $item_id && ! empty( $option[ $item_id ] ) ) {
+				return $option[ $item_id ];
+			}
+
+			return $option;
+		}
+
+		// Fresh data has been retrieved, so remove the option and populate with fresh data.
+		delete_option( $option_name );
```
ashleyfae, line 76: I don't think this `delete_option()` is needed, as we update it on line 97 anyway.

ashleyfae:
When the API request fails, I get these errors:

```
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-recurring.php on line 102
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-invoices.php on line 80
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-reviews.php on line 99
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-reviews.php on line 99
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-recurring.php on line 102
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-invoices.php on line 80
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-reviews.php on line 99
[08-Dec-2021 15:40:32 UTC] PHP Warning:  Trying to access array offset on value of type bool in easy-digital-downloads/includes/admin/extensions/product-education/class-reviews.php on line 99
```

To reproduce:

1. Delete all `edd_extension_%` options.
2. Edit this line: https://github.com/awesomemotive/easy-digital-downloads/blob/try/product-education/includes/admin/extensions/class-extensions-api.php#L53 and add:

```php
$request = new \WP_Error();
```

cklosowski:
I really like the look of `detailed-2col` @robincornett. That is exactly what I envisioned after spending some time in there playing with a few designs.

I should have the copy for you by monday.

robincornett:
@ashleyfae I think I've hopefully accounted better for an API request failure in e80d652 and then 961af36:
- I created a private property for the product data (it maybe doesn't need to be a property...just felt like I might prefer that instead of calling it multiple times in the product classes).
- Each product class now returns early if there is no product data so we won't end up with a blank screen.
- Additionally, the `is_plugin_active` method now takes either a string (the plugin basename) or an array (the product data) and that method handles the check for whether the basename is defined and then runs its code.

Other notes:
- 85a54c2: I discovered that the `wp_localize_script` was actually adding the JS variable once for each product class, so I reworked the script to be registered just once, and then enqueued instead of printed. This also separates the CSS out into its own file so it's only loaded as needed.
- 9174584: Created a helper method to determine if the current screen is an EDD settings screen or something else...needed so that the Reviews configuration could change based on the screen/context and figured I should make that a thing for any product to use.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,302 @@
...
+	 * @var int
+	 */
+	protected $item_id;
+
+	/**
+	 * The product data.
+	 *
+	 * @var boolean
+	 */
+	private $product_data = false;
```
ashleyfae, line 23:
```suggestion
	protected $product_data = false;
```

`private` means only this abstract class can access it. Because of that, each of the child classes are actually dynamically declaring a new public property. Changing to `protected` allows them to access it properly.

ashleyfae: Actually building on this, I'm not quite sure I like how this change sets the product data in the constructor. Example: https://github.com/awesomemotive/easy-digital-downloads/blob/try/product-education/includes/admin/extensions/product-education/class-email-marketing.php#L20-L23

I think it would actually be really nice to only run the API requests when absolutely needed. (Meaning: both the cache has expired AND you're currently viewing the relevant settings page.) Right now this will be doing API requests if the cache has expired and you're viewing ANY admin page at all. That's bad for both the customer and for us.

includes/admin/extensions/product-education/class-invoices.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	/**
+	 * Adds the Invoices Payments section to the settings.
+	 *
+	 * @param array $sections
+	 * @return array
+	 */
+	public function add_section( $sections ) {
+		if ( $this->is_activated() ) {
```
ashleyfae, line 65:
```suggestion
		if ( ! edd_is_admin_page( 'settings' ) ) {
			return $sections;
		}

		if ( $this->is_activated() ) {
```

If we add this to each extension class that checks `is_activated()` then we can prevent ever doing an API request unless the person is on the EDD settings page.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,207 @@
...
+		$key = $this->array_key_first( $body );
+		// The option name is created from the first key/value pair of the API "body".
+		$option_name = sanitize_key( "edd_extension_{$key}_{$body[ $key ]}_data" );
+		$option      = get_option( $option_name );
+		$is_stale    = ! empty( $option['timeout'] ) && time() > $option['timeout'];
+
+		// If the data is "fresh" and what we want exists, return it.
+		if ( ! $is_stale ) {
+			if ( $item_id && ! empty( $option[ $item_id ] ) ) {
+				return is_array( $option[ $item_id ] ) ? $this->product_data->fromArray( $option[ $item_id ] ) : $option[ $item_id ];
```
robincornett, line 44: @ashleyfae This and line 70 do not necessarily have to stay this way--they are in so that anyone testing this PR won't have explosions over the change from the array to the `ProductData` object. Once I can safely assume that anyone would be getting fresh/new options created, this should just be able to be:

```php
return $option[ $item_id ];
```

If I do that, then I could remove the product data property from the class and constructor and just instantiate the product data class down in the `get_item_data` method, I think, and skip the whole array to object conversion, I think?

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,207 @@
...
+
+	/**
+	 * Gets the product data as needed for the extension manager.
+	 *
+	 * @since 2.11.x
+	 * @param object $item
+	 * @return ProductData
+	 */
+	private function get_item_data( $item ) {
+		return $this->product_data->fromArray( array(
```
robincornett, line 149: @ashleyfae I did this assuming that it would be better, now that we have it, to store the product data object directly instead of storing it as an array. I did want to note that the `$features` property is then saved for every product, even though it isn't populated at this time (and maybe that was wrong), so I did wonder if it would be better to save the data as an array and convert it to the object on retrieval....so theoretically properties that exist in the class but aren't populated for whatever reason won't be saved to the database.

robincornett: Actually I decided to go ahead and do that in 7623366 but can revert it if you think I should.

includes/admin/extensions/class-extension-product-data.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @var array
+	 */
+	public $features = array();
+
+	/**
+	 * Take array and return object.
+	 *
+	 * @since 2.11.x
+	 * @param array $array
+	 * @return ProductData
```
ashleyfae, line 76:
```suggestion
	 * @return ProductData
	 * @throws \InvalidArgumentException
```

(Assuming you accept my next suggestion.)

includes/admin/extensions/class-extension-product-data.php
```diff
@@ -0,0 +1,111 @@
...
+	 */
+	public function fromArray( $array ) {
+		$expected_keys  = array( 'title', 'slug', 'description', 'basename' );
+		$array_to_check = array_intersect_key( $array, array_flip( $expected_keys ) );
+
+		if ( empty( $array_to_check ) ) {
+			throw new \Exception(
```
ashleyfae, line 83:
```suggestion
			throw new \InvalidArgumentException(
```

I kind of like `InvalidArgumentException` here because it is a problem with the arguments passed in. Just a bit more specific.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+			$this->get_link_parameters( $product_data ),
+			$this->get_configuration( $product_data )
+		);
+	}
+
+	/**
+	 * Gets the product data for a specific extension.
+	 *
+	 * @param false|int $item_id
+	 * @return bool|\ProductData|array False if there is no data; product data object if there is, or possibly an array of arrays.
```
ashleyfae, line 89:
```suggestion
	 * @return bool|ProductData|array False if there is no data; product data object if there is, or possibly an array of arrays.
```

This is in the same namespace so we don't want the backslash.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+	 */
+	public function get_product_data( $item_id = false ) {
+		require_once EDD_PLUGIN_DIR . 'includes/admin/extensions/class-extensions-api.php';
+		$api          = new ExtensionsAPI();
+		$body         = $this->get_api_body();
+		$api_item_id  = $item_id ? $item_id : $this->item_id;
```
ashleyfae, line 95:
```suggestion
		$api_item_id  = $item_id ?: $this->item_id;
```

Optional (don't know how you feel about this syntax), but my editor picked it up. This has the exact same effect.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		return $product_data;
+	}
+
+	/**
+	 * Gets the custom configuration for the extension.
+	 *
+	 * @since 2.11.x
+	 * @param \EDD\Admin\Extensions\ProductData $product_data Optionally allows the product data to be parsed inthe configuration.
```
ashleyfae, line 116:
```suggestion
	 * @param ProductData $product_data Optionally allows the product data to be parsed in the configuration.
```

Having the full namespace isn't a problem, but it's unnecessary since `ProductData` is in the same namespace as this class.

(Also fixed a typo.)

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		return $type;
+	}
+
+	/**
+	 * Gets the button parameters.
+	 * Classes should not need to replace this method.
+	 *
+	 * @param \EDD\Admin\Extensions\ProductData $product_data The extension data returned from the Products API.
```
ashleyfae, line 183:
```suggestion
	 * @param ProductData $product_data The extension data returned from the Products API.
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+	 * @param int|false                         $item_id      Optional: the item ID.
+	 * @return array
+	 */
+	protected function get_button_parameters( $product_data, $item_id = false ) {
```
ashleyfae, line 187:
```suggestion
	protected function get_button_parameters( ProductData $product_data, $item_id = false ) {
```

We're actually allowed to type hint classes! (Assuming it has to be that class, which it looks like it does.)

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	/**
+	 * Gets the custom configuration for the extension.
+	 *
+	 * @since 2.11.x
+	 * @param \EDD\Admin\Extensions\ProductData $product_data Optionally allows the product data to be parsed inthe configuration.
+	 * @return array
+	 */
+	protected function get_configuration( \EDD\Admin\Extensions\ProductData $product_data ) {
```
ashleyfae, line 119:
```suggestion
	protected function get_configuration( ProductData $product_data ) {
```

Optional. Full namespace works, but isn't necessary.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+
+		return $button;
+	}
+
+	/**
+	 * Gets the upgrade URL for the button.
+	 *
+	 * @todo add UTM parameters
+	 * @since 2.11.x
+	 * @param \EDD\Admin\Extensions\ProductData $product_data The product data object.
```
ashleyfae, line 234:
```suggestion
	 * @param ProductData $product_data The product data object.
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+	 * @param int                               $item_id      The item/product ID.
+	 * @param bool                              $has_access   Whether the user already has access to the extension (based on pass level).
+	 * @return string
+	 */
+	private function get_upgrade_url( $product_data, $item_id, $has_access = false ) {
```
ashleyfae, line 239:
```suggestion
	private function get_upgrade_url( ProductData $product_data, $item_id, $has_access = false ) {
```

Because we can. :heart:

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+		$url            = 'https://easydigitaldownloads.com';
+		$tab            = ! empty( $_GET['tab'] ) ? sanitize_text_field( $_GET['tab'] ) : '';
+		$slug           = ! empty( $product_data->slug ) ? $product_data->slug : '';
```
ashleyfae, line 242: If you set the default inside the `ProductData` class, then this wouldn't be necessary.

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+			$utm_parameters,
+			$url
+		);
+	}
+
+	/**
+	 * Gets the array of parameters for the link to configure the extension.
+	 *
+	 * @since 2.11.x
+	 * @param \EDD\Admin\Extensions\ProductData  $product_data  The product data object.
```
ashleyfae, line 269:
```suggestion
	 * @param ProductData  $product_data  The product data object.
```

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,312 @@
...
+	 * @return array
+	 */
+	protected function get_link_parameters( $product_data ) {
```
ashleyfae, line 272:
```suggestion
	protected function get_link_parameters( ProductData $product_data ) {
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,489 @@
...
+	 *
+	 * @since 2.11.x
+	 * @var int
+	 */
+	private $required_pass_id;
+
+	/**
+	 * Pass Manager class
+	 *
+	 * @var \EDD\Admin\Pass_Manager
```
ashleyfae, line 29:
```suggestion
	 * @var Pass_Manager
```

Full namespace isn't necessary since you import it further up.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,489 @@
...
+				'extension_manager_nonce'  => wp_create_nonce( 'edd_extensionmanager' ),
+			)
+		);
+	}
+
+	/**
+	 * Outputs a standard extension card.
+	 *
+	 * @since 2.11.x
+	 * @param \ProductData $product             The product data object.
```
ashleyfae, line 79:
```suggestion
	 * @param ProductData $product             The product data object.
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,489 @@
...
+	 * @param array        $inactive_parameters The array of information to build the button for an inactive/not installed plugin.
+	 * @param array        $active_parameters   The array of information needed to build the link to configure an active plugin.
+	 * @param array        $configuration       The optional array of data to override the product data retrieved from the API.
+	 * @return void
+	 */
+	public function do_extension_card( $product, $inactive_parameters, $active_parameters, $configuration = array() ) {
```
ashleyfae, line 85:
```suggestion
	public function do_extension_card( ProductData $product, $inactive_parameters, $active_parameters, $configuration = array() ) {
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,489 @@
...
+		if ( ! $product ) {
+			return;
+		}
```
ashleyfae, line 88: Don't think this is necessary? Certainly not if the type hint is added, as then it could never get to this point.

includes/admin/extensions/class-extension-product-data.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @param  array       $configuration The custom configuration data.
+	 * @return ProductData
+	 */
+	public function mergeConfig( ProductData $product, array $configuration ) {
+		foreach ( $configuration as $key => $value ) {
+			$product->{$key} = $value;
+		}
+
+		return $product;
+	}
```
ashleyfae, line 110: I think it would be better like this:

```suggestion
	public function mergeConfig( array $configuration ) {
		foreach ( $configuration as $key => $value ) {
			$this->{$key} = $value;
		}

		return $this;
	}
```

There isn't really any need to have to create a new object since we already have the current one.

(This will of course require reworking your usage in the Extension_Manager.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,489 @@
...
+			</div>
+		</div>
+		<?php
+	}
+
+	/**
+	 * Gets the CSS classes for the single extension card.
+	 *
+	 * @since 2.11.x
+	 * @param \ProductData $product The product data object.
```
ashleyfae, line 139:
```suggestion
	 * @param ProductData $product The product data object.
```

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,489 @@
...
+	 */
+	public function is_plugin_installed( $plugin ) {
+		return array_key_exists( $plugin, $this->get_plugins() );
+	}
+
+	/**
+	 * Whether a given plugin is active or not.
+	 *
+	 * @since 2.11.x
+	 * @param false|string|The product data object. $basename_or_data The path to the main plugin file, eg 'my-plugin/my-plugin.php', or the product data object.
```
ashleyfae, line 481:
```suggestion
	 * @param string|ProductData $basename_or_data The path to the main plugin file, eg 'my-plugin/my-plugin.php', or the product data object.
```

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,199 @@
...
+
+class ExtensionsAPI {
+
+	/**
+	 * Gets the product data from the EDD Products API.
+	 *
+	 * @since 2.11.x
+	 * @param array $body    The body for the API request.
+	 * @param int   $item_id The product ID, if querying a single product.
+	 * @return false|array|\ProductData
```
ashleyfae, line 15:
```suggestion
	 * @return false|array|ProductData
```

includes/admin/extensions/class-extensions-api.php
ashleyfae: I do prefer saving it as an array and converting on retrieval. :+1:

includes/admin/extensions/class-extensions-api.php
ashleyfae: I think this is outdated after https://github.com/awesomemotive/easy-digital-downloads/pull/8993#discussion_r766967065 ?

ashleyfae:
This error needs sorting:

```
[13-Dec-2021 11:42:28 UTC] PHP Warning:  foreach() argument must be of type array|object, bool given in easy-digital-downloads/includes/admin/extensions/product-education/class-email-marketing.php on line 109
[13-Dec-2021 11:42:28 UTC] PHP Warning:  foreach() argument must be of type array|object, bool given in easy-digital-downloads/includes/admin/extensions/product-education/class-email-marketing.php on line 81
```

To reproduce:

1. Delete all option caches.
2. In `ExtensionsAPI::get_all_product_data()` manually set `$request = new \WP_Error()` to simulate an error.
3. Visit the email marketing page.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,199 @@
...
+
+		return $all_product_data;
+	}
+
+	/**
+	 * Gets the product data as needed for the extension manager.
+	 *
+	 * @since 2.11.x
+	 * @param object $item
+	 * @return ProductData
```
ashleyfae, line 138:
```suggestion
	 * @return array
```

includes/admin/extensions/product-education/class-reviews.php
```diff
@@ -0,0 +1,120 @@
...
+
+	/**
+	 * Gets the custom configuration for Reviews.
+	 *
+	 * @since 2.11.x
+	 * @param \EDD\Admin\Extensions\ProductData $product_data The product data object.
+	 * @return array
+	 */
+	protected function get_configuration( \EDD\Admin\Extensions\ProductData $product_data ) {
+
```
ashleyfae, line 56:
```suggestion
```

includes/admin/admin-pages.php
```diff
@@ -279,6 +279,11 @@ function edd_is_admin_page( $passed_page = '', $passed_view = '', $include_non_e
 						$found = true;
 					}
 					break;
+				case 'marketing':
```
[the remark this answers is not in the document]
ashleyfae: Yes I also agree it's horrendous. Something that's also bothered me for months is that `'reports'` is in there twice.

I am in favour of refactoring, but maybe >= 3.0 because I think 3.0 touched this too and the merge conflicts would kill me. :sweat_smile:

ashleyfae:
I think you need to unset the `'timeout'` key in two other places in the `ExtensionsAPI::get_product_data()`.

I'm testing failures. So here's what I did:

Set `$is_stale` to `true` here: https://github.com/awesomemotive/easy-digital-downloads/blob/c1648df25147733a18c753426b7f5365695a11b6/includes/admin/extensions/class-extensions-api.php#L28

This forces it to keep doing the API request.

Set `$request = new \WP_Error()` here: https://github.com/awesomemotive/easy-digital-downloads/blob/c1648df25147733a18c753426b7f5365695a11b6/includes/admin/extensions/class-extensions-api.php#L118

This simulates a failure.

Then I went to the `Marketing` tab. On first refresh, the `Email Marketing` section did not appear, as expected. But when I refreshed again, it did appear. Upon clicking on it, I get this error:

```
PHP Fatal error:  Uncaught TypeError: array_intersect_key(): Argument #1 ($array) must be of type array, int given in easy-digital-downloads/includes/admin/extensions/class-extension-product-data.php:81
```

What's happening was that the marketing class ran this: https://github.com/awesomemotive/easy-digital-downloads/blob/c1648df25147733a18c753426b7f5365695a11b6/includes/admin/extensions/product-education/class-email-marketing.php#L85

`get_product_data()` included the `timeout` in it. So `$item_id` ended up being `'timeout'`. So it passed that to `do_single_extension_card()`, which called `get_product_data( 'timeout' )`, which returned the timeout _timestamp_ (the int in the error message).

You unset the timeout here: https://github.com/awesomemotive/easy-digital-downloads/blob/c1648df25147733a18c753426b7f5365695a11b6/includes/admin/extensions/class-extensions-api.php#L38

But not here: https://github.com/awesomemotive/easy-digital-downloads/blob/c1648df25147733a18c753426b7f5365695a11b6/includes/admin/extensions/class-extensions-api.php#L65

or here: https://github.com/awesomemotive/easy-digital-downloads/blob/c1648df25147733a18c753426b7f5365695a11b6/includes/admin/extensions/class-extensions-api.php#L88

robincornett:
Note: I went back to deleting the option in 5744d5a because I think if we were to remove an extension from our site, and we are querying a tag/category, just overwriting the products with the new data won't remove a deleted product. eg if we removed the Mailchimp extension, we need to make sure it gets removed from here as well.

ashleyfae:
> Note: I went back to deleting the option in [5744d5a](https://github.com/awesomemotive/easy-digital-downloads/commit/5744d5a81bb1941a2a69650ecfee195b3dbb004e) because I think if we were to remove an extension from our site, and we are querying a tag/category, just overwriting the products with the new data won't remove a deleted product. eg if we removed the Mailchimp extension, we need to make sure it gets removed from here as well.

I'm not seeing how that's necessary because it's deleted here: https://github.com/awesomemotive/easy-digital-downloads/blob/d1097d238decef6229bc447da1de00a01b3f3a96/includes/admin/extensions/class-extensions-api.php#L70

But then it's _always_ updated here: https://github.com/awesomemotive/easy-digital-downloads/blob/d1097d238decef6229bc447da1de00a01b3f3a96/includes/admin/extensions/class-extensions-api.php#L90

So no matter what it's always going to be overwritten with `$value`. I'm not seeing any kind of possible `return` between those two points, unless I'm missing something?

robincornett:
Ugh, you are correct--with the added wrinkle of merging data into the stale option before that, that's what I was thinking of, but you are right, we aren't merging at this point.

ashleyfae:
## Testing

### API failures

Testing when there's no option to fall back on. Ensure graceful handling.

- [x] Email marketing
- [x] Reviews
- [x] Recurring
- [x] Invoices

Testing when there is a valid option to fall back on. Ensure stale cache is shown.

- [x] Email marketing
- [x] Reviews
- [x] Recurring
- [x] Invoices

ashleyfae:
Email marketing has gone a bit weird for me I think.

In `ExtensionsAPI::get_product_data()` I manually set `$is_stale = true` and it looks like this (perfect):

Screenshot from 2021-12-14 15-40-39

Then I removed that so it's back to the default (checking the timeout) and it looks like this:

Screenshot from 2021-12-14 15-40-25

includes/admin/extensions/abstract-extension.php
```diff
@@ -0,0 +1,322 @@
...
+			$button['button_text'] = sprintf( __( 'Activate %s', 'easy-digital-downloads' ), $product_data->title );
+		}
+
+		return $button;
+	}
+
+	/**
+	 * Gets the upgrade URL for the button.
+	 *
+	 * @todo add UTM parameters
```
ashleyfae, line 243: This is done now, right?

includes/admin/extensions/product-education/class-wpsmtp.php
```diff
@@ -0,0 +1,228 @@
...
+			// If the plugin is active, but not configured, the button will send them to the setup wizard.
+			$button = $this->get_link_parameters();
+		}
+
+		return $button;
+	}
+
+	/**
+	 * Gets the array of parameters for the link to configure WP Mail SMTP.
+	 * @todo maybe this should go to the settings page instead.
```
robincornett, line 124: @cklosowski should this button go to the setup wizard or to the WPSMTP settings page please? I was looking at their code to see if there was an easy rule to make that determination but not really seeing anything.

The settings page does have a link to the setup wizard so that may make more sense than just launching the wizard. Otherwise, if we are going to launch the wizard then I think the button should say "Launch the WP Mail SMTP Wizard" instead of "Configure..". Just wanted to check with you first.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,495 @@
...
+		if ( ! is_wp_error( $activated ) ) {
+
+			/**
+			 * Fire after plugin activating via the EDD installer.
+			 *
+			 * @since 2.11.x
+			 *
+			 * @param string $plugin_basename Path to the plugin file relative to the plugins directory.
+			 */
+			do_action( 'edd_plugin_activated', $plugin_basename );
```
ashleyfae, line 396: Is this hook needed for anything?

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,201 @@
...
+
+	/**
+	 * Gets all of the product data, either from a tranient or an API request.
+	 *
+	 * @since 2.11.x
+	 * @return object|false
+	 */
+	private function get_all_product_data() {
+		// Possibly all product data is in a transient. If it is, return it.
+		$all_product_data = get_transient( 'edd_all_extension_data' );
```
ashleyfae, line 101: Just curious if there was a reason for this one to use a transient when the other uses options. I think we tend to prefer options at least when making API requests to EDD's site because of annoyances we've had with transients in the past (Software Licensing).

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,201 @@
...
+			$url,
+			array(
+				'timeout'   => 15,
+				'sslverify' => true,
+			)
+		);
+
+		// If there was an API error, set timeout for 1 hour and return false.
+		if ( is_wp_error( $request ) || ( 200 !== wp_remote_retrieve_response_code( $request ) ) ) {
+			return false;
```
ashleyfae, line 123: Wondering if we should still cache this for 1 hour.

The individual products will be cached, but if this fails, then each product will still try again. So if you delete all your options, make sure the transient isn't set, then visit the necessary pages to trigger each of the 4 promos, that will actually make 4 API requests. Once the 4 are made, it won't make any more, but I still feel like we only need to be making 1.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		// Fresh data has been retrieved, so populate with fresh data.
+		$all_product_data = json_decode( wp_remote_retrieve_body( $request ) );
+
+		// Store all of the product data for one hour.
+		set_transient( 'edd_all_extension_data', $all_product_data, HOUR_IN_SECONDS );
```
ashleyfae, line 130: Wondering if this should be longer than an hour. :thinking:

includes/admin/extensions/product-education/class-wpsmtp.php
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	/**
+	 * Register the setting to show the WP SMTP installer if it isn't active.
+	 *
+	 * @param array $settings
+	 * @return array
+	 */
+	public function register_setting( $settings ) {
+		if ( $this->is_smtp_configured() ) {
```
robincornett, line 52: Wondering if this early return should be removed, so that the SMTP field always shows. That would be consistent with Recapture. I put the early return in to be consistent with Reviews/Invoices/Recurring but this is maybe a separate scenario.

includes/admin/extensions/class-extensions-api.php
robincornett: It is cached for an hour--not here, but in lines 48-59. So a failed request should happen for the first product, and then products 2-4 would use the stale data, or no data, depending on what was stored for the option.

includes/admin/extensions/class-extensions-api.php
robincornett: I like transients because I don't have to manage the expiration and I'm a little bit lazy? I can switch if you want, but I like transients, I confess.

includes/admin/extensions/class-extensions-api.php
robincornett: Chris asked me yesterday what was left for this PR and what I needed decisions from him on. I listed all of the "timings" that I've set up and asked him to review and ensure that each is a reasonable time in his mind. Specifically, I told him:

>
> * there is a transient on the `extension_data` request that expires after an hour. initially that was because the settings checks weren't working quite right and I was having four requests all at once. this way if someone goes to the marketing tab, the request for email marketing will also work for Reviews, for example
> * the extension data is saved for one week
> * if an extension data request fails but there is existing data, it's not deleted, but the expiration time is reset for an hour, so that stale data is used instead of having no data
> * if there is no extension data and the request fails then there is a timeout for an hour
> That first one is a transient but the specific extension data is saved as an option.

So I'm totally fine to change this to be longer but maybe we should see what he thinks. He is looking at caching the extension data request on the marketing site side as well.

includes/admin/extensions/class-extension-manager.php
robincornett: I had it in back when I wanted to delete the saved options, and it was in the original code I was using as an example/starting point.

ashleyfae: If we don't have a use case for it right now then I vote we remove it. Can always add it back in later, but once it's in there we can't really take it out. :(

includes/admin/extensions/class-extensions-api.php
ashleyfae: The use case I'm looking at is if there's no stale data at all and the API request fails. In my tests it's then doing 5 API requests. Logic is like this:

- Email Marketing goes "give me product data". There's no saved option, so it calls `get_all_product_data()` and does an API request. This fails, so marketing data is saved as empty for 1 hour. No more marketing API requests.
- Invoices goes "give me product data". There's no saved option for it, so it calls `get_all_product_data()` and does an API request. This fails, so invoices data is saved as empty for 1 hour. No more Invoices-triggered API requests.
- Rinse and repeat for each extension.

So each individual one does get cached for 1 hour, but each extension _individually_ triggers another call to `get_all_product_data()` if their option doesn't already exist.

To test:

1. Delete transient.
2. Delete all options.
3. Right here https://github.com/awesomemotive/easy-digital-downloads/blob/3a97fae5867a1b5fd195ad00cfc73c3684025e79/includes/admin/extensions/class-extensions-api.php#L120 add these lines:

```
error_log('did API request');
$request = new \WP_Error();
```

Then navigate to pages that would trigger this and you'll see it trigger 5 times (assuming all extensions it's advertising are not active).

includes/admin/extensions/class-extensions-api.php
ashleyfae: Okay I'm fine with whatever number he's happy with.

includes/admin/extensions/class-extensions-api.php
robincornett: Okay, see what you think now--I've set it to set the hour long transient as an empty array if the request fails, and updated the conditional checks accordingly (because `get_transient` will return `false` if it doesn't exist).

ashleyfae: Works great now. :raised_hands:

includes/admin/extensions/product-education/class-wpsmtp.php
ashleyfae: I kind of like the early return.

ashleyfae:
@robincornett I think you can pretty confidently change all the `@since 2.11.x` to `@since 2.11.4`.

includes/admin/extensions/class-extension-manager.php
```diff
@@ -0,0 +1,486 @@
+<?php
+
+namespace EDD\Admin\Extensions;
+
+use \EDD\Admin\Pass_Manager;
+use \EDD\Admin\Extensions\ProductData;
```
ashleyfae, line 6:
```suggestion
```

Not used/needed because it's in the same namespace as `ProductData` already.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,205 @@
+<?php
+
+namespace EDD\Admin\Extensions;
+
+use \EDD\Admin\Extensions\ProductData;
```
ashleyfae, line 5:
```suggestion
```

Not used/needed, as it's in the same namespace.

includes/admin/extensions/class-extensions-api.php
cklosowski: My only concern here is the transient, if transients are turned off, we run the risk of sending a lot of traffic our way, even with it restricted to specific pages, it is easy for that to get forgotten in future releases. I'm going to try and setup a caching layer into this so it never touches the server, but even that will expire at some point and require hitting the web backend.

includes/admin/extensions/product-education/class-wpsmtp.php
```diff
@@ -0,0 +1,232 @@
...
+	protected function is_smtp_configured() {
+
+		if ( ! $this->is_smtp_activated() || ! class_exists( '\\WPMailSMTP\\Options' ) ) {
+			return false;
+		}
+
+		$phpmailer = $this->get_phpmailer();
+
+		$mailer             = \WPMailSMTP\Options::init()->get( 'mail', 'mailer' );
+		$is_mailer_complete = empty( $mailer ) ? false : wp_mail_smtp()->get_providers()->get_mailer( $mailer, $phpmailer )->is_mailer_complete();
```
ashleyfae, line 156:
```suggestion
		$is_mailer_complete = ! empty( $mailer ) && wp_mail_smtp()->get_providers()->get_mailer( $mailer, $phpmailer )->is_mailer_complete();
```

Inconsequential, but we can simplify this a bit.
I

includes/admin/extensions/product-education/class-wpsmtp.php
cklosowski: I'm okay with it too

includes/admin/settings/settings-compatibility.php
```diff
@@ -0,0 +1,165 @@
+<?php
+
```
ashleyfae, line 2:
```suggestion
```

I don't like the empty line. :stuck_out_tongue:

includes/admin/settings/settings-compatibility.php
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Settings Compatibility Functions
+ *
+ * For managing settings compatibility in a reorganized settings structure.
+ *
+ * @package     EDD
+ * @subpackage  Settings Compatibility
+ * @copyright   Copyright (c) 2021, Easy Digital Downloads
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       2.11.3
```
ashleyfae, line 12:
```suggestion
 * @since       2.11.4
```

includes/admin/extensions/class-extensions-api.php
robincornett: Okay, I have updated the API class to not use any transients at all. I also adjusted the timings:
- `edd_all_extension_data` has a timeout of 1 hour if the API request fails; 4 hours if it succeeds
- Individual product options have a timeout of 1 hour if the API request fails; 1 week if it succeeds

If the individual product options have "expired", the stale data is merged with an updated 1 hour timeout. I did not do this for the all extension data, just set the timeout.

cklosowski: I like that a lot better, should protect us from rouge checks.

includes/admin/extensions/class-extensions-api.php
```diff
@@ -0,0 +1,222 @@
...
+
+	/**
+	 * Checks whether a given option has "expired".
+	 *
+	 * @since 2.11.4
+	 * @param array|false $option
+	 * @return bool
+	 */
+	private function option_has_expired( $option ) {
+		return ! empty( $option['timeout'] ) && time() > $option['timeout'];
```
ashleyfae, line 220:
```suggestion
		return empty( $option['timeout'] ) || time() > $option['timeout'];
```

Problem I'm having now is that this is returning `false` if `'timeout'` doesn't exist (my option is `false` due to not having nay data yet).

ashleyfae (changes requested):

includes/admin/extensions/class-extensions-api.php
robincornett: I probably should update line 32 to be `if ( $option && ! $is_stale ) {` and the same for the all product data--that would have fixed this, too.

ashleyfae (approved):
Well I've done everything I can think of to break this and it's not breaking. :joy: 

I'm happy with it. Nice work @robincornett !

ashleyfae (approved):
Re-approving this as I just tested on a fresh install using PHP 5.6 and confirmed everything worked.

cklosowski (approved):

PR #8996, opened 2021-11-22
Simplify start & end logic to always use last day of month #8995

### LINKED ISSUE #8995: Incorrect date range when exporting earnings results in invalid data
### Expected behavior
When I export earnings, I expect to get accurate numbers.

### Actual behavior
The numbers can be lower than expected.

Here's an example of what one of the queries looks like:

```mysql
SELECT SUM(meta_value) AS total, COUNT(DISTINCT posts.ID) AS count, posts.post_status AS status
FROM wp_posts AS posts
         INNER JOIN wp_postmeta ON posts.ID = wp_postmeta.post_ID
WHERE posts.post_type IN ('edd_payment')
  AND wp_postmeta.meta_key = '_edd_payment_total'
  AND posts.post_date >= '2021-08-01 00:00:00'
  AND posts.post_date < '2021-09-01 23:59:59'
GROUP BY YEAR(posts.post_date), MONTH(posts.post_date), posts.post_status
ORDER by posts.post_date ASC
```

The problem is this:

`AND posts.post_date < '2021-09-01 23:59:59'`

That actually kicks us into the next month, which is NOT what we want. This ends up tripping up the `GROUP BY`, which can result in having multiple results for one status if you just so happened to get sales on September 1st. For example, here are the results from a site:

| total | count | status |
| ------ | ------- | -------- |
| 15592.8 | 320 | edd_subscription |
| 9524.199999999997 | 130 | publish |
| 2659.35 | 31 | refunded |
| 1113 | 11 | abandoned |
| 641 | 3 | refunded | 
| 315.5 | 8 | edd_subscription |
| 15 | 1 | publish |

Note how some statuses appear more than once. This can cause the export to record `1` "complete" payment, instead of recording more like `130`.

PHP Version:

EDD Version (or branch): 2.11.3

WordPress Version:

ashleyfae:
Fixes #8995

Proposed Changes:
Simplify the start and end date logic. The end date is now always set to the last day of the start date's month (using the `t` date format). So instead of this:

```mysql
  AND posts.post_date >= '2021-08-01 00:00:00'
  AND posts.post_date < '2021-09-01 23:59:59'
```

we should be seeing this:

```mysql
  AND posts.post_date >= '2021-08-01 00:00:00'
  AND posts.post_date <= '2021-08-31 23:59:59'
```

This ensure we only ever work within 1 month.

To account for this, you'll note I also changed the query comparison to `<=` instead of `<`.

To test, I honestly just found it easiest to log this query: https://github.com/awesomemotive/easy-digital-downloads/blob/b48c077c245b1329e0244aa30d5ff3728bedce4c/includes/admin/reporting/export/class-batch-export-earnings-report.php#L265-L274 while running a few exports. Ensure the date ranges make sense.

### DIFF
--- a/includes/admin/reporting/export/class-batch-export-earnings-report.php
+++ b/includes/admin/reporting/export/class-batch-export-earnings-report.php
@@ -249,21 +249,12 @@ class EDD_Batch_Earnings_Report_Export extends EDD_Batch_Export {
 		$data = array();
 
 		$start_date = date( 'Y-m-d 00:00:00', strtotime( $this->start ) );
-
-		if ( $this->count() == 0 ) {
-			$end_date = date( 'Y-m-d 23:59:59', strtotime( $this->end ) );
-		} else {
-			$end_date = date( 'Y-m-d 23:59:59', strtotime( 'first day of +1 month', strtotime( $start_date ) ) );
-		}
+		$end_date   = date( 'Y-m-t 23:59:59', strtotime( $this->start ) );
 
 		if ( $this->step > 1 ) {
-			$start_date = date( 'Y-m-d 00:00:00', strtotime( 'first day of +' . ( $this->step - 1 ) . ' month', strtotime( $start_date ) ) );
-
-			if ( date( 'Y-m', strtotime( $start_date ) ) == date( 'Y-m', strtotime( $this->end ) ) ) {
-				$end_date = date( 'Y-m-d 23:59:59', strtotime( $this->end ) );
-			} else {
-				$end_date = date( 'Y-m-d 23:59:59', strtotime( 'first day of +1 month', strtotime( $start_date ) ) );
-			}
+			$start_timestamp = strtotime( 'first day of +' . ( $this->step - 1 ) . ' month', strtotime( $start_date ) );
+			$start_date      = date( 'Y-m-d 00:00:00', $start_timestamp );
+			$end_date        = date( 'Y-m-t 23:59:59', $start_timestamp );
 		}
 
 		if ( strtotime( $start_date ) > strtotime( $this->end ) ) {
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/export/class-batch-export-earnings-report.php
```diff
@@ -249,21 +249,12 @@ public function get_data() {
 		$data = array();
 
 		$start_date = date( 'Y-m-d 00:00:00', strtotime( $this->start ) );
-
-		if ( $this->count() == 0 ) {
-			$end_date = date( 'Y-m-d 23:59:59', strtotime( $this->end ) );
-		} else {
-			$end_date = date( 'Y-m-d 23:59:59', strtotime( 'first day of +1 month', strtotime( $start_date ) ) );
-		}
+		$end_date   = date( 'Y-m-t 23:59:59', strtotime( $this->start ) );
```
robincornett, line 252:
```suggestion
		$end_date   = date( 'Y-m-t 23:59:59', strtotime( $this->end ) );
```

ashleyfae: `start` is actually correct here. (Unless you're seeing an error in the data?)

This `get_data()` method is meant to get records one month at a time. So if you select the range:

Jan 2021
to
March 2021

Then `$this->start` will be January 2021, and `$this->end` will be March 2021. However, `get_data()` only gets one month at a time. We're interested in data for the month of January only. So the local `$start_date` should be `2021-01-01 00:00:00` and the local `$end_date` should be `2021-01-31 23:59:59` -- not `2021-03-31 23:59:59`

robincornett: No, looking at it again, your way is getting correct data--I guess I am just confused and/or need more coffee.

robincornett (approved):

PR #9000, opened 2021-11-24
Add webp to supported file, image types

### LINKED ISSUE #8999: Support webp images in the allowed file types of .htaccess restrictions.
## Enhancement Request
Currently our `.htaccess` file allows access access to the following files as not protected:
`'jpg', 'jpeg', 'png', 'gif', 'mp3', 'ogg'`

We should add `webp` images to this. This needs to also support `png.webp`.

### Justification or use case
With adoption of `webp` images on the rise, we need to make it so these images can be viewed in page content.

robincornett:
Fixes #8999

Proposed Changes:
1. Adds `webp` to `edd_protected_directory_allowed_filetypes`.
2. Updates `edd_string_is_image_url` to include `webp` and condense logic.
3. Adds `webp` to `edd_get_file_ctype`.
4. Adds `webp` assertion to `test_string_is_image_url`.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -148,21 +148,18 @@ function edd_string_is_image_url( $str ) {
 	$ext = edd_get_file_extension( $str );
 
 	switch ( strtolower( $ext ) ) {
-		case 'jpg';
-			$return = true;
-			break;
-		case 'png';
-			$return = true;
-			break;
-		case 'gif';
-			$return = true;
+		case 'jpg':
+		case 'png':
+		case 'gif':
+		case 'webp':
+			$is_image = true;
 			break;
 		default:
-			$return = false;
+			$is_image = false;
 			break;
 	}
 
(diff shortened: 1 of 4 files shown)


includes/misc-functions.php
```diff
@@ -148,21 +148,18 @@ function edd_string_is_image_url( $str ) {
...
+			$is_image = true;
 			break;
 		default:
-			$return = false;
+			$is_image = false;
 			break;
 	}
 
-	return (bool) apply_filters( 'edd_string_is_image', $return, $str );
+	return (bool) apply_filters( 'edd_string_is_image', $is_image, $str );
```
ashleyfae, line 162: Love the changes you've made here.

Given that we only really care about yes or no, maybe we should refactor this some more to just set up an array of image extensions, then check if `in_array()` ?

robincornett: Works for me. I also went ahead and added `jpeg` to the image extensions.

ashleyfae (approved):

PR #9002, opened 2021-11-24
Refactor edd_get_settings_tabs

### LINKED ISSUE #9001: Consider refactoring edd_get_settings_tabs
I would like to consider refactoring the `edd_get_settings_tabs` function, mainly to not use the `edd_get_registered_settings` function.

### Justification or use case
I'd like to use the `edd_get_settings_tabs` function elsewhere to check if a tab exists (we are adding a new one for product education) and I can't because there are varying levels of errors that happen, if Reviews is active, or if the Amazon gateway is active. From what I can tell, the issue is because of calling the `edd_get_registered_settings` function, which also feels expensive.

I have pondered a couple of thoughts about how to tinker with this. One would be, instead of getting _all_ of the settings in order to see if there are settings registered for extensions or licenses, to replicate their individual `apply_filters` calls and just check for them:

```php
	$extensions = apply_filters( 'edd_settings_extensions', array() );
	if ( ! empty( $extensions ) ) {
		$tabs['extensions'] = __( 'Extensions', 'easy-digital-downloads' );
	}
	$licenses = apply_filters( 'edd_settings_licenses', array() );
	if ( ! empty( $licenses ) ) {
		$tabs['licenses'] = __( 'Licenses', 'easy-digital-downloads' );
	}
```
This feels a little more viable approach because those are the only two things that are being checked by pulling in the entire `$settings` array.

Another thought I had was to make a new function called `edd_get_default_settings_tabs` and have that return _just_ the array of EDD registered tabs, and then this function can pull in that array and do the extensions/licenses check and apply its filters as it does now. Then I could use the new function to check for the tab I want to find.

Personally, long term I would love to separate the arrays in `edd_get_registered_settings` into possibly separate views type files, maybe, or at least just get the long array of arrays out of the main register-settings.php file because I find it overwhelming...but setting them up as separate files, I could use an `include` to pull just the extensions and licenses arrays to add those tabs.

robincornett:
Fixes #9001

Proposed Changes:
1. Do not get all of the registered settings within `edd_get_settings_tabs`.
2. Update `edd_options_page`:
    - Remove duplicate `edd_get_settings_tab_sections`.
    - Remove duplicate `edd_get_settings_tabs`.
    - Evaluate the settings tabs against the registered settings, which have already been pulled, and remove any tabs which have no settings.

### DIFF
--- a/includes/admin/settings/display-settings.php
+++ b/includes/admin/settings/display-settings.php
@@ -55,10 +55,9 @@ function edd_options_page() {
 	$all_settings = edd_get_registered_settings();
 
 	// Remove tabs that don't have settings fields.
-	$tabs_with_settings = array_keys( $all_settings );
-	foreach ( $tabs_with_settings as $tab_with_settings ) {
-		if ( empty( $all_settings[ $tab_with_settings ] ) ) {
-			unset( $settings_tabs[ $tab_with_settings ] );
+	foreach ( array_keys( $settings_tabs ) as $settings_tab ) {
+		if ( empty( $all_settings[ $settings_tab ] ) ) {
+			unset( $settings_tabs[ $settings_tab ] );
 		}
 	}
 
--- a/includes/admin/settings/display-settings.php
+++ b/includes/admin/settings/display-settings.php
@@ -50,11 +50,17 @@ function edd_options_page() {
 		$key = key( $sections );
 	}
 
-	$registered_sections = edd_get_settings_tab_sections( $active_tab );
-	$section             = isset( $_GET['section'] ) && ! empty( $registered_sections ) && array_key_exists( $_GET['section'], $registered_sections ) ? sanitize_text_field( $_GET['section'] ) : $key;
+	$section = isset( $_GET['section'] ) && ! empty( $sections ) && array_key_exists( $_GET['section'], $sections ) ? sanitize_text_field( $_GET['section'] ) : $key;
 
-	// Unset 'main' if it's empty and default to the first non-empty if it's the chosen section
 	$all_settings = edd_get_registered_settings();
+
+	// Remove tabs that don't have settings fields.
+	$tabs_with_settings = array_keys( $all_settings );
+	foreach ( $tabs_with_settings as $tab_with_settings ) {
+		if ( empty( $all_settings[ $tab_with_settings ] ) ) {
+			unset( $settings_tabs[ $tab_with_settings ] );
+		}
+	}
 
 	// Let's verify we have a 'main' section to show
 	$has_main_settings = true;
@@ -75,6 +81,7 @@ function edd_options_page() {
 	}
 
 	$override = false;
+	// Unset 'main' if it's empty and default to the first non-empty if it's the chosen section
 	if ( false === $has_main_settings ) {
 		unset( $sections['main'] );
 
@@ -95,7 +102,7 @@ function edd_options_page() {
 		<h1><?php esc_html_e( 'Settings', 'easy-digital-downloads' ); ?></h1>
 		<h2 class="nav-tab-wrapper">
 			<?php
-			foreach ( edd_get_settings_tabs() as $tab_id => $tab_name ) {
+			foreach ( $settings_tabs as $tab_id => $tab_name ) {
 				$tab_url = add_query_arg( array(
 					'settings-updated' => false,
 					'tab'              => $tab_id,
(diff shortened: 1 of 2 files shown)


includes/admin/settings/display-settings.php
```diff
@@ -50,12 +50,17 @@ function edd_options_page() {
 		$key = key( $sections );
 	}
 
-	$registered_sections = edd_get_settings_tab_sections( $active_tab );
-	$section             = isset( $_GET['section'] ) && ! empty( $registered_sections ) && array_key_exists( $_GET['section'], $registered_sections ) ? sanitize_text_field( $_GET['section'] ) : $key;
+	$section = isset( $_GET['section'] ) && ! empty( $sections ) && array_key_exists( $_GET['section'], $sections ) ? sanitize_text_field( $_GET['section'] ) : $key;
```
ashleyfae, line 53:
```suggestion
	$section = ! empty( $_GET['section'] ) && ! empty( $sections[ $_GET['section'] ] ) ? sanitize_text_field( $_GET['section'] ) : $key;
```

1. I prefer `! empty( $_GET['section'] )`  because we also don't want to end up with an empty one anyway.
2. Refactored the second conditional as that allows us to drop down from 3 checks to 2, which feels a bit simpler to me.

includes/admin/settings/register-settings.php
```diff
@@ -1364,26 +1364,17 @@ function edd_sanitize_html_class( $class = '' ) {
  */
 function edd_get_settings_tabs() {
```
ashleyfae, line 1365: Wonder if we should add a `@since` to this docblock to note that as of this version, empty tabs are filtered out on display.

cklosowski:
Have we tested this with a number of extensions including some 3rd party ones? I'm somewhat weary of changing out settings registration in a minor point release, but if we're positive we don't break any backwards compatibility with registering and saving settings we can move forward. We've just had some challenges with extensions integrating with settings before, which is why I think this is risky in a minor.

ashleyfae:
We're not changing anything about settings registration or saving. The only change of note is in `edd_get_settings_tabs()` and what that returns.

Previous behaviour:

- Only add the "Extensions" and "Licenses" tabs if we have settings for those. So in the event that there are no extensions on the site, `'extensions'` would not be a key returned in `edd_get_settings_tabs()`.
- New behaviour: Those tabs are always available in `edd_get_settings_tabs()`. If there are no extension settings, then `'edd_get_settings_tabs'` will still be returned from `edd_get_settings_tabs()` when previously it wouldn't have. We now filter out empty ones further down the chain during display.

This has no impact whatsoever on actual settings being registered or how they're saved. It's just a minor adjustment of including "Extensions" and "Licenses" as tabs in `edd_get_settings_tabs()` all the time. I'm not aware of any extensions that would rely on those **not** being in that array. I'd be more concerned if we were _removing_ something that used to always be there.

robincornett:
I proposed updating `edd_get_settings_tabs` in general because it would be nice to be able to use it to evaluate the array of settings tabs and make decisions based on whether they're registered. With it calling `edd_get_registered_settings`, which has a lot of filters, solely to maybe add the licensing/extensions tabs, it's expensive to run, and the function cannot be used predictably.

I want to use it in various email marketing extensions and Reviews, but doing so causes fatal errors when Reviews is active or when the Amazon gateway is active. Removing the settings check from this function allows me to use it, and since only EDD core currently calls the settings tabs function as part of outputting the options screens, it feels fairly safe. The options screen already retrieves the settings, so it's a little redundant to have it here as well, I think.

It looks like none of our extensions call the `edd_get_settings_tabs` function, and only FES runs the `edd_settings_tab` filter, which wouldn't be affected by this change. I currently have active with this branch with no issues:
- Commissions
- FES
- Reviews (with the updated code which depends on this change)
- ConvertKit (with the updated code which depends on this change)
- Aelia Currency Switcher (sets up its own settings screen, doesn't use EDD)
- Advanced Reports
- All Access
- Amazon S3
- Conditional Success Redirects
- EU VAT (Barn2)
- Git Updater
- Hide Download
- Invoices
- PayPal Commerce Pro
- Recurring
- Simple Shipping
- Software Licensing
- Stripe Pro

Not an exhaustive list, I know, and not thoroughly thoroughly tested, but it seems like EDD itself and FES are the only two likely to be impacted by the change, and they are not.

cklosowski:
Awesome, thank you for the info and comprehensive list @robincornett and @ashleyfae. Just knowing the history of the settings API in EDD, i'm always a little weary of changes like this, but sounds like we've run it through a lot of paces.

:+1:

ashleyfae (approved):

PR #9020, opened 2021-12-15
Append to log using FILE_APPEND flag #9019

### LINKED ISSUE #9019: Debug Log can cause 500 errors when it grows too large.
### Expected behavior
I expect the EDD Debug log to not cause performance issues on a store.

### Actual behavior
When the debug log is left on, it can cause logs to grow to large amounts, causing checkouts and actions on the site to slow down, due to having to open the debug log into memory for writing. Eventually this can cause memory issues and eventually 500 errors.

This is hard to 'reproduce' as it requires a large amount of data and a server configuration that runs out of memory. In the example I had, the debug log file was over 200MB.

PHP Version: All

EDD Version (or branch): 2.8.7+

WordPress Version: All

## Requirements
- [ ] #9056

## Other ideas that we may punt to later versions.
- Output debug log file size on the debug log view. This can help identify when your debug log file is increasing in size.
- Restrict Debug logging to only be enabled for a certain time? Auto deactivate after 24 hours? (Allow custom timeframes?)
- Rotate log file or automatically truncate after a certain file size.

@ashleyfae I'm really only requiring that first admin notice. The rest could be useful, but if we want, we can kick those to a later version of EDD Core.

I'd like to have this maybe go in a version after 2.11.4?

Possible wording for the admin notice:
"Easy Digital Downloads debug logging is enabled. It is not advisable to leave this running at all times, and should only be used when actively trying to solve a problem.  <view link>View Debug Log</view link> <link to disable debug logging>Disable Debug Logging</link to disable debug logging>"

We should make the link to disable debug logging actually disable it, instead of sending the user to the Misc Settings to then make 2 more clicks to uncheck the box and then save settings.

ashleyfae:
Fixes #9019

~~Branched against 2.11.4 for now.~~

Proposed Changes:
1. Use `FILE_APPEND` flag to append message instead of loading the file into memory.

To-do:
1. ~~Admin notice.~~ Edit: Opened a separate issue for this

Also see https://github.com/awesomemotive/easy-digital-downloads/pull/9021


ashleyfae:
I decided to have this standalone issue for fixing what I consider the main problem here. I opened a separate issue for the admin notice.

To test:

Basically just enable debug mode, do some stuff that will trigger logs, and ensure logs still get appended to the file correctly.

robincornett (approved):
- [x] Everything continued to work with my existing log, and data was added.
- [x] Clearing the log work, and then new data was added.

PR #9021, opened 2021-12-15
Rotate logs when they get large #9019

### LINKED ISSUE #9019: Debug Log can cause 500 errors when it grows too large.

ashleyfae:
Fixes #9019

Proposed Changes:
1. Rotates logs when they reach 5 MB.

This is in a separate PR from https://github.com/awesomemotive/easy-digital-downloads/pull/9020 in case we want to add it later / separately.

### DIFF
--- a/includes/class-edd-logging.php
+++ b/includes/class-edd-logging.php
@@ -51,8 +51,8 @@ class EDD_Logging {
 	public function setup_log_file() {
 
 		$upload_dir       = wp_upload_dir();
-		$this->filename   = wp_hash( home_url( '/' ) ) . '-edd-debug.log';
-		$this->file       = trailingslashit( $upload_dir['basedir'] ) . $this->filename;
+		$this->filename   = $this->make_log_file_name();
+		$this->file       = $this->get_log_directory_path() . $this->filename;
 
 		if ( ! is_writeable( $upload_dir['basedir'] ) ) {
 			$this->is_writable = false;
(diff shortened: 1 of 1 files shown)


cklosowski:
Lets go ahead and leave this one off the 2.11.5 release. If we are going to rotate, I want to be able to have the historical UI/UX to view and manage rotated log files.

PR #8457, opened 2021-03-08
Update EDD_Batch_Import to use str_getcsv

robincornett:
Fixes #8149, #8453

Proposed Changes:
1. Removes the `parseCSV` library.
2. Updates the batch import classes to use the CSV data in the array format instead of the object.
 
Note that if we need to make this completely backwards compatible, I think we just need to set `$this->csv` as an empty `stdClass` and then set the data property to be the returned CSV data, and update all usages accordingly. I searched and did not find any extensions using the `$this->csv` object outside of EDD core, but if we need to account for that, then it's pretty straightforward to make the changes.

PR #9034, opened 2021-12-29
Update batch importer to use str_getcsv instead of an external library

### LINKED ISSUE #9033: Update EDD_Batch_Import to use str_getcsv
This is related to #8149, which we've fixed already in 3.0, but we agreed today that we should try to bring those changes into 2.11, with the exception of actually removing the ParseCSV library file.

robincornett:
Fixes #9033

Proposed Changes:
This is nearly identical to #8457, with a couple specific differences:
1. Per Chris, we are not removing the ParseCSV library file in case any extensions may be trying to load it. It is removed in 3.0.
2. The since version on the `get_csv_file` method is marked as 2.11.5 instead of 3.0.
3. When importing payments, I encountered a job-stopping error when `$customer` and `$email` were not defined, and then `$customer_by_email`. I think that method is difficult to read but I did the minimum I could do to prevent the errors.

### DIFF
--- a/includes/admin/import/class-batch-import-downloads.php
+++ b/includes/admin/import/class-batch-import-downloads.php
@@ -75,11 +75,11 @@ class EDD_Batch_Downloads_Import extends EDD_Batch_Import {
 			unlink( $this->file );
 		}
 
-		if( ! $this->done && $this->csv->data ) {
+		if( ! $this->done && $this->csv ) {
 
 			$more = true;
 
-			foreach( $this->csv->data as $key => $row ) {
+			foreach( $this->csv as $key => $row ) {
 
 				// Skip all rows until we pass our offset
 				if( $key + 1 <= $offset ) {
--- a/includes/admin/import/class-batch-import-payments.php
+++ b/includes/admin/import/class-batch-import-payments.php
@@ -546,7 +546,10 @@ class EDD_Batch_Payments_Import extends EDD_Batch_Import {
 
 			foreach( $downloads as $key => $download ) {
 
-				$d   = (array) explode( '|', $download );
+				$d = (array) explode( '|', $download );
+				if ( ! array_key_exists( 1, $d ) ) {
+					continue;
+				}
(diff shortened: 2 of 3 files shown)


ashleyfae (commented):
Code looks good.

## Testing

- [ ] Download import
- [ ] Payments import

ashleyfae:
Unrelated to your PR, but I'm getting a fatal error when importing payments:

> PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string - int in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php:566

Can you sort that while we're here?

robincornett:
I haven't gotten that error myself but made an educated guess as to how to sort it.

ashleyfae:
Sorry a few more :sweat_smile: 

```
[03-Jan-2022 12:28:01 UTC] PHP Warning:  Undefined array key 1 in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php on line 550
[03-Jan-2022 12:28:01 UTC] PHP Warning:  Undefined array key 1 in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php on line 552
[03-Jan-2022 12:28:01 UTC] PHP Warning:  Undefined array key 1 in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php on line 558
[03-Jan-2022 12:28:01 UTC] PHP Warning:  Undefined array key 1 in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php on line 550
[03-Jan-2022 12:28:01 UTC] PHP Warning:  Undefined array key 1 in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php on line 552
[03-Jan-2022 12:28:01 UTC] PHP Warning:  Undefined array key 1 in easy-digital-downloads/includes/admin/import/class-batch-import-payments.php on line 558
```

I have a few rows where the "Products (Raw)" column is blank. I think that's probably from where I've done wallet purchases, which are just a fee and no download?

ashleyfae (approved):

PR #9038, opened 2022-01-04
Update download meta saving, bundled products

### LINKED ISSUE #9037: PHP Notice in download metabox
### Expected behavior
When I edit a download, I expect to not have any PHP notices.

### Actual behavior
I'm editing a private download on our marketing site (ID 1610), mainly because I wanted to look at the download for an old old order and make sure the migration went through correctly. I'm getting this PHP Notice:

> PHP Notice:  Undefined offset: 0 in \wp-content\plugins\easy-digital-downloads\includes\admin\downloads\metabox.php on line 631 (line 643 in release/3.0)

No idea how to make it happen, but it looks like the download has a value for `_edd_bundled_products`, even though it should not, because it's not a bundle. Then `edd_get_bundle_pricing_variations()` in line 630 is returning an empty array, and line 631 is assuming that there will be at least one item in the array--in this case, there is not.

PHP Version: 7.4.1
EDD Version (or branch): master, release/3.0
WordPress Version: 5.8.2

It seems like the first thing would be to not use the first array item if the array is empty. That would get rid of the notice. However, it seems like bundled products should maybe not be saved to the download at all if there are no bundled products--eg I should not have database rows like this:

image

The download I'm actually editing has a meta value of `a:1:{i:0;s:3:"518";}` but I believe that's only because 518 is the first download in the list of downloads and there isn't an empty option for the product selector. Which is maybe an issue as well.

robincornett:
Fixes #9037 (note: not currently milestoned)

Proposed Changes:
1. e4e9259 is the minimum needed to fix the issue/remove the PHP notice--this just checks if `$price_assignments[0]` exists before using it.
2. I've committed the other two changes separately, but both are required to truly fix this issue (and likely others):
    - ffa3594 updates the `edd_sanitize_bundled_products_save` to remove both the current product ID and the `0` product ID from the products array.
    - 6d753c5 updates the `edd_download_meta_box_save` condition to only update the post meta if `$new` has value after it's been filtered. If $new has no value, then the post meta will be deleted. I _think_ this makes sense to do but it's a fix with potentially far reaching consequences. Without this, the fix above would save the bundled product meta as `a:0:{}` instead of `a:1:{i:0;s:1:"0";}`. That, at least, would be consistent with the `edd_variable_prices` and `edd_download_files` meta keys, which I think also should not be stored if they are an empty array.


ashleyfae (commented):
I still need to test this but the code looks good and makes sense.

ashleyfae:
## Testing

- [ ] Confirmed saving a download without bundled products does not save a meta key.
- [ ] Confirmed saving a download with bundled products still saves correctly.

ashleyfae (approved):

PR #9043, opened 2022-01-06
Check if current product section can show before hiding submit button

### LINKED ISSUE #9042: EDD 2.11.4 incorrectly hides Recurring settings submit button
### Expected behavior
When I have the Recurring plugin installed and activated, I expect to see the submit button on the Recurring settings page.

### Actual behavior
It's not there.

Screenshot from 2022-01-06 12-36-58

1) Be on EDD 2.11.4.
2) Install and activate Recurring.
3) Go to Downloads > Settings > Payments > Recurring Payments.
4) Note the lack of submit button.

PHP Version:

EDD Version (or branch): 2.11.4

WordPress Version:

The problem is here: https://github.com/awesomemotive/easy-digital-downloads/blob/2.11.4/includes/admin/extensions/product-education/class-recurring.php#L42
https://github.com/awesomemotive/easy-digital-downloads/blob/2.11.4/includes/admin/extensions/abstract-extension.php#L339-L343

That `hide_submit_button()` method needs conditionals to only execute if the card is also going to display.

The reason the submit button appears for Invoices is because the section ID is actually incorrect in product education. It's different from the one the actual Invoices extension uses, which means product education is only impacting the "fake" section it sets up and not the real one.

Invoices hooks in here https://github.com/awesomemotive/easy-digital-downloads/blob/2.11.4/includes/admin/extensions/product-education/class-invoices.php#L42 via `edd_settings_tab_top_gateways_invoices`. But the hook the real extension uses is: `edd_settings_tab_top_gateways_edd-invoices`.

robincornett:
Fixes #9042
Currently based off of master.

Proposed Changes:
1. Adds `can_show_product_section` check to the `hide_submit_button` method.

### DIFF
--- a/includes/admin/extensions/abstract-extension.php
+++ b/includes/admin/extensions/abstract-extension.php
@@ -337,6 +337,9 @@ abstract class Extension {
 	 * @return void
 	 */
 	public function hide_submit_button() {
+		if ( ! $this->can_show_product_section() ) {
+			return;
+		}
 		?>
 		<style>p.submit{display:none;}</style>
 		<?php


mihaijoldis (approved):
Working as intended!

ashleyfae (approved):
- [x] Confirmed submit buttons do not appear when promo is showing.
- [x] Confirmed submit buttons **do** appear when extensions are activated.

PR #9051, opened 2022-01-10
Use currency from current order #9050

### LINKED ISSUE #9050: 3.0 - Order Admin View - Use decimal settings for order's currency
See https://github.com/awesomemotive/edd-multi-currency/issues/29

In short:

- Say your site currency is set to USD.
- You have an order that was made using HUF (a zero decimal currency).
- You view that HUF order.
- You expect it to show no decimals, but it does show decimals because it's using the settings for the global currency instead of the current order. We can change this easily.

ashleyfae:
Fixes #9050 

## Before

Screenshot from 2022-01-10 14-24-42

## After

Screenshot from 2022-01-10 14-24-32

Proposed Changes:
1. If we're on the individual order page, use the currency for that order.
2. Additionally, pass the currency into `edd_currency_decimal_filter()` so we get the decimal settings for the current currency.

### DIFF
--- a/templates/shortcode-receipt.php
+++ b/templates/shortcode-receipt.php
@@ -99,7 +99,7 @@ do_action( 'edd_order_receipt_before_table', $order, $edd_receipt_args );
 					?>
 					<tr>
 						<td><?php echo esc_html( $label ); ?></td>
-						<td><?php echo esc_html( edd_currency_filter( edd_format_amount( edd_negate_amount( $order_discount->total ) ), $order->currency ) ); ?></td>
+						<td><?php echo esc_html( edd_display_amount( edd_negate_amount( $order_discount->total ), $order->currency ) ); ?></td>
 					</tr>
 					<?php
 				}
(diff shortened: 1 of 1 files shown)


robincornett (commented):
This is working well for orders in the admin, but I think we need to check purchase receipts/history as well. It looks like multi-currency is maybe overriding on the front end in some places so you'll want to deactivate that to check core.

I think we need to update some of the original payment functions, eg `edd_payment_subtotal` is used in the purchase receipt. This works for the subtotal:

```php
function edd_payment_subtotal( $order_id = 0 ) {
	$subtotal      = edd_get_payment_subtotal( $order_id );
	$currency_code = edd_get_payment_currency_code( $order_id );

	return edd_currency_filter( edd_format_amount( $subtotal, edd_currency_decimal_filter( 2, $currency_code ) ), $currency_code );
}
```

ashleyfae:
@robincornett Slightly different fix, but same idea: https://github.com/awesomemotive/easy-digital-downloads/pull/9051/commits/100dd571234239f75cdcaf5ead324d35f243b9dc

templates/shortcode-receipt.php
```diff
@@ -324,7 +324,7 @@
 					<td><?php echo esc_html( $item->quantity ); ?></td>
 				<?php } ?>
 				<td>
-					<?php echo esc_html( edd_currency_filter( edd_format_amount( $item->total ), $order->currency ) ); ?>
+					<?php echo esc_html( edd_display_amount( $item->total, $order->currency ) ); ?>
```
robincornett, line 327: We also need to update lines for adjustments please:
- 127 (fees)
- 102 (discounts)
- 154 (credits)

robincornett (approved):

PR #9052, opened 2022-01-10
Generalize condition so it can apply to refund UI as well #9050

### LINKED ISSUE #9050: 3.0 - Order Admin View - Use decimal settings for order's currency

ashleyfae:
Fixes #9050

Proposed Changes:
1. Make the conditional a bit less strict so it can apply to the refund UI as well.

### DIFF
--- a/includes/admin/payments/class-refund-items-table.php
+++ b/includes/admin/payments/class-refund-items-table.php
@@ -367,6 +367,8 @@ class Refund_Items_Table extends List_Table {
 
 			if ( $order ) {
 				$currency_decimals = edd_currency_decimal_filter( 2, $order->currency );
+			} else {
+				$currency_decimals = 2;
 			}
 		}
 
--- a/includes/admin/payments/class-refund-items-table.php
+++ b/includes/admin/payments/class-refund-items-table.php
@@ -296,7 +296,17 @@ class Refund_Items_Table extends List_Table {
 				}
 				?>
 				<span class="edd-amount-control__input">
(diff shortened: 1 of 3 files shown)


robincornett (commented):
The initial data in the modal is not correct with this still:

image

Once checkboxes are checked, the updated numbers are okay (note max numbers underneath are unchanged):

image

Thinking maybe the `format_currency` method [needs to be updated here maybe](https://github.com/awesomemotive/easy-digital-downloads/blob/issue/9050-2/includes/admin/payments/class-refund-items-table.php#L221)?

robincornett:
Actually there are four uses of `edd_format_amount` in this file (Refund_Items_Table) which I think all need to be updated, possibly with a helper method similar to the `get_currency_symbol` method to get the currency decimals. I have something working and can commit it if you like.

ashleyfae:
> Actually there are four uses of `edd_format_amount` in this file (Refund_Items_Table) which I think all need to be updated, possibly with a helper method similar to the `get_currency_symbol` method to get the currency decimals. I have something working and can commit it if you like.

Go for it. :+1:

includes/admin/payments/class-refund-items-table.php
```diff
@@ -339,6 +351,28 @@ private function get_quantity_column( $item, $column_name, $item_id, $object_typ
...
+	 */
+	private function get_order_currency_decimals( $order_id ) {
+		static $currency_decimals = null;
+
+		if ( is_null( $currency_decimals ) ) {
+			$order = edd_get_order( $order_id );
+
+			if ( $order ) {
+				$currency_decimals = edd_currency_decimal_filter( 2, $order->currency );
+			}
```
ashleyfae, line 370: @robincornett Maybe we should have an `else` here that sets it to `2`. Unlikely to wind up there, but just in case.

robincornett: Yes, I agree. Committed.

ashleyfae:
https://github.com/awesomemotive/easy-digital-downloads/pull/9052/commits/63183cabf44aa88461b63fe0de45a4b76ecf83c9 adds a new `$currency` parameter to `edd_format_amount()`.

It wasn't working as expected before, because where you were passing the currency code was actually meant to be a boolean -- not the actual currency. This change fixes that.

robincornett (approved):

PR #9058, opened 2022-01-19
Add property tags to docblock #7248

### LINKED ISSUE #7248: Add @property tags to classes which use __get to access private methods.
Classes like EDD_Payment have tons of private properties that are expected to be directly accessed via magic __get methods.

Unfortunately this is breaking PHP Storm & other IDE inspections

image

Adding `@property array cart_details` to the `EDD_Payment` class comment resolves this correctly and also
- shows up for auto complete when you type `$payment->`
- alerts our IDE of what that property type will be, in this case an `array`

image

image

Heavily suggest investing a minimal amount of time into adding property comments to each of the main EDD core classes.

Happy to submit a PR myself with them slowly over time, but you guys could probably do it in a half hour.

ashleyfae:
Fixes #7248

Proposed Changes:
1. Adds property tags to the `EDD_Discount` class.
2. Adds property tags to the `EDD_Payment` class.

### DIFF
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -16,6 +16,25 @@ if ( ! defined( 'ABSPATH' ) ) exit;
  * EDD_Discount Class
  *
  * @since 2.7
+ *
+ * @property int    $ID
+ * @property string $name
+ * @property string $code
+ * @property string $status
+ * @property string $type
+ * @property float  $amount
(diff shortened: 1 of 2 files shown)


robincornett (approved):

spencerfinnell (approved):
🎉 Thanks @ashleyfae!

PR #9059, opened 2022-01-19
Add an admin notice if logging is enabled

### LINKED ISSUE #9056: Add a non-dismissible admin-notice when the EDD Debug Log is enabled.

robincornett:
Fixes #9056

Proposed Changes:
1. Adds a non-dismissible admin notice if the EDD debug log setting is enabled.
2. The notice does not appear if only the `EDD_DEBUG_MODE` constant is set.
3. The notice includes a link to view the debug log.
4. It also includes a button to clear the log and disable the setting.


ashleyfae (commented):
Can we not show the notice if `WP_DEBUG` is enabled?

Also I feel like we need some styling on the different action links. Right now I find it hard to differentiate them. They sort of blend together and look like one button (until you hover over one).

Screenshot from 2022-01-20 11-16-25

Bonus points if you ditch the jQuery. :stuck_out_tongue:

ashleyfae:
Sorry I'm being weird about these buttons lol. Wondering if I like this better:

Screenshot from 2022-01-20 13-57-23

Also, I'm thinking of adjusting the behaviour a bit when clicking the button to disable logging. Right now it does this:

Screenshot from 2022-01-20 13-59-03

What if instead:

1. Replace the _entire_ notice contents with the message.
2. Wait a few seconds.
3. Then remove the notice completely.

I at least want to do the first one, I think.

ashleyfae:
## Testing

- [x] `WP_DEBUG` off -- notice does not appear if debug logging not enabled.
- [x] `WP_DEBUG` off -- notice appears if debug logging is enabled.
- [x] `WP_DEBUG` on -- notice does not appear if debug logging is enabled.
- [x] Clicking button to disable disables debug mode.

includes/admin/class-edd-notices.php
```diff
@@ -310,6 +313,57 @@ public function show_notices() {
...
+	}
+
+	/**
+	 * Disables the debug log setting and deletes the existing log file.
+	 *
+	 * @since 2.11.5
+	 * @return void
+	 */
+	public function edd_disable_debugging() {
+		edd_update_option( 'debug_mode', false );
```
ashleyfae, line 361: Can you add a capability check here for `current_user_can( 'manage_shop_settings' )` ?

Also let's add a nonce.

includes/admin/class-edd-notices.php
```diff
@@ -310,6 +313,62 @@ public function show_notices() {
...
+	 * @return void
+	 */
+	private function show_debugging_notice() {
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+			return;
+		}
+		if ( ! current_user_can( 'manage_shop_settings' ) ) {
+			return;
+		}
+		if ( ! edd_get_option( 'debug_mode', false ) ) {
```
ashleyfae, line 330:
```suggestion
		if ( ! edd_is_debug_mode() ) {
```

Should this be used instead? Because what you have now isn't checking the `EDD_DEBUG_MODE` constant, but `edd_is_debug_mode()` will be.

ashleyfae: Never mind! Misunderstood something.

ashleyfae (approved):

cklosowski:
One last change, since we are performing 2 actions (disabling logging _and_ clearing the log file) the message should reflect that, right now it just mentions that we've disabled logging. It does not mention that we've also cleared the debug log.

cklosowski (changes requested):
We need to update the messaging from `Logging has been disabled` to `Debug log has been cleared and logging has been disabled` to better reflect what actions we took.

ashleyfae (approved):

cklosowski (approved):

PR #9071, opened 2022-02-09
Update stats get_earnings method to correctly query for a single download

### LINKED ISSUE #9070: 3.0 - EDD()->payment_stats->get_earnings() returns incorrect value when a download is specified
### Expected behavior
When I use `EDD()->payment_stats->get_earnings()` for a specific download, I expect to get the total earnings for the download over my specified date range.

### Actual behavior
When a download is specified, only one row is queried/returned from the database, so the total is incorrect.

1. Generate multiple sales for the current year for a specific download.
2. Run `EDD()->payment_stats->get_earnings( $download_id, 'this_year' );` for that download.
3. The amount returned will be one sale, not all of them.

PHP Version: 7.4
WordPress Version: 5.9

I think the database query was copied from a similar block for `get_sales`, but that does a `COUNT` in the query so it returns expected data. I think this one needs to use `get_results` instead.

robincornett:
Fixes #9070

Proposed Changes:
1. Updates the `get_earnings` method in the payment stats class to use `$wpdb->get_results()` and then add up the results ... I don't know if it's possible to get the `SUM` of two columns in a single query--if so, that might would be more efficient.
2. I also updated the stats unit tests to run two orders and get the data from them. This could be improved by generating a distinctly different order instead of just duplicating the one that's there. But with the current release code, `test_get_earnings_for_this_month_for_download_should_return_240` and `test_get_earnings_for_this_month_for_download_excluding_taxes_should_return_190` would fail, I think.

### DIFF
--- a/includes/payments/class-payment-stats.php
+++ b/includes/payments/class-payment-stats.php
@@ -276,8 +276,8 @@ class EDD_Payment_Stats extends EDD_Stats {
 				$statuses = apply_filters( 'edd_payment_stats_post_statuses', $statuses );
 				$statuses = "'" . implode( "', '", $statuses ) . "'";
 
-				$results = $wpdb->get_results( $wpdb->prepare(
-					"SELECT edd_oi.tax, edd_oi.total
+				$result = $wpdb->get_row( $wpdb->prepare(
+					"SELECT SUM(edd_oi.tax) as tax, SUM(edd_oi.total) as total
 					 FROM {$wpdb->edd_order_items} edd_oi
 					 INNER JOIN {$wpdb->edd_orders} edd_o ON edd_oi.order_id = edd_o.id
 					 WHERE edd_o.status IN ($statuses) AND edd_oi.product_id = %d {$date_query_sql}",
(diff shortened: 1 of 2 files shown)


includes/payments/class-payment-stats.php
```diff
@@ -276,7 +276,7 @@ public function get_earnings( $download_id = 0, $start_date = false, $end_date =
 				$statuses = apply_filters( 'edd_payment_stats_post_statuses', $statuses );
 				$statuses = "'" . implode( "', '", $statuses ) . "'";
 
-				$result = $wpdb->get_row( $wpdb->prepare(
+				$results = $wpdb->get_results( $wpdb->prepare(
 					"SELECT edd_oi.tax, edd_oi.total
```
cklosowski, line 280: You should be able to get the SUM of both of these with something like:
`SELECT SUM(edd_oi.tax) as tax, SUM(edd_oi.total) as total`

That would let you not have to do the `foreach` later on to do the math here. That would be much faster.

cklosowski (changes requested):

cklosowski (approved):

Issue #8003, opened 2020-06-10
3.0 - Update sample-products-import.xml


robincornett:
### Explain your enhancement (please be detailed)
The sample products import file should be updated to fix some issues for 3.0:
- failed imports
- duplicated pages

### Justification or use case
When I imported the sample products file into a new 3.0 install, I got the following failure notices:
-Failed to import edd_log_type file_download
-Failed to import edd_log_type gateway_error
-Failed to import edd_log_type sale

Additionally, the import created duplicates of two pages already created when EDD was activated:
- Checkout
- Transaction Failed (child of Checkout)

In reporting another issue where [product variation names were unable to be retrieved](https://github.com/easydigitaldownloads/easy-digital-downloads/issues/7972), it was mentioned that the price IDs for imported products may be inconsistent with EDD's current expected dataset.

cklosowski:
@robincornett that last part about the current expected data set, is likely due to previously the price IDs being `0` based instead of `1` based.

Personally I like to keep the `0` based items in the XML file because it allows us to simulate a store that has products with variations old enough to have a `0` price ID. If we remove that we'd have to hack together a `0` price ID in order to verify our changes work on both the new and old price ID formats.

PR #9077, opened 2022-02-17
Updating usernames #8003

cklosowski:


robincornett (approved):
Updated to remove one stray post meta and other creator names. My favorite part about this change is this:

image

PR #9078, opened 2022-02-18
Add unit tests for edd_is_cart_empty()

robincornett:
Proposed Changes:
1. Adds a new unit test, `test_cart_is_empty`, with two assertions.
2. Adds a test for `edd_is_cart_empty()` (true) to `test_empty_cart_is_array`.
3. Adds a test for `edd_is_cart_empty()` (false) to `test_get_cart_contents`.
4. Adds a test for `edd_is_cart_empty()` (false) to `test_restore_cart`.

### DIFF
--- a/tests/tests-cart.php
+++ b/tests/tests-cart.php
@@ -212,7 +212,6 @@ class Test_Cart extends EDD_UnitTestCase {
 
 		$this->assertInternalType( 'array', $cart_contents );
 		$this->assertEmpty( $cart_contents );
-		$this->assertTrue( edd_is_cart_empty() );
 	}
 
 	public function test_add_to_cart_multiple_price_ids_array() {
@@ -263,7 +262,6 @@ class Test_Cart extends EDD_UnitTestCase {
 		);
 
 		$this->assertEquals( $expected, edd_get_cart_contents() );
-		$this->assertFalse( edd_is_cart_empty() );
 	}
 
 	public function test_get_cart_content_details() {
@@ -597,7 +595,6 @@ class Test_Cart extends EDD_UnitTestCase {
 		edd_update_option( 'enable_cart_saving', '1' );
 		$this->assertTrue( edd_restore_cart() );
 		$this->assertEquals( edd_get_cart_contents(), $saved_cart );
-		$this->assertFalse( edd_is_cart_empty() );
 	}
 
 	public function test_generate_cart_token() {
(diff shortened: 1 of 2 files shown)


tests/tests-cart.php
```diff
@@ -212,6 +212,7 @@ public function test_empty_cart_is_array() {
 
 		$this->assertInternalType( 'array', $cart_contents );
 		$this->assertEmpty( $cart_contents );
+		$this->assertTrue( edd_is_cart_empty() );
```
cklosowski, line 215: This is an unnecessary assertion as we should be isolating the testing of the wrapper function in it's own test.

tests/tests-cart.php
```diff
@@ -262,6 +263,7 @@ public function test_get_cart_contents() {
 		);
 
 		$this->assertEquals( $expected, edd_get_cart_contents() );
+		$this->assertFalse( edd_is_cart_empty() );
```
cklosowski, line 266: Same as above, it's unnecessary to test this alongside other assertions.

tests/tests-cart.php
```diff
@@ -595,6 +597,7 @@ public function test_restore_cart() {
 		edd_update_option( 'enable_cart_saving', '1' );
 		$this->assertTrue( edd_restore_cart() );
 		$this->assertEquals( edd_get_cart_contents(), $saved_cart );
+		$this->assertFalse( edd_is_cart_empty() );
```
cklosowski, line 600: This can get removed as well.

tests/tests-cart.php
```diff
@@ -721,4 +724,12 @@ public function test_negative_fees_cart_tax() {
 
 		edd_update_option( 'enable_taxes', false );
 	}
+
+	public function test_cart_is_empty() {
+		edd_add_to_cart( self::$download->ID, array( 'price_id' => 0 ) );
+		$this->assertFalse( edd_is_cart_empty() );
+
+		edd_empty_cart();
+		$this->assertTrue( edd_is_cart_empty() );
```
cklosowski, line 733: These two should get broken off into their own 'test'. This test can be renamed to `test_cart_is_empty_false`, and then add these two lines to a method called `test_cart_is_empty_true`.

Isolating assertions to as few as possible in a 'test' make it easier to debug when things go wrong. When a single assertion fails, the entire test is marked as failed, even if something else in it passes.

cklosowski (changes requested):

cklosowski (approved):

PR #9082, opened 2022-02-22
Update get_most_valuable_order_items to query only sales

### LINKED ISSUE #9080: 3.0 - Refunded order increments number of sales in reports
### Expected behavior
Refunding a purchase should not increase the number of sales in the downloads report

### Actual behavior
Refunding a purchase increase the number of sales by 1.

1) Purchase a product
2) Check Reports > Downloads, note the number of Sales. In my case, it's 1.
3) Refund full amount from purchase
4) Return to Downloads report. In my case, number of sales is now 2.
Markup on 2022-02-18 at 17:43:17

PHP Version: 7.4.16

EDD Version (or branch): `3.0-rc1`

WordPress Version: 5.9

robincornett:
Fixes #9080

Proposed Changes:
1. This works similarly to the changes made for #8659. The `get_most_valuable_order_items` was counting all rows in the order items table for a product ID, but that table doesn't know if it's a sale or a refund. This adds an `INNER JOIN` to only get order items tied to sales.


robincornett:
Changing to draft--based on discussion for #9087, this needs to be updated to show net information, including quantity sold, not number of sales.

robincornett:
Marking this as ready for review again. It's a different implementation than what's currently in the release branch because getting net sales is a different query than getting net earnings. So here are the high points:
1. In `get_most_valuable_order_items`:
    - the query_vars have been updated to exclude taxes (because net). To do this, I had to remove the `column` and set `exclude_taxes` to `true` because of how things are parsed in `pre_query`.
    - The query is updated to use the same SQL query as the new download model to get the product totals, and rank them by that. The count/sales query has been removed because it's not accurate.
    - When setting up the `$value` object to return, the sames are calculated now using the download model `get_net_sales` method.
2. In order to use the download model, I had to update the constructor to include:
    - an optional `price_id`
    - optional dates (since the stats class is including dates). I set up just an array for optional arguments as we may need to do similar things in the future.
3. The download model was so complicated when I added in the price ID query that I ended up creating multiple `generate_*_query_sql` methods to handle variables like the product ID, price ID, and statuses, which also makes things like the status query generator more DRY.
4. The date generator is just copied from the stats class.
5. The top selling downloads list table also needed to be updated to specify that the earnings number is net.
6. I also fixed the price ID checks for the name and price columns as those did not work with a `0` price ID.

cklosowski (approved):

PR #9083, opened 2022-02-22
Update amazon gateway refunds

### LINKED ISSUE #7078: 3.0 - Orders: Refunding an Amazon payment throws fatal error.
### Expected behavior
When refunding a payment made with Amazon, it should refund at the gateway.

### Actual behavior
A fatal error happens stating that the `refund` method doesn't exist.

1) MAke a purchase with Amazon
2) Go to edit the order
3) Set the status to refunded, and save the order.

PHP Version: All

WordPress Version: All

robincornett:
Fixes #7078

Proposed Changes:
1. On the edit order screen, marks orders placed through the Amazon gateway as not refundable.
2. Removes the call for and deprecates the `process_refund` method in the Amazon payments class.

This will direct store admins to process refunds at the gateway and manually mark the payment as refunded.


includes/admin/views/tmpl-order-actions.php
```diff
@@ -49,6 +55,8 @@ class="button button-secondary"
 			<span class="edd-help-tip dashicons dashicons-unlock" title="<?php esc_attr_e( 'The refund window for this Order has passed; however, you have the ability to override this.', 'easy-digital-downloads' ); ?>"></span>
 		<?php elseif ( false === $is_refundable && true === $is_window_passed ) : ?>
 			<span class="edd-help-tip dashicons dashicons-lock" title="<?php esc_attr_e( 'The refund window for this Order has passed.', 'easy-digital-downloads' ); ?>"></span>
+		<?php elseif ( 'amazon' === $order->gateway ) : ?>
+			<span class="edd-help-tip dashicons dashicons-lock" title="<?php esc_attr_e( 'Orders placed through the Amazon gateway must be refunded through Amazon.', 'easy-digital-downloads' ); ?>"></span>
```
cklosowski, line 59: We should maybe add that after they have processed the payment as 'refunded' at the gateway, they need to come back here and manually set the payment status to 'Refunded'

cklosowski (changes requested):

includes/admin/views/tmpl-order-actions.php
robincornett: I thought about that--feels a little problematic because it's really all hidden in the tooltip. I'd prefer not to use a tooltip at all, honestly, at least not for this message.

includes/admin/views/tmpl-order-actions.php
```diff
@@ -13,6 +13,12 @@
 $is_override      = edd_is_order_refundable_by_override( $order->id );
 $is_window_passed = edd_is_order_refund_window_passed( $order->id );
 
+// EDD does not support initializing refunds through the Amazon gateway.
```
robincornett, line 16: @cklosowski I'm reconsidering having this set here--it's not necessary to prevent the call to Amazon, but does prevent a store owner from issuing a partial refund at the gateway and updating the order here accordingly. What if we allowed the refund form to still initialize, but use the notice area to show that it's a two part process?

robincornett:
@cklosowski after thinking it over, I've adjusted my approach here:
- I'm not disabling the "Initialize Refund" button. That way, if the store admin processes a partial refund at the gateway, they can do the same thing in EDD. It's fine to do this, I think, since the refund function is unhooked anyway.
- I made the notice look more like a WP admin notice, just so it would stand out a little more.
- Updated the styles a little bit due to getting some flex-wrap action on a big monitor.

cklosowski (approved):

cklosowski:
@robincornett I'm good with this approach.

PR #9091, opened 2022-02-23
Improve download sales/earnings statistics

### LINKED ISSUE #9087: 3.0 - Downloads Sales/Earnings numbers are not consistent
### Expected behavior
I expect the sales/earnings numbers for a download to be accurate.

### Actual behavior
Depending on what actions were taken most recently, the numbers can be very different.

Not really steps to reproduce the behavior, but noting the different times and places and ways this data is calculated:
1. When an order is created, the sales and earnings numbers are increased.
    - The number of _sales_ here is actually the _quantity_ of an item purchased.
    - The _earnings_ factor in the quantity as well.
 2. Based on the work done for #8659, a philosophical change was made for the Download Recount Stats tool. With that tool, recounting the stats for a download:
     - For _sales_, the number of times the download was sold is what's counted. Quantity is ignored.
     - For _earnings_, the sum of all totals for that download is used. So quantities are included.
     - Based on that same PR, only sales are included, not refunds.
3. For #9080, it was noted that refunds are being included in the download reports, which is skewing those numbers. The open PR uses the same logic as the recount stats.

With all this in mind, you can end up with wildly different numbers based on what's happened most recently, especially when refunds are involved. Fixing the open refunds issues (also #9031 which I haven't mentioned yet) will alleviate some of these, but it is still possible to end up with reports which show different numbers (based on a direct DB query) than what a download will show (based on incremental updates).

As an example, I have a download which is $47.
- I recounted the stats and the download shows 9 sales, $893 in earnings (one order had a quantity of 2 and one had a quantity of 10).
- I checked out the issue/9085 branch so I could create a manual order without stats being duplicated and created one order with a quantity of 10 for that download. The new stats show 19 sales and $1363 in earnings. Note that the Reports > Downloads numbers here will be inconsistent with what shows for the download elsewhere.
- I run the recount tool again and the sales are now 10 and $1363.

PHP Version: 7.4
EDD Version (or branch): various 3.0 branches
WordPress Version: 5.9

My gut inclination is that we really need to add a `recalculate_stats` method to the download class and use that instead of the increase/decrease methods. With the order items and orders tables, that's actually fairly straightforward to implement and would make significant strides to make the numbers throughout EDD consistent. We did this for the customer purchase count and value data in #7947; making this change for downloads would be consistent with that.

One question, though, is what numbers are expected and what are correct?
- Is the number of sales of a download really the number of _times_ it was sold, or is it the overall quantity sold of that download?
- Should download earnings/sales change when an order is refunded? Meaning, which is correct, the current behavior (decreasing sales/earnings) or the recount tool? Note that part of refactoring the recount tool included changing the label for earnings to be "Gross Revenue" to clarify this, but it did introduce a disconnect.

robincornett:
Fixes #9087

Proposed Changes:
1. This creates two new methods within the download class:
    - `recalculate_net_sales_earnings` ... updates the existing download sales/earnings meta. This one is a little complicated in order to account for partial refunds.
    - `recalculate_gross_sales_earnings` ... updates new gross sales/earnings meta
2. Refactors the recount tools to use the new methods. Note that `EDD_Tools_Recount_All_Stats` no longer extends `EDD_Tools_Recount_Download_Stats`.
3. Deprecates the `edd_` ledger functions.


includes/payments/actions.php
```diff
@@ -302,6 +298,22 @@ function edd_record_order_status_change( $old_status, $new_status, $order_id ) {
 	}
 }, 10, 3 );
 
+/**
```
robincornett, line 301: @cklosowski I _finally_ remembered that we have the transition hooks in 3.0 so I've removed most of the calls to recalculate and made this new function. I haven't named it for now but I think this is way better than going through arrays of cart details (which was causing some issues during the purchase process anyway while order items were still pending). I'm wondering if we have need of an `order_items` directory with its own functions/actions file(s).

includes/payments/class-edd-payment.php
```diff
@@ -1110,6 +1013,14 @@ public function save() {
 		$customer = new EDD_Customer( $this->customer_id );
 		$customer->recalculate_stats();
 
+		$order_items = edd_get_order_items( array(
```
robincornett, line 1016: Download recalculations are moved here with the customer recalculations--the new download calculations have to happen after the order items have been saved. This might actually be covered now with the `edd_transition_order_item_total` hook being added in but I've left it for now.

tests/orders/tests-edd-payment.php
```diff
@@ -549,7 +549,7 @@ public function test_modify_cart_item_price() {
 		$this->assertEquals( 1, $this->payment->cart_details[0]['item_price'] );
 
 		$download = new \EDD_Download( $this->payment->cart_details[0]['id'] );
-		$this->assertEquals( 41, $download->get_earnings() );
+		$this->assertEquals( 1.00, $download->get_earnings() );
```
robincornett, line 552: I did a lot of logging during tests and I think the changes here are correct--it seems like the original modifications were adding to the download earnings instead of changing. So my understanding is that this one payment goes through:
1. Starts with a `40.00` order item total so that's the download [net] earnings.
2. The cart item price is changed to `1.00` so the download [net] earnings should be `1.00`.
3. The cart item price is changed to `3.00` in the next test so the [net] earnings should be `3.00`.

tests/tests-api.php
```diff
@@ -299,10 +299,10 @@ public function test_get_product_stats() {
 		$this->assertArrayHasKey( 'sales', $out['products'][0]['stats']['monthly_average'] );
 		$this->assertArrayHasKey( 'earnings', $out['products'][0]['stats']['monthly_average'] );
 
-		$this->assertEquals( '60', $out['products'][0]['stats']['total']['sales'] );
-		$this->assertEquals( '229.430000', $out['products'][0]['stats']['total']['earnings'] );
-		$this->assertEquals( '60', $out['products'][0]['stats']['monthly_average']['sales'] );
-		$this->assertEquals( '229.430000', $out['products'][0]['stats']['monthly_average']['earnings'] );
+		$this->assertEquals( '1', $out['products'][0]['stats']['total']['sales'] );
```
robincornett, line 302: I do not know where the original values came from for these--from what I see in this class, there's just one payment to evaluate and the product in question has one sale at `100.00`. But that's so different from the `229.430000` number that we maybe need to look into this more.

tests/tests-downloads.php
```diff
@@ -133,7 +132,7 @@ public function test_edd_download() {
 		$this->assertEquals( $prices, $download3->prices );
 		$this->assertEquals( $prices, $download3->get_prices() );
 		$this->assertEquals( 6, $download3->sales );
-		$this->assertEquals( 120.50, $download3->earnings );
+		$this->assertEquals( 120.00, $download3->earnings );
```
robincornett, line 135: All download ledger tests are removed--those functions/methods have been updated to run recalcuations if called, but the incrementing is ignored. But doing things like `$download3->increase_earnings( '0.50' );` is now nonsensical and shouldn't actually affect the earnings, because it's an artificial adjustment.

includes/class-edd-download.php
```diff
@@ -666,19 +666,13 @@ public function get_sales() {
 	 */
 	public function increase_sales( $quantity = 1 ) {
 
-		$quantity    = absint( $quantity );
-		$total_sales = $this->get_sales() + $quantity;
+		_edd_deprecated_function( __METHOD__, '3.0', 'EDD_Download::recalculate_net_sales_earnings()' );
+		edd_recalculate_download_sales_earnings( $this->ID );
 
-		if ( $this->update_meta( '_edd_download_sales', $total_sales ) ) {
+		$this->get_sales();
```
robincornett, line 672: I've updated these to all be marked as deprecated, and run the recalculation functions as needed. The action hooks are still available.

includes/admin/tools/class-edd-tools-recount-download-stats.php
```diff
@@ -77,45 +77,13 @@ public function get_data() {
...
-		}
-
 		if ( ! empty( $this->download_id ) && is_numeric( $this->download_id ) ) {
-			$conditions[] = $wpdb->prepare(
-				"oi.product_id = %d",
-				$this->download_id
-			);
-		}
-		$conditions = ! empty( $conditions ) ? ' AND ' . implode( ' AND ', $conditions ) : '';
+			edd_recalculate_download_sales_earnings( $this->download_id );
```
robincornett, line 81: I removed a lot here as this tool can really only run on one download. The original code added a condition for the download ID if it wasn't empty but then updated the download meta potentially on a null ID.

includes/admin/tools/class-edd-tools-recount-all-stats.php
```diff
@@ -13,22 +13,39 @@
...
-	require_once EDD_PLUGIN_DIR . 'includes/admin/tools/class-edd-tools-recount-download-stats.php';
-}
-
 /**
  * EDD_Tools_Recount_All_Stats Class
  *
  * @since 2.5
  */
-class EDD_Tools_Recount_All_Stats extends EDD_Tools_Recount_Download_Stats {
+class EDD_Tools_Recount_All_Stats extends EDD_Batch_Export {
```
robincornett, line 21: With the helper function, this class doesn't need to extend the `EDD_Tools_Recount_Download_Stats` class anymore--more simplification.

robincornett:
@cklosowski I think I'm ready for you to review this for real now. I've been tinkering with the net sales/earnings queries quite a bit because the quantities there are tricky. I think I finally got the calculations correct in [9c82720](https://github.com/awesomemotive/easy-digital-downloads/pull/9091/commits/9c82720c1db0e58de682ecb338a297749ebd49b0), but wanted to try to get it into one query because we have an open issue/PR about reports (#9080) which should use the same data as this does.

In [763dca6](https://github.com/awesomemotive/easy-digital-downloads/pull/9091/commits/763dca6f618d877b80db96a97b9a5dbac65d6cc5) it is reduced to one query (well if you don't count the subqueries separately).

The tricky part is that partial refunds can basically null out the quantities, because (for example) the sale and the refund will both have a quantity of 1 (or -1 for the refund), although the totals will be different. I ended up putting together a very simple set of orders for which I could calculate the net revenue/sales manually:

image

With these orders, the net quantity sold is 3 and the net revenue is $24.98. What I came up with feels garly but seems to generate the correct numbers.

includes/payments/actions.php
cklosowski: Yeah @robincornett lets go ahead and create a new file to hold all the transitions related to payments. That will make things much easier to maintain later on.

cklosowski:
@robincornett I think we might have a bit further to go. I'm digging into the reporting views and I found this little bit of clarification we'll need to make. 

The tiles (a couple examples here) are not clarifying if these are net or gross (which we now have the ability to report on).
image

The first item builds from this:
```
		$reports->register_endpoint( 'overview_time_period_data', array(
			'label' => __( 'Sales / Earnings', 'easy-digital-downloads' ),
			'views' => array(
				'tile' => array(
					'data_callback' => function () use ( $dates, $exclude_taxes, $currency ) {
						$stats = new EDD\Stats( array(
							'range'         => $dates['range'],
							'exclude_taxes' => $exclude_taxes,
							'currency'      => $currency,
							'output'        => 'formatted',
						) );

						return $stats->get_order_count() . ' / ' . $stats->get_order_earnings();
					},
					'display_args'  => array(
						'comparison_label' => $label,
					),
				),
			),
		) );
```

If we follow the `$stats->get_order-count()` we end up here: 
`$this->query_vars['status'] = edd_get_gross_order_statuses();`

It appears most tiles are actually set as 'gross' following through to the `$stats` class. The graph below is `net` as we have two graphs, one for net and one for gross.

The top graph is clearly set for `net`
`$statuses = array( 'complete', 'publish', 'revoked', 'refunded', 'partially_refunded' );`

So we need to determine if we're going to either update these tiles to pull `net` as I think that is the 'expected' default stat, or if we should just label these to be `gross` where applicable.

robincornett:
Per discussion today:

- Gross download stats need to be calculated off of the order item subtotals--before discounts--and include taxes as well.
- A long term dream (of mine anyway) is to see if we can make an interface which will handle database queries and calculations for downloads, reporting, and stats. Currently, each of these handle their own queries, and we are finding that each handle things slightly differently. If we can determine a method of querying which will work universally, it will make these consistent, and if we can handle the querying in one place, it will simplify long term maintenance.

cklosowski:
@robincornett 

| Gross download stats need to be calculated off of the order item subtotals--before discounts--and include taxes as well.

Taxes are not included in Gross revenue usually.

robincornett:
> Taxes are not included in Gross revenue usually.

Great, because I just realized I forgot to add them in.

robincornett:
One item that's not factored in here [yet?] is fees....

It seems like fees on order items should be accounted for in the net earnings, which then brings in the `_edd_order_adjustments` table and introduces another query/join. At that point, since there is already a subquery to handle quantities for net sales, it seems like it would be more practical to separate sales and earnings in the database query and get them separately--noting that the second subquery for net earnings/sales is really only getting sales, not earnings.

However, thinking about gross earnings, I feel like fees are a little trickier there. If a fee is a discount (Discounts Pro), it should _not_ be included in the gross earnings, but if it's an actual fee (like a shipping charge for simple shipping), then it seems like it should be part of the gross earnings. Again, this is a join that affects earnings, but not sales.

Note that some fees are levied against the order, not an individual order item, so those would not be an issue, but the two extensions noted (Discounts Pro and Simple Shipping) apply their fees to the order item.

cklosowski:
Where relevant, `Gross` and `Net` labels have been added, to make it explicitly clear what stat we are giving:
image

cklosowski (approved):

PR #9095, opened 2022-02-25
Make customer "add note" button text translation ready

### LINKED ISSUE #9090: Hardcoded text string that shoulb be available to translate
There is a text string that we hardcoded and it should be available for translation here 
https://github.com/awesomemotive/easy-digital-downloads/blob/master/includes/admin/customers/customers.php#L618

robincornett:
Fixes #9090 (currently branched off `master`)

I'm going ahead and making a PR for this but noting that the issue is present only in EDD 2.x, and is fixed already in 3.0.


cklosowski (approved):
Confirmed rerunning the `build` adds this line to the `.pot` file.

PR #9096, opened 2022-02-25
Disable PayPal button until required fields have value

### LINKED ISSUE #9094: Consider disabling PayPal buttons until required fields are completed
> ...on the EDD site itself if you click on the PayPal buttons then close the dialog a few issues happen:
>
> 1. You get a bunch of EDD errors about fields being incorrect (for example if you have not filled in any of the fields at the checkout).
> 2. You get an endless spinner/error that does not go away until you change gateways again.

_Originally posted by @davidsherlock in https://github.com/awesomemotive/easy-digital-downloads/issues/9092#issuecomment-1050740229_

Looking over the PayPal developer docs, I think we could set up a listener for our required fields and disable the PayPal button until they have been filled out, which would be a less disruptive user experience, I think: https://developer.paypal.com/docs/checkout/standard/customize/validate-user-input/

robincornett:
Fixes #9094 (currently branched off master)

Proposed Changes:
1. Gets all fields marked as required on the checkout form.
2. Adds JS functions for `onInit` and `onClick` for PayPal:
    - On initialization, disables the PayPal button and tallies up how many required inputs have value. The PayPal button is disabled until the required inputs all have a value (did it this way as a regular check out may have two required inputs--email and first name--but if taxes are enabled, that jumps to five). If the error message (from `onClick`) is displaying, it will clear as soon as all inputs have a value.
    - On click, if not all of the inputs have a value, displays an error message indicating that some required fields are missing data. This doesn't list the specific fields like the process does now.

The net effect of this PR is that the PayPal flow is not initialized until all required data is present.


robincornett:
Closing this as it needs to be reworked off of 3.0 as a starting point.

PR #9101, opened 2022-03-02
Remove Mailchimp version check from product education

### LINKED ISSUE #9100: Mailchimp configuration link in product education can be incorrect
### Expected behavior
If I activate the Mailchimp extension from the Downloads > Settings > Marketing > Email Marketing screen, I expect the configuration button to take me to the correct settings screen.

### Actual behavior
If I activate Mailchimp and then click the button without reloading the page, I'm taken to the Extensions tab even if I'm running a version of the Mailchimp extension which has the settings in the proper location.

1. Install Mailchimp (latest version), but do not activate it.
2. Navigate to Downloads > Settings > Marketing > Email Marketing and activate Mailchimp.
3. Without reloading the page, click the new configuration button.

PHP Version: 7.4
WordPress Version: 5.9

The configuration button URL is defined before the plugin is activated. It has a version check to change the extension tab if the newest version of Mailchimp is not installed, but the version check always fails if the extension isn't active. We need to just remove this check  and expect that users are either installing Mailchimp for the first time or have the latest version installed.

robincornett:
Fixes #9100

### DIFF
--- a/includes/admin/extensions/product-education/class-email-marketing.php
+++ b/includes/admin/extensions/product-education/class-email-marketing.php
@@ -72,9 +72,6 @@ class EmailMarketing extends Extension {
 			/* translators: the product name */
 			$configuration['title'] = sprintf( __( 'Get %s Today!', 'easy-digital-downloads' ), $product_data->title );
 		}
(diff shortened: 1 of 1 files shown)


cklosowski (approved):
Confirmed incorrect URL destination on `release/2.11.6`
Confirmed redirected to Mailchimp settings page when on `issue/9100`

PR #9111, opened 2022-03-08
Improve edd_validate_discount logic

davidsherlock:
Fixes #

The logic in the validation logic.

Proposed Changes:
1.
2.
3.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1489,7 +1489,7 @@ add_action( 'init', 'edd_apply_preset_discount', 999 );
 /**
  * Validate discount code.
  *
- * @since 3.0
+ * This is a 'fork' of the original validate_discount function but fixes the incorrect logic.
  *
  * @param int   $discount_id  Discount ID.
  * @param array $download_ids Array of download IDs.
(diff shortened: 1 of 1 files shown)


robincornett:
@davidsherlock I've confirmed that the function may return inconsistent results when the product condition is set to `any` and you swap out different download IDs. I need to test more, but I have a favor to ask: your PR shows more changes than there really are because the function has been reformatted to indent with 2 spaces instead of a tab. Could you update the PR code please for formatting?

I think we'll also want to add some unit tests to cover this function--I do not see that we have any yet. I'll mark it accordingly--I can work on that part if it's okay with you. Thank you!

davidsherlock:
@robincornett I tried to fix that but it didn't work as I wanted. I don't personally use GitHub (other than posting issues) because it slows down my development speed, so if you just copy and paste this section of code and replace it with your formatting that will work best:

```
if ( ! empty( $product_requirements ) ) {

    $matches = count( array_intersect( $product_requirements, $download_ids ) );

    switch ( $discount->get_product_condition() ) {
      case 'all':
        $is_valid = $matches === count( $product_requirements );
        break;
      default:
        $is_valid = 0 < $matches;
    }

  } else {
    $is_valid = true;
  }

  if ( ! empty( $excluded_products ) ) {
    $is_valid = false === (bool) array_intersect( $excluded_products, $download_ids );
  }
```

robincornett:
@davidsherlock I'm going to go ahead and close this PR in favor of the other one--thank you so much for getting the ball rolling on this!

PR #9112, opened 2022-03-08
Prevent notifications from showing if JS is not fully initialized

robincornett:
Proposed Changes:
1. If Alpine is not properly initialized for some reason, the notifications button and panel remain hidden.


cklosowski:
@robincornett do we have a reliable way to replicate this yet? Even if it's editing a line of code to trigger the situation?

cklosowski:
@robincornett do we have a reliable way to replicate this yet? Even if it's editing a line of code to trigger the situation?

robincornett:
I don't know if it's an accurate representation of what is actually happening when it's happening to someone, but I was able to sort of replicate by adding an early return after line 4 here: https://github.com/awesomemotive/easy-digital-downloads/blob/try/fix-notifications-conflict/assets/js/admin-notifications.js#L4. I got some fun console errors and the panel did misbehave, so I worked backwards from there.

robincornett:
Closing and reworking for EDD 3.0.

Issue #8757, opened 2021-07-15
3.0 - Product Requirement flat rate discount amount is incorrect when scope is specified


davidsherlock:
### Expected behavior
I expect flat rate discount amounts to be correctly distributed across all `exclusive` cart items when specific rather than applied directly to the amount zeroing it out. This happens when non-exclusive, but not exclusive as the amount value is just passed to the discount class method for basic calculation.

### Actual behavior
Flattens amounts rather than distributes the flat rate amount across the exclusive items.

### Steps to reproduce the behavior
1) Create a discount, and add product requirements for 2-3 products. Set condition to all and scope to exclusive.
2) Make it a flat rate amount that is generally higher than the product prices.
3) Do not add excluded products and add the items to the cart, you can add additional items too.
4) Apply the discount and you will see the flat rate amount is not distributed, but applied on a per-item basis.

### Information (if a specific version is affected):
PHP Version: 8.0

EDD Version (or branch): 3.0-beta3

WordPress Version: latest

davidsherlock:
This can essentially be fixed by forking the else condition logic and moving it to a separate function to keep things DRY, then passing in an array of download ID and cart items to discount. This will then run the logic but only on the exclusive items, keeping the amounts distributed. This would obviously need to exclude products, etc.

davidsherlock:
This actually results in the customer getting significantly more off also, see this screenshot for example:

<img width="681" alt="Screenshot 2021-07-15 at 18 53 55" src="https://user-images.githubusercontent.com/22668877/125818634-65c0c2f3-2efe-4e30-8f77-8e61c2df5850.png">

This was a $10 off flat rate discount, but $20 has been applied because it has applied the $10 to both items.

ashleyfae:
Is this definitely different behaviour in 3.0 than master branch?

davidsherlock:
I haven't tested with the older one, but I noticed most of the functions were re-factored for 3.0 so assumed it is the same logically speaking.

I can test it tomorrow and report back if you want?

ashleyfae:
Yes please. :+1: If this is also the behaviour in master then I'm not sure if we'll change it. (Or at least it definitely won't be changed for 3.0 and we'd have to carefully consider it for later.) It's quite significant to go from a flat discount across _all items_ versus a flat discount on the cart.

davidsherlock:
Yeah, I agree. For now it's fine, but worth mentioning because it's not "expected" behaviour per-say...

[...]

PR #9113, opened 2022-03-09
Fix and improve discount validation logic

### LINKED ISSUE #9110: 3.0 - edd_validate_discount product requirements logic is incorrect
### Expected behavior
When validating a discount, the `foreach` loop logic is incorrect for product requirements. See the following code:

```
	if ( ! empty( $product_requirements ) ) {
		foreach ( $product_requirements as $download_id ) {
			if ( empty( $download_id ) ) {
				continue;
			}

			$download_id  = absint( $download_id );
			$has_download = in_array( $download_id, $download_ids, true );

			switch ( $discount->get_product_condition() ) {
				case 'all':
					$is_valid = false !== $has_download;
					break;
				default:
					$is_valid = $has_download;
			}
		}
	} else {
		$is_valid = true;
	}
```

When the `default` case is hit, once the final array element is validated it could end up being false instead of true (like the other items), breaking the logic for the check.

Additionally because we are only dealing with IDs (and not variable priced products), using an `array_intersect` is a about 10 times faster than a foreach. This could also apply to the exclusions also, which would increase performance. 

### Actual behavior
If the last item product requirement is NOT in the cart, this will result in a false which is wrong. As it should be true for ANY of the items.

I will submit a PR to fix this issue and improve the performance of this function.

robincornett:
Fixes #9110

This is a revision/extension of the work @davidsherlock did in #9111. The product requirements/exclusions logic are all his; I just did a bit of logic reorganization for improved readability.

Proposed Changes:
1. Updates the doc block for `edd_validate_discount` to better explain what the function does and does not cover.
2. Moves/updates the initial `$is_valid` variable assignment as it was always being overwritten.
3. Updates the discount class' `is_product_requirements_met` method to match this logic and use the array_intersect approach so that the class method is consistent with this new function.
4. Adds unit tests to cover `edd_validate_discount`.

Note that if you copy the unit tests to the release/3.0 branch, they do fail, which I believe is an additional confirmation of David's findings.

David, it looks like I cannot officially add you to this as a reviewer, but I would greatly appreciate your review. Thank you for the PR!


davidsherlock:
Thanks, @robincornett! Looks good to me. Code also looks more beautiful now.

As a side note, on a related note to discounts there is a pretty serious issue with flat rate discounts I posted a while back. I need to find 15 minutes or so to write up a proper fix for this, but you can see the issue I opened here: #8757 

Because I am building a bunch of new EDD 3.0 extensions, I am uncovering some small issues like this. Though the discount one is pretty serious as the flat rate amount is essentially added multiple times.

tests/discounts/tests-discounts.php
```diff
@@ -393,6 +393,57 @@ public function test_discount_is_product_requirements_met() {
...
+		$args     = array(
+			'product_reqs'      => $products,
+			'product_condition' => 'any',
+			'max_uses'          => 10000,
+		);
+
+		edd_update_discount( self::$discount_id, $args );
+		$this->assertTrue( edd_validate_discount( self::$discount_id, $products ) );
+		$this->assertTrue( edd_validate_discount( self::$discount_id, array( self::$download->ID ) ) );
+		$this->assertTrue( edd_validate_discount( self::$discount_id, array( 100 ) ) );
```
robincornett, line 410: I'm feeling like the tests should also cover sending in a product which is _not_ part of the product requirements. But before I get too excited about adding in more assertions, I want to check with you, @cklosowski about it because on a recent PR you said we should limit the number of assertions per test. We do, however, have quite a few tests which have multiple (and even many) assertions, so my question is:
1. If the assertions cover an identical discount setup (like the discount should not change as in this method), is it okay to have multiple assertions?
2. I updated the discount object in this method and then again in both of the following methods, but I'm wondering if the modified discount is what's used in each. I wrote the update args assuming that was the case both ways.
3. It seems like the `test_discount_is_product_requirements_met` method should also test for an expected false return but it doesn't.

Anyway I would hypothetically like to add assertions for a hypothetical product which is not listed in the requirements, eg: 
```php
		$products[] = 3467;
		$this->assertFalse( edd_validate_discount( self::$discount_id, array( 3467 ) ) );
		$this->assertTrue( edd_validate_discount( self::$discount_id, $products ) );
```

cklosowski (approved):

PR #9114, opened 2022-03-10
Add new doc link for pass help, and include MonsterInsights in notice list

cklosowski:
Proposed Changes:
1. Includes the URL to the new pass doc.
2. Adds logic to also suggest MonsterInsights if AffilaiteWP license is active.

### DIFF
--- a/includes/admin/promos/notices/class-license-upgrade-notice.php
+++ b/includes/admin/promos/notices/class-license-upgrade-notice.php
@@ -134,12 +134,7 @@ class License_Upgrade_Notice extends Notice {
 		}
 
 		$mi_license = \MonsterInsights::$instance->license->get_license_key();
-
-		if ( empty( $mi_license ) ) {
-			return false;
-		}
-
-		return true;
+		return ! empty( $mi_license );
 	}
 
 	/**
--- a/includes/admin/promos/notices/class-license-upgrade-notice.php
+++ b/includes/admin/promos/notices/class-license-upgrade-notice.php
@@ -94,7 +94,8 @@ class License_Upgrade_Notice extends Notice {
 			if (
 				$this->pass_manager->has_pass() &&
 				Pass_Manager::pass_compare( $this->pass_manager->highest_pass_id, Pass_Manager::EXTENDED_PASS_ID, '>=' ) &&
-				$this->has_affiliate_wp_license()
+				$this->has_affiliate_wp_license() &&
+				$this->has_mi_license()
 			) {
 				return false;
 			}
@@ -118,6 +119,27 @@ class License_Upgrade_Notice extends Notice {
 		}
 
 		return (bool) affiliate_wp()->settings->get( 'license_key' );
+	}
+
+	/**
+	 * Determines whether or not MonsterInsights is installed and has a license key.
+	 *
+	 * @since 2.11.6
+	 *
+	 * @return bool
+	 */
+	private function has_mi_license() {
+		if ( ! class_exists( 'MonsterInsights' ) ) {
+			return false;
+		}
+
+		$mi_license = \MonsterInsights::$instance->license->get_license_key();
+
+		if ( empty( $mi_license ) ) {
+			return false;
+		}
+
+		return true;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/admin/promos/notices/class-license-upgrade-notice.php
```diff
@@ -175,13 +199,22 @@ protected function _display() {
...
+					/* Translators: %1$s opening anchor tag; %2$s closing anchor tag */
+						__( 'Grow your business and make more money with affiliate marketing. %1$sGet AffiliateWP%2$s', 'easy-digital-downloads' ),
+						'<a href="' . esc_url( add_query_arg( $this->query_args( 'extended-pass', $source ), 'https://affiliatewp.com/?ref=743' ) ) . '" target="_blank">',
+						'</a>'
+					);
+				} elseif( ! $this->has_mi_license() ) {
+					printf(
+					/* Translators: %1$s opening anchor tag; %2$s closing anchor tag */
+						__( 'Gain access to powerful insights to grow your traffic and revenue. %1$sGet MonsterInsights%2$s', 'easy-digital-downloads' ),
+						'<a href="' . esc_url( 'https://www.monsterinsights.com?utm_campaign=xsell&utm_source=eddplugin&utm_content=top-promo' ) . '" target="_blank">',
```
robincornett, line 214:
```suggestion
						'<a href="' . esc_url( 'https://monsterinsights.com?utm_campaign=xsell&utm_source=eddplugin&utm_content=top-promo' ) . '" target="_blank">',
```
Minor, but following the link, there is no www.

includes/admin/promos/notices/class-license-upgrade-notice.php
```diff
@@ -120,6 +121,27 @@ private function has_affiliate_wp_license() {
...
+	 * @return bool
+	 */
+	private function has_mi_license() {
+		if ( ! class_exists( 'MonsterInsights' ) ) {
+			return false;
+		}
+
+		$mi_license = \MonsterInsights::$instance->license->get_license_key();
+
+		if ( empty( $mi_license ) ) {
```
robincornett, line 138: Could we simplify/reduce this to:

```php
return ! empty( $mi_license );
```

includes/admin/promos/notices/class-license-upgrade-notice.php
```diff
@@ -148,8 +170,10 @@ protected function _display() {
 				// No license keys active at all.
 				printf(
 				/* Translators: %1$s opening anchor tag; %2$s closing anchor tag */
-					__( 'You are using the free version of Easy Digital Downloads. %1$sPurchase a pass%2$s to get email marketing tools and recurring payments.', 'easy-digital-downloads' ),
+					__( 'You are using the free version of Easy Digital Downloads. %1$sPurchase a pass%2$s to get email marketing tools and recurring payments. %3$sAlready have a Pass?%4$s', 'easy-digital-downloads' ),
 					'<a href="' . esc_url( add_query_arg( $this->query_args( 'core', $source ), 'https://easydigitaldownloads.com/pricing/' ) ) . '" target="_blank">',
+					'</a>',
+					'<a href="' . esc_url( add_query_arg( $this->query_args( 'core', $source ), 'https://easydigitaldownloads.com/what-is-an-edd-pass' ) ) . '" target="_blank">',
```
robincornett, line 176: Should this link be added to the next condition as well, if someone has a licensed product but not a pass?

robincornett (commented):
Overall this is working, just some fiddly comments. Also should a test for MI be added after `test_aap_wins_over_personal_pass`?

includes/admin/promos/notices/class-license-upgrade-notice.php
cklosowski: Thanks, that's the link they gave me :P

includes/admin/promos/notices/class-license-upgrade-notice.php
cklosowski: No because once their key is a pass, the entire system should identify that at the next check.

robincornett: When I upgraded to a pass, it had an entirely new license key, so I _think_ if you have a single product key entered and buy a pass but don't enter the key, you'll still see the single product message.

cklosowski:
@robincornett For now no, just because we'd have to start introducing adding AffiliateWP to the test process which I think is a much larger challenge.

robincornett (approved):

PR #9117, opened 2022-03-23
Add filter to edd_get_discount_by_code

davidsherlock:
Fixes #

Proposed Changes:
1.
2.
3.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -153,12 +153,23 @@ function edd_get_discount( $discount_id = 0 ) {
  * @since 1.0
  * @since 2.7 Updated to use EDD_Discount object
  * @since 3.0 Updated to call use new query class.
+ * @since 3.0 Updated to include a filter.
  *
  * @param string $code Discount code.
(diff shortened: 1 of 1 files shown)


davidsherlock:
@cklosowski I have built 10 commercial extensions for EDD 3.0 with full reporting, multi-currency support, REST endpoints, etc. A few of the plugins use discounts and I want to make them more efficient (and cleaner) in how this is handled.

Rather than clog up the discounts table with dynamically generated discounts, I want to create them on the fly when they are entered at the checkout. The exact use case is referrals, I want that a customer can enter a friend's email address into the discount code box and that if valid (exists as a customer), a discount will be automatically generated and applied to the session.

This specific function is called by almost every single discount function, which makes sense to add a filter so I can pass back a newly created discount to achieve what I want.

davidsherlock:
Any chance of getting this one merged for 3.0? Would help me a lot as an extension developer!

cklosowski (approved):

davidsherlock:
Thanks @cklosowski! Much appreciated :)

PR #9121, opened 2022-04-05
Moves moderation to checkout and adds refunds to gateways #9120

### LINKED ISSUE #9120: 3.0 - Setting adjustments
### Expected behavior
Moderation and Refunds should fall under the newer 'Payments' tab.

### Actual behavior
they are in the 'Store' tab.

cklosowski:
Fixes #9120 

Proposed Changes:
1. Shifts the 'moderation' settings to the 'checkout' settings.
2. Moves 'Refunds' submenu to 'Payments' tab

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -489,52 +489,6 @@ function edd_get_registered_settings() {
 						'placeholder' => '.'
 					),
 				),
-				'moderation' => array(
-					'moderation_settings' => array(
-						'id'   => 'moderation_settings',
-						'name' => '<h3>' . __( 'Moderation', 'easy-digital-downloads' ) . '</h3>',
-						'desc' => '',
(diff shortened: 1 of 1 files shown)


robincornett (approved):
Approving, but just mentioning that I would love to see some revisions on how some of the settings are presented. Specifically:
- I think it might be clearer for the `refundability` setting to read as a positive rather than a negative, eg. something like "This will be the store default setting; it can be overridden on a per-product basis".
- For the heading "settings", if the text is important, I think we should consider just adding the text, without resorting to a tooltip. If it's not important, is it necessary?

cklosowski:
@robincornett I've updated the text for those settings, however updating the tooltips for 'headers' callbacks is a much larger project that I think we should iterate on separately after 3.0.

robincornett (approved):

PR #9126, opened 2022-04-11
Update order stats class to use null, remove empty checks

### LINKED ISSUE #9122: 3.0 - Order Stats class does not account for price IDs correctly
### Expected behavior
I expect to be able to run order stats for variably priced products which include a price ID of `0`.

### Actual behavior
The Orders Stats class uses check like `empty`, rather than checking whether a value is null or numeric for price IDs. Example:

https://github.com/awesomemotive/easy-digital-downloads/blob/c441740dd998c3fbe651534aac2f4e51a3057975/includes/class-stats.php#L577-L579

The above will not query for the `0` variation of a product.

1. If you start with the EDD demo content, you'll have a product with a `0` price ID.
2. Create orders for that product/price ID.
3. Check the Reports > Downloads screen. The price label will not be correct, for example, but I expect the stats to be incorrect in places as well.

PHP Version: 7.4
WordPress Version: 5.9.2

Encountered this while reading through the class code when working on #9080.

robincornett:
Fixes #9122

Proposed Changes:
1. Sets the `price_id` default to `null` instead of an empty string.
2. Creates `generate_price_id_query_sql` method to check the price ID and generate the SQL query string.
3. In places where a price ID is cast to an absolute value, check if it's numeric first--if numeric, cast as before; if not, set to `null` instead.

### DIFF
--- a/tests/orders/tests-stats.php
+++ b/tests/orders/tests-stats.php
@@ -43,6 +43,23 @@ class Stats_Tests extends \EDD_UnitTestCase {
 		for ( $i = 0; $i < 2; $i++ ) {
 			self::$refunds[] = edd_refund_order( self::$orders[ $i ] );
 		}
+
+		// Add a variable download with price ID 0 to order 3.
+		edd_add_order_item(
+			array(
+				'order_id'     => 3,
+				'product_id'   => 2,
+				'price_id'     => 0,
+				'product_name' => 'Variable Test Download Product - Simple',
+				'status'       => 'complete',
+				'amount'       => 20,
+				'subtotal'     => 20,
+				'discount'     => 0,
+				'tax'          => 0,
+				'total'        => 20,
+				'quantity'     => 1,
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


cklosowski:
@robincornett Any chance we can get a unit test or two around this?

tests/orders/tests-stats.php
```diff
@@ -173,6 +173,48 @@ public function test_get_order_item_refund_count_with_range_this_year_should_be_
 		$this->assertSame( 2, $count );
 	}
 
+	/**
+	 * @covers ::get_order_item_count
+	 */
+	public function test_get_order_item_count_null_price_id_should_be_2() {
```
cklosowski, line 179: @robincornett Let's also get a unit test around a `0` price ID. We've captured `null` and `2` but we haven't done a check to make sure `0` still works.

cklosowski (changes requested):

cklosowski (approved):
Approve this as it seems to work fine, @robincornett you can merge this once we get a unit test in around `0` price IDs.

PR #9127, opened 2022-04-11
Create edd_get_download_taxonomies helper function

### LINKED ISSUE #9124: 3.0 - consider filtering download taxonomies
In at least some of the places where EDD uses `get_object_taxonomies( 'download' )`, it would be great to be able to filter the resulting list.

### Justification or use case
In EDD 3.0, there is an "Earnings By Taxonomy" table which pulls all terms from all download taxonomies. In our own store, we have custom taxonomies set up for gateways (countries, features, and currencies). These probably shouldn't be included in this table, or in the Customers export, but currently there is no way to remove them, so they're there.

robincornett:
Fixes #9124

Proposed Changes:
1. Creates the `edd_get_download_taxonomies` helper function, which will, by default, return only the `download_category` and `download_tag` taxonomy names.
2. Applies a filter to the array so a developer can manually add a custom taxonomy to the array.
3. Updates the customer export and the taxonomy earnings table to use this new helper function.


cklosowski (approved):

cklosowski:
Verified that this worked by adding a custom post type taxonomy using the CPT UI plugin, and adding it's slug to the filter shows it in the reports

PR #9132, opened 2022-04-12
Merge/2.11.6

cklosowski:
Proposed Changes:
1. Brings changes from `master` into `release/3.0`
2. Updates NPM packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "easy-digital-downloads",
-	"version": "3.0.0",
+	"version": "2.11.6",
 	"description": "The easiest way to sell digital products with WordPress.",
 	"private": true,
 	"author": "Easy Digital Downloads",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "easy-digital-downloads",
-	"version": "2.11.5",
+	"version": "3.0.0",
```
robincornett, line 3: I don't know why, but we've been consistently keeping the package.json version number at the master version number, so should this be this instead?
```suggestion
	"version": "2.11.6",
```

cklosowski: We need to move it forward to be `3.0.0` from now on. As this is what is going to be used as we get to string lock for translations (it is pulled from this) so that we can make sure the translation files are for specific versions.

robincornett (approved):

package.json
cklosowski: Let's just use `2.11.6` for now and as we enter string lock we'll update it to `3.0.0`

PR #9135, opened 2022-04-13
Allow old payment objects to be accessed while migration is pending

### LINKED ISSUE #9125: 3.0 - Create an option which indicates that a store migration is in process
We should create a general option which can be checked to determine if a store migration to EDD 3.0 is in process.

### Justification or use case
Thinking through this in our dev call today, for a large store which is remaining open during the migration. For example, if a renewal payment comes in via webhook, the original payment needs to be retrieved. Currently, 3.0 will look for the new order object, but if that specific payment hasn't yet been migrated, it would be helpful to allow the store to look up the original payment (post) object manually as a fallback, but only until the migration is completed.

robincornett:
Fixes #9125

Proposed Changes:
1. When EDD checks for old payments in `admin_init`, if payments exist, adds a `edd_v3_migration_pending` option to the database (so this does require being in the admin).
2. When the migration is complete, but before legacy data is deleted, deletes the option. Since we are only dealing with post/payment objects at this time, it could be moved to the order migration step, but not sure it matters significantly.
3. In the `EDD_Payment` class, if an order object is not retrieved and this option exists, `setup_compat_payment` runs to set up the payment object from the original post data.
    - This method loads the file for the `EDD_Payment_Compat` class and updates the `EDD_Payment` object from that data.
    - The `EDD_Payment_Compat` is an extremely stripped down version of the `EDD_Payment` class from EDD 2.x, and includes only what is needed for the setup. Although I'm expecting to only use it with the payment ID, I did leave the original constructor intact.
4. Not sure how extensive we want the unit tests to be but I set them up in their own class.

Note that this will not show payments in the admin or front end, as those all use order functions. However, it does allow a subscription to be manually renewed (I haven't tested with a webhook yet). Everything looks correct for the new order in the database, although not everything displays correctly in the admin. The order is also ID 1, which is a little weird, but not unexpected.

We should test that for sure, along with an API request for licensing, to make sure updates can still be delivered.

We will want to communicate early and often that stores should NOT delay on migrating.


cklosowski:
`The order is also ID 1, which is a little weird, but not unexpected.`

@robincornett This seems off, we will need to make sure the `auto_increment` value is properly set if we find existing payments in the posts table, so that we don't stomp on an existing order when we're migrating.

robincornett:
@cklosowski so I don't think it's an issue with the `auto_increment`, but this does expose a very real source of conflict. Basically, there is currently nothing in place to set the initial order number for the new orders table. So if an order is placed before any orders are migrated, the order number will be ID `1`. That's awkward but potentially not a problem as it's highly unlikely to have an actual `edd_payment` ID that low (since posts, pages, etc. will have been created first).

However, where it gets sticky is if a migration is actively running, or was started and then stopped. For example, I started a migration via CLI and then stopped it. My last _payment_ number is 17446. I started generating some random orders via CLI. The next order number was 17447, as expected--there is an existing `WP_Post` for that ID, but it's a sales log (expected). So I kept going. The new order numbers keep incrementing and do match up with existing post objects...at 17455, I hit an `edd_payment`, and once I restart the migration, that's going to cause a real problem. So this is an issue that's been present all along, I think, but if we want to allow stores to be open during migration, we need to resolve this.

One thing I'm thinking is to set the migration_pending value to the last payment number, instead of just a boolean, and then (minimally if the option exists) manually set the preferred order ID using that value +1. Or we could create a separate option that's the final payment ID, and check for that, and use that for the first new order, and then delete the option once the initial order has been placed--I expect the auto increment would properly take over from there? Meaning if the last payment ID is `5465412`, then the first order ID could be set to `5465413` and the migration could keep percolating in the background, and the second new order would be `5465414` due to the magic of auto increment.

cklosowski:
@robincornett and I walked through this on a call and found that this call is causing the problem:
`$result = $this->get_db()->get_var( "SELECT ID FROM {$this->get_db()->prefix}posts WHERE post_type = 'edd_payment' ORDER BY ID DESC LIMIT 1;" );`

It is also happening in adjustments for `edd_discount` post types. We at some point adjusted the prefix property to not include the `_` in favor of a prefix helper function.

robincornett:
@cklosowski still digging on this one...`{$this->get_db()->prefix}` is returning `wp_` with the underscore, so it's correct, but didn't work correctly for me. Deleting tables and trying again here.

robincornett:
@cklosowski I think I'm ready for you to eyeball this. I haven't been able to replicate the initial order number being correct. Key changes:

- A migration pending option is added to the database when the new orders table is created, if a final EDD Payment ID is found. The option is _named_ `edd_v3_migration_pending`, but it's an integer (the final ID), not a boolean. I'm not actually using the integer anywhere but I'd rather have it and not need it than need it and not have it.
- I updated the migration functions and payments classes to _not_ use `edd_get_order` as they ended up in an infinite loop, once I added backwards compatibility to that function. They now instantiate the orders class directly.
- The shimmed order data is not complete, I expect, but it's possible that it cannot be.
- Both migrations use a new function to check if the data migration part of the update process is complete. If it is, that function deletes the migration pending option and returns true (otherwise returns false).
- A closure has been added to the new `edd_pre_order_history` hook to show a message if it's determined that there is a migration pending.
- The shimmed order object is super minimal (All Access checks for the order completed date, for example, so I wanted that), and I did add some very basic order items, but did not do anything for adjustments or metadata.
- The payment object is probably going to be more accurate until the given order is migrated.

I've tested that the option is created for an existing site when 3.0 is activated, and that the option is deleted in both migration scenarios.

We need to do more checks. The ones I can think of for before or during the migration process:
- downloading a file should still be possible
- licensing API requests should work
- downloading an update file via SL should work
- processing a renewal order should work and the migration should continue without error

includes/payments/functions.php
```diff
@@ -1832,3 +1832,31 @@ function edd_filter_where_older_than_week( $where = '' ) {
...
 }
+
+/**
+ * Gets the payment ID from the final edd_payment post.
+ * This was set as an option when the custom orders table was created.
+ * For internal use only.
+ *
+ * @return false|int
+ */
+function edd_get_final_payment_id() {
```
cklosowski, line 1843: This is another great candidate for an `_` function. We are labeling it for `internal use only`, so let's prefix it in a common way that identifies that. Which will make it easier for us to remove later as well.

While we are doing this issue, please make an issue for 3.1 that mentions all of these methods/functions so we can mark them as deprecated in 3.1. If we log them all now, we have less of a chance that we'll forget about them and end up maintaining them long term.

cklosowski (changes requested):

cklosowski:
@robincornett On a code level this all looks good. Going to setup a new migration site and manually set the migration pending option so I can just test the following:
- [] Stripe Renewal webhook
- [] Software Licensing Activation
- [] Software Licensing Deactivation
- [] Software Licensing Get Version
- [] Software Licensing Check License
- [] Frontend messaging

robincornett:
The option should be set as soon as the tables are created, as long as there is at least one existing `edd_payment` post. Also note that if you manually set the option but have run the upgrades already you'll have to comment out the line that deletes the option because it will get deleted.

cklosowski:
In testing, we have an issue with All Access, which we will need for the EDD site during the migration and other users who are upgrading.

The issue lies in the way that All Access pulls if a pass is valid or not.

In the class we run the following:
```
// Get the list of ids in this payment that have previously been set to "active" (if any).
$all_access_active_ids = edd_get_payment_meta( $this->payment->ID, '_edd_aa_active_ids', true );
$all_access_active_ids = empty( $all_access_active_ids ) ? array() : $all_access_active_ids;
```
Because this calls the `get_meta` method, it is not finding the meta values in the order. We can solve this I think on a case by case basis, by maybe setting a 'state' property of `EDD_Payment` by saying when we are or aren't in the back compat mode. If we are, we can short the `EDD_Payment::get_meta` and just call `get_post_meta` directly?

If we aren't in backwards compatibility we can just go forward with the order meta.
We could add a flag here:
```
if ( ! $this->order || is_wp_error( $this->order ) ) {
	$this->is_back_compat = true;
	return _edd_get_final_payment_id() ? $this->_setup_compat_payment( $payment_id ) : false;
}

$this->is_back_compat = false;
```

Then in the `get_meta` method we could do something like:
```
public function get_meta( $meta_key = '_edd_payment_meta', $single = true ) {
	if ( $this->is_back_compat ) {
		return get_post_meta( $this->ID, $meta_key, $single );
	}
	
	$meta = edd_get_order_meta( $this->ID, $meta_key, $single );
```

I'm not sure if this has other implications, but it would allow us access to the primary object via the back compat loading, as well as the meta calls. I don't think we have to go much further than that.

cklosowski:
@robincornett I'm not married to this solution exactly but I think a way to identify when we are in 'back compat' mode would be an ideal solution to address this and any other issues we run into.

robincornett:
Your suggestion seems reasonable to me, and it works (added code and unit test in 6633cd7). Confirmed that it does not work if the flag is not set.

cklosowski:
This is working as I expect. I was able to get renewals to come in from Stripe via actual webhooks (using Stripe Test Clocks).

cklosowski (approved):

PR #9139, opened 2022-04-14
Make sure custom object meta is deleted along with the object

### LINKED ISSUE #9138: Order meta should be deleted when an order is deleted
Looking at our core code, `edd_destroy_order` doesn't have any hooks, and neither does `edd_delete_order`, which `edd_destroy_order` calls.

However, looking at the BerlinDB code for deleting an item, it looks like in 3.0, all order meta should be deleted when an order is deleted from the database, so EDD core _should_ be handling this automatically. However, it looks like none of the order metadata is being deleted from the database, so we need to look into why that's not happening.

So I think we need to look into core to figure out why the order metadata isn't being deleted, and provide a hook for extensions to be able to manage things like this.

_Originally posted by @robincornett in https://github.com/awesomemotive/edd-upload-file/issues/55#issuecomment-1099164960_

Edited to add: logging in the BerlinDB `delete_all_item_meta` method, it looks like everything is fine until EDD calls the `delete_metadata_by_mid` function. That's returning `false` and the order metadata isn't deleted when the order is.

robincornett:
Fixes #9138

Proposed Changes:
1. Runs the meta type parameter through `add_prefix` before attempting to delete the metadata.
2. Adds a unit test to ensure that order meta is deleted when an order is deleted.
3. Adds `edd_order_destroyed` action hook.

Note: this is already fixed in BerlinDB (https://github.com/berlindb/core/issues/84) and I'm thinking EDD is on version 1.0.x of Berlin. This PR is pretty minimal--and actually looking at what was done to resolve the issue in Berlin, probably incomplete.

### DIFF
--- a/includes/orders/functions/orders.php
+++ b/includes/orders/functions/orders.php
@@ -294,7 +294,14 @@ function edd_destroy_order( $order_id = 0 ) {
 		}
 	}
 
-
+	/**
+	 * Action hook for developers to do extra work when an order is destroyed.
+	 *
+	 * @since 3.0
+	 * @param int  $order_id  The original order ID.
+	 * @param bool $destroyed Whether the order was destroyed.
+	 */
+	do_action( 'edd_order_destroyed', $order_id, $destroyed );
 
 	return $destroyed;
 }
--- a/includes/orders/functions/orders.php
+++ b/includes/orders/functions/orders.php
@@ -242,6 +242,14 @@ function edd_delete_order( $order_id = 0 ) {
  */
 function edd_destroy_order( $order_id = 0 ) {
 
+	/**
+	 * Action hook for developers to do extra work when an order is destroyed.
+	 *
+	 * @since 3.0
+	 * @param int  $order_id  The original order ID.
+	 */
+	do_action( 'edd_pre_destroy_order', $order_id );
+
 	// Delete the order
 	$destroyed = edd_delete_order( $order_id );
 
--- a/includes/database/engine/class-query.php
+++ b/includes/database/engine/class-query.php
@@ -2397,9 +2397,12 @@ class Query extends Base {
 			return;
 		}
 
+		// Maybe apply the prefix to the meta type.
+		$meta_type = $this->apply_prefix( $this->item_name );
+
 		// Delete all meta data for this item ID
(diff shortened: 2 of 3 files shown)


includes/orders/functions/orders.php
```diff
@@ -294,7 +294,14 @@ function edd_destroy_order( $order_id = 0 ) {
...
 
-
+	/**
+	 * Action hook for developers to do extra work when an order is destroyed.
+	 *
+	 * @since 3.0
+	 * @param int  $order_id  The original order ID.
+	 * @param bool $destroyed Whether the order was destroyed.
+	 */
+	do_action( 'edd_order_destroyed', $order_id, $destroyed );
```
robincornett, line 304: @cklosowski I think what I would prefer to do over this is to add a more extendable hook within the BerlinDB query class `delete_item` method. This, for example:

```php
do_action( $this->apply_prefix( "{$this->item_name}_deleted" ), $item_id, $result );
```

Running `edd_delete_order` would then automatically generate the following hooks straight away:
- `edd_order_deleted`
- `edd_order_item_deleted`
- `edd_order_address_deleted`

Since that is in a BerlinDB class, though, I'm not confident it would be safe to do, as it doesn't exist as a hook in BerlinDB at this time. If you think it's worth doing, though, I can submit a PR there.

cklosowski: I think it is an interesting idea and worth pursuing. Let's go ahead and open a PR on BerlinDB for it, but keep the code in EDD core as you have done it here for the time being.

We've already had discussions about getting up to the latest BerlinDB for EDD 3.1, and if we can work to get this into a future BerlinDB release, then we can take advantage of it.

cklosowski:
Testing this, but pushed a small addition to the tests to verify that `non-unique` meta removes all meta values.

It should but it triggers a different layer of the DB when you have a non-unique.

@robincornett I also adjusted the order of the `assertEquals` call in `test_delete_should_delete_metadata()`. The `assertEquals` (and other compare methods) should be written with the order of `expected, actual`:
https://www.geeksforgeeks.org/phpunit-assertequals-function/

cklosowski (approved):

PR #9144, opened 2022-04-19
Make editing order dates affect 'Completed Date', not 'Created Date'

### LINKED ISSUE #9053: 3.0 - Order dates need to be reviewed/evaluated
### Expected behavior
Per @cklosowski in slack, the order dates should be:
- `date_created`: Saved the date created as the date it was added to the DB (thats a historical reference point)
- `date_completed`: Set the date completed to when the payment was actually recieved
- `date_modified`: And editing it should have edited the date <del>completed</del> modified (rgc, I think modified is correct)

Invoices, for example, uses the completed date to display the order date on the invoice.

### Actual behavior
It looks like EDD core uses the `date_created` column for displaying, querying, ordering, creating, and editing orders.

1. Manually create an order in EDD 3.0. Make sure to set an order date that is different from today.
2. View the order in the database. The `date_created` column will be the order date, while the `date_completed` column will be today's date.
3. Activate Invoices. View the invoice for the order. It will show today's date for the order.

This came up in Slack because a manually added subscription order was showing today's date on the invoice. It looks like 3.0 is consistent in using the `date_created` for everything related to saving, modifying, and displaying an order. However, we intentionally used `date_completed` in Invoices, for example, because our understanding was that it was, in fact, the order's completed date.

Note that updating a payment in EDD 2.x changes the object `post_date`, not the `_edd_completed_date` metadata.

cklosowski:
Fixes #9053 

This PR has some substantial changes...so here goes.

# Proposed Changes:
1. Introduces 6 new functions and 3 new filters:
 * `edd_complete_order_status_keys()` with filter `edd_complete_order_statuses`
 * `edd_complete_order_statuses()`
 * `edd_incomplete_order_status_keys() with filter `edd_incomplete_order_status_keys`
 * `edd_incomplete_order_statuses()`
 * `edd_recoverable_order_status_keys()` with filter `edd_recoverable_order_status_keys`
 * `edd_recoverable_order_statuses()` <-- We can change what this returns since it was introduced in 3.0. However we'll keep the filter in the `edd_recoverable_order_status_keys()` in case a developer has coded against that so far.
2. Updates the Order Details metabox to have the `Date Completed` and `Date Created` fields shown, but the `Date Created` is read-only/disabled and has a tool tip defining the difference of the two.
* There is JS to handle showing/hiding the fields for the `Date Created` based off the selection of `status`. If the status selected is a 'complete' state, we show it, otherwise we hide it. It is subsequently only entered into the DB when we are in a complete status.
* image (1)
4. Ensures `date_refundable` is added to the column when a `date_completed` is added/updated.
5. Adds any manually added order to default to `pending` so that we can run `edd_complete_purchase` on manually created orders, which is what triggers refundable status.
6. Updates checks for 'completed', 'recoverable' statuses (usually a hard coded array) to use the new helper functions.
7. Adds unit tests around new helper functions.
8. Adds a `wp_localize_script` output on the add/edit order screens with the collection of order status keys
```
	// Add order statuses for use in jQuery
	wp_localize_script(
			'edd-admin-orders',
			'edd_order_statuses',
			array(
					'completed'   => edd_complete_order_status_keys(),
					'incomplete'  => edd_incomplete_order_status_keys(),
					'recoverable' => edd_recoverable_order_status_keys(),
			)
	);
```
# Things to test:
1. Add new order
* Pending
* Complete
2. Test editing an existing order
* Move from Pending > Complete
* Adjust Completed Date
3. Test actual purchases
* Stripe
* Off-site Gateway (PayPal)
* Free Downloads


includes/orders/functions/actions.php
```diff
@@ -119,29 +119,34 @@ function edd_add_manual_order( $args = array() ) {
...
+		$minute = sanitize_text_field( $order_data['edd-payment-time-min'] );
+
+		// Restrict to our high and low
+		if ( $minute > 59 ) {
+			$minute = 59;
+		} elseif ( $minute < 0 ) {
+			$minute = 00;
+		}
+
+		if ( ! empty( $date ) ) {
```
robincornett, line 144: I think this is where the fatal error is coming from...it's maybe happening because of the hour/minute fields being blank. The `$hour` and `$minute` variables are both empty for me.

includes/admin/reporting/export/class-batch-export-earnings-report.php
```diff
@@ -212,7 +212,7 @@ public function print_csv_rows() {
 			}
 
 			// Allows extensions with other 'completed' statuses to alter net earnings, like recurring.
-			$completed_statuses = apply_filters( 'edd_export_earnings_completed_statuses', array( 'complete', 'revoked' ) );
+			$completed_statuses = apply_filters( 'edd_export_earnings_completed_statuses', edd_complete_order_statuses() );
```
robincornett, line 215: `edd_complete_order_statuses` is returning a different kind of array than what you're replacing.

includes/orders/functions/orders.php
```diff
@@ -1375,42 +1375,125 @@ function edd_get_net_order_statuses() {
...
+	return array_intersect_key( edd_get_payment_statuses(), array_flip( edd_complete_order_status_keys() ) );
+}
+
+/**
+ * Get the order status array keys which are considered incomplete.
+ *
+ * @since 3.0
+ * @return array An array of order status keys which are considered incomplete.
+ */
+function edd_incomplete_order_status_keys() {
```
robincornett, line 1438: Two comments on this (and the new functions in general):
- I would prefer for them to be named `edd_get_` just because that makes it really clear that it is intended to return something. That would also make it consistent with the net/gross order status functions we already introduced (`edd_get_net_order_statuses`, eg).
- In that same vein, I would love for `edd_get_incomplete_order_statuses` to be consistent with the net/gross status functions, and return the same kind of array (sans labels).
- Do we _need_ the functions which return the statuses with labels? I feel like that introduces some extra complexity which may not be necessary.

includes/orders/functions/orders.php
```diff
@@ -1375,42 +1375,125 @@ function edd_get_net_order_statuses() {
...
+	return apply_filters( 'edd_recoverable_order_status_keys', $statuses );
+}
+
+/**
+ * Get the order statuses which are considered non-completed.
+ *
+ * @since 3.0
+ * @return array An array of order statuses which are considered not completed.
+ */
+function edd_recoverable_order_statuses() {
```
robincornett, line 1497: Because this changes the array to include labels, it's breaking the order `is_recoverable` method. I am very strongly inclined to not have `_keys` methods.

includes/admin/payments/orders.php
```diff
@@ -1031,17 +1034,19 @@ function edd_order_details_attributes( $order ) {
...
-				<div class="edd-admin-box-inside">
+				<?php $hide_date_completed = empty( $date_completed ) ? ' hidden' : ''; ?>
+				<div class="edd-admin-box-inside completed-date-wrapper <?php echo $hide_date_completed; ?>">
 					<div class="edd-form-group">
-						<label for="edd-payment-date" class="edd-form-group__label"><?php esc_html_e( 'Date', 'easy-digital-downloads' ); ?>
+						<label for="edd-payment-date" class="edd-form-group__label"><?php esc_html_e( 'Date Completed', 'easy-digital-downloads' ); ?>
 						</label>
 						<div class="edd-form-group__control">
-							<input type="text" id="edd-payment-date" class="edd-form-group__input edd_datepicker" name="edd-payment-date" value="<?php echo esc_attr( $order_date->format( 'Y-m-d' ) ); ?>"placeholder="<?php echo esc_attr( edd_get_date_picker_format() ); ?>"/>
+							<?php $date_completed_value = ! empty( $date_completed ) ? esc_attr( $date_completed->format( 'Y-m-d' ) ) : ''; ?>
```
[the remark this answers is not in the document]
robincornett: I think the reason it's working more generally on the release branch is that the date/time fields are populated by default. I can actually replicate the fatal error on the release branch by clearing out the fields (on that branch they have a value by default when the screen loads).

includes/orders/functions/orders.php
robincornett: Note: we have no tests covering the `is_recoverable` method and we probably should.

robincornett (changes requested):
In addition to the individual comments, I'm noting some things that would still need to be changed:
- the `date_created` needs to be updated in:
    - the purchase history template
    - `edd_customers_view` (Recent Orders)
    - the payments table (`column_default`)
    - several of the exporters
    - `setup_payment` in the `EDD_Payment` class
    - I think in the payments query class as well
    - the shortcode receipt

I _think_ the upgrade process should be okay and actually not need to be changed, but I still need to verify. I am wondering if we need to be concerned about existing installs which may already have significantly different created/completed values based on adding new orders.

includes/admin/reporting/export/class-batch-export-earnings-report.php
cklosowski: @robincornett Right this is one of those things I updated, I need to change it to the keys.

includes/orders/functions/orders.php
```diff
@@ -1375,42 +1375,125 @@ function edd_get_net_order_statuses() {
...
+ */
+
+/**
+ * Get the order status array keys which are considered completed.
+ *
+ * @since 3.0
+ * @return array An array of order status keys which are considered completed.
+ */
+function edd_complete_order_status_keys() {
+	$statuses = array( 'publish', 'complete', 'completed', 'partially_refunded', 'revoked', 'refunded' );
```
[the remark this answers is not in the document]
cklosowski: These have different purposes.
`edd_get_gross_order_statuses` is related to reporting where as this is used to determine if an order is at it's final state. While they are similar in what they return, if we combine them we are mixing the checks for reporting with the checks for state, which would not allow us to expand upon these statuses later, if the time comes.

includes/orders/functions/actions.php
cklosowski: Okay we can solve that, just need to set a default `hour` and `minute`

includes/admin/payments/add-order.php
```diff
@@ -25,12 +25,23 @@ function edd_add_order_page_content() {
...
+			)
+	);
+
 	// Create empty order object to pass to callback functions.
 	$order = new Order( array(
 		'id'              => 0,
 		'parent'          => 0,
 		'order_number'    => 0,
-		'status'          => 'complete',
+		'status'          => 'pending',
```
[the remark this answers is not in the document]
cklosowski: @robincornett That's fair, I was playing around with it while testing and I was mostly testing around pending, and it started to make sense to start all orders as `pending`. We can revert that.

includes/admin/payments/orders.php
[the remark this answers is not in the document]
cklosowski: We should put some validation on those.

includes/admin/payments/orders.php
```diff
@@ -1051,17 +1056,34 @@ function edd_order_details_attributes( $order ) {
...
 								<?php esc_html_e( 'Minute', 'easy-digital-downloads' ); ?>
 							</label>
-							<input type="number" class="edd-form-group__input small-text" min="0" max="59" step="1" name="edd-payment-time-min" id="edd-payment-time-min" value="<?php echo esc_attr( $order_date->format( 'i' ) ); ?>" />
+							<?php $minute_completed_value = ! empty( $date_completed ) ? esc_attr( $date_completed->format( 'i' ) ) : ''; ?>
+							<input type="number" class="edd-form-group__input small-text" min="0" max="59" step="1" name="edd-payment-time-min" id="edd-payment-time-min" value="<?php echo $minute_completed_value; ?>" />
 						</div>
 					</fieldset>
 				</div>
 
+				<div class="edd-admin-box-inside">
```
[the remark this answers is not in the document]
cklosowski: It seems very awkward to not show any dates at all when an order is pending:

If you are attempting to trouble shoot a pending payment, based on this the only way get any idea of when things happened is to view the notes:

image

We could maybe move it to the `order extras` area, as it's not really part of it's 'status' but still allows for proper troubleshooting?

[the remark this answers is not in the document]
cklosowski: @robincornett I'd still like to keep it on there, as I think it is very confusing without it, does moving it here make more sense?
image

includes/orders/functions/actions.php
cklosowski: This is fixed now and has validation to make sure when the fields are required, they are filled out.

includes/admin/payments/orders.php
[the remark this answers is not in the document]
robincornett: Sure, I can run with that. I don't think I care for hiding the date completed inputs. Currently, if the order status is set to pending, even if the inputs have values (as they do in the release branch), the `date_completed` is properly set to `null`. Of course, in the release branch, the order date does still show (using the created date).

PR #9149, opened 2022-04-20
Register customer emails/addresses tabs/views

### LINKED ISSUE #9148: 3.0 - Move customer emails/addresses to tabs on customer details screen
We should register new tabs for customer addresses and (I think) emails on the customer details screen.

### Justification or use case
For a store which does not collect customer addresses at all, it's been reported that it seems off to display physical addresses above recent orders. Even for a storefront which does collect addresses (taxes), that's not necessarily "primary" information.

Since every customer does have an email address, that's maybe less important to move; however, the primary address is already displayed at the top of the main details screen. Additionally, creating an Emails tab would be consistent with the order details tabs:

image

robincornett:
Fixes #9148

Proposed Changes:
1. Registers emails and addresses tabs for the customer view.
2. If there are no physical addresses stored for a customer, the addresses tab is unset and not displayed at all.
3. If there are no physical addresses, the view just returns early, since it should not be called at all.
4. The "add new email address" has been moved out of the condition so that if a customer has no emails, it will be possible to add a new address. Currently the release branch only shows the "add new" if a customer does have emails saved already.


cklosowski:
When adding a new email address from the email address view, the page redirects back to the customer overview. We need to return to the same interface with the fragment `#edd_general_emails`

Otherwise this looks good to me.

cklosowski (changes requested):
Just one change noted, on how the page redirects when a new email address is added.

cklosowski:
Actually one other change I just noticed @robincornett We need to carry the header to the sub-sections:

Existing View
image

New Emails tab:
image

We have no indication of what customer we're even working with here.

cklosowski (changes requested):
@robincornett one more thing I missed as I was testing with a set of users that didn't apparently have physical addresses, is that when we 'delete' a physical address, it returns us back to the primary customer screen, not the Addresses screen. So we need to do the same thing we did with email addresses on those action links.

cklosowski (approved):

PR #9151, opened 2022-04-21
Reset date string if invalid

### LINKED ISSUE #9150: 3.0 - fatal error when an invalid/incomplete date string is passed to date function
### Expected behavior
When I save a new order with an empty date, I expect no fatal errors.

### Actual behavior
When I save a new order with an empty date, I get this:

```
[21-Apr-2022 12:59:03 UTC] PHP Fatal error:  Uncaught Exception: DateTime::__construct(): Failed to parse time string ( ::00) at position 0 (:): Unexpected character in \wp-content\plugins\easy-digital-downloads.30\includes\libraries\Carbon.php:291
Stack trace:
#0 \wp-content\plugins\easy-digital-downloads.30\includes\libraries\Carbon.php(291): DateTime->__construct(' ::00', Object(DateTimeZone))
#1 \wp-content\plugins\easy-digital-downloads.30\includes\utils\class-date.php(38): Carbon\Carbon->__construct(' ::00', Object(DateTimeZone))
#2 \wp-content\plugins\easy-digital-downloads.30\includes\class-utilities.php(239): EDD\Utils\Date->__construct(' ::00', Object(DateTimeZone))
#3 \wp-content\plugins\easy-digital-downloads.30\includes\orders\functions\actions.php(143): EDD\Utilities->date(' ::00', 'GMT+0', false)
#4 C:\ in \wp-content\plugins\easy-digital-downloads.30\includes\libraries\Carbon.php on line 291
```

1. Visit `/wp-admin/edit.php?post_type=download&page=edd-payment-history&view=add-order` to create a new order.
2. Clear out the date, hour, and minute fields, but otherwise create an order.
3. Save the order.

PHP Version: 7.4
WordPress Version: 5.9.3

Encountered in the process of testing/pondering #9053. We discussed making sure that a full/complete date string be constructed when building an order--we can improve that--but I think we also need to make sure that the utility `date` function is [working with a valid date string here](https://github.com/awesomemotive/easy-digital-downloads/blob/release/3.0/includes/class-utilities.php#L223). 

Since `strtotime` returns false if a string cannot be converted to a time, I think adding that as a check and resetting the `$date_string` variable to `'now'` if it returns false seems to cover the most scenarios.

robincornett:
Fixes #9150

Proposed Changes:
1. In `EDD()->utils->date`, adds a check for whether the incoming string is a valid date/time string. If not, resets it to `'now'`.
2. Adds a unit test which tests an invalid date string to ensure it returns a valid date object.

### DIFF
--- a/includes/admin/payments/actions.php
+++ b/includes/admin/payments/actions.php
@@ -85,29 +85,16 @@ function edd_update_payment_details( $data = array() ) {
 
 	$order_update_args = array();
 
-	$unlimited  = isset( $data['edd-unlimited-downloads'] ) ? '1' : null;
-	$new_status = sanitize_key( $data['edd-payment-status'] );
-	$date       = sanitize_text_field( $data['edd-payment-date'] );
-	$hour       = sanitize_text_field( $data['edd-payment-time-hour'] );
-
-	// Restrict to our high and low
-	if ( $hour > 23 ) {
-		$hour = 23;
-	} elseif ( $hour < 0 ) {
-		$hour = 00;
-	}
-
-	$minute = sanitize_text_field( $data['edd-payment-time-min'] );
-
-	// Restrict to our high and low
-	if ( $minute > 59 ) {
-		$minute = 59;
-	} elseif ( $minute < 0 ) {
-		$minute = 00;
-	}
+	$unlimited   = isset( $data['edd-unlimited-downloads'] ) ? '1' : null;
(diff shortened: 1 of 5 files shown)


cklosowski:
@robincornett Are we handling this too far down the chain? The original bug was identified when editing a payment and not completing the form fields without one of the fields being populated.

If we default any invalid date/time to `now` I believe we're making an assumption. Should we not throw an exception and handle that?

cklosowski (approved):

PR #9153, opened 2022-04-22
Create order status helper functions

### LINKED ISSUE #9152: 3.0 - Add additional order status arrays
Per dev discussion today, we are going to create additional `edd_get_*_statuses` helper functions, in addition to the existing gross, net, and recoverable functions.

### Justification or use case
This will allow us to make sure that the order statuses used throughout EDD are consistently defined.

This is to cover part of what we encountered while working on #9053.

robincornett:
Fixes #9152

Proposed Changes:
1. Creates a statuses.php file and adds new order status functions, `edd_get_complete_order_statuses` and `edd_get_incomplete_order_statuses`. Existing similar functions have been moved there as well.
2. Each of the new functions accepts an optional parameter--if set to `true`, it will return the statuses with their appropriate labels.
3. Replaces most manually defined status arrays with the new functions.
4. Note the exception specifically called out in `edd_complete_purchase`.


robincornett:
@cklosowski I think that's a reasonable question. I think we would have to go with `gross`, at least for now, since querying net gets a little more complex due to refunds. It might be worth opening an issue to review/update stats to handle net queries to match what we've done with the individual download calculations.

cklosowski:
@robincornett Based off the work you did for the other reporting, worth it for 3.0? or do we want to make that a 3.1 enhancement?

cklosowski:
The benefit of the 3.1 or later is we can add a full interface and method of viewing all reports in `gross` or `net` as a user based setting.

robincornett:
I think using `gross` for these for now is okay for 3.0. To refactor all stats/reports to get net is something I'd like to sit with and see if we can build some models for to set all reports and all status up to use the same query logic, instead of building it out multiple times in multiple places, so that's something to push for for 3.1.

cklosowski:
:+1: Once CI passes for this i'll approve and merge.

cklosowski (approved):

Issue #9179, opened 2022-05-13
3.0 - Create an order item helper method to indicate that the file(s) can be downloaded


robincornett:
### Explain your enhancement (please be detailed)
We should add a method to the `Order_Item` class to indicate somehow that files, for example, for an order item, can be downloaded.

### Justification or use case
Making a helper method will allow us to be consistent in applying the logic. For example, our download history template right now checks if an order item status is `complete` or `partially_refunded`, but the order item template for the order details screen just checks if the status is `complete`, and will not show links for partially refunded items.

This code works but `is_downloadable` is likely not the language we want long term:

```php
	/**
	 * Checks the order item status to determine if it is downloadable.
	 *
	 * @since 3.0
	 * @return bool
	 */
	public function is_downloadable() {
		return in_array( $this->status, array( 'complete', 'partially_refunded' ), true );
	}
```

cklosowski:
After some discussion, we decided to go with the verbiage `is_deliverable` to determine if an order item is at the point where it can be provided to a customer.

PR #9192, opened 2022-05-16
Update download file URL, access evaluation

### LINKED ISSUE #9182: 3.0 - Bundle downloads failing with Error 102: You do not have permission to download this file
### Expected behavior
I expect to be able to download all the files from bundle purchases, but in the recent 3.0 branch these are failing.

### Actual behavior
I get the following error:

`Error 102: You do not have permission to download this file`

This does not happen with normal downloads, just bundles. I wil

1) Create a bundle product. In my case I had two products assigned
2) Make sure each product has a single file associated with it (or multiple).
3) Purchase the bundle and try to download the files. You should see the above error.

PHP Version: 8.0.0

EDD Version (or branch): 3.0 (latest)

WordPress Version: 5.9

robincornett:
Fixes #9182 
Branched from #9179 due to using new `edd_get_deliverable_order_item_statuses` function.

Proposed Changes:
This ended up being a bit thornier to sort through than initially intended due to previously ignoring `0` price IDs as actual price IDs. Also I think Berlin is doing something that maybe we don't want it to do and I need to look into that.
1. The primary change was to add a secondary set of checks to the `edd_order_grants_access_to_download_files` method to check for items being downloaded as part of a bundle, since the download URL will not contain the correct order item/price ID. This secondary check loops through all order items and checks if the current download/price ID combo matches a defined bundle item.
2. The `edd_process_signed_download_url` function has been updated to get the order object and use the properties from that instead of going through the payment meta backwards compatibility process.
3. That function was changing the price ID to an empty string before passing it to the "grants access" function if the price ID was empty, but that's incorrect since a price ID can legitimately be 0.
4. However, the "grants access" function needs to unset the price ID if it's `null` before handing off to Berlin, because it looks like Berlin casts that to `AND edd_oi.price_id = 0` instead of doing `IS NULL`.
5. In related places, I've checked through functions which use the `edd_get_download_file_url` function and made sure they correctly set the price ID to `null` as needed, and then changed that function to not add the price ID to the `eddfile` parameter if it is null, to remove potential casting of `null` to `0`. I expect this particular change to need to be scrutinized quite a bit, and there may still need to be a check if a `0` value is passed through when it should be `null`. Possibly the final bundle check in "grants access" could do a price ID check too since it's already looping through the items?
6. I've added a couple new unit tests and tweaked an existing one to cover some of the price ID gaps.

We should make sure all of the following still work:
- [ ] file download URLs in emails
- [ ] file download URLs in purchase receipts
- [ ] file download URLs in download history
- [ ] file download URLs in all access
- [ ] file download URLs copied from order tails
- [ ] file download URLs for single price items
- [ ] file download URLs for variably priced items
- [ ] file download URLs for bundled items
- [ ] SL download URLs
- [ ] free downloads URLs
Note: due to changes in how price IDs are handled, single price downloads should specifically be tested. Variable ones as well, but because we've used `0` interchangeably as both "no price ID" and an actual price ID this is the most likely potential failure point, I think.

### DIFF
--- a/includes/emails/tags.php
+++ b/includes/emails/tags.php
@@ -288,7 +288,6 @@ function edd_email_tag_download_list( $payment_id ) {
 		return '';
 	}
 
-	$payment_data  = $payment->get_meta();
 	$download_list = '<ul>';
 	$cart_items    = $payment->cart_details;
 	$email         = $payment->email;
@@ -331,12 +330,12 @@ function edd_email_tag_download_list( $payment_id ) {
 
 					if ( $show_links ) {
 						$download_list .= '<div>';
-							$file_url = edd_get_download_file_url( $payment_data['key'], $email, $filekey, $item->product_id, $item->price_id );
-							$download_list .= '<a href="' . esc_url_raw( $file_url ) . '">' . edd_get_file_name( $file ) . '</a>';
-							$download_list .= '</div>';
+						$file_url       = edd_get_download_file_url( $order->payment_key, $email, $filekey, $item->product_id, $item->price_id );
+						$download_list .= '<a href="' . esc_url_raw( $file_url ) . '">' . edd_get_file_name( $file ) . '</a>';
+						$download_list .= '</div>';
 					} else {
 						$download_list .= '<div>';
-							$download_list .= edd_get_file_name( $file );
+						$download_list .= edd_get_file_name( $file );
 						$download_list .= '</div>';
 					}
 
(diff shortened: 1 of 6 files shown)


cklosowski:
- [x] file download URLs in emails (Bundle download type, product itself is a variable priced product)
- [x] file download URLs in purchase receipts
- [ ] file download URLs in download history
- [ ] file download URLs in all access
- [ ] file download URLs copied from order tails
- [x] file download URLs for single price items
- [x] file download URLs for variably priced items
- [x] file download URLs for bundled items
- [ ] SL download URLs
- [ ] free downloads URLs

includes/emails/tags.php
```diff
@@ -288,7 +288,6 @@ function edd_email_tag_download_list( $payment_id ) {
 		return '';
 	}
 
-	$payment_data  = $payment->get_meta();
```
robincornett, line 291: @cklosowski noting specifically that I removed this because I'm now using the order's payment key directly and there is no need to get the payment meta.

robincornett:
Both bundle URLs (in the email tag and receipt shortcode) were incorrectly passing the parent item price ID to the `edd_get_download_file_url` function. With the stricter check against the order item instead of the general order status, this could definitely have caused issues.

cklosowski (approved):

PR #9210, opened 2022-05-25
Adding an index for object_type and object_id to the transactions table #9209

### LINKED ISSUE #9209: 3.0 - Order Transactions produces a slow query on large sites.
### Expected behavior
I expect the order transactions table to not be slow when queried by object_type and object_id (used in the `is_recoverable` checks)

### Actual behavior
When we run `is_recoverable` on a large `wp_edd_order_transactions` table, we end up with slow queries.

1) Have a large set of transactions, and view the purchase history.

PHP Version: All

WordPress Version: all

Since we use this combination of `object_type` and `object_id` in core, we should add an in

cklosowski:
Fixes #9209 

Proposed Changes:
1. Adds an index of `object_type, object_id` to the `edd_order_transactions` table.

### DIFF
--- a/includes/database/tables/class-order-transactions.php
+++ b/includes/database/tables/class-order-transactions.php
@@ -38,7 +38,7 @@ final class Order_Transactions extends Table {
 	 * @since 3.0
 	 * @var int
 	 */
-	protected $version = 202105291;
+	protected $version = 202205241;
 
 	/**
 	 * Array of upgrade versions and methods
@@ -51,6 +51,7 @@ final class Order_Transactions extends Table {
 		'202002141' => 202002141,
 		'202005261' => 202005261,
 		'202105291' => 202105291,
+		'202205241' => 202205241,
 	);
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/database/tables/class-order-transactions.php
```diff
@@ -139,4 +141,21 @@ protected function __202105291() {
...
+	 *
+	 * @since 3.0
+	 * @return bool
+	 */
+	protected function __202205241() {
+		if ( $this->index_exists( 'object_type_object_id' ) ) {
+			return true;
+		}
+
+		$this->get_db()->query( "ALTER TABLE {$this->table_name} ADD INDEX object_type_object_id (object_type, object_id)" );
```
robincornett, line 156: Looking at the other upgrade functions in the class, shouldn't this be this instead?

```php
		return $this->is_success(
			$this->get_db()->query( "ALTER TABLE {$this->table_name} ADD INDEX object_type_object_id (object_type, object_id)" )
		);
```
and remove the final `return true;`?

robincornett (changes requested):

includes/database/tables/class-order-transactions.php
cklosowski: Every time we're altering for an index, we use the `return true` method of doing this. I pulled this from other upgrade routines that specifically worked around altering to add an index, not a column specifically, so there might be a difference here.

PR #9214, opened 2022-05-25
Make sure a product has variable prices before making a price ID part of the validation

cklosowski:


includes/process-download.php
```diff
@@ -867,7 +867,7 @@ function edd_process_signed_download_url( $args ) {
 	}
 
 	$order_parts = explode( ':', rawurldecode( $_GET['eddfile'] ) );
-	$price_id    = isset( $order_parts[3] ) ? (int) $order_parts[3] : null;
+	$price_id    = edd_has_variable_prices( $order_parts[1] ) && isset( $order_parts[3] ) ? (int) $order_parts[3] : null;
```
robincornett, line 870:
```suggestion
	$price_id    = isset( $order_parts[3] ) && edd_has_variable_prices( $order_parts[1] ) ? (int) $order_parts[3] : null;
```
I think if we change the order then the edd function would only be called if the `isset()` is true, right? And might save a microsecond somewhere?

cklosowski: @robincornett I don't believe it will make a difference, as it is always there and sets to `0` as a default for `null` (the old issue showing it's head). But yeah we can reverse it.

robincornett: I think it could make a difference, as the `edd_get_download_file_url` function now will not set the fourth order part if the download is not a variable product and the price ID is properly set to `null` (which it will be for new and migrated orders) -- that was changed in #9179. Tiny but feels worth doing.

robincornett:
@cklosowski I can approve this, but I'm thinking through this more, and wondering if it might instead be better to move the `has_variable_prices` check to the `edd_order_grants_access_to_download_files` function, in the section of code where we're currently checking for a bundle item. If we add an extra check immediately after the download object has already been retrieved, it seems like we could catch any/all scenarios where the price ID is `0` incorrectly.

We could update the fallback check in that function to be more along these lines:
```php
	$product_to_check = isset( $args['price_id'] ) && is_numeric( $args['price_id'] ) ? "{$args['product_id']}_{$args['price_id']}" : $args['product_id'];
	foreach ( $order_items as $product_id ) {
		$download = edd_get_download( $product_id );
		if ( ! $download instanceof EDD_Download ) {
			continue;
		}

		// Check if the requested download is not variably priced but incorrectly included a price ID.
		if ( empty( $args['price_id'] ) && $product_id == $args['product_id'] && ! $download->has_variable_prices() ) {
			return true;
		}

		// Check if the requested download is part of a bundle.
		if ( 'bundle' === $download->type && in_array( $product_to_check, $download->get_bundled_downloads() ) ) {
			return true;
		}
	}
```

robincornett (approved):

robincornett:
Updated PR changes:
1. Fixed the `edd_ajax_generate_file_download_link` function to not always cast the price ID to an integer, as this was breaking admin download links for all non-variable products (free/paid irrelevant).
2. Reverted the original change for fixing the download URL if a price ID was incorrectly passed for a non-variable product. I don't necessarily disagree with making that change but not sure we need it, because...
3. Updated the `edd_order_grants_access_to_download_files` function to include non-variable products and an empty (`0`) price ID if the original check fails, to run after the bundle product check. I did this because it should catch devs incorrectly passing a `0` price ID to the download URL generation function somehow.

PR #9217, opened 2022-05-26
Change admin notice if a CLI migration is running

### LINKED ISSUE #9216: 3.0 - When running an upgrade via CLI, prevent admin notice which can allow starting a migration via UI
### Expected behavior
When I start the EDD 3.0 upgrade via WP CLI, I expect to be notified in some way in the UI of the admin area that it is running.

### Actual behavior
When I'm running the migration to EDD 3.0 via CLI, the notice in admin to start the migration is still available, which could lead to potentially running the migration in parallel, which might cause some issues.

1) Start the EDD 3.0 Migration via WP CLI
2) View the admin area, and see that you can also start the migration via UI.

PHP Version: All

WordPress Version: All

robincornett:
Fixes #9216

Proposed Changes:
1. If a CLI migration has been started, adds an option to the database.
2. Loading an admin screen while this option exists will show an admin message "Easy Digital Downloads is performing a database migration via WP-CLI. This message will be removed when the migration is complete." instead of the default EDD migration notice/link. This notice has no link to the upgrade screen.
3. The CLI option is deleted before the legacy data prompt shows.

### DIFF
--- a/includes/admin/upgrades/upgrade-functions.php
+++ b/includes/admin/upgrades/upgrade-functions.php
@@ -210,71 +210,77 @@ function edd_show_upgrade_notices() {
 			if ( $has_orders ) {
 				?>
 				<div class="updated">
-					<p>
-					<?php
-					printf(
-						__( 'Easy Digital Downloads needs to upgrade the database. %sLearn more about this upgrade%s.', 'easy-digital-downloads' ),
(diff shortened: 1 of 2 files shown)


cklosowski (approved):

robincornett:
@cklosowski thinking about this more and wondering: what happens if the CLI migration is interrupted? Or if one admin begins a UI migration and then another logs in? Or if someone starts a CLI migration and wanted to finish via UI?

PR #9221, opened 2022-05-27
Try/net stats

cklosowski:


cklosowski:
@robincornett I'm ready to have this looked at a bit, and get some ideas on where we go next with it. Ideally this adds two things:
1. The main graph becomes 'net' instead of 'gross'
2. Adds a quick tile for 'net' status for the period shown, to help assess net vs gross, which is something I use on the EDD site a lot.

includes/admin/reporting/reports-callbacks.php
```diff
@@ -56,9 +56,24 @@ function edd_overview_sales_earnings_chart() {
 	$statuses = apply_filters( 'edd_payment_stats_post_statuses', $statuses );
 	$statuses = "'" . implode( "', '", $statuses ) . "'";
 
-	$results = $wpdb->get_results(
+	$earnings_results = $wpdb->get_results(
 		$wpdb->prepare(
-			"SELECT COUNT(id) AS sales, SUM({$column}) AS earnings, {$sql_clauses['select']}
+			"SELECT SUM({$column}) AS earnings, {$sql_clauses['select']}
```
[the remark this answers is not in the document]
cklosowski: Thinking about this further, we aren't looking for quantity sold in this report, just number of orders defined. So we can use a 'count' here since we aren't looking for number of items sold but number of orders processed.

This is different than a units report.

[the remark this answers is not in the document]
cklosowski: @robincornett the `$column` variable is handling removing taxes as our reporting allows the exclusion of taxes on all reports:

`$column       = Reports\get_taxes_excluded_filter() ? '(total - tax)' : 'total';`

For the purposes of the reports, we should keep that checkbox from being the decision maker on when we exclude or include taxes for 'net' purposes.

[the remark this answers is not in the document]
cklosowski: Positive fees seem to already be accounting for it:
Screen Shot 2022-06-24 at 9 34 04 AM
Screen Shot 2022-06-24 at 9 33 45 AM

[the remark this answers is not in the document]
robincornett: Oh, I think that's working as it's going off of the total, not the subtotal, and in downloads we had to work from the subtotal.

robincornett (approved):

includes/class-stats.php
```diff
@@ -3154,4 +3156,38 @@ private function set_date_ranges() {
...
+	 *
+	 * @since 3.0
+	 *
+	 * @return array The statuses of orders to use for the stats generation.
+	 */
+	private function get_revenue_type_statuses() {
+
+		$statuses = edd_get_gross_order_statuses();
+		if ( 'net' === $this->query_vars['revenue_type'] ) {
+			$statuses = edd_get_net_order_statuses();
```
robincornett, line 3171: If this is changed to a `return` statement, then line 3169 can be removed and the final return can just return gross statuses--no need to get them otherwise, and since there's not a filter, we can do an early return.

robincornett (commented):
This seems to be working well for me now! One interesting thing on the monthly/multi day graphs is that if my time zone is different from the store time zone, the points are not on the lines (and you can see that the start/end date is shifted a bit):

image

cklosowski:
@robincornett pull down the latest, I just pushed a lot of changes which should help fix that up. There some queries that were getting mixed up with date formats.

includes/class-stats.php
```diff
@@ -3149,9 +3023,85 @@ private function set_date_ranges() {
...
+	/**
+	 * Generates output for the report tiles when a relative % change is requested.
+	 *
+	 * @since 3.0
+	 *
+	 * @param int|float $total     The primary value result for the stat.
+	 * @param int|float $relative  The value relative to the previous date range.
+	 * @param bool      $reverse   If the stat being displayed is a 'reverse' state, where lower is better.
+	 */
+	private function generate_relative_markup( $total = 0, $relative = 0, $reverse = false ) {
```
robincornett, line 3078: The `$reverse` parameter is not being used in this function, although at least the `get_order_refund_amount` defines it.

I'm also seeing inconsistencies in whether the relative markup is displaying...sometimes it is, sometimes not, on the same tiles. Here's "last quarter":
image

And here is "this quarter":
image

includes/reports/reports-functions.php
```diff
@@ -975,6 +975,9 @@ function default_display_charts_group( $report ) {
 		<?php
 	}
 	?>
+		<div class="chart-timezone">
+			<?php printf( esc_html( 'Chart timezone: %s', 'easy-digital-downloads' ), edd_get_timezone_id() ); ?>
```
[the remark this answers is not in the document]
cklosowski: @robincornett relative styles fixed.

cklosowski:
@robincornett Okay I think the file downloads graph should be good now, and I've fixed the tooltip for negative values showing as a postiive.

includes/class-stats.php
```diff
@@ -3149,9 +3023,89 @@ private function set_date_ranges() {
...
+		$relative_markup  = '';
+
+		$total_output    = $this->maybe_format( $total );
+		$relative_output = '';
+
+		if ( ( floatval( 0 ) === floatval( $total ) && floatval( 0 ) === floatval( $relative ) ) || ( $total === $relative ) ) {
+			$relative_output = esc_html__( 'No Change', 'easy-digital-downloads' );
```
robincornett, line 3072:
```suggestion
			return $relative_markup;
```
Suggesting this as I think returning an empty string here might reduce confusion as to why sometimes a tile will show a trend/percentage change, sometimes "No Change", and sometimes nothing at all. If it only shows a trend or nothing at all, that seems potentially less confusing.

robincornett:
I pushed a couple commits to fix the failing unit test, and use the relative markup method for the file download count. Just one last question/comment....running Buildkite now.

robincornett (approved):

includes/class-stats.php
cklosowski: These are three distinct cases though, and I think `No Change` is actually important to show.

* Trend means we had data to compare against and there was a difference
* No Change means we had data to compare against and there was no difference
* Showing nothing at all means we didn't have any data to compare against.

Removing 'No Change' is factually incorrect. If we are going to change something I think it should be what we show when we have no data to compare against. We could just use an mdash or something similar to show that we don't have an entry for this.

includes/class-stats.php
```diff
@@ -256,23 +257,43 @@ public function get_order_earnings( $query = array() ) {
 		) );
 
 		if ( true === $this->query_vars['relative'] ) {
+
+			$initial_query = "SELECT IFNULL({$function}, 0) AS total
```
robincornett, line 261: I'm finding this a lot easier to read and have two questions:
1. Should the `$initial_query` match the default `else` query? I'm seeing differences (`IFNULL` and the else doesn't have a `where_sql`), but if they should be the same, maybe we can tinker with the logic to just set up that query once and then add the relative as needed.
2. I can't comment on line 250 directly, but I'm seeing `edd_subscription` in the array of statuses twice. Maybe we can run that line through `array_unique` after/outside of `apply_filters`?

robincornett: That would hypothetically change this entire block beginning at line 259 to:

```php
		$initial_query = "SELECT IFNULL({$function}, 0) AS total
			FROM {$this->query_vars['table']}
			WHERE 1=1
			{$this->query_vars['status_sql']}
			{$this->query_vars['type_sql']}
			{$this->query_vars['currency_sql']}
			{$this->query_vars['where_sql']}
			{$this->query_vars['date_query_sql']}";

		$initial_result = $this->get_db()->get_row( $initial_query );

		if ( true === $this->query_vars['relative'] ) {

			$relative_date_query_sql = $this->generate_relative_date_query_sql();

			$relative_query = "SELECT IFNULL({$function}, 0) AS total
				FROM {$this->query_vars['table']}
				WHERE 1=1
				{$this->query_vars['status_sql']}
				{$this->query_vars['type_sql']}
				{$this->query_vars['currency_sql']}
				{$this->query_vars['where_sql']}
				{$relative_date_query_sql}";

			$relative_result = $this->get_db()->get_row( $relative_query );
		}

		$total = null === $initial_result->total
			? 0.00
			: (float) $initial_result->total;

		if ( true === $this->query_vars['relative'] ) {
			$total = $this->generate_relative_markup( floatval( $total ), floatval( $relative_result->total ) );
		} else {
			$total = $this->maybe_format( $total );
		}
```

But that's assuming that the `IFNULL` and `where_query` are acceptable in both queries (non-relative as well as relative).

robincornett (approved):

PR #9224, opened 2022-06-10
Update widgets.php

arshidkv12:
Fixes #

Proposed Changes:
1. `echo esc_attr()` to `esc_attr_e()`


cklosowski:
@robincornett Will you take a look and see if we've already covered this in 3.0? If we haven't let's get it slated for a follow up release for the 3.0 release, 3.0.5 or the like.

cklosowski:
@arshidkv12 After our dev team reviewed this we noticed that you are running the translation version of `esc_attr`, `esc_attr_e` on a non-translatable string.

Is there a reason you are asking for these fields to be run through the translation functions, when there is nothing to translate?

PR #9227, opened 2022-06-15
Return early if download object isn't retrieved

### LINKED ISSUE #9226: 3.0 - fatal error when getting download name if the download doesn't exist
### Expected behavior
When I use the `edd_get_download_name` function, I expect it to not cause a fatal error.

### Actual behavior
When I use the function for a download that doesn't exist, I get this error:

```
[15-Jun-2022 04:52:34 UTC] PHP Fatal error:  Uncaught Error: Call to a member function get_name() on null in easy-digital-downloads\includes\download-functions.php:164
```

1. Run `edd_get_download_name` on a random ID, querying for a download that doesn't exist.

PHP Version: 8
WordPress Version: 6.0

The function returns early if the download ID is empty, or if it's not numeric, but there is no fallback for `edd_get_download()` not retrieving a download.

robincornett:
Fixes #9226

Proposed Changes:
1. If `edd_get_download` does not return a download object, return early.

Noting two things that seem incorrect about the function:
1. The doc block says that it always returns a string, but it does already return `false` if no download ID is provided. I feel like this could be problematic if the function is used without extra logic. We do things if we can't find a customer, for example, like say "customer missing". Without extra checks on this, the download name could just be empty where its used. Maybe that's fine but we should update the doc block.
2. I think the default price ID parameter should be `null` instead of `0` as that's technically correct for no price ID.

### DIFF
--- a/includes/download-functions.php
+++ b/includes/download-functions.php
@@ -147,11 +147,11 @@ function edd_is_free_download( $download_id = 0, $price_id = false ) {
  * @since 3.0
  *
  * @param int $download_id
- * @param int $price_id
+ * @param int|null $price_id
  *
- * @return string
+ * @return false|string
  */
-function edd_get_download_name( $download_id = 0, $price_id = 0 ) {
+function edd_get_download_name( $download_id = 0, $price_id = null ) {
 
 	// Bail if no download ID was passed.
 	if ( empty( $download_id ) || ! is_numeric( $download_id ) ) {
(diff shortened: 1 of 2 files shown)


cklosowski (changes requested):
Overall this looks good, but let's get a unit testing place to make sure we don't introduce this again by accident.

cklosowski (approved):

PR #9240, opened 2022-06-17
Move payment property assignments to order queries class

### LINKED ISSUE #9239: 3.0 - every order should have a currency assigned
### Expected behavior
When I create an order, I expect it to always have a currency assigned, even if the order is free.

### Actual behavior
When using the Free Downloads extension with email verification enabled, that extension creates a pending payment (not an order), and no currency is assigned, even when the order is completed.

1. Activate Free Downloads and make sure that the user has to confirm the download "purchase" by email verification.
2. Generate some free download orders.
3. Check the database. There are no currencies assigned.

PHP Version: 8.0
WordPress Version: 6.0

Although we could fix this in Free Downloads by explicitly setting the currency for the pending payment there, it's still possible to use the EDD Payment class to create an order with no currency assigned. I think that where the order queries class extends the BerlinDB `add_item` method, we should enforce the currency.

Some properties are already being enforced by the payment class, either by default (gateway always defaults to manual) or dynamically (the IP and payment key); we may want to consider moving those to the order queries class, if possible.

robincornett:
Fixes #9239

Proposed Changes:
1. Removes the fallback calculations for IP and payment key from the EDD_Payment class.
2. Extends the BerlinDB `add_item` method for orders and makes sure that every order has an IP, purchase key, and currency set when added to the database.

This needs to be tested with anything which manipulates/saves payments. Specifically, off the top of my head:
- [ ] Free Downloads
- [ ] Recurring
- [ ] Multi Currency

### DIFF
--- a/includes/admin/payments/class-payments-table.php
+++ b/includes/admin/payments/class-payments-table.php
@@ -450,7 +450,8 @@ class EDD_Payment_History_Table extends List_Table {
 		$timezone_abbreviation = edd_get_timezone_abbr();
 		switch ( $column_name ) {
 			case 'amount':
-				$value = edd_display_amount( $order->total, $order->currency );
+				$currency = ! empty( $order->currency ) ? $order->currency : edd_get_currency();
+				$value    = edd_display_amount( $order->total, $currency );
 				break;
 			case 'date':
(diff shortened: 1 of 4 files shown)


mihaijoldis:
This is working fine for me in all 3 cases. Currency is recorded in the database

cklosowski (changes requested):
I'm seeing things working correctly on inserting, but we also need to fix the orders list table to, when there I no gateway defined on the order, to use the store's default currency on display:
Screen Shot 2022-06-17 at 10 51 15 AM

Order 2139 was created on `release/3.0`
Order 2140 was created on `issue/9239`

This would prevent any issues where someone somehow creates an order without a currency assigned.

cklosowski (approved):

PR #9244, opened 2022-06-22
Improve settings title and admin bar menu

cklosowski:
Fixes some admin bar and header issues found during an audit.


includes/admin/settings/display-settings.php
```diff
@@ -23,6 +23,7 @@ function edd_admin_header() {
 	}
 	$numberNotifications = EDD()->notifications->countActiveNotifications();
 	$current_page        = ! empty( $_GET['page'] ) ? $_GET['page'] : '';
+	$is_single_view      = ! empty( $_GET['view'] ) ? true : false;
```
robincornett, line 26:
```suggestion
	$is_single_view      = ! empty( $_GET['view'] );
```

assets/css/admin/style.scss
```diff
@@ -93,12 +93,33 @@ body.post-type-download #screen-meta {
 	padding: 0 20px;
 }
 
+#edd-header-branding {
+	vertical-align: middle;
```
robincornett, line 97: I don't believe this rule actually does anything; think it can be removed.

assets/css/admin/style.scss
```diff
@@ -93,12 +93,33 @@ body.post-type-download #screen-meta {
...
+}
+
 #edd-header img {
-	display: block;
```
robincornett, line 97: Removing this line makes the image layout be inline, which visually adds a little extra space below the image. Setting it to block was intentional to have the padding on the top and bottom of the header be even.

assets/css/admin/style.scss
```diff
@@ -93,12 +93,33 @@ body.post-type-download #screen-meta {
...
 	max-width: 300px;
 	margin: 0 0 0 20px;
 }
 
+.edd-header-page-title-wrap {
+	margin-top: 7px;
+	font-size: 22px;
+	line-height: 28px;
+	vertical-align: top;
+	display: inline;
```
robincornett, line 110:
```suggestion
```

robincornett: This and the vertical-align can be removed because the parent element is using flexbox. `align-items: center;` should be added to that. Additionally, this rule could be simplified to this:

```css
.edd-header-page-title-wrap {
    margin-top: -5px;
    font-size: 22px;
    margin-right: auto;
    padding-left: 7px;
}
```

assets/css/admin/style.scss
```diff
@@ -93,12 +93,33 @@ body.post-type-download #screen-meta {
...
+	margin-right: auto;
+	padding-left: 7px;
+}
+
+.edd-header-seperator {
```
robincornett, line 115: If we are going to make this change can the rule and element be updated to use `separator` please (spelling)?

assets/css/admin/style.scss
```diff
@@ -93,12 +93,33 @@ body.post-type-download #screen-meta {
...
+	margin-top: -2px;
+}
+
+.edd-header-page-title {
+	font-weight: 400;
```
robincornett, line 120: This span is already 400; this rule seems redundant.

includes/admin/admin-bar.php
```diff
@@ -19,37 +19,111 @@
...
+		$count     = $notifications < 10 ? $notifications : '!';
+		$indicator = ' <div class="wp-core-ui wp-ui-notification edd-menu-notification-counter">' . $count . '</div>';
+	}
+
+	// Add the menu
+	$wp_admin_bar->add_menu(
+		array(
+			'id'    => 'edd-store-menu',
+			'title' => 'EDD' . $indicator,
+			'href'  => add_query_arg(
```
robincornett, line 45: Consider updating this to use `edd_get_admin_url()` instead:

```php
$url = edd_get_admin_url(
    array(
        'page' => 'edd-reports',
   )
);

File: includes/admin/admin-bar.php [hunk 8]
```diff
@@ -19,37 +19,111 @@
...
+				admin_url( 'edit.php' )
+			),
+		)
+	);
+
+	$wp_admin_bar->add_menu( array(
+		'parent' => 'edd-store-menu',
+		'id'     => 'edd-store-prodcuts',
+		'title'  => sprintf( __( 'All %1$s', 'easy-digital-downloads' ), edd_get_label_plural() ),
+		'href'  => add_query_arg(
```
robincornett, line 59: This can also use the `edd_get_admin_url()` function.

includes/admin/admin-bar.php
```diff
@@ -19,37 +19,111 @@
...
 			'tab'  => 'gateways'
 		) )
 	) );
 
+	$pass_manager = new \EDD\Admin\Pass_Manager();
+	if ( false === $pass_manager->has_pass() ) {
+		$wp_admin_bar->add_menu( array(
+			'parent' => 'edd-store-menu',
+			'id'     => 'edd-upgrade',
+			'title'  => esc_html__( 'Upgrade to Pro', 'wpforms-lite' ),
```
robincornett, line 118:
```suggestion
			'title'  => esc_html__( 'Upgrade to Pro', 'easy-digital-downloads' ),
```

includes/admin/settings/display-settings.php
```diff
@@ -21,12 +21,81 @@ function edd_admin_header() {
...
+			break;
+		case 'edd-customers':
+			$page_title = __( 'Customers', 'easy-digital-downloads' );
+			break;
+		case 'edd-tools':
+			$page_title = __( 'Tools', 'easy-digital-downloads' );
+			break;
+		default:
+			if ( ! empty( $_GET['page'] ) ) {
+				$page_title = ucfirst( str_replace( 'edd-', '', $current_page ) );
```
robincornett, line 50: Casting a noun to uppercase is problematic from a i18n perspective. Maybe the default should just be "Downloads".

includes/admin/settings/display-settings.php
```diff
@@ -21,12 +21,81 @@ function edd_admin_header() {
...
+
+	$page_title = apply_filters( 'edd_settings_page_title', $page_title, $current_page );
+	if ( ! empty( $page_title ) && empty( $is_single_view ) ) {
+		?>
+		<style>
+			.wrap > h1,
+			.wrap h1.wp-heading-inline,
+			.wrap h2,
+			.page-title-action {
+				display: none;
```
robincornett, line 66: This removes the heading from the screen, which is an accessibility issue. We intentionally un-hid all of the headings in a previous issue/PR. Is there another way we can accomplish this?

includes/admin/settings/display-settings.php
cklosowski: I'm updating this going forward, the point here is to just guess for now, and if we truly want i18n, we'll update all the extensions to load in their own using the filter below.

includes/admin/settings/display-settings.php
cklosowski: This makes all the 'section' labels an H1: https://github.com/awesomemotive/easy-digital-downloads/pull/9244/commits/aec57702a9a3d63c12cd01a96b0f25decc19f46d

assets/css/admin/style.scss
cklosowski: Now that we've converted this to an H1, we need to set it or else it ends up being defaulted to 600, like the admin styles for H1.

includes/admin/settings/display-settings.php
cklosowski: We'll be able to go through other extensions, and use this filter to apply our 'breadcrumb' in a future update:
`	$page_title = apply_filters( 'edd_settings_page_title', $page_title, $current_page );`

assets/css/admin/style.scss
```diff
@@ -99,6 +99,30 @@ body.post-type-download #screen-meta {
 	margin: 0 0 0 20px;
 }
 
+.edd-header-page-title-wrap {
+	margin-top: 7px;
```
robincornett, line 103: The margin-top and line-height become unnecessary if we can add `align-items: center;` to the `#edd-header-wrapper` rule (lines 90-94), and then the margin-top should be -7.

assets/css/admin/style.scss
```diff
@@ -99,6 +99,30 @@ body.post-type-download #screen-meta {
...
+
+.edd-header-separator {
+	margin-top: -2px;
+	opacity: 0.25;
+}
+
+.edd-header-page-title {
+	font-weight: 400;
+	font-size: 22px;
+	line-height: 28px;
```
robincornett, line 118: I'm not sure the line-height is necessary here....also what would you think of sizing the headings in `em` instead of `px`? `em` is what WP Core uses and it would be nice to start moving that way as we can.

includes/admin/admin-bar.php
```diff
@@ -19,48 +19,126 @@
...
 		return;
 	}
 
-	// String
+	$indicator     = '';
+	$notifications = EDD()->notifications->countActiveNotifications();
+
+	if ( $notifications ) {
+		$count     = $notifications < 10 ? $notifications : '!';
+		$indicator = ' <div class="wp-core-ui wp-ui-notification edd-menu-notification-counter">' . $count . '</div>';
```
robincornett, line 37: What would you think of only showing the admin bar notifications indicator if we are not on a screen with the EDD header active? Having both red circles right next to each other (this may be due to my screen size) feels like a lot:

image

If that's just me, though, it's fine and I can live with both present.

includes/admin/admin-bar.php
```diff
@@ -19,48 +19,126 @@
...
+			'title' => 'EDD' . $indicator,
+			'href'  => edd_get_admin_url(
+				array(
+					'page' => 'edd-reports',
+				)
+			),
+		)
+	);
+
+	$wp_admin_bar->add_menu( array(
```
robincornett, line 53: Since the top/parent menu item also goes to reports, is it redundant/necessary to have this as well? Wondering if the top should go somewhere else (main downloads list maybe) instead? Possibly the top one could not be a link, but that's problematic if it's showing that there are notifications.

assets/css/admin/style.scss
cklosowski: Without the line-height we end up with this:
Screen Shot 2022-06-23 at 10 13 43 AM

I'll adjust to using `em`

includes/admin/admin-bar.php
cklosowski: @robincornett  I think this is just the nature of the current environment you're in, I'm on a smaller screen and with less menu bar items they aren't as close.

keep in mind that you have the EDD Dev tools installed, as well as Query Monitor, so in a production environment, your EDD menu item would be all the way to the left in your screenshot, b/c the QM fields, Blog #, branch identifier and settings dropdown will all be gone.

cklosowski:
@robincornett I believe I've covered all the bases here, have a look.

includes/admin/settings/display-settings.php
```diff
@@ -21,12 +21,86 @@ function edd_admin_header() {
...
 		<div id="edd-header-wrapper">
-			<img class="edd-header-logo" alt="" src="<?php echo esc_url( EDD_PLUGIN_URL . '/assets/images/logo-edd-dark.svg' ); ?>" />
+			<span id="edd-header-branding">
+				<img class="edd-header-logo" alt="" src="<?php echo esc_url( EDD_PLUGIN_URL . '/assets/images/logo-edd-dark.svg' ); ?>" />
+			</span>
+
+			<?php if ( ! empty( $page_title ) ) : ?>
+			<span class="edd-header-page-title-wrap">
+				<span class="edd-header-separator">/</span>
+				<h1 class="edd-header-page-title"><?php echo esc_html( $page_title ); ?></h1>
```
robincornett, line 101: Probably this needs to be an `h1` or `span` conditionally, maybe, depending on if the `$single_view` variable is true. On the individual customer details screen, or add/edit orders, now I have two `h1` and that's not ideal either.

cklosowski: @robincornett Done.

includes/admin/settings/display-settings.php
```diff
@@ -98,7 +98,8 @@ function edd_admin_header() {
 			<?php if ( ! empty( $page_title ) ) : ?>
 			<span class="edd-header-page-title-wrap">
 				<span class="edd-header-separator">/</span>
-				<h1 class="edd-header-page-title"><?php echo esc_html( $page_title ); ?></h1>
+				<?php $element = true === $is_single_view ? 'span' : 'h1'; ?>
+				<<?php echo $element; ?> class="edd-header-page-title"><?php echo esc_html( $page_title ); ?></<?php echo $element; ?>>
```
robincornett, line 102: I know this isn't modifiable, as it's defined just before, but should it be echoed through `esc_attr`?

cklosowski: @robincornett I don't think there's any real risk of not escaping it, but sure why not :) just to be safe.

robincornett (approved):

PR #9248, opened 2022-06-23
Improve/reports

cklosowski:
Adjusts the tiles on the overview to be more clear and actionable to store owners.


assets/css/admin/style.scss
```diff
@@ -2414,6 +2414,45 @@ span.edd-to-and-from--separator {
...
+	left: 3px;
+	margin-left: 0px;
+	color: #999;
+}
+
+.edd-reports-tile .tile-relative span.dashicons-arrow-down {
+	color: #FF6666;
+}
+
+.edd-reports-tile .tile-relative span.dashicons-arrow-up {
```
robincornett, line 2444: Minor thing but since we have four rules which are really two pairs of rules could they be combined/reduced to two? Also would love if we used existing WP Core colors or defined/used a red/green ourselves. We have a `$wp-red-50` and a `$wp-green-50` already defined, for example.

includes/admin/reporting/reports.php
```diff
@@ -1052,14 +931,12 @@ function edd_register_refunds_report( $reports ) {
 				'tile' => array(
 					'data_callback' => function () use ( $dates, $exclude_taxes, $currency ) {
 						$stats  = new EDD\Stats();
-						$amount = $stats->get_order_refund_amount( array(
+						return $stats->get_order_refund_amount( array(
```
robincornett, line 934: This was previously escaped upon return--is that not necessary?

robincornett (commented):
I need to review the PHP more in depth but so far this is looking good in the admin.

Curious about the decision to remove the aspect ratio and setting the height on the graphs.

includes/admin/reporting/reports.php
```diff
@@ -2398,33 +2275,17 @@ function edd_register_customer_report( $reports ) {
...
-						) );
-					},
-				),
-			),
-		) );
-
-		$reports->register_endpoint( 'average_customer_value', array(
-			'label' => __( 'Average Value', 'easy-digital-downloads' ),
+		$reports->register_endpoint( 'average_revenue_per_customer', array(
+			'label' => __( 'Average Revenue per Customer', 'easy-digital-downloads' ),
```
robincornett, line 2279: I feel like the triangle on this tile should be colored the same way the triangles are on the overview tiles...right now it's just gray.

includes/admin/reporting/reports.php
```diff
@@ -2384,9 +2261,9 @@ function edd_register_customer_report( $reports ) {
 			'priority'  => 40,
 			'endpoints' => array(
 				'tiles'  => array(
-					'lifetime_value_of_customer',
+					'average_revenue_per_customer',
 					'average_number_of_orders_per_customer',
-					'average_customer_value',
+					'average_customer_revenue'
```
robincornett, line 2266: I don't think a tile exists for either `average_customer_value` or `average_customer_revenue`. I'm seeing three tiles registered but only two display ("Average Revenue Per Customer" and "Average Number of Orders").

Additionally, the average number of orders seems to be incorrect. My top five customers is showing 52 for the highest number of sales for a customer, but the average number of orders is 222.

includes/class-stats.php
```diff
@@ -658,26 +666,82 @@ public function get_order_item_refund_count( $query = array() ) {
...
+		$this->query_vars['date_query_column'] = 'date_created';
+
+		if ( empty( $this->query_vars['function'] ) ) {
+			$this->query_vars['function'] = 'SUM';
+		}
 
 		/*
-		 * Request raw output so we can run `abs()` on the value.
-		 * But save the original value so we can restore it later.
+		 * By default we're checking sales only and excluding refunds. This gives us gross order earnings.
```
robincornett, line 680: This comment is confusing me because the `type` is explicitly set to refund. Should the wording be inverted instead?

robincornett:
It looks like we have the whole logic to modify the `$total` output based on relative change repeated in this class six times. I would love to see if we can extract that into a private helper method to make things a little more dry and reduce possible differences/errors in implementation.

cklosowski:
@robincornett I would like to as well, but I think we need to save that for a later release..there are a number of factors I ran into when attempting it:
1. Some are absint, some are float.
2. Some have reversed logic around what is positive (green) and what is negative (red)
3. Why is the stats class handling this output anyway? I feel like we need to middleware that handles the output of graphing data from the stats classes delivery.

All that said, I think it's fine for now, this isn't something I Introduced with this, we've been living with this relative logic for some time now, so we can live with it through the release.

Since these are specific to tiles, what I'd like to do is save the DRY implementation for an actual abstraction of it into a sort of API instead of being baked into the stats class, so that the Stats class returns stats, not formatted HTML.

robincornett (approved):

PR #9262, opened 2022-07-05
Fixing list table date ranges, and fixing report graph groupings to account for UTC Offset

cklosowski:
@robincornett have a look at this for me. It makes a couple small tweaks to reporting graphs and the list table to account for UTC offset when doing queries for grouping.


robincornett (approved):
If this is working the way you want it to, I'm good to approve (I made one tiny tweak). The orders list table is returning more accurate data for me now.

PR #9267, opened 2022-07-07
Updated debug mode contextual help

### LINKED ISSUE #9044: 3.0 - Change Debug mode checkbox description text
On a fresh 3.0 release install I wanted to activate Debug mode but the description text next to the checkbox seems a bit too vague.
https://a.supportally.com/hYZGXN

I feel like we should rephrase that to say "Check this box to Enable Debug mode", or something like that.

zkawesome:
Fixes #9044

Small change in the contextual help for Debug Mode.

### DIFF
--- a/includes/admin/settings/register-settings.php
+++ b/includes/admin/settings/register-settings.php
@@ -905,7 +905,7 @@ function edd_get_registered_settings() {
 						'id'    => 'debug_mode',
 						'name'  => __( 'Debug Mode', 'easy-digital-downloads' ),
 						'check' => __( 'Enabled',    'easy-digital-downloads' ),
-						'desc'  => sprintf( __( 'Check this box to enable Debug Mode. %s', 'easy-digital-downloads' ), $debug_log_link ),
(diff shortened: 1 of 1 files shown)


includes/admin/settings/register-settings.php
```diff
@@ -905,7 +905,7 @@ function edd_get_registered_settings() {
 						'id'    => 'debug_mode',
 						'name'  => __( 'Debug Mode', 'easy-digital-downloads' ),
 						'check' => __( 'Enabled',    'easy-digital-downloads' ),
-						'desc'  => sprintf( __( 'Check this to enable logging of certain behaviors to a file. %s', 'easy-digital-downloads' ), $debug_log_link ),
+						'desc'  => sprintf( __( 'Check this box to enable Debug Mode. %s', 'easy-digital-downloads' ), $debug_log_link ),
```
robincornett, line 908: I feel like tying the link into the translatable string is unnecessary and something we could go ahead and change now. What would you think of removing the `sprintf` and just appending the link to the string? It feels like less work for translators.

robincornett (approved):

PR #9268, opened 2022-07-08
Fixes adding more than one additional customer email in admin

### LINKED ISSUE #9254: 3.0 - Adding more than one additional customer email in admin causes issues
### Expected behavior
I expect to be able to add more than one additional email address while on the customer edit/screen.

### Actual behavior
While this is not a "common" workflow, if you try and add more than 1 additional email on the customer edit/view screen (under the emails tab), when you get to the second one it does not add until refreshed and infinitely "spins" without refreshing the page. This can be resolved by physically refreshing the browser but somewhat confusing for non-technical users.

1) Goto a customer then the "Emails" tab
2) Add an email (don't make it primary), just an additional address.
3) It should add fine.
4) Now don't refresh and add another one, it breaks...

PHP Version: 8.0

EDD Version (or branch): latest (3.0)

zkawesome:
Fixes #9254

When user is adding an additional customer email in the admin, the JS will refresh the page as per `response.redirect` parameter from the backend. It will work the first time because the URL of the page will change, but the page reload will stop working for any subsequent new emails, because the URL of the second refresh has not changed. If `window.location.href` is the same as the current page (with the # paramater), the page will not reload.

Proposed Changes:
- Added a unique parameter to each redirect response with the customer email ID, so that JS will always reliably reload the page
- Changed the `add_email` function inside `EDD_Customer` to return ID instead of boolean

### DIFF
--- a/includes/admin/customers/customer-actions.php
+++ b/includes/admin/customers/customer-actions.php
@@ -222,12 +222,13 @@ function edd_add_customer_email( $args = array() ) {
 		);
 
 	} else {
-		$email       = sanitize_email( $args['email'] );
-		$customer_id = (int) $args['customer_id'];
-		$primary     = 'true' === $args['primary'] ? true : false;
-		$customer    = new EDD_Customer( $customer_id );
+		$email                  = sanitize_email( $args['email'] );
+		$customer_id            = (int) $args['customer_id'];
+		$primary                = 'true' === $args['primary'] ? true : false;
+		$customer               = new EDD_Customer( $customer_id );
+		$customer_email_address = $customer->add_email( $email, $primary );
 
-		if ( false === $customer->add_email( $email, $primary ) ) {
+		if ( false === $customer_email_address ) {
 
 			if ( in_array( $email, $customer->emails, true ) ) {
 				$output = array(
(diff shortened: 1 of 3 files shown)


includes/admin/customers/customer-actions.php
```diff
@@ -245,10 +246,11 @@ function edd_add_customer_email( $args = array() ) {
...
 				array(
-					'page'        => 'edd-customers',
-					'view'        => 'overview',
-					'id'          => urlencode( $customer_id ),
-					'edd-message' => 'email-added',
+					'page'         => 'edd-customers',
+					'view'         => 'overview',
+					'id'           => urlencode( $customer_id ),
+					'edd-message'  => 'email-added',
+					'edd-email-id' => $customer_email_id,
```
robincornett, line 253: This variable should be wrapped in `urlencode` or `absint` (since you know it's an ID I think the latter is fine, but the former would be consistent with the customer ID).

robincornett (approved):
This is working well for me and I think a solid solution. I'm approving but not merging as @cklosowski will need to determine if we can update the return value for `add_email`.

I think it's worth doing to be consistent with the general BerlinDB functions; I do not think we can necessarily just switch this to using `edd_add_customer_email_address()` directly, although tempting, since the `add_email()` method does register an action hook.

includes/class-edd-customer.php
```diff
@@ -369,7 +369,7 @@ public function update( $data = array() ) {
 	 * @param string $email The email address to remove from the customer.
```
cklosowski, line 369: Please add a `since` tag for `3.0.1` that mentions the return type change.

cklosowski (changes requested):

cklosowski:
@robincornett once that docblock is updated, feel free to merge.

PR #9269, opened 2022-07-09
Allow filtering the order item name on the view order details and refund details #9265

### LINKED ISSUE #9265: 3.0 - Missing filters for order item name and if possible other data in order view
### Expected behavior
What about adding filters to customize the order view?

eg. It seems we don't have an easy way to customize the order item name. 

### Actual behavior

Please see https://github.com/awesomemotive/easy-digital-downloads/blob/60a205cfab9136a3994b548d78f6bb2f5913628f/includes/orders/classes/class-order-item.php#L234-L236

By adding a filter there, it would be a lot easier when trying to add custom data to the order item name (eg. data from custom price option settings)

It would also be useful to add filters to other data that is sent to the order view.

If it already exists and I missed it, please let me know. Thank you.

EDD Version (or branch): 3.0

cklosowski:
Fixes #9265 

Proposed Changes:
1. Adds a new filter to be able to modify the order item name in the view order/refund details
2. New filter is `edd_order_details_item_name`
3. Passes in the name, and the `Order_item` object.

to test this filter I am using the following:
```
function ck_test_order_item_row( $item_name, $item ) {
	$item_name .= ' - ' . $item->product_id;

	return $item_name;
}
add_filter( 'edd_order_details_item_name', 'ck_test_order_item_row', 10, 2 );
```

### DIFF
--- a/includes/orders/classes/class-order-item.php
+++ b/includes/orders/classes/class-order-item.php
@@ -232,7 +232,15 @@ class Order_Item extends \EDD\Database\Rows\Order_Item {
 	 * @return string The product name including any price ID name.
 	 */
 	public function get_order_item_name() {
-		return $this->product_name;
+		return is_admin() ?
+			/**
+			 * Allow the product name to be filtered within the admin.
+			 * @since 3.0
+			 * @param string $product_name  The order item name.
+			 * @param EDD\Orders\Order_Item The order item object.
+			 */
+			apply_filters( 'edd_order_details_item_name', $this->product_name, $this ) :
+			$this->product_name;
 	}
 
 	/**
--- a/includes/orders/classes/class-order-item.php
+++ b/includes/orders/classes/class-order-item.php
@@ -232,15 +232,17 @@ class Order_Item extends \EDD\Database\Rows\Order_Item {
 	 * @return string The product name including any price ID name.
 	 */
 	public function get_order_item_name() {
-		return is_admin() ?
+		if ( is_admin() && ( function_exists( 'edd_doing_ajax' ) && ! edd_doing_ajax() ) ) {
 			/**
 			 * Allow the product name to be filtered within the admin.
 			 * @since 3.0
 			 * @param string $product_name  The order item name.
 			 * @param EDD\Orders\Order_Item The order item object.
 			 */
-			apply_filters( 'edd_order_details_item_name', $this->product_name, $this ) :
-			$this->product_name;
+			return apply_filters( 'edd_order_details_item_name', $this->product_name, $this );
+		}
+
+		return $this->product_name;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


robincornett (changes requested):
The order item name should also be modified in the refund modal table for this to be consistent.

Some thoughts:
- I'm not a fan of having identical `apply_filters` calls in multiple places--this is currently two and to get the refund modal consistent would be a third.
- I'm not sure why the `get_order_item_name` method was introduced as it's just returning a property directly.
- However, the method is only used in EDD for the order details screens, and (incorrectly, I believe, as it's unique to the test and apparently not in the main code) one unit test on emails. Currently, none of our extensions use it.
- Therefore, what would you think about updating the method to include the filter, and update the doc block to indicate that the method is only intended to be used for admin facing order details? That would make the method _do_ something other than just return the property, and clarify why it exists, and add the filter in just one location. We could even potentially update the filter name to include `admin` possibly.

cklosowski:
@robincornett I do not like it being part of the class itself b/c the class itself can be called from anywhere and all the 'admin' references in the world aren't going to prevent a user from using the method on the front end. Just b/c 'nothing else uses it' now, before our final release, doesn't mean that as soon as it's live our integrations and 3rd party integrations won't start picking it up. We need to determine scope of it now.

The main reason for it being part of the class and a property would be for inheritance purposes. Allowing it to be overloaded with an extended class of sorts. If it were a property, you could not be able to extended it without magic getters and setters, but with a single method, we can simple extend this class and override the logic.

If we make this return the same content for front end and back end, then we run the risk of allowing a developer to show sensitive information to a front facing customer (which is why the reporter had to use 3 hooks/filters to get this to work how they wanted). https://github.com/awesomemotive/easy-digital-downloads/issues/9265#issuecomment-1179345338

We need to take the approach that _we_ are handling the scope, and that means either:
1. Using 1 filter (or 3 filters) that are specific the the admin area
2. Create another method (or alter the existing method) to allow filtering based on the scope of which the method is used.

robincornett:
I think my vote, then, would be to use the method as we have it now, and apply the filter conditionally within the message with `edd_is_admin_page()`. I would much prefer that than applying one filter in three different places and having to maintain that.

I can update the PR.

cklosowski:
@robincornett sounds good, please commit your changes that you think will work.

includes/orders/classes/class-order-item.php
```diff
@@ -232,7 +232,15 @@ public function get_fees() {
 	 * @return string The product name including any price ID name.
 	 */
 	public function get_order_item_name() {
-		return $this->product_name;
+		return is_admin() ?
```
robincornett, line 235: Note: `edd_is_admin_page()` returns `false` for the refunds table so I stuck with `is_admin()` instead.

cklosowski:
I can't approve this b/c I made it originally but looks good on my end.

PR #9275, opened 2022-07-13
Fixed ambiguous date_created column inside taxed orders reports

### LINKED ISSUE #9165: 3.0 - Report exports fail when "All Statuses" and a country/region is selected
### Expected behavior
I expect to be able to export the reports with All Statuses (the default) enabled while also selecting a country/region.

### Actual behavior
When I select a country or region with "All Statuses" selected in the "Export Taxed Orders", I get no data found. When I select the actual status like "Complete" it works fine.

zkawesome:
Fixes #9165

Proposed change:
- The field `date_created` is ambiguous when you export taxed orders while selecting date filter + country filter. I have made a change inside `class-batch-export-taxed-orders.php` so that Berlin will append the table alias to the `date_created` field.

### DIFF
--- a/includes/admin/reporting/export/class-batch-export-taxed-orders.php
+++ b/includes/admin/reporting/export/class-batch-export-taxed-orders.php
@@ -98,7 +98,7 @@ class EDD_Batch_Taxed_Orders_Export extends EDD_Batch_Export {
 		);
 
 		if ( ! empty( $this->start ) || ! empty( $this->end ) ) {
-			$args['date_query'] = $this->get_date_query();
+			$args['date_created_query'] = $this->get_date_query();
 		}
 
 		if ( 'any' === $args['status'] || 'all' === $args['status'] ) {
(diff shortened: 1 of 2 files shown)


includes/admin/reporting/export/class-batch-export-taxed-orders.php
```diff
@@ -98,7 +98,7 @@ public function get_data() {
 		);
 
 		if ( ! empty( $this->start ) || ! empty( $this->end ) ) {
-			$args['date_query'] = $this->get_date_query();
+			$args['date_created_query'] = $this->get_date_query();
```
robincornett, line 101: I think this should be applied to the `get_percentage_complete` method in this class too, the same way you did for file downloads?

robincornett (commented):
I think just one more tweak that I can see. Well done on figuring this out!

cklosowski (approved):

robincornett (approved):

PR #9278, opened 2022-07-14
Improved hourly graphs in reporting for custom daterange

### LINKED ISSUE #9263: 3.0 - Possibly display graphs as hourly for custom date ranges
If I set a custom date range for a report, and the date range is one day (or possibly even two), it would be great if it displayed an hourly graph (I would like `get_dates_filter_hour_by_hour` to return `true` in this instance).

### Justification or use case
I put a one day range into the reports filter and the graph just had one point on it. I would like for it to look like a today/yesterday graph when the date range is so short.

zkawesome:
Fixes #9263

For custom date range mode in reporting, if the request date range is not more than two days, we should switch to hourly graphs.

Proposed Changes:
- Added custom date range logic inside `get_dates_filter_hour_by_hour()` method to calculate if the requested date range is not more than two days
- Improved graph rendering logic, since it was reversed -`hour by hour` check should always be first, followed by `day by day` check

### DIFF
--- a/includes/admin/reporting/reports-callbacks.php
+++ b/includes/admin/reporting/reports-callbacks.php
@@ -47,12 +47,12 @@ function edd_overview_sales_earnings_chart() {
 	$sql_clauses['orderby'] = 'MONTH(date_created)';
 
 	// Now drill down to the smallest unit.
-	if ( $day_by_day ) {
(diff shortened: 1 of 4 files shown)


cklosowski (approved):
@zkawesome this looks good to me, @robincornett if you are good with it and all the adjusted reports work as you expect, approve and merge.

robincornett (approved):

PR #9282, opened 2022-07-14
Fix/migration oom

cklosowski:
Fixes an issue where the CLI command to migrate logs can hit the Linux OOM process.

Proposed Changes:
1. Batches the logs migration
2. Avoids a full count of logs, so we do not have a progress bar.


robincornett:
@cklosowski just updating--John said that the PR worked for the customer who was testing it and he was able to migrate his logs.

cklosowski:
@robincornett @JJJ So I went through and adjusted most of the longer running processes to use a stepped (offset, limit) approach. I just tested this locally on about 8,000 order records. I made a new branch off `master` (2.11.7 base) called `legacy/notes-generator`. So I basically did the following:

1. Imported products from the `.xml` file, set the Test Purchase as the default gateway.
2. `wp edd payments create --number=8000 --date=random --range=720 --generate_users=1`
3. `wp edd download_logs create --number=15000`
4. `wp edd generate_customer_notes create --number=1000`
5. `wp edd generate_payment_notes create --number=10000`

After that I did the `wp edd v30_migration` command.

One thing I didn't get was the addition of additional email addresses or customer addresses. I need to work up a generator for those still, so we can generate a mass amount of those to test. The migration did however, migrate the primary email addresses correctly for the customers.

includes/class-edd-cli.php
```diff
@@ -1085,30 +1184,79 @@ public function migrate_customer_notes( $args, $assoc_args ) {
...
+				}
 			}
 
 			$progress->finish();
 
-			WP_CLI::line( __( 'Migration complete: Customer Notes', 'easy-digital-downloads' ) );
+			if ( 0 === $step ) {
+				WP_CLI::line( __( 'No customer notes found.', 'easy-digital-downloads' ) );
+			} else {
+				WP_CLI::line( __( 'Migration complete: Cusotmer Notes', 'easy-digital-downloads' ) );
```
robincornett, line 1252:
```suggestion
				WP_CLI::line( __( 'Migration complete: Customer Notes', 'easy-digital-downloads' ) );
```

robincornett:
@cklosowski noting that [0adb0d8](https://github.com/awesomemotive/easy-digital-downloads/pull/9282/commits/0adb0d82de40bba528ebddbbf912789cf173d3ff) needs to be added to the 3.1 release regardless as it's a hook AA uses.

I've merged release/3.1 into this branch and rebased the PR. Currently running a migration to test again.

robincornett (approved):
I think we can move forward with this. Additional emails and addresses are, I think, less of a concern than the logs and general order data.

cklosowski:
Thanks @robincornett

PR #9292, opened 2022-07-16
Fix/mysql 55

JJJ:
Re-adds support for older versions of MySQL.

Proposed Changes:
1. Remove `CURRENT_TIMESTAMP`
2. Use `0000-00-00 00:00:00` again
3. Remove some now-irrelevant upgrade routines & queries
4. Bumps table versions
5. Modifies CREATE TABLE queries and default column values in Schemas

### DIFF
--- a/includes/database/schemas/class-adjustments.php
+++ b/includes/database/schemas/class-adjustments.php
@@ -186,7 +186,7 @@ final class Adjustments extends Schema {
 		array(
 			'name'       => 'date_created',
 			'type'       => 'datetime',
-			'default'    => '', // Defaults to current time in query class
+			'default'    => '0000-00-00 00:00:00',
 			'created'    => true,
 			'date_query' => true,
 			'sortable'   => true
@@ -196,7 +196,7 @@ final class Adjustments extends Schema {
 		array(
 			'name'       => 'date_modified',
 			'type'       => 'datetime',
-			'default'    => '', // Defaults to current time in query class
+			'default'    => '0000-00-00 00:00:00',
 			'modified'   => true,
 			'date_query' => true,
 			'sortable'   => true
--- a/includes/database/NotificationsDB.php
+++ b/includes/database/NotificationsDB.php
@@ -252,7 +252,16 @@ class NotificationsDB extends \EDD_DB {
 	    KEY dismissed_start_end (dismissed, start, end)
 		) DEFAULT CHARACTER SET {$wpdb->charset} COLLATE {$wpdb->collate};" );
 
+		// Update `date_created`.
+		$wpdb->query(
+			"ALTER TABLE {$this->table_name} MODIFY COLUMN `date_created` datetime NOT NULL default '0000-00-00 00:00:00'"
+		);
+
+		// Update `date_updated`.
+		$wpdb->query(
+			"ALTER TABLE {$this->table_name} MODIFY COLUMN `date_updated` datetime NOT NULL default '0000-00-00 00:00:00'"
+		);
+
 		update_option( $this->table_name . '_db_version', $this->version );
 	}
-
 }
--- a/includes/database/NotificationsDB.php
+++ b/includes/database/NotificationsDB.php
@@ -24,7 +24,7 @@ class NotificationsDB extends \EDD_DB {
 
 		$this->table_name  = $wpdb->prefix . 'edd_notifications';
 		$this->primary_key = 'id';
-		$this->version     = '1.0';
+		$this->version     = '1.0.1';
 
 		add_action( 'edd_daily_scheduled_events', array( $this, 'schedule_daily_notification_checks' ) );
 
@@ -246,8 +246,8 @@ class NotificationsDB extends \EDD_DB {
 	    start datetime DEFAULT NULL,
 	    end datetime DEFAULT NULL,
 	    dismissed tinyint(1) UNSIGNED NOT NULL DEFAULT 0,
-	    date_created datetime NOT NULL DEFAULT CURRENT_TIMESTAMP(),
-	    date_updated datetime NOT NULL DEFAULT CURRENT_TIMESTAMP(),
+	    date_created datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
+	    date_updated datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
 	    PRIMARY KEY (id),
 	    KEY dismissed_start_end (dismissed, start, end)
 		) DEFAULT CHARACTER SET {$wpdb->charset} COLLATE {$wpdb->collate};" );
--- a/includes/database/schemas/class-customer-addresses.php
+++ b/includes/database/schemas/class-customer-addresses.php
@@ -145,7 +145,7 @@ class Customer_Addresses extends Schema {
 		array(
 			'name'       => 'date_created',
 			'type'       => 'datetime',
-			'default'    => '', // Defaults to current time in query class
+			'default'    => '0000-00-00 00:00:00',
 			'created'    => true,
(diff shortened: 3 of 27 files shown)


cklosowski:
Hey @robincornett, John and I were talking about the 'missing database tables' we mentioned in support and he pointed out that this use of the `CURRENT_TIMESTAMP` drastically limits our MySQL version support.

I believe we've pre-maturely dropped support for a widely used version of PHP in the WordPress ecosystem. We need to test  this using different MySQL versions (5.5 and up)

includes/database/schemas/class-adjustments.php
```diff
@@ -186,7 +186,7 @@ final class Adjustments extends Schema {
 		array(
 			'name'       => 'date_created',
 			'type'       => 'datetime',
-			'default'    => '', // Defaults to current time in query class
+			'default'    => '0000-00-00 00:00:00',
```
robincornett, line 189: I'm not sure we can change this default as we intentionally moved away from `0000-00-00 00:00:00` due to MySQL errors  in higher versions of MySQL, at least as I read the issue: #7578; wouldn't this change reintroduce that issue?

I'm wondering if the solution mentioned here ([changing the type to `timestamp`](https://dba.stackexchange.com/questions/132951/cant-default-date-to-current-timestamp-in-mysql-5-5)) might be an option? I see in the linked docs that it can only be used for one column per table in MySQL 5.5, but maybe it's an option?

cklosowski: @robincornett we have multiple columns that need that type though, so if it is limited to a single column per table, it can't be used.

robincornett: I understand that we have multiple columns that need that type--my concern is that we intentionally moved away from the `0000-00-00 00:00:00` default because it's not allowed with the strict MySQL mode in more current versions of MySQL, which is presumably something set at a hosting level, so we're introducing a regression--so I'm pondering what the alternatives might be (including selecting a primary column to have the `timestamp` type).

Would an option be defining the database version in the base `Schema` and `Table` classes from `$wpdb->db_version()`, and conditionally setting the default to the zeros or `CURRENT_TIMESTAMP` based on that? That way, users with more current versions of MySQL get the advantages of the current timestamp, we avoid the strict issue, and users on older versions have a default that works for them and is the expected format.

JJJ:
Because CURRENT_TIMESTAMP currently isn't supported by the minimum allowed MySQL version for WordPress (MySQL 5.0+) it looks like about 30% of EDD sites may be at risk of failing to create the object tables, failure to migrate data, etc...

When it is safe to drop support for older MySQL versions, `CURRENT_TIMESTAMP` (and the synonyms for `NOW()`) still have the _potential_ to introduce data inconsistencies that are impossible to identify without relatively deep forensic investigation.

Essentially, if the MySQL server, the web server (Apache/Nginx), and/or the PHP runtime have different time zones configured in their stacks – (system/application/connection) – when EDD/Berlin do not explicitly supply a default or updated value via what is known to be UTC in PHP, the MySQL server would supply what it thinks `NOW()` is which is not guaranteed to equal precisely what PHP thinks `time()` is.

These values could be off by microseconds or years depending how careful or careless someone was when setting everything up and connecting them together.

When MySQL deprecates zero-dates entirely, we (WordPress) will need an alternative, and I do not think that will be a NOW() synonym either, but I'm also not sure what my recommendation would be... maybe UNIX epoch?

includes/database/schemas/class-adjustments.php
robincornett: Looking at how the Core WP columns are set up for `post_modified` and `post_modified_gmt`, these changes are consistent with the posts table, so I am willing to withdraw my concerns based on that. I just want to make sure that the other changes made in the PR which changed to `CURRENT_TIME` don't need to be reverted as well--and that migrated and new orders have the correct dates assigned, etc.

JJJ: > I'm wondering if the solution mentioned here ([changing the type to `timestamp`](https://dba.stackexchange.com/questions/132951/cant-default-date-to-current-timestamp-in-mysql-5-5)) might be an option?

The range for `TIMESTAMP` values is:
> `1970-01-01 00:00:01.000000` to `2038-01-19 03:14:07.999999`

I'll be 59 years old in 2038 and will either have forgotten all of this or still be working on fixing it 👴🏻 💖

JJJ:
The Sequel Ace app for macOS (which is the new fork of Sequel Pro) provides a Time Zone picker as part of the Database connection. This ends up bing a neat way to alter the session connection and watch my `CURRENT_TIMESTAMP` concerns in real time.

jjj-on-2022-07-17-at-13-46-17@2x

See: https://dev.mysql.com/doc/refman/8.0/en/datetime.html

> MySQL converts TIMESTAMP values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval. (This does not occur for other types such as DATETIME.) By default, the current time zone for each connection is the server's time. The time zone can be set on a per-connection basis. As long as the time zone setting remains constant, you get back the same value you store. If you store a TIMESTAMP value, and then change the time zone and retrieve the value, the retrieved value is different from the value you stored. This occurs because the same time zone was not used for conversion in both directions. The current time zone is available as the value of the [time_zone](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_time_zone) system variable. For more information, see [Section 5.1.15, “MySQL Server Time Zone Support”](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html).

You can disconnect and reconnect with a different Time Zone, insert & select rows that have `CURRENT_TIMESTAMP` columns, and watch the default values change based on its offset. Wee!

----

The SQL to counteract a lossy conversion probably looks something like this...
```mysql
SELECT date_created FROM whatever WHERE UNIX_TIMESTAMP( CONVERT_TZ( date_created, '+00:00', @@global.time_zone ) )
```
...but `CONVERT_TZ` has the same range limitations that `TIMESTAMP` data types do:

> On 32-bit platforms, the supported range of values for this function is the same as for the [TIMESTAMP](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) type (see [Section 11.2.1, “Date and Time Data Type Syntax”](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-type-syntax.html), for range information). On 64-bit platforms, beginning with MySQL 8.0.28, the maximum supported value is '3001-01-18 23:59:59.999999' UTC.

...so barf if we need to do that for every datetime column 🤮 and still come back to it in 2038 (or 3001 🕰)

----

> because it's not allowed with the strict MySQL mode in more current versions of MySQL, which is presumably something set at a hosting level, so we're introducing a regression

This is nearly correct!

The important missing bit is that all MySQL versions (including MariaDB) allow for client connections (like WordPress via `wpdb`) to override all of the modes – specifically in our case:

* `STRICT_ALL_TABLES`
* `STRICT_TRANS_TABLES`
* `NO_ZERO_DATE`

WordPress, by default, via `wpdb::set_sql_mode()`, explicitly removes those modes (and a few others) from the MySQL server configuration. It does this every single time it makes a database connection, which means it is never seeing strict mode and never has a problem with zero-value datetimes.

It is safe to assume sometime in the future that support for `NO_ZERO_DATE` will be deprecated; the MySQL documentation says so.

jjj-on-2022-07-17-at-15-31-07@2x

But that does not necessarily mean that the concept of an "empty" or "minimum" datetime value will be deprecated too. Perhaps we will need to decide it is UNIX epoch, or the earliest datetime value of `1000-01-01 00:00:00`. I think, perhaps, that might be up to us to decide together! 🛼

Issue #9270, opened 2022-07-11
show_option_all check for HTML select method is not PHP8 compatible


robincornett:
### Expected behavior
When I have the "all" option selected and saved to my [post] metadata, I expect it to show as selected when using the `EDD()->html->select()` method.

### Actual behavior
On a site running PHP8, the "all" option does not show as selected.

### Steps to reproduce the behavior
This is how we encountered the issue:
1. Activate All Access on a site running PHP8.
2. Create an All Access enabled product and select "All Products" for the allowed categories.
3. Save the product.
4. Check the categories select for All Access--nothing appears to be selected.

### Information (if a specific version is affected):
PHP Version: 8.0
EDD Version (or branch): master, release/3.0
WordPress Version: 6.0

Any other relevant information:
Here's how EDD handles the `show_option_all` select option:

https://github.com/awesomemotive/easy-digital-downloads/blob/ce9a083056826cd64cc7b231ed0e57dcb26134f3/includes/class-edd-html-elements.php#L720-L727

The key is in the `in_array` check, specifically:

```php
in_array( 0, (array) $args['selected'] )
```
Which omits the third parameter to make this a strict check. I expect that was intentional, and it works (or appears to work) in PHP versions lower than 8.0, because apparently an array of strings will return `true` for `0` when it's not a strict check, even though a string shouldn't evaluate to an empty value. This is fixed in PHP 8, so what appears to work for EDD in lower PHP versions will no longer appear to work.

What happens in All Access is that the metadata which is properly saved as:
```
Array
(
    [all_access_categories] => Array
        (
            [0] => all
        )

    [all_access_duration_number] => 1
    [all_access_duration_unit] => never
    [all_access_download_limit] => 0
    [all_access_download_limit_time_period] => per_day
    [all_access_number_of_price_ids] => 0
)
```
incorrectly shows an empty select field for the product categories.

I can understand doing some gymnastics to show this as selected if there is no value saved to the metadata, or if it's a select which is not tied to the database (for the export screen, for example), but if it is tied to metadata somehow, and the `all` value is saved, then it seems that the `all` value should be used, as that's what's set for the value and what's saved to the database when something is saved.

Related: https://github.com/awesomemotive/edd-all-access/issues/534

PR #9294, opened 2022-07-18
Adds support for a 'global' tax rate using the All Countries and * as the value.

cklosowski:
After many reports of people who were using the global tax rate, I went down the path of looking how other eCommerce platforms support this and found that we can do it with a pretty simple solution, we just use `*` as the Country.

If a country/region match is found for the customer it uses the defined tax rate. Then if no rates are found for that user, it checks for a `global` rate, identified by the `*` character. This is then applied.

This solves the problem that we had originally where we needed a tax rate to tie to purchases, but a 'global' rate wasn't a real rate. By doing this we can also set it up to only support a single 'global' tax rate thanks to the work done to prevent 'duplicates'.


robincornett:
If default rate was set previously and we are going to support it in 3.0 then yes, I think we need to make sure it gets migrated to the new table.

I don't love that the `all` option is having to be shimmed over to `*` but I suspect that is due to #9270 rearing its head here?

cklosowski:
@robincornett Yeah we'll migrate the default tax rate into the first slot.

cklosowski:
This is ready for another go, this time it includes migration utilities
1. Migrates the default tax rate
2. while migrating orders detects when we should use the current default tax rate, and if the order has a tax rate that we cannot find, it creates an inactive global tax rate.

cklosowski:
Things left to do:
Adjust the notice as well.

includes/admin/upgrades/v3/class-data-migrator.php
```diff
@@ -657,12 +657,56 @@ public static function orders( $data = null ) {
 
 		$set_tax_rate_meta = false;
 
-		if ( ! empty( $tax_rate ) && ! empty( $order_address_data['country'] ) ) {
```
robincornett, line 660: I think I would revert all of the changes made here. With the global added correctly to `edd_get_tax_rate_by_location`, if a migrated order has the global tax rate applied, it's assigned automatically. If the tax rate doesn't match the retrieved tax rate, we have just added it to the order as metadata, and that happens automatically too.

Creating an inactive global rate to assign to an order would just be confusing, I think, and be inconsistent with how we are handling old orders with a country or regional tax rate which differs from the current country/regional rate. In those cases, we are adding the tax rate as order meta; we should do the same for the global rate.

Since the `$order_address_data['region']` is definitely set here, although it may be an empty string, instead of completely reverting, we could update the code to be less redundant:

```php
			$tax_rate_object = edd_get_tax_rate_by_location(
				array(
					'country' => $order_address_data['country'],
					'region'  => $order_address_data['region'],
				)
			);
```

includes/admin/upgrades/v3/class-tax-rates.php
```diff
@@ -41,6 +41,22 @@ public function __construct( $step = 1 ) {
 	public function get_data() {
 		$offset = ( $this->step - 1 ) * $this->per_step;
 
+		if ( 1 === $this->step ) {
+			$default_tax_rate = edd_get_option( 'tax_rate', false );
+			if ( ! empty( $default_tax_rate ) ) {
+				edd_add_adjustment(
+					array(
+						'name'        => '',
```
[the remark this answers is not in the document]
cklosowski: Fixed in 45ba2bf1338006706415b7b3410c777376a7ca93

includes/admin/views/tmpl-tax-rates-table-row.php
```diff
@@ -16,17 +16,25 @@
...
+	<# } else { #>
+		*
+		<input type="hidden" name="tax_rates[{{ data.id }}][country]" value="{{ data.country }}" />
+	<# } #>
 </td>
 
 <td data-colname="<?php esc_attr_e( 'Region', 'easy-digital-downloads' ); ?>">
 	<# if ( data.global ) { #>
 	&mdash;
 	<input type="hidden" name="tax_rates[{{ data.id }}][global]" value="{{ data.global }}" />
```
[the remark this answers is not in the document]
cklosowski: This is what I was pointing out on the call when we were discussing what to put in the 'scope' column in the Database, that we would end up with two `global` values. The 'scope' of other tax rates are actually:
`country` or `region`

`country` which ends up being `global` in the UI here
`region` which means we have a regional tax rate.

I'll have to rewrite quite a bit of the front end to get this to be more clear, if we think it's worth it in the long run. We have to rebuild the model in backbone to understand what the three 'scopes' are and not change their meaning.

includes/admin/upgrades/v3/class-data-migrator.php
cklosowski: I can understand the logic of if we find an order that doesn't match the currently set regional tax rate (or global now) that we just add meta, and not an inactive rate. However I do think that we should, if we find an order that is not a regional order, but matches the fallback tax rate, assigns to the newly created default tax rate ID.

includes/tax-functions.php
```diff
@@ -452,6 +453,18 @@ function edd_get_tax_rate_by_location( $args ) {
...
+		$tax_rates = edd_get_tax_rates(
+			array(
+				'name'   => '',
+				'scope'  => 'global',
+				'status' => 'active',
+			),
+			OBJECT
+		);
+	}
+
```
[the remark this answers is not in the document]
cklosowski: Fixed in 96ff71ba3e2d82817b8ccef151a123dc1b08b096

includes/admin/upgrades/v3/class-data-migrator.php
robincornett: It would do that--basically whatever tax rate that's found that matches will get assigned to the order, as long as the tax rate amount matches. The behavior in 3.0 currently is to match orders with existing (imported) tax rates--with the fix in `edd_get_tax_rate_by_location` then the global could end up matching--and assign the ID if there is a match; otherwise the rate is added as metadata. Reverting to this makes that possible with the global rate (at least that's what it's doing for me locally; the code changes were never assigning the default rate.

The current iteration is working for orders, but for an order which doesn't match an existing tax rate but does have tax collected, no inactive tax rate is being added to the database, and I would expect for it to be.

I think part of the issue is that when `$tax_rate_object` is defined in line 663 and 699, both of those are using a function which returns an array of tax rate objects, not a single tax rate object, so the metadata is assigned anyway and an inactive tax rate is not created. Since I don't think one should be created, I'm fine with that, but just noting that the code is not working the way I think it's intended as written.

cklosowski: @robincornett We'll have to update `edd_get_tax_rate_by_location()` if we take this approach, as that function returns early if the `country` is empty:

```
function edd_get_tax_rate_by_location( $args ) {

	$rate = false;
	if ( empty( $args['country'] ) ) {
		return $rate;
	}
```

robincornett: Possibly, but I think we force any store collecting taxes to collect address data, and the country/region are required. This `$args['country']` is coming from the order address, so any order with tax should have a country. If it doesn't, that feels like a different issue?

cklosowski: I've removed it in 8b3ca3a95aa233d9e39b65d7f0613220847173a4 and it seems that everything seems to work as I expect, will you give that a try @robincornett. I just moved the empty check to around the first search and that allows it to go down to search further.

robincornett (commented):
So I've just found something that's not working quite correctly. My scenario is set up the same way the unit tests are written:
- fallback rate 10%
- US rate 15%
- US/TN rate 9.25%

If I create a manual order for a customer with an existing address, when I go to _select_ the address, the form updates and although the country is properly selected, the state is not. So my existing address for TN wants to have the US rate applied.

Also....if I _save_ the order without selecting a state, I'm allowed to do that, which seems problematic, at least when taxes are involved.

This is _not_ happening in the `release/3.0.2` branch, interestingly.

cklosowski:
@robincornett I'm seeing this behavior on 3.0.2, intermittently, so I don't think this is unique to the taxes branch.

Screen Shot 2022-07-21 at 11 52 35 PM

cklosowski:
Yeah I've confirmed this is a but in 3.0.2 currently.

The issue is that We're updating the country (which triggers the first update to the tax rates) and then we're updating the region which triggers another tax change, I believe a race condition is causing that tax rate listener to sometimes cause the region update to revert. Seeing if I can fix it.

cklosowski:
@robincornett I can't get the JS triggers to behave correctly after 30 minutes of tinkering. If you want to take a stab at it, go for it. But since this exists in `main` even, I'm going to say I'm okay punting fixing that JS interaction for now.

robincornett:
In 804cead, I updated the admin order localization to look for a global tax rate and to use that by default, unless/until an address is selected:

image

Note that I checked with Zan and I think that with this change, the UI is behaving as currently intended. He selected a customer, but did not select the customer address. Since no address was associated with the order, it makes sense that the global rate would have been used instead of the preferred country rate. We probably could improve our messaging on that, to ensure that the person creating the order knows to select an address. Alternatively, we could default to the customer's primary address, if there is one, which would be consistent with the front end behavior for logged in users, but that's something to look into in a different issue.

robincornett (dismissed):

robincornett (approved):

PR #9297, opened 2022-07-19
Refactor lost password flow

robincornett:
Proposed Changes:
1. Updates `edd_get_lostpassword_url` to add parameters to the redirect parameter for the lost password URL, and also adds a query arg to the returned URL.
2. Adds a new function, `edd_set_lostpassword_session`, to set the correct redirect as session data.
3. Updates `edd_login_register_error_message` to conditionally get the correct redirect URL from the session. If the session data is not available, then the message is modified to be a little more generic.
4. The session variable is unset before outputting the errors.

### DIFF
--- a/includes/login-register.php
+++ b/includes/login-register.php
@@ -179,20 +179,15 @@ add_filter( 'wp_login_errors', 'edd_login_register_error_message', 10, 2 );
  * @return void
  */
 function edd_login_register_error_message( $errors, $redirect ) {
-	$edd_action_is_reset = ! empty( $_GET['edd_reset_password'] ) && 'done' === sanitize_text_field( $_GET['edd_reset_password'] );
-	if ( ! $edd_action_is_reset ) {
+	$redirect_url = EDD()->session->get( 'edd_forgot_password_redirect' );
+	if ( empty( $redirect_url ) ) {
 		return $errors;
 	}
-	$redirect_url = EDD()->session->get( 'edd_forgot_password_redirect' );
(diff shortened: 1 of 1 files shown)


includes/login-register.php
```diff
@@ -210,17 +216,37 @@ function edd_login_register_error_message( $errors, $redirect ) {
...
+
+	return add_query_arg(
 		array(
-			'checkemail'         => 'confirm',
-			'edd_reset_password' => 'confirm',
-			'edd_redirect'       => urlencode( $url ),
+			'edd_forgot_password' => 'confirm',
 		),
-		wp_login_url()
+		wp_lostpassword_url(
```
robincornett, line 224: @cklosowski I'm kind of inclined to remove the redirect parameter from the `wp_lostpassword_url()` call here and just manage the login errors filter based solely on the `edd_forgot_password_redirect` being available in the session. What would you think of that? It lets WP Core handle the redirect_to automatically instead of us forcing it to be the login URL.

robincornett:
@cklosowski after thinking on this somewhat and discussing with Zan, I'm removing the parameter from the `wp_lostpassword_url` function so that we're not overriding any custom code. If the session is set based on the query parameter we're adding, then our custom error message will show; if not, we'll just return the default error messages early. If the session fails to set, same thing--whatever the default is will happen.

cklosowski:
I think that's reasonable for sure @robincornett.

cklosowski (approved):

PR #9298, opened 2022-07-19
Fixed wrong capability check when adding manual order note

### LINKED ISSUE #9293: Non Shop Manager roles cannot add notes to orders (and other objects)
### Expected behavior
As a shop worker or shop accountant, I expect to be able to add notes to orders, customers, and discounts.

### Actual behavior
When attempting to add a note as these roles, my note is not saved.

1) Create and login as a user with the Shop Worker or Shop Accountant role
2) View an order
3) Attempt to add a note.

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

This is due to a capability check done [here](https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/admin/notes/note-actions.php#L26)

We should allow adding notes for a role that can `edit_shop_payments`. If a role is allowed to edit payments, they should be allowed to add notes.

zkawesome:
Fixes #9293

Proposed Changes:
- We should allow adding notes for a role that can`edit_shop_payments`. If a role is allowed to edit payments, they should be allowed to add notes.

### DIFF
--- a/includes/admin/notes/note-actions.php
+++ b/includes/admin/notes/note-actions.php
@@ -23,7 +23,7 @@ function edd_admin_ajax_add_note() {
 	check_ajax_referer( 'edd_note', 'nonce' );
 
 	// Bail if user cannot delete notes
-	if ( ! current_user_can( 'manage_shop_settings' ) ) {
+	if ( ! current_user_can( 'edit_shop_payments' ) ) {
 		wp_die( -1 );
 	}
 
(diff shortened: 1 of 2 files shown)


cklosowski (dismissed):

cklosowski:
@robincornett could we go a step further and just not show the box if the user cannot edit shop payments?

robincornett:
If a user can _view_ the payments then I think they should be allowed to see the notes as well. I considered whether we should remove the notes section/box for users who can't edit, but that seems extreme. But the form and ajax capabilities should certainly match, so my feeling is just adding the same capability to the form should be sufficient.

robincornett:
Also just thinking that since we can guarantee that the new note addition will fail if certain conditions are not met then the corresponding function to show the form should have the same checks (outside of the referer).

PR #9301, opened 2022-07-20
Added support for bundle products in history downloads shortcode

### LINKED ISSUE #9288: Bundle purchases don't show in the history-download short-code
Spotted the usage of `$download` in the template file here https://github.com/awesomemotive/easy-digital-downloads/blob/main/templates/history-downloads.php#L63-L64, however that variable is not set anywhere.

To replicate:
Simply purchase a bundle download that has download files attached to the bundled products.
Create a page with the [download_history] code on it and view it. You should see the download files from the downloads in the bundle but it shows nothing.

zkawesome:
Fixes #9288

Proposed Changes:
-  Introduced new method to `class-order.php` to retrieve bundled products as ordinary order items, so that we can correctly show bundled downloads inside `[download_history]` shortcode without breaking the layout or removing old actions.

### DIFF
--- a/includes/orders/classes/class-order.php
+++ b/includes/orders/classes/class-order.php
@@ -363,7 +363,7 @@ class Order extends Rows\Order {
 	 *
 	 * @return Order_Item[] Order items.
 	 */
-	public function get_deliverable_items() {
+	public function get_items_with_bundles() {
 		$items = $this->get_items();
 		foreach ( $items as $index => $item ) {
 			if ( edd_is_bundled_product( $item->product_id ) ) {
--- a/templates/history-downloads.php
+++ b/templates/history-downloads.php
@@ -49,7 +49,7 @@ if ( $orders ) :
 		</thead>
 		<?php
 		foreach ( $orders as $order ) :
-			foreach ( $order->get_deliverable_items() as $key => $item ) :
+			foreach ( $order->get_items_with_bundles() as $key => $item ) :
 				?>
 
 				<tr class="edd_download_history_row">


templates/history-downloads.php
```diff
@@ -47,24 +47,20 @@
...
 			</tr>
 		</thead>
-		<?php foreach ( $orders as $order ) :
-			foreach ( $order->get_items() as $key => $item ) :
-
-				// Skip over Bundles. Products included with a bundle will be displayed individually
+		<?php
+		foreach ( $orders as $order ) :
+			foreach ( $order->get_items_with_bundles() as $key => $item ) :
+				// We have to skip bundle parent.
```
robincornett, line 53: I am not _sure_ that the bundle parent is actually returned in the `get_items_with_bundles()` array--it doesn't seem to be for me--but it doesn't hurt to leave this in.

includes/orders/classes/class-order.php
```diff
@@ -355,6 +355,37 @@ public function get_items() {
...
+					$order_item_args = array(
+						'order_id'     => $this->ID,
+						'status'       => $item->status,
+						'product_id'   => edd_get_bundle_item_id( $bundle_item ),
+						'product_name' => edd_get_bundle_item_title( $bundle_item ),
+						'price_id'     => edd_get_bundle_item_price_id( $bundle_item ),
+					);
+					$new_items[]     = new \EDD\Orders\Order_Item( $order_item_args );
+				}
+				array_splice( $items, $index, 1, $new_items );
```
robincornett, line 382: I think it's highly unlikely that we could end up with `$new_items` being empty, but I'm wondering if we should only do the `array_splice` if it's not empty? Just as a safety check?

robincornett (commented):
Two minor questions but overall I'm good with this!

robincornett (dismissed):
I like it! Nice work @zkawesome ... @cklosowski waiting to merge this so you can look it over--adding the new order method is the solution Zan and I decided made the most sense to be consistent with the 2.x behavior and to be least invasive with the template. It's working for me, but I know you will want to have final eyes on it.

includes/orders/classes/class-order.php
```diff
@@ -355,6 +355,39 @@ public function get_items() {
...
 
+
+	/**
+	 * Retrieve all the items in the order and transform bundle items into regular items.
+	 *
+	 * @since 3.0.2
+	 *
+	 * @return Order_Item[] Order items.
+	 */
+	public function get_items_with_bundles() {
```
cklosowski, line 366: I don't love the name here. I feel like we could lock ourselves into it just being used for `bundles`.

Let's try and not limit ourselves in the future. Maybe name it something like `get_deliverable_items`.

This would give us the freedom (by name only, but make it more clear) that we are returning all products that this order gives access to, leaving room for filters later in the future.

cklosowski (changes requested):

includes/orders/classes/class-order.php
robincornett: My concern with get_deliverable_items is that we have a similarly named helper method which returns specific order item statuses. Using this name implies that we are returning items which have a deliverable status, I think.

I'm wondering if we could add a parameter to the get_items method and add this functionality in there instead? If we allowed an array as the parameter, then we could (for today) set a flag to include bundled items. Eventually we could include a status flag, to return only deliverable items. Not sure this is the right way to go but throwing out options.

cklosowski: @robincornett I wondered about that too but didn't want to drastically change that function immediately as those are actual items purchased, this is a different context than just getting the actual order items.

Let's just stick with this now and we can maybe work on adjusting at a later date.

cklosowski (dismissed):

cklosowski (approved):

PR #9312, opened 2022-07-22
Update tax functions, unit tests

robincornett:
This started with me trying to update the general taxes unit tests and discovering that they were failing when using registered tax rates instead of the EDD options.

Proposed changes:
1. Updates the `edd_get_tax_rates` to make the logic a little less gnarly and account for the global (not country and not regional) tax rate.
2. Updates `edd_get_tax_rate` to not use `edd_get_option`.
3. Then I discovered that if multiple regional rates were registered for a country (or probably just one), but no country wide rates, that `edd_get_tax_rate_by_location` would not correctly return the fallback tax rate, so that function has been updated to always get the fallback rate, and then return that if an otherwise matching rate could not be found. This is part of the reason the unit tests were failing, because this function wasn't quite right.
4. Then I updated the unit tests. I split some of them into singles while I was trying to diagnose what was wrong. Additionally, I'm not convinced that the calculation tests were testing for the correct values, so you'll see that those have been updated (I checked the math by hand and my math agrees with the unit tests).


includes/tax-functions.php
```diff
@@ -453,22 +446,24 @@ function edd_get_tax_rate_by_location( $args ) {
...
+		array(
+			'name'   => '',
+			'scope'  => 'global',
+			'status' => 'active',
+		),
+		OBJECT
+	);
 
-	if ( empty( $tax_rates ) ) {
+	if ( empty( $tax_rates ) && empty( $fallback ) ) {
```
robincornett, line 458: Really, if we removed this early return and only got the fallback if `$rate` is still false after going through the `foreach`, that might save us a database call.

robincornett:
Updated to group the tax rate tests in one folder and to properly deactivate taxes after the general tax class has run. Also, added a TN tax rate and a couple tests for that because it's 9.25% and we didn't yet have a test for rates with that many decimal places. I haven't noticed issues with it in EDD core, but I know we fixed an issue in Recurring with it.

cklosowski:
@robincornett yes, once we get through this merge, we can fix that in the main branch, this one is focused more on the underlying functionality, not the UI/UX components.

includes/database/queries/class-adjustment.php
```diff
@@ -167,4 +167,75 @@ class Adjustment extends Query {
...
+
+	/**
+	 * Handles adding an adjustment to the database.
+	 * Tax rates are unique in that only one can be active for one unique country/region combination,
+	 * so if the adjustment is a tax rate, we need to do some extra checking.
+	 *
+	 * @param array $data
+	 * @return int|false
+	 */
+	public function add_item( $data = array() ) {
```
cklosowski, line 179: This whole section feels really awkward and off. We're in the `class-adjustment.php`, yet we're calling things by `$existing_tax_rate_id`. I know the goal is to not allow duplicates, but we also shouldn't be 'updating' tax rates. Those are intended to be immutable, yet if we detect an existing tax rate we'll be able to modify it with the adjustment class. We need strong check around this `item_call` that we check the type before we determine if we allow updating or adding, and we should rename the methods to be more about being adjustments in general, and call out types of adjustments needing different treatments with comments.

cklosowski (changes requested):
Overall I think this does well, but I'm not in love the section I commented on.

It feels like we're treating the symptom, not curing the problem by making everything there all about taxes. We need a more robust check since tax rates shouldn't be allowed to be added if they create a duplicate, and for sure should not 'update' an existing tax ID because we need those to be immutable.

includes/database/queries/class-adjustment.php
robincornett: That's fair, and merits discussion. The reason the logic is in here to potentially update a tax rate is because that's what is currently in the tax compatibility class (if `update_option` is used to add tax rates). So I worked through this here to be consistent with that behavior. I agree, though, it seems off.

But the question is, then, is that wrong? And if so, how should this be handled? Should a new tax rate be registered as active and the original marked as inactive? Or should it silently fail?

If the former behavior is what it should be, then we can look at `edd_maybe_add_customer_address` as a starting point, and pull all of this logic out of the adjustment class and into a new `edd_add_tax_rate` function, and any time a tax rate is added, run it through that instead of directly through `edd_add_adjustment`. That would have some advantages in that we could ensure that tax rates are always complete (I found, for example, that I could add a tax rate with missing parameters--I think the status was one, and possibly even type?).

robincornett: Currently working to update this using a new `edd_add_tax_rate` function; I think that approach makes the most sense. It would still be possible to get into trouble using `edd_add_adjustment` directly but we can treat tax rates as specific kinds of objects.

cklosowski: @robincornett I think taking our queues from the `edd_maybe_add_customer_address` is a decent avenue to pursue.

PR #9314, opened 2022-07-24
Update orders table search

### LINKED ISSUE #9280: Searching the orders table no longer works as it used to
### Expected behavior
Searching for orders by ID/Purchase key and other variables should work

### Actual behavior
Right now searching by order ID or purchase key does not return any data for me.

1) Go to Downloads > orders 
2) Enter an order ID or the purchase key and try a search

PHP Version: 8.0

EDD Version (or branch): main

WordPress Version: latest

In 2.x version and in our in admin help text and docs we list the possible search terms below:
```
This search accepts:

    The purchase ID
    The 32-character purchase key
    The customer’s email address
    The customer’s name or ID prefixed by user:
    The Download ID prefixed by #
    The Discount Code prefixed by discount:
    A transaction ID prefixed by txn:
```
I feel like if possible we should allow all of those terms and if not clarify exactly what can be searched.

robincornett:
Fixes #9280

Proposed Changes:
1. Registers a new private method (`parse_search`) for the payments table class. This takes the search term and updates the array of orders query parameters based on the search string:
   - `:txn` searches by a transaction ID. This required adding a new filter to the order query class so that we could to an inner join to the order transaction table.
   - If the search string is an email address, then add the search term as an `email` parameter.
   - '#' searches by a product or download ID.
   - A numeric value is considered to be an order ID search.
   - `customer:` searches by a customer ID or a string...if a string is provided, then we search for customer IDs first and then send the customer IDs to the order parameters as `customer_id__in`. This type of search is new because I felt like searching by a "user" ID or name was actually kind of confusing, as not all stores have users, but all stores do have customers.
   - `user:` sets the `user_id` parameter. So this is technically different from what it was in 2.x.
   - `discount:` queries the adjustments table for a discount ID (by code) and then sets that as the `discount_id` parameter. This also required a new filter on the order clauses in the order query class.
2. Updates the contextual help to add the new search term and also removes the `<code></code>` fragments from the translatable strings because they should not be translated.
3. Since I added in the `discount_id` filter, I added a link to the discount row actions to show all orders placed with a given discount code. This has been something I've thought for a while would be nice to have and since I was 90% of the way there I figured I'd add it in.


robincornett:
@zkawesome I'm not able to replicate--I can search for a customer by first, last, or full name and it seems to find the right person. Actually searching by a partial email address works for me too. Can you give me more details about the customer name you're looking for and not finding (also what is their email)? I'll set up a copy of your customer and try it here.

robincornett:
Okay, thanks, that helps--let me see what I can sort out for that scenario.

cklosowski:
I have a few changes I want to see, I'll code them up and push to the branch.
1. If there is no customer name, use the email address of the order for the primary data for the column.
2. If there is a customer name, show the email address below the name (multiple complaints about this missing as they use it to help search)
3. If there is no customer attached (which shouldn't happen but just in case) add the order email address as the primary data for the column.
4. Adjust some of the column widths so the email addresses don't look awkward wrapping as much

Something I'm not sure if we want to retain, but once searched, the filters at the top retain the search parameters.

cklosowski:
View with adjusted widths and email display:
Screen Shot 2022-07-25 at 11 36 55 AM

robincornett:
I think the search being retained is the result of the search form using the referer instead of the base URL. It would be nice to use the base URL instead if we can because it's full of potentially empty filter parameters and you can end up with a URL that crashes the browser.

cklosowski:
> 

Okay I'll go ahead and fix that up.

robincornett:
If the referer could be sorted out then I think that would be fabulous; I think we could also tackle it in a separate issue tbh. I'm good with this PR with your changes, @cklosowski, as it stands, if you want to defer.

cklosowski:
Let's go ahead and just leave the referer as it is at this point. There is a way back by just clicking on the 'Orders' main menu item.

cklosowski (approved):

PR #9318, opened 2022-07-27
Change the upgrade message if the orders table could not be created

robincornett:
This adds an additional check to the 3.0 upgrade notice, checking for the orders table. If it was not created, an error notice is displayed instead of the upgrade notice. The error shows one of two messages:
1. If the database version doesn't meet the minimum 5.6 they're instructed to contact their host for help with upgrading that.
2. If the database version did meet the minimum then they're instructed to contact our support instead.

### DIFF
--- a/includes/admin/upgrades/upgrade-functions.php
+++ b/includes/admin/upgrades/upgrade-functions.php
@@ -206,8 +206,9 @@ function edd_show_upgrade_notices() {
 
 			// The final EDD Payment ID was recorded when the orders table was created.
 			$needs_migration = _edd_needs_v3_migration();
+			global $wpdb;
 
-			if ( $needs_migration ) {
+			if ( $needs_migration && ! empty( $wpdb->edd_orders ) ) {
 				?>
 				<div class="updated">
 					<?php if ( get_option( 'edd_v30_cli_migration_running' ) ) { ?>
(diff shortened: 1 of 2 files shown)


includes/admin/upgrades/upgrade-functions.php
```diff
@@ -281,6 +282,32 @@ function edd_show_upgrade_notices() {
...
+
+				// The orders database table is missing (we assume all primary tables have failed to create).
+				$message          = __( 'Easy Digital Downloads was unable to create the necessary database tables. Your site may not meet the minimum requirements for EDD 3.0.', 'easy-digital-downloads' );
+				$database_version = $wpdb->db_version();
+
+				// The database version is the problem.
+				if ( version_compare( $database_version, '5.6', '<' ) ) {
+					$message .= ' ' . sprintf(
+						/* translators: MySQL database version, do not translate */
+						__( 'Please contact your host and ask them to upgrade your MySQL version (your current version is %s).', 'easy-digital-downloads' ),
```
cklosowski, line 295: We can't leave them without what to upgrade to. Telling them that they need to upgrade their version could lead them to contact our support asking what version. We should link off to our recommended wordpress hosting page, which lists our requirements.

cklosowski (changes requested):

includes/admin/upgrades/upgrade-functions.php
robincornett: Sure, I can link to that...it makes me think that there is somewhere else that we want to link to something similar and it goes to the general EDD overview doc...which does have the minimum requirements, but all the way at the bottom. Updating that to use the same link would be better. I just have to remember where it was.

includes/admin/upgrades/upgrade-functions.php
```diff
@@ -281,6 +282,34 @@ function edd_show_upgrade_notices() {
 					?>
 				</div>
 				<?php
+			} else {
+
+				// The orders database table is missing (we assume all primary tables have failed to create).
+				$message          = __( 'Easy Digital Downloads was unable to create the necessary database tables. Your site may not meet the minimum requirements for EDD 3.0.', 'easy-digital-downloads' );
```
cklosowski, line 288:
```suggestion
				$message          = __( 'Easy Digital Downloads was unable to create the necessary database tables to complete this update. Your site may not meet the minimum requirements for EDD 3.0.', 'easy-digital-downloads' );
```

cklosowski (changes requested):

cklosowski:
Just one minor adjustment after seeing it live, and this is good to go.

cklosowski (approved):

PR #9333, opened 2022-08-01
Show any edd_template files copied to the theme for possible customization #7827

### LINKED ISSUE #7827: Phase 1: Add template override information to system information file
## Enhancement Request
Add "template override information to the system information file provided through support requests. 
It would be helpful to support to see any/all template overrides that are being applied by the user.

cklosowski:
Fixes #7827 

Proposed Changes:
1. Adds new `edd_get_theme_edd_templates` function, that finds any directories and files in the parent and child theme that are in the `edd_templates` directory and returns them as an array.
2. Adds the results of `edd_get_theme_edd_templates` to the system info.

### DIFF
--- a/includes/template-functions.php
+++ b/includes/template-functions.php
@@ -1172,3 +1172,43 @@ function edd_pagination( $args = array() ) {
 		</div>
 	<?php endif;
 }
+
+/**
+ * Return a list of theme files found in both parent and child themes.
+ *
+ * This allows us to detect when a core template is being overridden by a theme.
+ *
+ * @since 3.1
+ *
+ * @return array List of files that the parent and child themes are overriding, relative.
+ */
+function edd_get_theme_edd_templates() {
+	$parent_theme_dir = trailingslashit( get_template_directory() ) . 'edd_templates/';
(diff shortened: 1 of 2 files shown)


includes/template-functions.php
```diff
@@ -1172,3 +1172,43 @@ function edd_pagination( $args = array() ) {
...
+	$theme_edd_templates = array_map(
+		function( $file ) {
+			return str_replace( get_theme_root() . '/', '', $file );
+		},
+		array_merge( $found_templates, $found_child_templates )
+	);
+
+	return $theme_edd_templates;
+}
+add_action( 'admin_init', 'edd_get_theme_edd_templates' );
```
robincornett, line 1214:
```suggestion
```
I'm thinking we really only need to call this when it's wanted, right? eg in the system info function?

robincornett (changes requested):

includes/template-functions.php
cklosowski: Yeah @robincornett that's correct.

robincornett (approved):

PR #9335, opened 2022-08-02
Update base.js::onChange method to account for minification throwing JS errors in Safari #9331

### LINKED ISSUE #9331: Toggle button to "Automatically calculate amounts" not working in Safari
### Expected behavior
When I am adding a new Download into the Order, I expect the toggle button `Automatically calculate amounts` to show a form below it.

### Actual behavior
Toggling the button is not showing the form and there is an error thrown to the console.

1) Create new order
2) Click to `Add download`
3) Toggle `Automatically calculate amounts`

Browser: **SAFARI** !

cklosowski:
Fixes #9331 

Proposed Changes:
1. Adjust the `onChange` method in `base.js` to move the `target.selectionStart` check into the `try`, which avoids compilation from including it in the previously if statement, at which the value is not of a valid type check.


cklosowski:
@spencerfinnell This was something I believe you originally wrote up. Does this make sense? Basically the compiling was causing both of these if statements to be included into a single if statement checking for the 'positive' state, and the type error would get thrown on the checkboxes.

robincornett (approved):

spencerfinnell:
Strange, I haven't seen the compiler get it wrong like that before. Confirmed previous unexpected behavior on `main` and it working as expected on `fix/sfari-add-order-amounts`.

PR #9340, opened 2022-08-04
Update price ID checks for reports

### LINKED ISSUE #9337: Individual Download report does not show data in the sales tile
### Expected behavior
Filtering by individual downloads to show me the actual sales data

### Actual behavior
Amounts stay at 0

1) Go to Downloads > Reports and filter by any specific Download
2) The Sales tile is 0 https://a.supportally.com/68BZbZ

PHP Version: 7+

EDD Version (or branch): main

WordPress Version: latest

This was reported on .org but the screenshot above is from my local https://wordpress.org/support/topic/new-v3-0-reports/

robincornett:
Fixes #9337

Proposed Changes:
1. Updates the price ID checks for reporting queries.
2. Updates the sales/earnings overview tiles to set up the parameters when the class is instantiated rather than when calling the individual method.

### DIFF
--- a/includes/admin/reporting/reports.php
+++ b/includes/admin/reporting/reports.php
@@ -407,7 +407,7 @@ function edd_register_downloads_report( $reports ) {
 		if ( $download_data ) {
 			$download = edd_get_download( $download_data['download_id'] );
 			$prices   = $download->get_prices();
-			if ( $download_data['price_id'] ) {
+			if ( isset( $download_data['price_id'] ) && is_numeric( $download_data['price_id'] ) ) {
 				$args       = array( 'price_id' => $download_data['price_id'] );
 				$price_name = edd_get_price_name( $download->ID, $args );
 				if ( $price_name ) {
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/reports.php
```diff
@@ -277,14 +281,7 @@ function edd_register_overview_report( $reports ) {
...
 						$stats = new EDD\Stats();
-						return $stats->get_order_earnings( array(
-							'function'      => 'AVG',
-							'output'        => 'formatted',
-							'relative'      => true,
-							'range'         => $dates['range'],
-							'exclude_taxes' => $exclude_taxes,
-							'currency'      => $currency,
-						) );
+						return $stats->get_order_earnings();
```
cklosowski, line 284: This doesn't seem right, moving towards just getting order earnings when we're actually trying to get the average order value for the period queried...

cklosowski: This is producing this tile now:
Screen Shot 2022-08-08 at 1 54 25 PM

Instead of this:
Screen Shot 2022-08-08 at 1 54 47 PM

cklosowski (changes requested):

includes/admin/reporting/reports.php
robincornett: Whoops, maybe because I _deleted_ the array without replacing it. Fixed!

cklosowski (approved):

PR #9353, opened 2022-08-11
Improve discount validation for product requirements

robincornett:
Proposed Changes:
1. Updates the product requirements check to evaluate independently of the current "is met" status.
2. Moves the error message logic.

### DIFF
--- a/includes/class-edd-discount.php
+++ b/includes/class-edd-discount.php
@@ -1445,7 +1445,7 @@ class EDD_Discount extends Adjustment {
 		$cart_ids = array_values( $cart_ids );
 
 		// Ensure we have requirements before proceeding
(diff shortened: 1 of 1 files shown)


cklosowski:
Things to test:
1. Single product requirement
2. Multiple product requirements
3. If 'any' are in the cart
4. if 'all' are in the cart
5. Applies to entire cart
6. Applies to just selected items in the cart.

We should probably go through and also ensure that if you have any products in the exclusion list, that it works as expected as well, so that it doesn't reduce the price of an excluded item.

cklosowski:
This seems to be working as I expect now. @zkawesome let me know what your testing results are.

mihaijoldis:
Seems good here too!

cklosowski (approved):

Issue #9341, opened 2022-08-04
3.0 - Customer tables may not upgrade correctly in specific scenarios


robincornett:
### Expected behavior
When I update from EDD 2.x to 3.0, I expect the existing customer and customer meta tables to be updated to the expected 3.0 format.

### Actual behavior
If I set up my site and the database tables had a certain prefix (for example, `wp_` as the WordPress suggested default), and then changed them later to a custom prefix, these two tables (which have been in EDD for a long time) will not update correctly.

### Steps to reproduce the behavior
1. Set up an EDD 2.11 site.
2. However you want to do it, change the prefix for the database tables. (Some security plugins will do this for you and will suggest it as a way to "harden" your site.)
3. Update your site to EDD 3.0. The customer and customer meta tables will not upgrade.

### Information (if a specific version is affected):
PHP Version: 8.0
EDD Version (or branch): main (up to 3.0.2.1)
WordPress Version: 6.0

Any other relevant information:
In EDD 2.x, the customer tables versions were saved with the table prefix as part of the option name. Once the prefixes change, that option is not findable with our current upgrade check.

PR #9358, opened 2022-08-15
Add destroy option to re-run migration for orders

zackkatz:
Use case: I ran a migration that had issues. I want to re-run this migration without bulk-deleting the migrated orders. This makes the migration delete orders when the `--destroy` argument is passed to the CLI.

Related to #9341

### DIFF
--- a/includes/class-edd-cli.php
+++ b/includes/class-edd-cli.php
@@ -1321,6 +1321,10 @@ class EDD_CLI extends WP_CLI_Command {
 		$force = isset( $assoc_args['force'] )
 			? true
 			: false;
+		
+		$destroy = isset( $assoc_args['destroy'] )
+			? true
+			: false;
 
 		$upgrade_completed = edd_has_upgrade_completed( 'migrate_orders' );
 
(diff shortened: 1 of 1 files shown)


cklosowski:
Hey @zackkatz 3.0.3 has been baking over the weekend on our site and basically hit the release state, so this won't make 3.0.3 just b/c we need to get it out the door with fixes that have been tested already. I'll re-base it for what the next milestone would be.

cklosowski:
@robincornett It looks like `edd_destroy_order` should cover most of the main items here like customer data, discount usage, adjustments, order items, etc.

A couple concerns I have with this:
1. I feel like if we're running `--destroy` we should require that `--force` is also passed. By making it two arguments, you make it harder to 'accidentally' run this.
2. We've already had some people's larger sites hit the Linux Out of Memory Killer (OOM Killer) trigger when the addresses and other tables get a high number of alterations (which force a reindex). By running a `edd_destroy_order` it is basically doubling the actions for each order to migrate, which could result in a hefty tax on the server...maybe we just warn on that and make a 'confirm' dialog when `--destroy` is passed in that this is a server intensive process?

robincornett:
I think adding an additional confirmation of some type seems reasonable, whether it's requiring `--force` or doing an extra confirmation--probably the latter, actually, as I expect you would already be adding `--force` in this scenario.

One thing I would love to see, possibly in conjunction with this, but could be separate, would be a flag for order IDs. I have had scenarios where I've wanted to re-migrate specific orders (due to metadata issues, possibly), and being able to run a destroy/migrate on just a set of specific orders could, I think, be useful.

zackkatz:
> One thing I would love to see, possibly in conjunction with this, but could be separate, would be a flag for order IDs. I have had scenarios where I've wanted to re-migrate specific orders (due to metadata issues, possibly), and being able to run a destroy/migrate on just a set of specific orders could, I think, be useful.

Yes! Big agreement.

robincornett:
Closing this as we are moving forward with #9523, which makes use of this. Thank you @zackkatz!

Issue #9347, opened 2022-08-08
Review order of operations in stats class


robincornett:
Looking into this and tiles in general today and I'm seeing a few issues:
- Not specific to this tile, but if a `revenue_type` is passed to the `get_order_count` method in the stats class, it's ignored and the tiles use whatever was set up when the class was instantiated. So the `overview_sales` endpoint is setting a `revenue_type` of net, but `gross` is actually what's being used because that's the default.
- In the reporting functions, there are a mix of ways that the stats class is instantiated. Sometimes it's instantiated with the parameters that are meant to be used in the query; sometimes it's instantiated with no parameters and they're set in the query instead, and (as in the case of the order count) the parameters there may or may not be properly accounted for.

I think all of the tiles need to be reviewed, and the order of operations in the stats class as well.

_Originally posted by @robincornett in https://github.com/awesomemotive/easy-digital-downloads/issues/9337#issuecomment-1205238136_

PR #9362, opened 2022-08-16
Update Stats API Endpoint to use proper stats class and date adjustments #9361

### LINKED ISSUE #9361: Since EDD 3.0, the Stats API returns data that is inconsistent with the reports interface.
### Expected behavior
When I look at the iOS app or the `edd-api/stats` endpoints, I expect to see data that is consistent with what is in my admin Reports.

### Actual behavior
Because the stats API is still using old functions for dates and calculations, it is returning a set of data that is inconsistent with what is being shown in the WordPress admin.

1) Have a store with sales spanning multiple days/months.
2) View Reports in admin for Today,Yesterday,This Month, Last Month
3) Now view those in the API `yoursite.com/edd-api/stats?key=XXXXX&token=YYYY`

All endpoints were affected, but even just viewing the main endpoint with the default stats should expose the issue.

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

cklosowski:
Fixes #9361 

Proposed Changes:
1. Many changes but primarily removing calls to `get_sales/earnings_range` function calls in favor of using the stats class.
2. Stop using the old date formatting, and rely on the Store timezone adjusted UTC `date_created` column so the queries are correct when using dates and ranges.

### DIFF
--- a/includes/api/class-edd-api.php
+++ b/includes/api/class-edd-api.php
@@ -13,10 +13,14 @@
  * @copyright   Copyright (c) 2018, Easy Digital Downloads, LLC
  * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
  * @since       1.5
+ * @since       3.0.4 Refactored to use the new stats API, returns same formatting as 2.x API.
  */
 
 // Exit if accessed directly
 defined( 'ABSPATH' ) || exit;
+
+use EDD\Reports;
+use EDD\Stats;
 
 /**
  * EDD_API Class
@@ -156,7 +160,6 @@ class EDD_API {
 
 		// Setup EDD_Stats instance
 		$this->stats = new EDD_Payment_Stats;
-
 	}
 
 	/**
@@ -1245,17 +1248,33 @@ class EDD_API {
 					$start_date = $dates['year'] . '-' . $dates['m_start'] . '-' . $dates['day_start'];
 					$end_date   = $dates['year_end'] . '-' . $dates['m_end'] . '-' . $dates['day_end'];
 
-					$stats = EDD()->payment_stats->get_sales_by_range( 'other', true, $start_date, $end_date );
+					// Force the data for the reports API.
+					$_GET['filter_from'] = $start_date;
+					$_GET['filter_to']   = $end_date;
+					$_GET['range']       = 'other';
 
-					foreach ( $stats as $sale ) {
-						$key                    = $sale['y'] . $sale['m'] . $sale['d'];
-						$sales['sales'][ $key ] = (int) $sale['count'];
-					}
+					$stats   = new EDD\Stats();
+					$dates   = EDD\Reports\parse_dates_for_range();
 
-					$start_date = date( 'Y-m-d', strtotime( $start_date ) );
-					$end_date   = date( 'Y-m-d', strtotime( $end_date ) );
+					$total_sales = $stats->get_order_count(
+						array(
+							'start'        => $dates['start']->format( 'Y-m-d H:i:s' ),
+							'end'          => $dates['end']->format( 'Y-m-d H:i:s' ),
+							'revenue_type' => 'net',
+						)
+					);
+					$start_date  = date( 'Y-m-d', strtotime( $start_date ) );
+					$end_date    = date( 'Y-m-d', strtotime( $end_date ) );
 
 					while ( strtotime( $start_date ) <= strtotime( $end_date ) ) {
+
+						// Force the data for the reports API.
+						$_GET['filter_from'] = $start_date;
+						$_GET['filter_to']   = $start_date;
+						$_GET['range']       = 'other';
+
+						$dates = EDD\Reports\parse_dates_for_range();
+
 						$d = date( 'd', strtotime( $start_date ) );
 						$m = date( 'm', strtotime( $start_date ) );
 						$y = date( 'Y', strtotime( $start_date ) );
@@ -1263,7 +1282,13 @@ class EDD_API {
 						$key = $y . $m . $d;
 
 						if ( ! isset( $sales['sales'][ $key ] ) ) {
-							$sales['sales'][ $key ] = 0;
+							$sales['sales'][ $key ] = $stats->get_order_count(
+								array(
+									'start'        => $dates['start']->format( 'Y-m-d H:i:s' ),
+									'end'          => $dates['end']->format( 'Y-m-d H:i:s' ),
+									'revenue_type' => 'net',
+								)
+							);
 						}
 
 						$start_date = date( 'Y-m-d', strtotime( '+1 day', strtotime( $start_date ) ) );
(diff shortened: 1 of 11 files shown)


cklosowski:
@robincornett @zkawesome I'm sure this is far from perfect, so in your reviews, please feel free to adjust as needed.

This should be a good starting point to get through the API and iOS tickets that we keep getting in support.

includes/api/class-edd-api.php
```diff
@@ -1245,72 +1248,96 @@ public function get_stats( $args = array() ) {
...
-						$sales['sales'][ $key ] = (int) $sale['count'];
-					}
+					$stats   = new EDD\Stats();
+					$dates   = EDD\Reports\parse_dates_for_range();
+
+					$total_sales = $stats->get_order_count(
+						array(
+							'start'        => $dates['start']->format( 'Y-m-d H:i:s' ),
+							'end'          => $dates['end']->format( 'Y-m-d H:i:s' ),
+							'revenue_type' => 'net',
```
robincornett, line 1263: This needs to be moved to where the class is instantiated in line 1256, I think, due to #9347.

includes/api/class-edd-api.php
```diff
@@ -1245,72 +1248,96 @@ public function get_stats( $args = array() ) {
...
 
 						$key = $y . $m . $d;
 
 						if ( ! isset( $sales['sales'][ $key ] ) ) {
-							$sales['sales'][ $key ] = 0;
+							$sales['sales'][ $key ] = $stats->get_order_count(
+								array(
+									'start'        => $dates['start']->format( 'Y-m-d H:i:s' ),
+									'end'          => $dates['end']->format( 'Y-m-d H:i:s' ),
+									'revenue_type' => 'net',
```
robincornett, line 1290: If line 1256 is updated then I think this can be removed, as it's the same instance of the class? But noting that setting the `revenue_type` here may not currently work as expected.

includes/api/class-edd-api.php
```diff
@@ -1338,96 +1365,102 @@ public function get_stats( $args = array() ) {
...
-					if ( cal_days_in_month( CAL_GREGORIAN, $dates['m_end'], $dates['year'] ) < $dates['day_end'] ) {
-						$date_end = $dates['year_end'] . '-' . $dates['m_end'] . '-' . cal_days_in_month( CAL_GREGORIAN, $dates['m_end'], $dates['year'] );
-					} else {
-						$date_end = $dates['year_end'] . '-' . $dates['m_end'] . '-' . $dates['day_end'];
-					}
+					$total_earnings = $stats->get_order_earnings(
+						array(
+							'start'         => $dates['start']->format( 'Y-m-d H:i:s' ),
+							'end'           => $dates['end']->format( 'Y-m-d H:i:s' ),
+							'revenue_type'  => 'net',
```
robincornett, line 1383: See line 1376.

robincornett (changes requested):

cklosowski:
Okay updated those @robincornett

robincornett (commented):
The results for `get_order_earnings` are not formatted--should they be?

cklosowski:
@robincornett This is ready for another look.

includes/api/class-edd-api.php
```diff
@@ -2260,12 +2301,37 @@ private function get_default_sales_stats() {
...
+			array(
+				'range'         => 'last_month',
+				'exclude_taxes' => ! $args['include_tax'],
+				'revenue_type'  => 'net',
+				'output'        => 'typed',
+			)
+		);
+		$earnings['earnings']['last_month'] = $stats->get_order_earnings();
+
+		$earnings['earnings']['totals'] = edd_get_total_earnings( $args['include_tax'] );
```
robincornett, line 2334: Since this is an independent function, I think it needs to be run through `edd_format_amount` as well in order to get the proper format, or else maybe the `edd_get_total_earnings` function should be run through that instead of what it is now. Anyway, the `totals` for me is still showing too many decimal places, although everything else looks correct now.

cklosowski: @robincornett I'm not sure how...b/c that function is run through here:
`return apply_filters( 'edd_total_earnings', round( $total, edd_currency_decimal_filter() ) );`

So it should only ever respond in 2 decimals. Or it is cached for some reason with more decimals since that uses a transient still.

robincornett: I think `edd_get_total_earnings` should be updated to return its value using `edd_format_amount`. I'm getting inconsistent results with `round` even though `edd_currency_decimal_filter()` is returning 2.

For example, today, the `totals` in the API request is showing as `5645`, which is exactly what the transient is. The current code is not formatting the amount to include the decimal places, but updating the function to use `edd_format_amount` instead properly displays `"5645.00"`.

I manually updated the transient to match what it was previously: `5336.8`. The `totals` amount is now showing as `5336.8000000000002` instead of the expected `"5336.80"`.

cklosowski: I think updating `edd_get_total_earnings` has far more reaching side affects, as your data type changed to a string from a float. If we just make the API format the totals using the `maybe_format_amount` method we should be in the clear.

this would affect the API data only, and not bring changes to the other areas of EDD calling `edd_get_total_earnings`.

cklosowski: I'll play around with this and commit when I have something I'm comfortable with for this.

cklosowski:
@robincornett this is ready for another look over with the changes we discussed on a call.
1. Improving the data type output.
2. fixing up `edd_get_total_earnings` to use a transient
3. fixing up `edd_get_total_earnings` to use the stats class so it is aware of things like renewals.

robincornett (dismissed):

cklosowski:
This is causing a bunch of tests to fail, looking into if the tests are incorrectly written.

cklosowski:
Okay the unit tests surrounding this PR are fixed @robincornett the failures are from the other PR we have open for the user attachment only. If you can re-approve I'll merge this in.

includes/currency/class-money-formatter.php
```diff
@@ -113,6 +113,37 @@ public function format_for_display( $decimals = true ) {
...
+		/**
+		 * Filter number of decimals to use for formatted amount
+		 *
+		 * @since unknown
+		 *
+		 * @param int        $number        Default 2. Number of decimals.
+		 * @param int|string $amount        Amount being formatted.
+		 * @param string     $currency_code Currency code being formatted.
+		 */
+		$decimals = apply_filters( 'edd_format_amount_decimals', $decimals ? $this->currency->number_decimals : 0, $amount, $this->currency->code );
```
robincornett, line 139: I meant to ask this yesterday but we got derailed by the formatting thing in Chrome.

Can we pull this `apply_filters` out into a separate private method here, since it's now applied identically in this method and `format_for_display`? I don't love how much duplication there is between the two methods, and wondered initially if the latter could use the new function and then return the customized output, but the final filters seem to make that not possible.

includes/currency/class-money-formatter.php
```diff
@@ -113,6 +113,37 @@ public function format_for_display( $decimals = true ) {
...
+
+		// Format amount using decimals and separators (also rounds up or down).
+		$formatted = number_format( (float) $amount, $decimals, $this->currency->decimal_separator, $this->currency->thousands_separator );
```
robincornett, line 142: I think this is actually not correct--shouldn't this be assigned to `$this->amount` instead of just a local variable? I'm thinking the current assignment actually means that the result is never used.

If I change it locally to `$this->amount` then the browser results are actually what I expect to see, although they are strings, not floats. But I think right now this method isn't actually modifying the end result.

robincornett (changes requested):

includes/currency/class-money-formatter.php
cklosowski: Done in https://github.com/awesomemotive/easy-digital-downloads/pull/9362/commits/e3ec422b9bcf853764fd5c3e9a71220d40d2d2ef

includes/currency/class-money-formatter.php
cklosowski: @robincornett Updated, unit tests still pass, have a look at the change in https://github.com/awesomemotive/easy-digital-downloads/pull/9362/commits/e3ec422b9bcf853764fd5c3e9a71220d40d2d2ef

robincornett (approved):

PR #9363, opened 2022-08-17
Improved relative date ranges and reports UI

### LINKED ISSUE #9338: EDD 3.0 Reports - "Relative" date ranges do not account for partially completed ranges vs previous range.
### Expected behavior
On the Reports page, the "Relative" arrows (and percentages) should correctly compare the current time slice with the previous time slice.

If `This Month` is set and it's August 3rd, then the "Relative" stats should compare the first 3 days of August with the first 3 days of July.

### Actual behavior
Currently it's comparing all of August (i.e. the first 3 days) with all of July. So at the beginning of a month, the stats are always down by 90%, skewing the stats.

1) Browse to the Reports page
2) See the "Relative" arrows / percentages below each statistic

PHP Version: 8.1

EDD Version (or branch): 3.0.2.1

WordPress Version: 6.0.1

zkawesome:
Fixes #9338

Proposed Changes:
- Made changes inside `set_date_ranges` method - I have separated parsing of relative date fields out of `Stats` class and into the `Reports` class so that it's consistent with what we have with `parse_dates_for_range`. We are now getting date and relative date ranges from a single source.
- Changed how relative periods are calculated. Everything that is `this_` [`this_month`, `this_week`,..] is now using today as the end date
- Added `other` to the relative date ranges, so that we can support `Custom` date range - it will compare custom dates to the month before them.
- UI changes inside reports. I wanted user to be able to see exactly what date range is active and against which dates they are comparing the data to:

Annotation on 2022-08-17 at 16-34-13


JJJ:
I like this better 👍

cklosowski:
@zkawesome Something seems off on these dates here. I'm not sure 100% of it's just the label or if the actual reporting is off.

I'm pulling this on what should be August 31st (00:17am on August 31st) however the date range for last 30 days shows until September 1st.

Also the dates overlap as in we're starting at August 1st, but we're also comparing up to August 1st. 
Screen Shot 2022-08-30 at 10 16 55 PM

cklosowski:
@zkawesome I'm really liking where this is heading, can we open it up for a 'Ready for Review'?

assets/css/admin/reports/_filter.scss
```diff
@@ -15,10 +15,228 @@
...
+			display: flex;
+			align-items: center;
+			border: 1px solid #8c8f94;
+			border-left: none;
+			color: #2c3338;
+			padding: 4px 10px;
+			margin-left: -5px;
+			border-top-right-radius: 4px;
+			border-bottom-right-radius: 4px;
+			cursor: pointer;
```
robincornett, line 54:
```suggestion
			cursor: pointer;
			gap: 4px;
```
And then you can remove the rule for `.edd-date-main-icon` below.

assets/css/admin/reports/_filter.scss
```diff
@@ -15,10 +15,228 @@
...
+			background-color: #fff;
+			border: 1px solid #8c8f94;
+			border-radius: 4px;
+			-webkit-box-shadow: 0px 2px 5px 0px rgba(0,0,0,0.25);
+			-moz-box-shadow: 0px 2px 5px 0px rgba(0,0,0,0.25);
+			box-shadow: 0px 2px 5px 0px rgba(0,0,0,0.25);
+			display: none;
+
+			.spinner {
+				display: none;;
```
robincornett, line 119:
```suggestion
				display: none;
```

assets/css/admin/reports/_filter.scss
```diff
@@ -15,10 +15,228 @@
...
+			z-index: 99;
+			width: 420px;
+			left: 50%;
+			top: 100%;
+			margin-top: 10px;
+			transform: translateX( -50% );
+			background-color: #fff;
+			border: 1px solid #8c8f94;
+			border-radius: 4px;
+			-webkit-box-shadow: 0px 2px 5px 0px rgba(0,0,0,0.25);
```
robincornett, line 113: I think webpack handles the prefixing--I'm not sure you need to have that in the .scss files.

assets/css/admin/reports/_filter.scss
```diff
@@ -15,10 +15,228 @@
...
+				transform: translateX( -50% );
+				border-style: solid;
+				border-width: 10px 10px 0 10px;
+				border-color: #8c8f94 transparent transparent transparent;
+			}
+			ul li {
+				display: flex;
+				align-items: center;
+				padding: 2px 10px;
+				opacity: 0.85;
```
robincornett, line 151:
```suggestion
				opacity: 0.85;
				gap: 20px;
```

assets/css/admin/reports/_filter.scss
```diff
@@ -15,10 +15,228 @@
...
+				&:hover, &.active {
+					cursor: pointer;
+					color: var( --wp-admin-theme-color );
+					opacity: 1;
+				}
+				.date-range-name {
+					margin-right: 20px;
```
robincornett, line 158:
```suggestion
```

includes/reports/reports-functions.php
```diff
@@ -507,6 +510,37 @@ function get_dates_filter_options() {
 	return apply_filters( 'edd_report_date_options', $options );
 }
 
+
```
robincornett, line 513: We're ending up with three blank lines here instead of just one--

includes/reports/reports-functions.php
```diff
@@ -507,6 +510,37 @@ function get_dates_filter_options() {
...
+	}
+
+	/**
+	 * Filters the list of key/label pairs of relative date filter options.
+	 *
+	 * @since 3.1
+	 *
+	 * @param array $date_options Date filter options.
+	 */
+	return apply_filters( 'edd_report_relative_date_options', $options );
```
robincornett, line 541: Do we need this filter at this time?

includes/reports/reports-functions.php
```diff
@@ -1053,50 +1192,131 @@ function default_display_charts_group( $report ) {
...
+		<!-- DATE RANGES -->
+		<div class="edd-date-range-dates">
+			<span class="dashicons dashicons-calendar edd-date-main-icon"></span>
+			<div class="edd-date-range-selected-date">
+				<?php
+				foreach ( $range_options as $range_key => $range_name ) :
+					$range_dates          = \EDD\Reports\parse_dates_for_range( $range_key );
+					$selected_range_class = ( $selected_range !== $range_key ) ? 'hidden' : '';
+					?>
+					<span class="<?php echo esc_attr( $selected_range_class ); ?>" data-range="<?php echo esc_attr( $range_key ); ?>"><?php echo esc_html( edd_get_edd_timezone_equivalent_date_from_utc( $range_dates['start'] )->format( $date_format ) ); ?> - <?php echo esc_html( edd_get_edd_timezone_equivalent_date_from_utc( $range_dates['end'] )->format( $date_format ) ); ?></span>
```
robincornett, line 1268: This is a minor quibble, but if the start and end dates are identical (because I picked today or yesterday or a single date), could this show just the one date instead of the `from - to`? I feel like that would be nice:

image

Also, and I don't know if this is possible, but when I change this date selector to something that's a different range, it would be nice if the "compared to" dropdown reset to the previous period instead of staying whatever it was--I switched from a quarter range to today and the comparison was still showing "previous quarter".

robincornett (commented):
I think this is looking amazing overall--just a few questions here and there.

robincornett (approved):

cklosowski (approved):

PR #9366, opened 2022-08-18
Fix string that is not translatable #9365

### LINKED ISSUE #9365: "Personal info" string on checkout is not running through gettext.
### Expected behavior
The 'Personal info' string should be translated on checkout.

### Actual behavior
The 'Personal info' string is not running through translation functions.

1) Install the following language pack, into `/wp-content/languages/plugins/` : https://a.supportally.com/VZbEsD
2) Set your site's language to French (France)
3) View Checkout
Screen Shot 2022-08-18 at 3 36 17 PM

The string should read `Informations personnelles`

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

cklosowski:
Fixes #9365 

Proposed Changes:
1. Updates the 'Personal info' text on checkout to be run through gettext.

### DIFF
--- a/includes/checkout/template.php
+++ b/includes/checkout/template.php
@@ -176,7 +176,7 @@ function edd_user_info_fields() {
 	$customer = array_map( 'sanitize_text_field', $customer );
 	?>
 	<fieldset id="edd_checkout_user_info">
-		<legend><?php echo apply_filters( 'edd_checkout_personal_info_text', esc_html__( 'Personal info', 'easy-digital-downloads' ) ); ?></legend>
+		<legend><?php echo apply_filters( 'edd_checkout_personal_info_text', esc_html_e( 'Personal info', 'easy-digital-downloads' ) ); ?></legend>
 		<?php do_action( 'edd_purchase_form_before_email' ); ?>
 		<p id="edd-email-wrap">
 			<label class="edd-label" for="edd-email">


robincornett (changes requested):
I'm not sure this is necessary. With the language pack installed, the default text is translated on the main branch:
image

Was the issue really with CFM? I saw that you set up a PR there for this string as well. Additionally, this PR is just introducing a duplicate echo, from what I can see.

includes/checkout/template.php
```diff
@@ -176,7 +176,7 @@ function edd_user_info_fields() {
 	$customer = array_map( 'sanitize_text_field', $customer );
 	?>
 	<fieldset id="edd_checkout_user_info">
-		<legend><?php echo apply_filters( 'edd_checkout_personal_info_text', esc_html__( 'Personal info', 'easy-digital-downloads' ) ); ?></legend>
+		<legend><?php echo apply_filters( 'edd_checkout_personal_info_text', esc_html_e( 'Personal info', 'easy-digital-downloads' ) ); ?></legend>
```
JJJ, line 179: Agree with @robincornett that:
* `esc_html__()` was the correct function
* `esc_html_e()` results in 2 `echo`'s

If an addon is filtering `edd_checkout_personal_info_text`, its return value will need to be translated using its own dedicated `textdomain` as registered. If it's expecting to use `easy-digital-downloads` for its `textdomain` it will need to perform additional work to do so.

cklosowski:
It 100% was CFM, I'll just close this as a non issue and just have the Checkout Fields one do it.

PR #9367, opened 2022-08-19
When adding a new customer via manual order, attach to a found user with the same email address #9339

### LINKED ISSUE #9339: Creating an order manually with a new customer but with an existing WP User account should link the user to customer
### Expected behavior
If I create a new order and select to create a new customer with an email that is already registered as a WP User I expect the Customer and User to be connected

### Actual behavior
The Customer is created but you have to manually edit it again to attach the User with the same email to it.

1) Go to Dashboard > Users and create a new user account.
2) Go to Orders > Add new and select create new Customer and use the same email as the newly created user account above.
3) Once the order is created click on View Customer details
4) The Customer has no User attached to it but it should be linked to the existing User with that email.

PHP Version: 7+

EDD Version (or branch): main

WordPress Version: 6+

issue reported here https://secure.helpscout.net/conversation/1963437726/2080861?folderId=5762562
This is a regression from 2.11 with Manual Payments extension. In 2.11 with manual purchases, the Customer that was created when the new payment is logged is attached to the existing user account automatically.

cklosowski:
Fixes #9339 

Proposed Changes:
1. When manually adding an order, search for existing users with the same email address.
2. If there is a user, get the User ID and attach it to the customer record.
3. Fixing some PHPCS issues with comments.


includes/orders/functions/actions.php
```diff
@@ -71,29 +71,39 @@ function edd_add_manual_order( $args = array() ) {
 	// Create a new customer record.
 	if ( isset( $order_data['edd-new-customer'] ) && 1 === absint( $order_data['edd-new-customer'] ) ) {
```
robincornett, line 72: This is not technically part of this issue, but I experimented with selecting "add a new customer" and then entering an email address for an existing customer and that was kind of fun:

```
WordPress database error Duplicate entry 'testpaypal@edd.local' for key 'email' for query INSERT INTO `wp_edd_customers` (`user_id`, `email`, `name`, `status`, `purchase_value`, `purchase_count`, `date_created`, `date_modified`, `uuid`) VALUES (241, 'testpaypal@edd.local', 'Test PayPal', 'active', '0', '0', '2022-08-19 13:48:42', '2022-08-19 13:48:42', 'urn:uuid:d8d0609e-ed10-4492-bde2-3a68ebe57759') made by require_once('wp-admin/admin.php'), require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, edd_post_actions, do_action('edd_add_order'), WP_Hook->do_action, WP_Hook->apply_filters, edd_add_manual_order, edd_add_customer, EDD\Database\Query->add_item
[19-Aug-2022 13:48:42 UTC] PHP Warning:  Attempt to read property "id" on bool in \easy-digital-downloads\includes\orders\functions\actions.php on line 241
```

So I'm feeling like we should harden this check to make sure that a customer doesn't already exist.

Otherwise this is working correctly, so if you want to handle that in a separate issue then I'm okay with that.

robincornett (approved):

includes/orders/functions/actions.php
cklosowski: @robincornett I can harden this, to that if it finds a customer with that email address it attaches it to the existing customer instead.

We can open an issue to improve the UI to do a customer check and suggest the existing customer if it's found.

cklosowski: @robincornett Take a look at these latest changes to see if you like this. I had to update the `edd_get_customer_by` function to account for non-primary email addresses.

If you think this is too much for 3.0.4 we can roll it back and just look for primary email addresses as well.

robincornett (commented):
I'm not sure where this is going sideways, but now, if I select the option to create a new user and enter an email address that I _know_ doesn't already exist for a user, somehow it's attempting to maybe add the customer twice?

```
WordPress database error Duplicate entry 'ki@edd.local' for key 'email' for query INSERT INTO `wp_edd_customers` (`user_id`, `email`, `name`, `status`, `purchase_value`, `purchase_count`, `date_created`, `date_modified`, `uuid`) VALUES (0, 'ki@edd.local', 'ki@edd.local', 'active', '0', '0', '2022-08-22 18:27:12', '2022-08-22 18:27:12', 'urn:uuid:8cb8a0de-5af4-4fdc-a76f-830d391acf76') made by require_once('wp-admin/admin.php'), require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, edd_post_actions, do_action('edd_add_order'), WP_Hook->do_action, WP_Hook->apply_filters, edd_add_manual_order, edd_update_order_status, EDD_Payment->save, EDD_Payment->maybe_create_customer, EDD_Customer->create, edd_add_customer, EDD\Database\Query->add_item
```
The customer is created, but not correctly assigned to the new order, and the user ID assigned to the order seems to be mine.

Additionally, if I enter a customer's secondary email, it does properly assign the order to that customer, but the manually entered email address is not used. I know we cannot select a secondary email address for an existing user when the order is created, but it seems like if I manually enter a specific email address, that's the one that should be used for the order.

cklosowski:
@robincornett i'm reverting all the additional changes other than associating with an existing user. The rest is moving past the scope of the original report, and started to include some changes to the `edd_get_customer_by` that i'm not fully comfortable with (and I'm thinking is what was causing the issues we started to see).

So if we can just retest with the scope of when adding a customer to a manual order, attach it to a user if one is found by the email address provided.

robincornett (approved):
I think something went a little sideways with commit [b79ed44](https://github.com/awesomemotive/easy-digital-downloads/pull/9367/commits/b79ed445246c269c6b19fb835ffb6089b3e556d4) (the PR is showing unrelated changes now), but this is linking the new customer to the existing user. Adding an order for an entirely new customer works, as does adding an order for an existing customer.

robincornett (approved):

mihaijoldis:
This is working correctly here!

PR #9368, opened 2022-08-19
Swtich from linking customer/user on customer creation, not payment attachment #9344

### LINKED ISSUE #9344: 3.0 - Guest customers for existing users are not connected to the user
### Expected behavior
When an existing user who is not already a customer makes a purchase, I expect the customer to be attached to the user by `edd_connect_guest_customer_to_existing_user`.

### Actual behavior
That function is hooked into `edd_customer_post_attach_payment`, which is not running during the normal checkout process in EDD 3.0.

1. Make sure that guest checkout is enabled.
2. Make a purchase with an email address which belongs to a registered user who is not already a customer. This _should_ link the new customer automatically to the existing user, but it does not.

PHP Version: 8.0
EDD Version (or branch): 3.0 (main)
WordPress Version: 6.0

Looking at the customer `attach_payment` method, that's where this hook is registered. However, it never fires because there is an early return if the order is already attached to the customer, and the way the orders are created in 3.0, the customer ID is assigned to the order before this method is even called. It also might mean that the customer stats are not updated, but I'm not seeing that happen elsewhere, so I think they are being updated during the purchase process, just not here.

Not sure if the early return needs to be updated (I assume it was present in 2.x but with the improved order creation process it's not working the same way), or if the function needs to be connected to a different hook, but marking this as a regression.

cklosowski:
Fixes #9344 

Proposed Changes:
1. Switches the hook for linking a customer to a user from `edd_customer_post_attach_payment` to `edd_customer_post_create` as that is more accurate of when we should link them.

### DIFF
--- a/includes/user-functions.php
+++ b/includes/user-functions.php
@@ -385,40 +385,48 @@ function edd_validate_username( $username ) {
 }
 
 /**
- * Attach the customer to an existing user account when completing guest purchase
+ * Attach the customer to an existing user account when the customer record is created.
  *
- * This only runs when a user account already exists and a guest purchase is made
+ * This only runs when a user account already exists and a customer is created
  * with the account's email address
  *
  * After attaching the customer to the user ID, the account is set to pending
  *
  * @since  2.8
- * @param  bool   $success     True if payment was added successfully, false otherwise
- * @param  int    $payment_id  The ID of the EDD_Payment that was added
- * @param  int    $customer_id The ID of the EDD_Customer object
- * @param  object $customer    The EDD_Customer object
+ * @since  3.0.4 This is now hooked into the customer creation process instead.
+ *
+ * @param int   $customer_id If created successfully, the customer ID.
+ * @param array $args        Contains customer information such as payment ID, name, and email.
+ *
  * @return void
  */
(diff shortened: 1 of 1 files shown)


robincornett (commented):
This is working for me, but I question why it doesn't also fix #9339. I mean, I know WHY, because the manual order process uses `edd_add_customer()` and a frontend payment uses `$customer->create()`, which then calls `edd_add_customer()`.

I'm left wondering if the hooks should be moved to that function instead, although I think the `post_create` would not work with the parameters available there.

I think I would love to see a hook in the `edd_add_customer` function and use that instead, or if there was one in Berlin on `add_item`, that feels like it would be useful. 🤷🏼‍♀️

cklosowski:
@robincornett I'd prefer to keep the the hooks/filters a bit further away from Berlin if possible, as that would let us update and manage berlin's hooks/filters independently. I believe Berlin is intentionally sparse on hooks/filters to avoid interfering with the actual data process, and all 'shaping' should happen before we call it.

I don't like the idea of using a hook in `edd_add_customer` as it's only job is to add the customer and return the customer created. This allows us to very intentionally build in hooks where we want instead of always running a hook when a customer is added.

robincornett (approved):

PR #9370, opened 2022-08-19
Use a transient to only check the shortcode-receipt template once a day #9354

### LINKED ISSUE #9354: Update receipt template check/modification
We should review the `update_receipt_template` method.

### Justification or use case
A customer has reported that it's running more than it ought. Looking at the code, it looks like it might be updating the template file even if it doesn't need to (if it's already been done). We should consider changing that so that it doesn't run if the strings to replace are not found; additionally, maybe this should run on a cron event hook.

Alternatively, this could be run as an upgrade routine and only run once. The concern there is if a store updates the template to revert back to using `get_post` instead of `edd_get_payment`, that the receipt template will be broken.

cklosowski:
Fixes #9354 

Proposed Changes:
1. Adds a transient of `edd-sc-receipt-check` when checking the `shortcode-receipt.php` template file to only run the check once a day.

### DIFF
--- a/includes/compat/class-template.php
+++ b/includes/compat/class-template.php
@@ -55,6 +55,11 @@ class Template extends Base {
 	public function update_receipt_template() {
 		$access_type = get_filesystem_method();
 
+		$last_checked = get_transient( 'edd-sc-receipt-check' );
+		if ( false !== $last_checked ) {
+			return false;
+		}
+
 		// Retrieve the path to the template being used.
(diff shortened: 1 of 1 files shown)


robincornett (commented):
I guess I feel like we should return early if `$get_post_call_exists` is false, and avoid running the `str_replace` and `put_contents` calls altogether in that case. Would that not also be worth doing?

cklosowski:
@robincornett I agree, this should do the trick.

robincornett (approved):

PR #9372, opened 2022-08-19
Reporting groupby should determine the difference between months in multiple years #9371

### LINKED ISSUE #9371: A report with date ranges that span more than 1 year, will group stats by month only, not year as well.
### Expected behavior
When I choose a report that is From January 1, 2020 - December 31, 2021, I expect the graph to be populated for all months where there were sales.

### Actual behavior
Because of this line:
```
	$sql_clauses['groupby'] = Reports\get_groupby_date_string( 'MONTH', 'date_created' );
	$sql_clauses['orderby'] = 'MONTH(date_created)';
```
The graph is grouping only by month, not a combination of month and year:
Screen Shot 2022-08-19 at 12 41 39 AM

1) Have a store with sales that span 2 years 2020-2021
2) In the Reports Dashboard set a custom range that will have the same month in both years on the graph. 
3) The first occurence of the month in the graph will get all stats for that month from both years.

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

cklosowski:
Fixes #9371 

Proposed Changes:
1. Runs `$function` through `strtoupper` as MySQL functions are always upper case.
2. When the `$function` is `MONTH`, adjusts the query to include `YEAR` and `MONTH` in order to avoid grouping the same month from different years as the same data point.

### DIFF
--- a/includes/reports/reports-functions.php
+++ b/includes/reports/reports-functions.php
@@ -795,18 +795,28 @@ function get_dates_filter_day_by_day() {
  * Given a function and column, make a timezone converted groupby query.
  *
  * @since 3.0
+ * @since 3.0.4 If MONTH is passed as the function, always add YEAR and MONTH
+ *              to avoid issues with spanning multiple years.
  *
- * @param string $function The function to run the value through, like DATE, HOUR, MONTH
+ * @param string $function The function to run the value through, like DATE, HOUR, MONTH.
  * @param string $column   The column to group by.
  *
  * @return string
  */
 function get_groupby_date_string( $function = 'DATE', $column = 'date_created' ) {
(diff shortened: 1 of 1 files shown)


robincornett (approved):

mihaijoldis:
Just had confirmation from a user that this worked for their issue!

sc0ttkclark:
Me ^^ and it works!

cklosowski:
@mihaijoldis @sc0ttkclark We made a few more adjustments to this in #9387 as I found a similar thing was happening with the `hour by hour` graphs.

sc0ttkclark:
I noticed that so I took the file from the release/3.0.4 branch. Thx for letting me know just in case though

PR #9378, opened 2022-08-22
Removing old W3TC hooks for discounts when they were a post type #9375

### LINKED ISSUE #9375: Persistent cache can cause discount codes to not activate/deactivate
### Expected behavior
When running a persistent object caching layer (like memcache or redis), using the 'activate' and 'deactivate' actions for a discount should update the cache.

### Actual behavior
when a persistent object caching layer is active, when you click 'Deactivate', the license key remains to appear active (and is), however the 'inactive' count is increased.

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

cklosowski:
Fixes #9375

Proposed Changes:
1. Removes hooks for discount changes to hook into W3TC

### DIFF
--- a/includes/class-edd-cache-helper.php
+++ b/includes/class-edd-cache-helper.php
@@ -75,18 +75,6 @@ class EDD_Cache_Helper {
 				}
 			}
 		}
-
-		add_action( 'edd_pre_update_discount',         array( $this, 'w3tc_suspend_cache_addition_pre' ) );
-		add_action( 'edd_pre_insert_discount',         array( $this, 'w3tc_suspend_cache_addition_pre' ) );
-		add_action( 'edd_pre_delete_discount',         array( $this, 'w3tc_suspend_cache_addition_pre' ) );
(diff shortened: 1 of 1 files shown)


includes/class-edd-cache-helper.php
```diff
@@ -133,22 +121,4 @@ public function notices() {
 		}
 
 	}
-
-	/**
-	 * Prevents W3TC from adding to the cache prior to modifying data.
-	 *
-	 * @since 1.7
-	 */
-	public function w3tc_suspend_cache_addition_pre() {
```
robincornett, line 142: I'm wondering if it would be safer to leave these in, and possibly just empty them, since they're public functions--it seems like someone could be using them in custom code.

cklosowski: @robincornett A search on all of GitHub for the function names results in ONLY EDD Forks.

I'm considering that safe to remove.

robincornett (approved):

PR #9387, opened 2022-08-23
Hour by Hour graph grouped only by hour, not day

cklosowski:
Fixes #

Proposed Changes:
1. If 'HOUR' is used as the grouping, ensures that we group by 'hour' and 'day' to fix a 2 day hour by hour report.

Two replicate, select a graph that reports hour by hour, like a 2 day span.

With Changes:
image

Without Changes:
Screen Shot 2022-08-23 at 1 30 07 PM


cklosowski:
@robincornett I updated this to be a little more DRY.

robincornett (approved):

PR #9391, opened 2022-08-25
Allow using a constant EDD_TEST_MODE to force test_mode to true #7098

### LINKED ISSUE #6712: Checkmark 'Test Mode' when using a filter to turn on test mode

When using the `edd_is_test_mode` filter to enable test mode, the checkmark in Payment Gateways > General can still be toggled on/off. This checkmark should be disabled and showed as checked to show that the filter is in effect.

### LINKED ISSUE #7098: Feature request: Allow Test Mode to be activated in an environment via a constant in wp-config.php
Hi, 

I would like to be able to put my shop in test mode when I am using the staging copy.

So far I have found this code: 
```
/**
 * Is Test Mode
 *
 * @since 1.0
 * @return bool $ret True if test mode is enabled, false otherwise
 */
function edd_is_test_mode() {
	$ret = edd_get_option( 'test_mode', false );
	return (bool) apply_filters( 'edd_is_test_mode', $ret );
}
```
But that means I need to install a custom plugin and activate it on my staging. Can be done but, 
would be possible to have an easier way to do this. 
For example, 2 suggestions: 
A) having a file (.maintenance or similar) on the edd plugin folder that when is there, the shop is considered "test mode". Removing the file will remove the test mode status. 
B) having a variable on wp-config, that will set the shop on test mode, with true or false. 
So it will be easy to add that to the staging creating process.

What do you think? 
Is this something that can be added in the future? 

Thanks

cklosowski:
Fixes #7098
Fixes #6712 

Proposed Changes:
1. Adds support for the `EDD_TEST_MODE` constant.
2. If `true` is the value, puts the store in test mode, disables the setting and always returns `true` for calls to `edd_is_test_mode()` and `edd_get_option( 'test_mode' )`
3. Adds a new `edd_is_test_mode_forced()` which uses the constant and the filter `edd_is_test_mode` (when using `__return_true`) to determine if we're forcing test mode so we can set the setting to disabled and readonly.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -60,7 +60,7 @@ function edd_is_test_mode() {
 	$ret = edd_get_option( 'test_mode', false );
 
 	// Override any setting with the constant.
-	if ( defined( 'EDD_TEST_MODE') && true === EDD_TEST_MODE ) {
+	if ( edd_is_test_mode_forced() ) {
 		$ret = true;
 	}
 
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -58,6 +58,13 @@ function edd_get_admin_url( $args = array() ) {
  */
 function edd_is_test_mode() {
 	$ret = edd_get_option( 'test_mode', false );
+
+	// Override any setting with the constant.
+	if ( defined( 'EDD_TEST_MODE') && true === EDD_TEST_MODE ) {
+		$ret = true;
+	}
+
+	// At the end of the day, the filter still has the final say.
 	return (bool) apply_filters( 'edd_is_test_mode', $ret );
 }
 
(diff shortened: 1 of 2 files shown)


includes/misc-functions.php
```diff
@@ -58,6 +58,13 @@ function edd_get_admin_url( $args = array() ) {
  */
 function edd_is_test_mode() {
 	$ret = edd_get_option( 'test_mode', false );
+
+	// Override any setting with the constant.
+	if ( edd_is_test_mode_forced() ) {
+		$ret = true;
+	}
+
+	// At the end of the day, the filter still has the final say.
```
robincornett, line 67: I kind of don't love allowing the filter to override the constant, but I see it's consistent with the `EDD_DEBUG_MODE` constant behavior. (Note: it would be neat if that constant got the same treatment as this.)

robincornett (approved):

includes/misc-functions.php
cklosowski: @robincornett Yeah I thought about that too, but my thought process was...

Since adding a filter requires some development work if it's in there already (prior to this release) it is in there for a reason, and since adding a filter is harder than a constant, it will prevent someone from overriding a deliberate decision by a developer to have test mode set through the filter.

PR #9393, opened 2022-08-25
Added support for multiple Y-Axis on charts

### LINKED ISSUE #9355: Reports Dashboard Graph uses a single Y-Axis for both charts
### Expected behavior
When viewing the Reports Dashboard, I expect the 'orders' line graph to be useful and provide a view that can be displayed.

### Actual behavior
Right now, the entire graph has a single Y-Axis for scaling by default (if you click on the legend at the top to disable the revenue, it gets better), however if we add a 2nd Y-Axis for the orders, the graph should result in something more useful to the store owner.

1) Have a store with sales where the revenue is higher than the orders (easily achieved with higher priced products over $100)
2) View Reports Dashboard
image

You can see in the graph that there are 'orders' on the orange line, however since it is using the y-Axis of the revenue, they are nearly zero'd out at scale.

Whereas if you remove the 'Earnings' graph, the scale corrects itself:
image

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

This might take a little refactoring of the way we instantiate Chart.js. Right now the line.js is hard coded to a single axes for Y, but we can overload it and maybe look for expanding upon the defaults.

zkawesome:
Fixes #9355

Proposed Changes:
- `$reports->register_endpoint` now supports `scales` property in the chart options.  With scales array, we can now have complete control over options of X and Y axes. Datasets can now connect to the specific Y axis via the `yAxisID` property that has to match the `id` of the Y-Axis scale.
- `get_chart_options()` in the `class-manfiest.php` is now setting the default options for X and Y axes to separate variables - the reason for that is, that if chart endpoint requested custom Y axes, **without any X axes**, we can push the defaults of the missing axis into the array. We cannot do that via `array_merge`, because both axes are in the `scales` array, and merging the array will override the defaults and we would be losing the axis options.
- All of our Y axis ticks are currently formatted via `NumberFormat`, which means that we are showing formatted numbers in the charts, where an integer would be more appropriate:
image

- ⬆️ That is why Y axis options now also support the `ticks` configuration straight from the PHP array. Currently, the ticks configuration was added in the JS. I have moved that into PHP and added a new property called `formattingType` which enables us to control the tick formatting on a per axis basis.
- I have added a new `utils.js` helper function called `attachAxisTickFormattingCallback` - this function allows us to attach a formatting callback to the ticks configuration based on the option that came from the `formattingType` property. Callback is then attached to the Y Axes in the `line.js`.
Current formatting options: `[raw, integer, format]`
- I have added `onClick` event listener to the dataset legend item so that we can hide the corresponding Y axis when certain dataset is hidden.

---------

Charts that were updated:
1.  📈 Sales and Earnings
- Added Y axis for the `Sales` dataset with integer formatting. Moved the `Earnings` legend to the left side, so that it matches the Y axis position.
2. 📈 Refunds
- Added Y axis for the `Number` dataset with integer formatting.
3. 📈 File downloads
- Changed formatting of Y axis to integer
3. 📈 New Customers
- Changed formatting of Y axis to integer

---------

Screenshots for historical reference:

Chart Before:
edd-localhost test_wp-admin_edit php_post_type=download page=edd-reports view=overview (1)

Chart After:
image


assets/js/admin/reports/charts/utils.js
```diff
@@ -2,6 +2,7 @@
  * Internal dependencies
  */
 import { Currency } from '@easy-digital-downloads/currency';
+import { NumberFormat } from '@easy-digital-downloads/currency';
```
robincornett, line 5: Can these two imports be combined into one?

assets/js/admin/reports/charts/utils.js
```diff
@@ -132,3 +133,31 @@ export const toolTipBaseConfig = {
...
+ * Attach formatting callback to axis ticks.
+ *
+ * @param {Object} ticks Axis ticks configuration.
+ */
+ export const attachAxisTickFormattingCallback = ( ticks ) => {
+	const number = new NumberFormat();
+
+	ticks.callback = function( value, index, values ) {
+		switch ( ticks.formattingType ) {
+			case 'raw':
```
robincornett, line 147: Should raw and the default be combined? I'm a little concerned that there isn't a default...although actually defining this is a little redundant so maybe we just need integer and format cases, and don't bother defining anything for raw/default?

includes/reports/data/charts/v2/class-manifest.php
```diff
@@ -449,34 +480,22 @@ public function get_chart_options() {
...
-							'position' => 'left',
-						),
-					),
+				'scales'    => array(
+					'xAxes' => array(),
+					'yAxes' => array(),
 				),
 			);
+
+			// Check if specific axes are missing from the endpoint options and load them from defaults.
```
robincornett, line 489: Wondering if the `$default_xAxes` and `$default_yAxes` definitions could be moved down here as this is where they're used. Or actually move them to a private method to be called if needed?

robincornett (changes requested):
I think this looks AMAZING.

Note you have a failing unit test:
```
EDD\Reports\Data\Charts\v2\Manifest_Tests::test_get_chart_options_default_non_pie_options_should_match_and_include_the_endpoint_label
--- Expected
+++ Actual
@@ @@
+                'ticks' => Array (...)

/tmp/wordpress-tests-lib/includes/abstract-testcase.php:852
/tmp/gJleWNpIHoR5qzcEDn2vxuneQ5BSVGup/tests/reports/data/charts/v2/tests-manifest.php:394
```

One question I had going in was whether this would break existing reports that use the new graphs (specifically thinking of Recurring and Licensing, at least) and I'm really impressed that they still work. We'll need to make corresponding issues in both of those repositories to account for the multiple y axes.

...it could be just me but I needed to rebuild my scripts for the graphs to display. Without rebuilding, I'm getting this console error:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'getBasePixel')
    at n.updateElement (chartjs.min.js?ver=1661433659:1:67136)
    at n.update (chartjs.min.js?ver=1661433659:1:66767)
    at n._update (chartjs.min.js?ver=1661433659:1:42102)
    at n.reset (chartjs.min.js?ver=1661433659:1:40588)
    at chartjs.min.js?ver=1661433659:1:105591
    at Object.each (chartjs.min.js?ver=1661433659:1:27185)
    at Ye.update (chartjs.min.js?ver=1661433659:1:105569)
    at Ye.construct (chartjs.min.js?ver=1661433659:1:102502)
    at new Ye (chartjs.min.js?ver=1661433659:1:101883)
    at edd-admin-reports.js?ver=1661433659:1:1200633
```
I would probably go ahead and `npm run dev` and push the compiled files, but you can wait for Chris to weigh in and see what he says. I can run it myself to test this.

robincornett (commented):
I think this looks amazing. One thing I'm seeing on the main sales/earnings graph that I really like is that the tallest point is not at the very top of the graph:

image

The File Downloads graph doesn't have the same breathing room though:

image

I know this doesn't have anything to do with the axes, but is this something we could tweak? The Refunds graph could use it as well.

Also, a line graph is rendered if you go to Reports > Downloads and then filter by an individual download. It should be updated to have the multiple axes, and if it could also get the breathing room at the top, that would be great.

image

cklosowski:
My only complaint at this point is in the colors (which isn't a result of this PR, but it highlights it).
Screen Shot 2022-08-30 at 10 22 37 PM

I feel like the dashed orange is VERY overbearing on the blue solid line.

Also the grid in the background needs to be adjusted a bit. It can get very confusing with the two different scales being used.

robincornett:
Do you think maybe the two lines could just both be solid, different colors (need to make sure they're colors that can still be differentiated by color blind users--I think red and green, for example, would not be a right choice?), or keep the orange as a dashed line and remove the fill?

cklosowski:
What do we think about this @robincornett @zkawesome ?
Screen Shot 2022-09-06 at 1 09 53 PM

cklosowski:
I'd still like to find a way to only show one set of 'y-axis' grids, so it is less confusing and has a more consistent spacing.

robincornett:
I like how the transparent background works there--I can certainly live with that.

spencerfinnell:
What about using a bar graph for sales, and hiding the y-axis ticks for that completely? Then you always show the values within the bars, or on hover like they are now.

https://www.chartjs.org/docs/latest/samples/other-charts/combo-bar-line.html
https://chartjs-plugin-datalabels.netlify.app/samples/charts/bar.html

cklosowski:
@spencerfinnell I think that might be a great thing to look at in the future. I don't feel like right now is the time to drastically change the reporting graphs as people are just getting used to them.

robincornett (commented):
This seems to be working well for me. I would still love to have some breathing room above the highest point on graphs that don't have it, eg Sales and Earnings for a specific product:

image

Also Refunds and the individual gateway, at least. The main graph has it, and File Downloads.

cklosowski:
@robincornett I'll see what I can do, these are auto calculated by the graph library and i think it has to do with the data sample sizes that makes the difference.

cklosowski:
@robincornett there isn't a great way to do this currently without refactoring a lot of items for backcompat. Maybe we can iterate on this in the future? right now it would require being able to calculate the max value in PHP, which not every graph can do currently in the way they are built out.

robincornett (approved):
Eventually I would love to see the padding above, but this is a huge improvement for me.

PR #9395, opened 2022-08-26
Disable PayPal button until required fields are populated

robincornett:
Fixes https://github.com/awesomemotive/easy-digital-downloads/issues/9094 (currently branched off 3.0.4)

Proposed Changes:

- Gets all fields marked as required on the checkout form.
- Adds JS functions for onInit and onClick for PayPal:
    - On initialization, disables the PayPal button and tallies up how many required inputs have value. The PayPal button is disabled until the required inputs all have a value (did it this way as a regular check out may have two required inputs--email and first name--but if taxes are enabled, that jumps to five). If the error message (from onClick) is displaying, it will clear as soon as all inputs have a value.
    - On click, if not all of the inputs have a value, displays an error message indicating that some required fields are missing data. This doesn't list the specific fields like the process does now.

The net effect of this PR is that the PayPal flow is not initialized until all required data is present.

Note: this currently won't work with reCAPTCHA v3 if that is tied to the HTML5 validation.

Marking as a draft because of that and we're not completely ready to handle this.


cklosowski:
@zkawesome Looks like @robincornett is trying to using some of the stuff we were talking about on our call about fixing recaptcha without having to release something for CFM

cklosowski (approved):

PR #9396, opened 2022-08-27
Update edd_*_customer functions to also trigger actions so we can take additional action

cklosowski:
This is kind of a big update, so I'll try and explain:

The initial root cause of all this was the fact that the `edd_get_user_by` function was not the value in the `email` column of the customers table. It ignored the customer email addresses table entirely.

Proposed Changes:
1.  Updates `edd_get_user_by` to be aware of the email addresses table when `email` is provided as the field
2. Fixes an issue in searching customers and customer emails in the admin that strips the `+` character before the query
3. Removes some additional processing from the `EDD_Customer::update` and `EDD_Customer::create` and moves them to `edd_process_update_customer` and `edd_process_add_customer`
4. Adds a 'transition' hook for when the primary email address changes on a customer, so that all other email addresses are moved to 'secondary'.
5. When a user ID changes on a customer, we associated the user email as the primary email and remove the old user's email address from the customer.
6. 


cklosowski:
So with this, my unit tests are passing, and I'm able to be logged in with a user that has an existing customer, use a different email at checkout, and that email address is associated with the right customer using both the store gateway and Stripe.

cklosowski:
Try with the following with both recurring and non-recurring purchases. Also Test with both Stripe and Store Gateway.`
1. Logged out, New Customer Guest purchase
2. Logged out, new customer with Auto Register
3. Logged in User, new customer
4. Logged in User, Existing customer

includes/customer-functions.php
```diff
@@ -199,7 +256,18 @@ function edd_get_customer_by( $field = '', $value = '' ) {
...
+	if ( 'email' === $field ) {
+		$customer_emails = new EDD\Database\Queries\Customer_Email_Address();
+		$customer_email  = $customer_emails->get_item_by( 'email', $value );
+
+		$customer = false;
+		if ( ! empty( $customer_email->customer_id ) ) {
+			$customer  = $customers->get_item_by( 'id', $customer_email->customer_id );
+		}
+	} else {
+		$customers = new EDD\Database\Queries\Customer();
```
robincornett, line 268:
```suggestion
```
This is already defined in line 258.

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
...
+			'type'        => 'primary',
+		)
+	);
+
+	if ( empty( $previous_primary_email_addresses ) ) {
+		return;
+	}
+
+	foreach ( $previous_primary_email_addresses as $previous ) {
+		edd_update_customer_address( $previous, array( 'type' => 'secondary' ) );
```
robincornett, line 120:
```suggestion
		edd_update_customer_email_address( $previous, array( 'type' => 'secondary' ) );
```

robincornett (changes requested):

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
  * @since       3.0
  */
 
+/**
+ * Is intended to be used when adding a customer direclty with edd_add_customer function.
```
robincornett, line 15:
```suggestion
 * Is intended to be used when adding a customer directly with edd_add_customer function.
```

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
...
+					'email'       => $previous_user->user_email,
+				)
+			);
+		}
+
+		// Update some payment meta if we need to.
+		$order_ids = edd_get_orders( array( 'customer_id' => $customer->id, 'number' => 9999999 ) );
+
+		foreach ( $order_ids as $order_id ) {
+			edd_update_order( $order_id, array( 'user_id' => $customer->user_id ) );
```
robincornett, line 86: Are you intending for the orders to still have the previous_user email address assigned? If not, this should also update the order email address.

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
...
+			'customer_id' => $customer_id,
+			'email'       => $data['email'],
+			'type'        => 'primary',
+		)
+	);
+}
+add_action( 'edd_customer_added', 'edd_process_customer_added', 10, 2 );
+
+/**
+ * Is intended to be used when updating a customer direclty with edd_add_customer function.
```
robincornett, line 35:
```suggestion
 * Is intended to be used when updating a customer directly with edd_add_customer function.
```

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
...
+/**
+ * Is intended to be used when adding a customer direclty with edd_add_customer function.
+ *
+ * Because the edd_add_customer function only directly interacts with the customers DB table, we may
+ * need to do some additional actions like adding items to the customer email addresses table.
+ *
+ * @since 3.0.4
+ */
+function edd_process_customer_added( $customer_id, $data ) {
+	// Make sure we add a new primary email address to the email addrsses table.
```
robincornett, line 23:
```suggestion
	// Make sure we add a new primary email address to the email addresses table.
```

robincornett (commented):
This appears to be working well for me. Some typos and one questions about updating orders after a user change.

includes/customers/customer-actions.php
cklosowski: @robincornett I'm not sure exactly honestly. We previously did not alter the email address on the order. I think since we use that email address for resending the purchase receipt as an option in the radio items, we should probably leave it on the order.

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
...
+	// Process a User ID change.
+	if ( intval( $customer->user_id ) !== intval( $previous_customer_data->user_id ) ) {
+		// Attach the User Email to the customer as well.
+		$user = new WP_User( $customer->user_id );
+		if ( $user instanceof WP_User ) {
+			edd_add_customer_email_address(
+				array(
+					'customer_id' => $customer->id,
+					'email'       => $user->user_email,
+					'primary'     => true, // Set this as the primary, which should transition any other emails.
```
robincornett, line 66:
```suggestion
					'type'     => 'primary', // Set this as the primary, which should transition any other emails.
```

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,117 @@
...
+				)
+			);
+		}
+
+		// Remove the old user email from this account.
+		$previous_user = new WP_User( $previous_customer_data->user_id );
+		if ( $previous_user instanceof WP_User ) {
+			edd_delete_customer_email_address(
```
robincornett, line 74: This function only accepts an ID as the parameter.

cklosowski: Should be fixed in https://github.com/awesomemotive/easy-digital-downloads/pull/9396/commits/9073682282320eda36af4828d4304f5fbe313fdd

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,124 @@
...
+
+/**
+ * Is intended to be used when updating a customer directly with edd_add_customer function.
+ *
+ * Because the edd_update_customer function only directly interacts with the customers DB table, we may
+ * need to do some additional actions like managing email addresses.
+ *
+ * @since 3.0.4
+ */
+function edd_process_customer_updated( $customer_id, $data, $previous_customer_data ) {
```
robincornett, line 42: I would really love to see the parameters documented here. I also think I would like for `$previous_customer_data` to be renamed to something that feels more like an object since it is actually the previous customer object, rather than an array of data, and we tend to name things `_data` when they're arrays.

robincornett (approved):
My final comment is about documentation/naming--functionally I can approve this.

includes/customers/customer-actions.php
cklosowski: Fixed in https://github.com/awesomemotive/easy-digital-downloads/pull/9396/commits/1d7a8828ec86ed0eb6151b6ea1da33e1ae59f227

includes/customers/customer-actions.php
```diff
@@ -11,6 +11,136 @@
...
+				'type'        => 'primary',
+			)
+		);
+	}
+	// Process a User ID change.
+	if ( intval( $customer->user_id ) !== intval( $prev_customer_obj->user_id ) ) {
+		// Attach the User Email to the customer as well.
+		$user = new WP_User( $customer->user_id );
+		if ( $user instanceof WP_User ) {
+			edd_add_customer_email_address(
```
robincornett, line 69: This is working, but when I switch users in the customer profile, the main customer email address in the customers table is not updated. So I'm ending up with a customer object with an email  of `originalemail@edd.local` and then two separate email addresses stored in the email addresses table. the `originalemail@edd.local` is deleted from the addresses table, but it's still present on the customer object.

cklosowski: @robincornett ready for another set of tests here.

robincornett (approved):

PR #9398, opened 2022-08-27
Add dropins to the system ifno #6405

### LINKED ISSUE #6405: Add checks for drop-ins into the system info
There are times when no object caching plugins appear however it is still used via things like `object-cache.php` drop-ins. We should include any drop-ins that exist in the `wp-content` directory for the system info.

cklosowski:
Fixes #6405

Proposed Changes:
1. Adds a section to the system info for dropins.

### DIFF
--- a/includes/admin/tools.php
+++ b/includes/admin/tools.php
@@ -1506,6 +1506,18 @@ function edd_tools_sysinfo_get() {
 		$return = apply_filters( 'edd_sysinfo_after_edd_templates', $return );
 	}
 
+	// Drop Ins
+	$dropins = get_dropins();
+	if ( count( $dropins ) > 0 ) {
+		$return .= "\n" . '-- Drop Ins' . "\n\n";
+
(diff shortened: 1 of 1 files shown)


cklosowski:
@zkawesome Because of the way that MU Plugins are used, they typically do not have all the same data available. Most MU Plugins don't even have a plugin header, and are just PHP files.

cklosowski:
I just did some looking at the standard DropIns and none of them even really have a 'version'. None of them even have headers that can be parsed. I think we should be fine just having the dropins listed by file, as most DropIns are actually for specific uses, unlike MU Plugins.

cklosowski:
I just pushed up some changes that make this a 'beast' of a PR to review from a line change perspective, but MOST of these are just indentation issues that I've corrected. If you want to just see the code changes now you can use:
https://github.com/awesomemotive/easy-digital-downloads/pull/9398/files?w=1

robincornett (approved):

PR #9405, opened 2022-08-31
Remove dev assets from the build #9277

### LINKED ISSUE #9277: Remove development asset files from build
We should remove the .scss and .js development files from the final build, and just deliver the compiled and minified files.

### Justification or use case
They just take up space/weight and since the config files for compiling them aren't included in the build, I don't think they provide any benefit.

cklosowski:
Fixes #9277 

Proposed Changes:
1. Adjusts the Grunt build process to only copy over the built assets, not the development assets.

As a note, this brought the final build .zip from 3.6MB to 3.2MB, which is a pretty good amount of savings.

To test:
1. Checkout `issue/9277`
2. Run `npm install`
3. Run `npm run biuld`
4. Copy the `build/easy-digital-downloads` folder to a clean WordPress install and activate EDD.
5. Ensure that Styles and JavaScript assets run, you can checkout, tax settings work, edit/add new orders work.


robincornett (changes requested):
File issues:
- assets/sample-products-import.xml (missing)
- assets/css/edd-admin.min.css.map (missing, maybe not needed?)
- assets/js/edd-admin-tax-rates.js.map (missing, maybe not needed?)
- assets/js/edd-admin-orders.js.map (missing, maybe not needed?)
- assets/js/frontend/edd-ajax.js (unminified version still present)
- assets/js/frontend/checkout/ (unminified files still present, I think this is all in the global checkout)
- assets/js/frontend/gateways/ (unminified files still present, just need paypal-checkout.js I think)
- assets/js/packages/ (unminified files still present, those went into the admin orders, I think)
- assets/js/utils/ (I believe these are only needed for development)

Unrelated but it seems like alpine.min.js should have gotten moved to the vendor directory when notifications were merged into 3.0.

cklosowski:
- [x] assets/sample-products-import.xml (missing)
- [x] assets/css/edd-admin.min.css.map - Not Needed
- [x] assets/js/edd-admin-tax-rates.js.map - Not Needed
- [x] assets/js/edd-admin-orders.js.map - Not Needed
- [x] assets/js/frontend/edd-ajax.js (unminified version still present)
- [x] assets/js/frontend/checkout/ (unminified files still present, I think this is all in the global checkout)
- [x] assets/js/frontend/gateways/ (unminified files still present, just need paypal-checkout.js I think)
- [x] assets/js/packages/ (unminified files still present, those went into the admin orders, I think)
- [x] assets/js/utils/ (I believe these are only needed for development)

cklosowski:
Ready for another check @robincornett

robincornett (approved):
This looks good. Installed from the build .zip and checked:
- [x] Front end pages load okay with expected scripts and styles
- [x] Notifications work as expected
- [x] EDD header shows in admin as expected
- [x] I can add tax rates
- [x] I can create orders
- [x] Reports display as expected
- [x] Various settings pages display as expected
- [x] Recount tool renders and runs as expected

I would still love to see the alpine.min.js file moved to the vendor directory, but that might also be true of the jQuery validate file and maybe even the polyfill file? Certainly not a deal breaker.

cklosowski:
@robincornett Yeah I'd like to look at moving those in the next major version in case someone is still looking for them there, for the time being.

PR #9406, opened 2022-09-01
Rock: Email Summary

zkawesome:
### Rock project: Email Summary
------------
- 🗓 Emaill Summaries are scheduled as one time events using `wp_schedule_single_event`. Scheduling of Email Summaries is done in two ways:
1. `edd_daily_scheduled_events` which is run daily, will check if Email Summaries have to be scheduled. 
2. Every time that `edd_options` or WordPress setting `start_of_week` are changed, email summary cron event is cleared and re-scheduled. These two options affect the dates therefore they have to update the schedule of emails.
----
- ⚙️ Summaries are isolated into separate hook called `edd_email_summary_cron`.


includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -0,0 +1,791 @@
...
+		$next_blurb   = $email_blurbs->get_next();
+
+		// Prepare email.
+		$email_subject    = $this->get_email_subject();
+		$email_recipients = $this->get_email_recipients();
+		$email_body       = $this->build_email_template( $next_blurb );
+		$email_headers    = array( 'Content-Type: text/html; charset=UTF-8' );
+
+		// If there is no email body, do not send an email.
+		if ( ! $email_body ) {
```
robincornett, line 777: Maybe getting the email body and doing this early return could be moved to the top of the method so that nothing else would run if it's empty.

includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -0,0 +1,791 @@
...
+	 * Prepare and send email.
+	 *
+	 * @since 3.1
+	 *
+	 * @return bool True if email was sent, false if there was an error.
+	 */
+	public function send_email() {
+		// Get next blurb.
+		$email_blurbs = new EDD_Email_Summary_Blurb();
+		$next_blurb   = $email_blurbs->get_next();
```
robincornett, line 768: Probably these variables can be moved down into the condition `if ( $email_sent )` since it looks like that's where they're actually used.

robincornett (commented):
My comments are mostly fiddly things--initial impression is that this is put together well and works well. Some thoughts in no particular order:
- On the icon files, such as icon-average, icon-gross, and a couple others, the circle has a white background. is that intentional? it doesn't matter if my browser is in light mode but I know I have issues with emails that have icons like this on my phone when it switches to dark mode. I'm wondering if we can pull the white background out and make them completely transparent.
- I think I would love to pull the entire email template, starting after the `ob_start()`, out into a separate view file so it's easier to think through and read, please. Also then that's several hundred lines of code to load only when actually needed.
- I already mentioned in dev tools but I would love to have the email schedule and testing on the settings screen in core.

Overall this looks great so far!

includes/emails/email-summary/edd-email-summary-template.php
```diff
@@ -0,0 +1,410 @@
+<!DOCTYPE html>
+		<html>
```
robincornett, line 2: It will affect the entire file, but not a big deal--but can we un-indent this one level?

robincornett (commented):
Thanks for the changes so far, especially moving the email template to its own file.

I'm now actually wondering if that template could be broken into separate files for each section--one for each part of the email (stats totals, top five products, and the blurb).

Also my email preview and the test email do not contain the blurb at all, which may be related to the PHP warning I got (the sample JSON link is throwing a 502 so that's maybe the issue I'm hitting).

includes/admin/emails/email-summary/class-edd-email-summary-admin.php
```diff
@@ -0,0 +1,99 @@
...
+
+		edd_redirect( $url );
+	}
+
+	/**
+	 * Build and output Email Summary template to the screen.
+	 *
+	 * @since 3.1
+	 */
+	public function preview_email_summaryy() {
```
robincornett, line 67:
```suggestion
	public function preview_email_summary() {
```
(Would need to update the call for it as well)

includes/admin/settings/register-settings.php
```diff
@@ -300,10 +300,36 @@ function edd_get_registered_settings() {
...
+		$email_summary_preview_url = wp_nonce_url(
+			edd_get_admin_url(
+				array(
+					'edd_action' => 'preview_email_summary',
+				)
+			),
+			'edd_preview_email_summary'
+		);
+		$email_summary_schedule      = wp_next_scheduled( EDD_Email_Summary_Cron::CRON_EVENT_NAME );
+		$email_summary_schedule_text = '<span><span class="dashicons dashicons-warning"></span> ' . esc_html( __( 'The summary email is not yet scheduled. Save the settings to manually schedule it.', 'edd-dev-tools' ) ) . '</span>';
```
robincornett, line 324:
```suggestion
		$email_summary_schedule_text = '<span><span class="dashicons dashicons-warning"></span> ' . esc_html( __( 'The summary email is not yet scheduled. Save the settings to manually schedule it.', 'easy-digital-downloads' ) ) . '</span>';
```

robincornett (commented):
Just a couple minor comments while we wait to hear back from Chris about the template/styles, and noting that unit tests are still failing because of this:

```
Fatal error: Call to undefined function jddayofweek() in /tmp/YNdYFC33QNYW1SePKkb0lKDeWfi0cHMC/includes/emails/email-summary/class-edd-email-summary-cron.php on line 78
```

includes/admin/settings/register-settings.php
```diff
@@ -809,6 +838,50 @@ function edd_get_registered_settings() {
 						'type' => 'checkbox',
 					),
 				),
+				'email_summaries' => array(
+					'email_summary_buttons' => array(
+						'id'   => 'email_summary_buttons',
+						'name' => '',
+						'desc' => '<a target="_blank" href="' . esc_url( $email_summary_preview_url ) . '" class="button">' . esc_html( __( 'Preview Email Summary', 'easy-digital-downloads' ) ) . '</a> <a href="' . esc_url( $email_summary_trigger_url ) . '" class="button">'. esc_html( __( 'Send Test Email', 'easy-digital-downloads' ) ) . '</a>',
```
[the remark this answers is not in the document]
cklosowski: When sending a `Test` email can we adjust the subject of the email to contain `[TEST]` so that the user isn't confused by the dates for the summary?

For instance I was a bit confused as to why my date ranges were not starting on my Week Start date until I realized that my report from Today - 7 days until Today. Adding the `[TEST]` to the subject will give the user a clear indication that this is something to test the visual aspects of devilry and what not.

cklosowski (changes requested):
Overall things seem to be working great.

There are some usability things to work out, as in the comments I've left, but overall the functionality itself of sending the email is working as I expect.

One thing we need to do here is compress the images we're including in the `icons` folder. We should be able to reduce the .png images by about 50% if we run them through a compression tool. If you don't have one, let me know and we can work out a good one to use for you.

cklosowski:
Side note @zkawesome 👏 for listening to the `start_of_week` change in the Main WP settings! That's not something I had in the requirements, but was a great touch and before I started looking at the code I went and change that setting, expecting it not to update the cron, but it totally did :)

cklosowski:
One other thing @zkawesome can we set a body width?

I know this is based on email client.
Screen Shot 2022-09-09 at 11 18 47 AM

Also, I noticed their TEST is purely made up data:
Screen Shot 2022-09-09 at 11 19 52 AM

They also don't show their 'did you know' section on tests.

includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -0,0 +1,348 @@
...
+	 *
+	 * @since 3.1
+	 *
+	 * @param  string $url Destination URL.
+	 * @param  array $blurb Blurb data.
+	 * @return string URL with tracking parameters.
+	 */
+	public function add_blurb_tracking_parameters( $url, $blurb ) {
+		// @todo - Define tracking parameters!
+		return add_query_arg(
```
cklosowski, line 304: As we talked on the call yesterday, we can remove these, as we're going to supply these in the JSON data.

includes/emails/email-summary/edd-email-summary-template.php
```diff
@@ -0,0 +1,410 @@
...
+					line-height: 20px;
+					background: #2794DA;
+					display: inline-block;
+					color: #FFFFFF;
+					text-decoration: none;
+				}
+			}
+		</style>
+
+		<body style="margin: 0px;">
```
cklosowski, line 158: @zkawesome We need to set a max width on this, so it doesn't span the whole body area in the email client. Also, once we do that we can set the margins to be `0px auto` so that it is centered in the email client.

In the Figma design for this the max width on desktop is `450px` and on mobile it is`320px`

We can use a media query on this for screen devices with a max width of `480px`

cklosowski (changes requested):

cklosowski:
@zkawesome looking much better! The last few things to clean up here I think:
1. The content width needs to be restricted, as per this comment: https://github.com/awesomemotive/easy-digital-downloads/pull/9406#discussion_r967367187
2. If the the Email Recipient changes, let's disable the 'test' button until the settings are refreshed, that way they know they have to save the settings in order to get the test email to the right email address(es)

assets/js/admin/settings/index.js
```diff
@@ -161,18 +161,64 @@ const EDD_Settings = {
...
-		// Toggle Email Recipient option
-		emailSummaryRecipient.on( 'change', function() {
-			if ( $( this ).val() === 'admin' ) {
-				emailSummaryCustomRecipients.prop( 'readonly', true );
-			} else {
-				emailSummaryCustomRecipients.prop( 'readonly', false );
-			}
-		} );
+		/**
+		 * Email Summaries
```
robincornett, line 165: I'm wondering if this code could be extracted into its own script file and enqueued only on the email summaries screen--it's all reasonable and needed code but it feels like it's ended up being a lot.

includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -69,7 +79,11 @@ public function get_email_subject() {
 		/* Translators: Site domain name */
 		$email_subject = sprintf( __( 'Easy Digital Downloads Summary - %s', 'easy-digital-downloads' ), $this->get_site_url() );
 
-		return apply_filters( 'edd_email_summary_subject', $email_subject );
+		if ( $this->test_mode ) {
```
robincornett, line 82: I'm not sure what happened but my test email came through with no subject at all.

assets/js/admin/settings/index.js
cklosowski: Let's leave it in here, as it makes more sense. The only things that have their own are taxes currently and that is because it is built on backbone.

includes/emails/email-summary/class-edd-email-summary.php
cklosowski: Interesting I haven't seen that yet.

cklosowski (changes requested):
One last change, can we 'reset' the test email button and notice after a few seconds?

So that if I click to 'test' the email and it comes back successful, it removes the notice and the button goes back to normal.

includes/emails/email-summary/class-edd-email-summary.php
robincornett: It could be just something weird with how Local/Mailhog work together. If it's only happening to me then we can ignore it. Tinkering with various modifications of the subject line got some options to work and some not.

includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -0,0 +1,791 @@
...
+
+					<!-- MAIN EMAIL CONTENT HOLDER -->
+
+					<div class="content-box" style="background: #FFF;">
+
+						<div class="content-holder" style="padding: 22px 31px 0px 31px;">
+
+							<h1 style="margin: 0px; color: #1F2937;font-weight: 700;font-size: 24px;line-height: 24px;"><?php echo esc_html( __( 'Your eCommerce Summary', 'easy-digital-downloads' ) ); ?></h1>
+
+							<div class="period-date pull-down-8" style="margin-top: 8px; font-weight: 400; font-size: 14px; line-height: 18px; color: #4B5563;">
```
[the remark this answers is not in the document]
cklosowski: Since it is in the email wrapper, they are necessary in order to maintain a consistent set of styles. Most email clients 'sandbox' the markup so we are just fine doing this.

cklosowski:
@zkawesome two last changes.
1. I think for the button animation lets do the following:
* When clicked does the spinning animation (already working)
* If it is successful, show a green checkmark (as it previously did, and that can go away in 3 seconds), no need to show the "The test Email Summary was sent successfully!" message.
* If it fails, set back to the default state immediately, and then show an error notice.

includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -0,0 +1,791 @@
...
+			<head>
+				<title><?php echo esc_html( $this->get_email_subject() ); ?></title>
+				<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+				<meta name="viewport" content="width=device-width, initial-scale=1.0">
+				<!--[if !mso]><!-->
+				<meta http-equiv="X-UA-Compatible" content="IE=edge">
+				<!--<![endif]-->
+
+				<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">
+				<link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@300;400;600;700&display=swap" rel="stylesheet">
```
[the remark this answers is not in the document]
cklosowski: Let's see how it plays out with the system stack:
`font-family: -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif;`

I don't love it as it runs into some style issues from time to time, but lets give it a shot. We use it on our website.

includes/emails/email-summary/class-edd-email-summary.php
```diff
@@ -0,0 +1,791 @@
...
+					h1 {
+						font-size: 18px !important;
+						line-height: 24px !important;
+					}
+					h2 {
+						font-size: 12px !important;
+						line-height: 16px !important;
+					}
+					p {
+						font-size: 10px !important;
```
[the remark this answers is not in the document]
cklosowski: @robincornett we cannot use anything other than PX b/c other units do not have full support in email clients. Important is necessary in order to make sure all clients (many of them browsers) don't stomp on our reports.

Since we aren't trying to 'play nice' with themes or anything this is completely in bounds in this type of scenario.

As for the font size @zkawesome let's try and bump it up a bit,  and try having the main content areas (tiles and table) match the sizes used in the desktop view please.

assets/css/admin/style.scss
```diff
@@ -2498,6 +2498,15 @@ span.edd-to-and-from--separator {
 	clear: both;
 }
 
+#edd-send-test-summary-save-changes-notice .notice p {
```
robincornett, line 2501: The CSS is okay but I'm not sure it's actually needed. If you add the `.notice` class to the `#edd-send-test-summary....` element you should be okay with just a paragraph inside, and save a little bit of markup. Probably do `notice inline notice-info` and it will have the blue border instead of the gray one.

includes/scripts.php
```diff
@@ -520,6 +520,7 @@ function edd_localize_admin_scripts() {
 		'hide_advanced_settings'  => __( 'Hide advanced settings', 'easy-digital-downloads' ),
 		'no_downloads_error'      => __( 'There are no downloads attached to this payment', 'easy-digital-downloads' ),
 		'wait'                    => __( 'Please wait &hellip;', 'easy-digital-downloads' ),
+		'test_email_save_changes' => __( 'You have to Save Changes to send the test email.', 'easy-digital-downloads' ),
```
robincornett, line 523:
```suggestion
		'test_email_save_changes' => __( 'You must save your changes to send the test email.', 'easy-digital-downloads' ),
```

robincornett (commented):
Everything looks great and seems to be working well! Just a couple requested adjustments to the email notice markup/wording and I think I'm good to go!

robincornett (approved):

Issue #3827, opened 2015-09-24
Build front-end "Forgot Password" form


shazahm1:
I seem to get this regularly... users wishing to purchase a renewal or making another purchase who forget their password. On the login screen of the checkout page, there should be an option to get and/or reset the password.

I already have in very plain sight (on the main site nav) a "My Account:" page which has the forgot password link but I think it would be beneficial to have it on the checkout/login page too.

Thanks for your consideration of this enhancement request!

pippinsplugins:
I agree. We should include one as part of the `[edd_login]` short code.

This was done recently in RCP: https://github.com/pippinsplugins/restrict-content-pro/pull/209

adampickering:
:+1:

adampickering:
Would be great if we could get this into 2.7 :) 👍

pippinsplugins:
@cklosowski this actually needs to go further than just adding a link to the checkout screen. I need a full, frontend reset password form.

cklosowski:
Ok, that wasn't part of his original statement:
`I already have in very plain sight (on the main site nav) a "My Account:" page which has the forgot password link but I think it would be beneficial to have it on the checkout/login page too.`

That mentions the link. We can build out a full reset form if we desire that.

pippinsplugins:
See https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3827#issuecomment-142969419

cklosowski:
Just to be clear, we'll have to do this in two places.

The original request was for a password reset on the checkout page, which doesn't use `[edd_login]` but a hand built form for the login process.

So we need to build a forgot password form for:
-[] [edd_login]
-[] Checkout Login Form.

pippinsplugins:
:+1:

cklosowski:
After some discussions with the team, we're going to punt this to 2.8 in need for #5398...we just don't have the proper means to guarentee an endpoint to show this interface, like RCP does...yet :)

adampickering:
:(

cklosowski:
Sorry @adampickering as I started working on it, we just don't have a spot we can guarentee will work on all sites, without significant changes to the current shortcodes, which will likely cause an uproar, since we'll be altering the front end for logged out users.

Basilakis:
That still has not happen - right?

pippinsplugins:
Correct and it is not a priority.

Basilakis:
So after a year on it, do we have any update? :)

[...]

PR #9415, opened 2022-09-06
Register blocks

robincornett:
Proposed Changes:
1. Adds a build script to pull EDD Blocks into core. Use `npm run blocks` on the command line to clone.
2. Some minor tweaks to edd.css (note: still need to add the template CSS files to the minification/build).
3. Moves existing user related functions for logging in, registering, to a users directory.
4. Adds new user related functions for managing lost password requests/password resets (#3827).
5. Adds a new "success errors" option for showing success messages when calling `edd_print_errors()` (needed specifically for the forgot password flow but useful anywhere).
6. Adds a `show_required()` method to the HTML elements class to display a "required" indicator with accessible markup.
7. Adds improved no-js/js handling so that "no-js" elements do not display on screen at all (currently basically a FOUC effect due to the EDD script properly loading in the footer).
8. Updates `edd_get_purchase_link()` to output a semantically correct button instead of a link for adding items to the cart via ajax.


cklosowski:
@robincornett The `templates/edd.css` file is processed by `grunt build` which is part of the `npm run build` script. Do you need it to run some other time?

robincornett:
After doing some digging, in [7486fa4](https://github.com/awesomemotive/easy-digital-downloads/pull/9415/commits/7486fa4272a33d236d720fec72b14a491a493615) I've decided to move and handle the front end CSS files the same way the admin CSS is managed--in the assets/css directory. Because those files aren't loaded the same way as PHP template files, they don't actually need to be in the templates directory. I've updated the final fallback path for the style in `edd_register_styles` accordingly.

Noting that as part of that, I updated the two background image files to use SVG instead, so they are slightly different visually but mostly as similar as I could make them.

Currently working on creating some unit tests.

PR #9416, opened 2022-09-06
Add order item hooks on add, update, delete; update calculations

robincornett:
Proposed Changes:
1. This follows the pattern introduced in #9396 and adds hooks to some `edd_*_order_item` methods.
2. The closures on `edd_transition_order_item_status` and `edd_transition_order_adjustment_type` have been removed.
3. They're replaced by the new `edd_recalculate_order_item_download` function, which listens to the new order item hooks (add, update, remove). This function does some evaluation before allowing the download values to be recalculated:
    - If the order item is being _added_, but with no status or the default status (pending), do not recalculate.
    - If the order item is being _updated_, check if the updated values are ones which affect stats (status, quantity, total, subtotal); if not, or if the values are not changed, do not recalculate.
4. Removes the download recalculation from the `EDD_Payment()->save()` method.
5. Note that the recalculation is manually triggered in the `test_download_earnings_gross_minus_net_equals_negative_fee` test as that involves adding a fee to an order item. This is because the order item is added directly to the database as `complete`, so the recalculation happens before the adjustment is added. Adding an order to the database as complete is not [currently] part of our normal workflow--orders and order items are created as pending and then transitioned to complete, generally. This does show a gap in this approach, though, although the adjustment would be factored in on the next order or manual recalculation. We could add a listener for the adjustment, but that would mean potentially running the recalculation twice on one order item, which is what we're trying to move away from.
6. The `edd_recalculate_order_item_download` function is unhooked during the orders migration. We do manually recalculate the download sales and earnings for all downloads at the end of the both the CLI and UI order migrations already.


includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * The hook is registered here because it's specifically for the cron job,
+ * but it calls the general download recalculation function.
+ */
+add_action( 'edd_recalculate_download_sales_earnings_deferred', 'edd_recalculate_download_sales_earnings' );
+
+/**
+ * Handles scheduling the recalculation for a download.
+ *
```
cklosowski, line 21: Can we add a longer description on this as well that highlights a few things for developers:
1. If `WP_Cron` is not disabled, this will run automatically within 5 minutes of any recalculations being scheduled.
2. This is a possibly very expensive query to run, so only do it if you absolutely need to, otherwise rely on the deferred cron so that it doesn't affect load times after the change.

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+ * @copyright   Copyright (c) 2022, Easy Digital Downloads, LLC
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ * @since       3.1
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * The hook is registered here because it's specifically for the cron job,
+ * but it calls the general download recalculation function.
```
cklosowski, line 15: Let's add a `since` tag on this one, so we can know when it was added in the future.

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+		time() + ( 5 * MINUTE_IN_SECONDS ),
+		'edd_recalculate_download_sales_earnings_deferred',
+		array( $download_id )
+	);
+}
+
+add_action( 'edd_order_item_added', 'edd_recalculate_order_item_download', 10, 2 );
+add_action( 'edd_order_item_updated', 'edd_recalculate_order_item_download', 10, 3 );
+add_action( 'edd_order_item_deleted', 'edd_recalculate_order_item_download' );
+/**
```
cklosowski, line 48:
```suggestion

/**
```

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+ * @param int $download_id
+ * @return void
+ */
+function edd_maybe_schedule_download_recalculation( $download_id ) {
+	$is_scheduled = wp_next_scheduled( 'edd_recalculate_download_sales_earnings_deferred', array( $download_id ) );
+	$bypass_cron  = apply_filters( 'edd_recalculate_bypass_cron', false );
+	if ( $is_scheduled && ! $bypass_cron ) {
+		edd_debug_log( 'Recalculation is already scheduled for ' . $download_id . ' at ' . edd_date_i18n( $is_scheduled, 'datetime' ) );
+		return;
+	}
```
cklosowski, line 32:
```suggestion
	}

```
This just feels 'cramped', can we give these 3 if statements a little breathing room?

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+	if ( $bypass_cron || ( defined( 'EDD_DOING_TESTS' ) && EDD_DOING_TESTS ) ) {
+		edd_recalculate_download_sales_earnings( $download_id );
+		return;
+	}
```
cklosowski, line 36:
```suggestion
	}

```

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+ * Handles scheduling the recalculation for a download.
+ *
+ * @since 3.1
+ * @param int $download_id
+ * @return void
+ */
+function edd_maybe_schedule_download_recalculation( $download_id ) {
+	$is_scheduled = wp_next_scheduled( 'edd_recalculate_download_sales_earnings_deferred', array( $download_id ) );
+	$bypass_cron  = apply_filters( 'edd_recalculate_bypass_cron', false );
+	if ( $is_scheduled && ! $bypass_cron ) {
```
cklosowski, line 29:
```suggestion

	if ( $is_scheduled && ! $bypass_cron ) {
```

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+		edd_debug_log( 'Recalculation is already scheduled for ' . $download_id . ' at ' . edd_date_i18n( $is_scheduled, 'datetime' ) );
+		return;
+	}
+	if ( $bypass_cron || ( defined( 'EDD_DOING_TESTS' ) && EDD_DOING_TESTS ) ) {
```
cklosowski, line 33: We should probably also check for the core constant `DISABLE_WP_CRON`. That is how most hosting companies who spawn cron on their own make it so that core and their own runners don't both fire cron.

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Attempts to schedule download sales and earnings when order items are changed.
+ *
+ * @since 3.1
+ * @param int $order_item_id                              The order item ID being added, updated, or deleted.
+ * @param array $data                                     The array of order item data being added or updated.
+ * @param bool\EDD\Orders\Order_Item $previous_order_item When updating an order, this is the original order item object.
+ * @return void
+ */
+function edd_recalculate_order_item_download( $order_item_id, $data = array(), $previous_order_item = false ) {
```
cklosowski, line 57: In this function (and in the changes overall) can we give ourselves a little breathing room after the logical blocks? Having a comment for the next if statement come right after the closing bracket just makes it harder to read and follow since the brain logically groups them together instead of considering them separate 'tasks'.

Example:
```
	}
	// If the order item data being updated doesn't affect sales/earnings, recalculations do not need to be run.
	if ( $previous_order_item instanceof EDD\Orders\Order_Item ) {
```

includes/downloads/recalculations.php
```diff
@@ -0,0 +1,120 @@
...
+	// Recalculations do not need to run when the order item is first being added to the database if it's pending.
+	if ( 'edd_order_item_added' === current_action() && ( empty( $data['status'] ) || 'pending' === $data['status'] ) ) {
+		return;
+	}
+	// If the order item data being updated doesn't affect sales/earnings, recalculations do not need to be run.
+	if ( $previous_order_item instanceof EDD\Orders\Order_Item ) {
+		$columns_affecting_stats = array( 'status', 'quantity', 'total', 'subtotal' );
```
cklosowski, line 64: Since we can affect more than just 'net' values here we should probably include all columns that effectively affect the revenue in reporting, this would include:
`discount, tax, and rate`

cklosowski (changes requested):

includes/downloads/recalculations.php
```diff
@@ -55,31 +64,41 @@ function edd_maybe_schedule_download_recalculation( $download_id ) {
...
 			return;
 		}
 	}
+
 	$order_item = edd_get_order_item( $order_item_id );
 	if ( empty( $order_item->product_id ) ) {
 		return;
 	}
+
 	edd_maybe_schedule_download_recalculation( $order_item->product_id );
```
cklosowski, line 102: If the `$order_item->product_id` changes, we probably need to trigger an update on that old `product_id` as well.

cklosowski (changes requested):

cklosowski (changes requested):
@robincornett This cron is working exactly as I expect, however I think something is off with the recalculations here.

I have 6 orders for this product
Screen Shot 2022-09-12 at 4 38 35 PM

1 is Refunded
Screen Shot 2022-09-12 at 4 38 47 PM

However the stats read 4 instead of 5:
Screen Shot 2022-09-12 at 4 39 17 PM

robincornett:
@cklosowski do you want me to reevaluate the download model in this PR or open a separate issue/PR for it? As this was intended primarily to account for timing issues.

cklosowski:
@robincornett let's go ahead and assess the recalculations in another issue.

cklosowski (approved):

PR #9421, opened 2022-09-07
Update object ID column for EDD logging class

### LINKED ISSUE #9313: 3.0 - $edd_logs->get_log_count doesn't work for file download logs
### Expected behavior
When I use this legacy code, I expect it to return accurate results:

```php
global $edd_logs;
$download_id = 942;
$count       = $edd_logs->get_log_count( $download_id, 'file_download' );
```

### Actual behavior
It returns the count of all file download logs, not just the one for that download ID.

Run the code above, which works in EDD 2.11 but not 3.0.

PHP Version: 8.0
EDD Version (or branch): main (3.0)
WordPress Version: 6.0.1

Looking at the `get_log_count` method, the `$object_id` parameter is set up as the `object_id` in the array of parameters used to query logs, but there isn't an `object_id` column in the file download logs table--I think that should be `product_id` instead. I think the same problem exists for the API requests logs table, but I'm not sure what the queried object ID might be there...looking at the method in 2.11, the object ID is the `post_parent` for the logs, but looking at our migration, the `post_parent` isn't migrated for API logs.

robincornett:
Fixes #9313

Proposed Changes:
1. Creates a private method `get_object_id_column_name_for_type` to get the correct "object ID" column for a given log type:
    - 'object_id` as a default
    - `product_id` for file downloads
    - `user_id` for API requests ... not sure on this one as it looks like API requests never assigned a post parent to that log type, so I'm not sure there was a way to actually query API logs in this manner. But user ID seemed like a reasonable option and object ID is certainly invalid.
2. Updates `get_log_count`, `delete_logs`, and `parse_args` methods to use the new helper method.
3. Adds a unit test to add a file download log and count it using the original `get_log_count` method.

You can test by generating some file download logs (CLI command: `wp edd download_logs create --number=10`) and then using the code in the original issue to query for logs.

### DIFF
--- a/includes/class-edd-logging.php
+++ b/includes/class-edd-logging.php
@@ -424,7 +424,7 @@ class EDD_Logging {
 	 */
 	public function get_log_count( $object_id = 0, $type = null, $meta_query = null, $date_query = null ) {
 		$r = array(
-			'object_id' => $object_id,
+			$this->get_object_id_column_name_for_type( $type ) => $object_id,
 		);
 
 		if ( ! empty( $type ) && $this->valid_type( $type ) ) {
@@ -462,7 +462,7 @@ class EDD_Logging {
 	 */
 	public function delete_logs( $object_id = 0, $type = null, $meta_query = null ) {
 		$r = array(
-			'object_id' => $object_id,
+			$this->get_object_id_column_name_for_type( $type ) => $object_id,
 		);
 
 		if ( ! empty( $type ) && $this->valid_type( $type ) ) {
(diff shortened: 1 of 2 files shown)


includes/class-edd-logging.php
```diff
@@ -553,11 +553,8 @@ private function parse_args( $args = array() ) {
 
 		// Back-compat for post_parent.
 		if ( ! empty( $r['post_parent'] ) ) {
-			if ( ! empty( $r['log_type'] && 'file_download' === $r['log_type'] ) ) {
-				$r['product_id'] = $r['post_parent'];
-			} else {
-				$r['object_id'] = $r['post_parent'];
-			}
+			$type                                        = ! empty( $r['log_type'] ) ? $r['log_type'] : '';
+			$r[ $this->get_object_id_column_name_for_type( $type ) ] = $r['post_parent'];
```
cklosowski, line 557: @robincornett we should probably unset the `post_parent` in the array `$r`, so it doesn't get passed in there, since this column doesn't exist.

cklosowski (changes requested):

includes/class-edd-logging.php
robincornett: It is unset right after this in a group of old keys beginning in line 566.

cklosowski: And so it is...thanks.

cklosowski (approved):

Issue #7603, opened 2020-02-27
Split Purchase Confirmation and View Receipt into 2 distinct templates


cklosowski:
### Explain your enhancement (please be detailed)
Carried over from #2304, as a starting point we need to look at splitting our Purchase Confirmation and View Receipt pages into two actual separete pages. They may contain some of the same content sections, but they need to trigger a different template.

### Justification or use case
Many times it is ideal to load retargeting or conversion tracking pixels/code on the purchase confirmation, but the problem right now is EDD uses the same template for both the initial purchase confirmation, or for when a user looks at their receipt anytime after purchase, which can then fire off the retargeting pixel and create inaccurate tracking.

Over the years there have been a number of ‘hacks’ around this in a custom template, but it is time to support two different templates, further allowing these two different contexts to be treated differently.

spencerfinnell:
Some other context: https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3391, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/3529, https://github.com/easydigitaldownloads/easy-digital-downloads/issues/5681

@cklosowski So the idea here is to introduce a completely separate/new page, setting, and shortcode? The shortcode will essentially duplicate `[edd_receipt]` for the first iteration?

spencerfinnell:
Also related https://github.com/easydigitaldownloads/easy-digital-downloads/issues/6678

cklosowski:
@spencerfinnell I'm not 100% sure if I want to rip the band-aid off yet, or not, but I feel like if we're going to do this, during 3.0 might be the time, possibly with some backwards compatibility for context.

I'm goign to try and code up what I think is the right solution, including backwards compatibility for this, then in 3.1 we can remove the back compat and move forward.

Issue #3529, opened 2015-06-26
Purchase Confirmation: Square peg in a round hole


chriscct7:
Right now, we use the purchase confirmation page for many things that it was never intended to do. For example, it handles the view for pending payments that come from the my account page. Even though the page says purchase confirmation, the order is pending and you can't actually finish it (something #2714 is about).

Purchase Confirmation is used for many more things than purchase confirmation, which is a problem. We need to make it so that the purchase confirmation is only used 1 time, for just completed purchases. Else it is confusing for users. We could even rename it "Thank You" or something.

I think there's a couple related changes that need to happen here.
- As suggested in #2714, eChecks and others should get a processing state.
- A new page should be made called order status.
  - The order status should take a url param called ?id=xx&cid=xx&&key=xx where id is the order ID, cid is the customer id and key is the payment key unique to the customer. If this id is not included, show a simple form that asks for an order number and payment key (both are on receipts) to see the status. Both the id, and key must be provided and the key must match the id in order to see the order information
    - If the order is processing (awaiting IPN callback from PayPal), show a waiting timer, where on timeout, show order status of pending
  - If the order is pending, show items in cart along with fees, and discounts and offer ability to pay for the item
  - If the order is abandoned, see pending
  - If the order is complete, show the totals, download links, etc. If a url param called "success" is present, show the thank you for purchasing notification
  - If the order is saved, see pending
- From the My Account page link to the order status page

The problem with "Purchase Confirmation" is we use it for many things where Purchase confirmation isn't actually a confirmation. This is confusing to people, particularly those who've added text to the page who think its only used for receipts.

Alternatively, we could introduce views on the my account page to handle complete, pending, saved, and abandoned, and views on the purchase confirmation page for processing and complete with success.

pippinsplugins:
We definitely need to re-work how this all works.

I haven't thought about it too much yet, but initially I like the idea of an order status page.

cklosowski:
This is the point of #3391 as well.

pippinsplugins:
This and #3391 moved to 2.6

pippinsplugins:
Too late in 2.6 cycle to get this.

cklosowski:
Related #2304

sunnyratilal:
Related #2304

pippinsplugins:
We're too close to 2.7 beta to get this in.

cklosowski:
[...]

Issue #2304, opened 2014-05-19
Improve "Purchase Confirmation" page layout


chriscct7:
_notes from @cklosowski_
I’ve updated this issue to be more about the design aspects of this. I’ve created #7603 to handle the actual splitting of the templates, so we can focus this issue specifically about improving the display and layout of them.

_original issue_

We need to look at renaming the `Purchase Confirmation` to `Receipt` or maybe `Purchase Receipt`.

We also need to build a way that viewing the receipt immediately after purchase has a different URL than viewing order details from the `Purchase History`.

This allows things like google analytics conversion tracking to only fire on the actual conversion, not reviewed receipts.

This would also solve #3391
Original issue:
> As begun in #1956 we should look at making the purchase confirmation page more intuitive. 
> 
> One of the ways suggested way splitting out some of the data, and rearranging some other data. Right now, the purchase confirmation page needs alot of work. 
>
>One of it's biggest weaknesses is that its original title "Purchase Confirmation" is a wording that by itself could either mean this is the confirmation of your purchase, or you need to click a button to finalize your purchase. This becomes even more confusing when translating that string to other languages. 
>
>While we can't change the name, we can look at ways of making the page's data better organized and enhance the "you're done" feeling (maybe change the "thanks for your purchase" to have a green background from the .edd_success CSS class).
>
>See https://github.com/easydigitaldownloads/Easy-Digital-Downloads/issues/1956#issuecomment-43524608

pippinsplugins:
Punting to 2.4

pippinsplugins:
Any objections to slimming this issue down to just renaming the title from `Purchase Confirmation` to `Receipt`?

cklosowski:
I'm good with that, but can we address #3391 along with it? It's nothing too major, just a different URL that loads the same template and is a much smaller scope.

pippinsplugins:
Yeah let's do it, both for 2.5.

cklosowski:
Related #3529

sunnyratilal:
@pippinsplugins Is the goal here, for now, just to rename `Purchase Confirmation` to `Receipt`?

cklosowski:
No this needs so much more. I'll outline it:
1) We need to build two different URLs, not just a single endpoint for both the purchase confirmation AND the view receipt link.
2) We need to rethink the phrasing on them

[...]

PR #9426, opened 2022-09-12
Rock/core blocks pages

robincornett:
This can be tested in conjunction with https://github.com/awesomemotive/edd-blocks/pull/88, but it should also be tested without any blocks at all to confirm that the expected current behavior still works.

Proposed Changes:
1. Creates a helper function `edd_get_required_pages` to define the array of required pages. There is a filter so that I can add a new page in the blocks plugin.
2. Updates the `edd_install_pages` function to use the new array. The changes actually make this process a little less convoluted and a little more DRY.
3. Updates `edd_can_view_receipt()` to allow anyone who can edit orders to view a receipt. Additionally, there is a new row action for those users to view any order receipt. 
4. Moved checkout page functions to a new file so I could keep better track of them:
    - `edd_get_success_page_uri` has been updated to include a check for the new confirmation page, if one is registered (will fall back to the receipt page if not). This check only happens if there are no query parameters passed to the function.
    - `edd_send_to_success_page` has been updated to pass the query string parameter to `edd_get_success_page_uri` for application instead of applying it directly.
    - a new `edd_get_confirmation_page_uri` function has been created to get the confirmation page from the option, or will fall back to the success page if there isn't a confirmation page.
    - a new `edd_get_receipt_page_uri` function has been created to build a custom receipt URI, using the success page, with the order ID and a hashed order parameter, similar to what we did in the Invoices extension.
5. The receipt page parameters have been added to the `edd_can_view_receipt` checks for guest orders and the shortcode.
6. The purchase history template has been updated to use the new receipt URL.

Related: #2304, #3529, #7603


cklosowski:
@robincornett  Overall it seems this is working as expected. We need some updates to the content of the page settings now though, if there are some technical limitations here, let me know, but currently the 'confirmation' and 'success' settings seem like they should be the same thing, so need to try and clarify that.
* Primary Checkout Page
```
This is the checkout page where buyers will complete their purchases.
The [download_checkout] shortcode must be on this page.
```

This should be updated to:
```
This is the checkout page where customers will complete their purchases.
The Checkout Block or [download_checkout] shortcode must be on this page.
```

* Confirmation Page
```
This page must include the EDD Confirmation block.
Using this separates the order confirmation from the receipt.
```

This should be updated to:
```
This page must include the EDD Confirmation block.
Use this page separate of your receipt page, to ensure proper conversion tracking.
```

* Success Page
Let's rename this to 'Receipt Page'
```
This is the page to show a detailed receipt for an order.
Use the EDD Receipt block here, to work with the confirmation page.
```

```
This is the page to show a detailed receipt for an order.
Use the EDD Receipt block or the [edd_receipt] shortcode, to work with the confirmation page.
```

robincornett:
Marking as ready for review--I'll update wording in the parent branch or blocks as appropriate.

cklosowski (approved):

PR #9428, opened 2022-09-12
Allow forcing login to download files

cklosowski:
Proposed Changes:
1. Adds a checkbox for requiring login to download files.
2. Adds a stop point in the process-download.php for this setting.
3. Stores the requested file args in a session variable.
4. Redirects to the login form or the wp-login
5. Upon successful login, redirects to the homepage with an action and a nonce
6. Starts the download, and empties the session data for the file download arguments.
7. Using `setTimeout` redirects to the homepage without the query strings for the action.


includes/admin/settings/register-settings.php
```diff
@@ -983,6 +983,14 @@ function edd_get_registered_settings() {
 					),
 				),
 				'file_downloads' => array(
+					'require_login_to_download' => array(
+						'id'            => 'require_login_to_download',
+						'name'          => __( 'Require Login', 'easy-digital-downloads' ),
+						'desc'          => __( 'Check this if you want to require a user to login before file download links deliver the file.', 'easy-digital-downloads' ),
```
robincornett, line 989:
```suggestion
						'desc'          => __( 'Require a user to login before file download links deliver the file.', 'easy-digital-downloads' ),
```
I would love to see us move away from "check this" language.

includes/process-download.php
```diff
@@ -1076,6 +1098,53 @@ function edd_local_file_location_is_allowed( $file_details, $schemas, $requested
 	return apply_filters( 'edd_local_file_location_is_allowed', $should_allow, $file_details, $schemas, $requested_file );
 }
 
+/**
+ * Detect downloading a file immeidatly after a forced login.
```
robincornett, line 1102:
```suggestion
 * Detect downloading a file immediately after a forced login.
```

includes/process-download.php
```diff
@@ -1076,6 +1098,53 @@ function edd_local_file_location_is_allowed( $file_details, $schemas, $requested
...
+ *
+ * When the store requires being logged in to download files, this handles the file download after logging in.
+ * We need this otherwise the file is downloaded immediatly after successfully logging in, but the page never changes.
```
robincornett, line 1105:
```suggestion
 * We need this otherwise the file is downloaded immediately after successfully logging in, but the page never changes.
```

includes/process-download.php
```diff
@@ -1076,6 +1098,53 @@ function edd_local_file_location_is_allowed( $file_details, $schemas, $requested
...
+function edd_redirect_file_download_after_login() {
+	$nonce = isset( $_GET['_nonce'] ) ? sanitize_text_field( $_GET['_nonce'] ) : false;
+
+	// No nonce provided, redirect to the homepage.
+	if ( empty( $nonce ) ) {
+		wp_safe_redirect( home_url() );
+	}
+
+	// Nonce verification failed, redirect to the homepage.
+	if ( wp_verify_nonce( $nonce, 'edd_process_file_download_after_login' ) ) {
```
robincornett, line 1118:
```suggestion
	if ( ! wp_verify_nonce( $nonce, 'edd_process_file_download_after_login' ) ) {
```
I think this is inverted? Although changing it is not allowing a download (either way I don't get my file after a successful login and I'm not seeing an error message for why).

robincornett (changes requested):
This isn't working for me at all--I think the nonce verification is inverted, but changing that doesn't get a successful download either, and no errors are thrown. The session data is being set. I am going through the wp-login.php, not a custom login form.

includes/process-download.php
cklosowski: Ah yep I was testing some things and committed my test.

cklosowski:
Okay @robincornett give that a shot.

cklosowski:
@zkawesome that means my `setTimeout` isnt' long enough. I think i have a new approach for this that should help is work as expected and not need a timeout.

With this current method, a LARGE file would actually get cut off due to the 2nd redirect.

cklosowski:
@zkawesome see if that change helps. I moved away from 'replace' and used 'assign', I'm wondering if the replace was overwriting the dom state of the download

cklosowski:
@robincornett what is your file download method? Forced or redirect?

cklosowski:
@robincornett @zkawesome Okay So I'm taking a different approach here. I was trying to use redirects, and window assignments, but that causes some race conditions. This approach adds a hidden link to the page, and then clicks it, then redirects.

robincornett (commented):
This worked for me this time! I also tried flubbing the login (putting in a wrong password) and then fixing it and the download still worked.

I'm a little confused by the sequence of events--I feel like there is an extra redirect happening that's a little disconcerting. Here's what I experienced:
1. I clicked the link from my purchase receipt.
2. I was redirected to the login page.
3. I logged in.
4. I was redirected to the front page (I do not have a login redirect page set). The URL had a lot of parameters in it, which was not unexpected.
5. The file downloaded successfully.
6. The URL was reset to the front page.
7. Then the site reloaded again. This is the one that I'm wondering if it's necessary? Is it happening automatically because of step 6?

robincornett (approved):
I'm still a little thrown by the final refresh but everything is working.

PR #9430, opened 2022-09-13
Update payment methods for incomplete data

### LINKED ISSUE #9157: 3.0 - using edd_insert_payment with incomplete data doesn't create a new order
### Expected behavior
When I use `edd_insert_payment`, I expect it to add a new order to the database, which it does in EDD 2.x. I also expect it to not throw notices.

### Actual behavior
When I use the function with incomplete data, I get multiple PHP notices (in both EDD 2.11.6 and release/3.0). However, in 2.11.6, the payment is added to the database; in 3.0, it is not.

Manually insert a payment object. Here's the code I used, just to kick things off:

```php
$payment_id = edd_insert_payment(
	array(
		'customer_id' => 1,
		'email'       => 'admin@edd.local',
		'user_id'     => 1,
		'gateway'     => 'manual',
		'total'       => edd_sanitize_amount( sanitize_text_field( 20 ) ),
		'status'      => 'pending',
	)
);
```

I know this is woefully incomplete; I was experimenting with something else.

Here's a screenshot of the PHP notices logged by Query Monitor (more concise than the actual log):

image

PHP Version: 7.4
WordPress Version: 5.9.3

I think there are potentially two issues:
* `edd_build_order` (what's called by `edd_insert_payment`) should maybe set up a default data array and improve checks to avoid the notices if data is missing.
* I believe that the order data is failing validation in Berlin DB (my guess is that something is empty which isn't allowed).

The question is, if the notices are fixed, should the order be created anyway? It seems like it should not, even though it's technically a behavior change from 2.x. Marking as a regression for now.

robincornett:
Fixes #9157

Proposed Changes:
1. Updates the `EDD_Payment` class to allow added downloads to set or inherit the order/payment status (this covers adding a payment and saving it directly as `complete` instead of as `pending` first).
2. Updates `edd_build_order` to account for missing/incomplete data.

### DIFF
--- a/includes/orders/functions/orders.php
+++ b/includes/orders/functions/orders.php
@@ -649,6 +649,17 @@ function edd_build_order( $order_data = array() ) {
 		return false;
 	}
 
+	$order_data = wp_parse_args(
+		$order_data,
+		array(
+			'user_info' => array(),
+		)
+	);
+
+	if ( empty( $order_data['user_info']['email'] ) ) {
+		return false;
+	}
+
 	/* Order recovery ********************************************************/
 
 	$resume_order   = false;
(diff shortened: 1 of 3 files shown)


cklosowski:
@robincornett Since this is primarily about custom integrations doing things a bit different, can we write a few unit tests to catch some of these and ensure things are working as expected?

cklosowski (approved):

PR #9431, opened 2022-09-14
Destroy order and customer when deleting a customer #9420

### LINKED ISSUE #9420: Bug: Deleting a customer & their orders turns into refunds
### Expected behavior

I expect that deleting a customer and all their records removes the data completely, and their numbers are removed from reporting.

### Actual behavior

Deleted orders are turned into refunds, they only appear in new 'refunds' tab. The deleted orders are counted as refunds in the reports. This is big issue when cleaning up test orders etc. 

One a side note: what's the point of the new Refunds tab? I'm confused to why there are refunds in two locations.

1) Go to a customer page and delete them ( ✅ Are you sure you want to delete this customer? ✅ Delete all associated payments and records?)
2) Customer orders are deleted, they don't appear in the standard orders list.
3) Somehow duplicate versions of the orders appear in the "Refunds" tab. These fake 'refunds' also show up in reports?

Screenshot for reference below

PHP Version: 7.4

EDD Version (or branch): Version 3.0.4

WordPress Version: 6.0.2

Screen Shot 2022-09-07 at 12 32 57
 relevant information:

cklosowski:
Fixes #9420 

Proposed Changes:
1. Swaps the delete customer action to use `destroy` functions instead.

### DIFF
--- a/includes/admin/customers/customer-actions.php
+++ b/includes/admin/customers/customer-actions.php
@@ -447,7 +447,7 @@ function edd_customer_delete( $args = array() ) {
 	if ( $customer->id > 0 ) {
 
 		$payments_array = explode( ',', $customer->payment_ids );
-		$success        = edd_delete_customer( $customer->id );
+		$success        = edd_destroy_customer( $customer->id );
 
 		if ( $success ) {
 
@@ -455,7 +455,7 @@ function edd_customer_delete( $args = array() ) {
 
 				// Remove all payments, logs, etc
 				foreach ( $payments_array as $payment_id ) {
-					edd_delete_purchase( $payment_id, false, true );
+					edd_destroy_order( $payment_id, false, true );
 				}
 
 			} else {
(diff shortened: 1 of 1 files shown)


includes/admin/customers/customer-actions.php
```diff
@@ -447,15 +447,15 @@ function edd_customer_delete( $args = array() ) {
...
+		$success        = edd_destroy_customer( $customer->id );
 
 		if ( $success ) {
 
 			if ( $remove_data ) {
 
 				// Remove all payments, logs, etc
 				foreach ( $payments_array as $payment_id ) {
-					edd_delete_purchase( $payment_id, false, true );
+					edd_destroy_order( $payment_id, false, true );
```
robincornett, line 458:
```suggestion
					edd_destroy_order( $payment_id );
```
This function only takes one parameter.

robincornett (approved):

PR #9433, opened 2022-09-14
When saving  the bundle product conditions, ensure that the array keys are re-indexed, starting at 1 #9343

### LINKED ISSUE #9343: The price option assignment of download files (products) are not showing correctly on the WP-Admin Download page when using Variable Pricing and Product Type Bundle
### Expected behavior
When I open an existing product on the WP-Admin Download page, which has a Product Type = Bundle, and I edit this product by changing the download files (products), I expect the price option assignments are correctly stored and displayed after refreshing the page.

### Actual behavior
The problem is that it is not correct after refreshing the page.

It shows "All" on the price option assignment for the newly created download file (product).

1) Create a new product with two price options (eg. Basic and Premium)
2) Change the product type to Bundle
3) Create a download file (product) and assign it to the Basic price option
4) Create another download file (product) and assign it to the Premium price option
5) Publish the product
6) Refresh the page if you are using Gutenberg
7) You will see everything is fine
8) Now create a new download file and assign it to the Basic price option
9) Remove the first download file which was also assigned to the Basic price option
10) Save the product
11) Refresh the page if you are using Gutenberg
12) See that you have a problem now

It might be that we don't need multiple price options, but I think that by using at least two price options can make this clearer.

PHP Version: latest

EDD Version (or branch): latest

WordPress Version: latest

It looks like the values are stored as expected, but it is not showing them correctly.

I think this is related to the indexes of these post meta (because they don't match):
```_edd_bundled_products```
```_edd_bundled_products_conditions```

There are places like the ```get_variable_priced_bundled_downloads``` function (```EDD_Download``` class) that uses the ```array_values``` function so that the indexes are reset and they can match later, see: https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/class-edd-download.php#L572

But there are places like the ```edd_render_products_field``` function that are not using the same rule, see: https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/admin/downloads/metabox.php#L650 

_Edited to fix post meta key_

### PR DESCRIPTION
**Pull request by cklosowski**
Fixes #9343 

Proposed Changes:
1. Adds a metabox save sanitization so that we ensure that we also re-index the bundle file conditions array as well.

### DIFF
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -184,6 +184,22 @@ function edd_sanitize_bundled_products_save( $products = array() ) {
 add_filter( 'edd_metabox_save__edd_bundled_products', 'edd_sanitize_bundled_products_save' );
 
 /**
+ * Sanitize bundled products conditions on save
+ *
+ * @since 3.1
+ *
+ * @param array $bundled_products_conditions
+ * @return array
+ */
+function edd_sanitize_bundled_products_conditions_save( $bundled_products_conditions = array() ) {
+	return array_combine(
+		range( 1, count( $bundled_products_conditions ) ),
+		array_values( $bundled_products_conditions )
+	);
+}
+add_filter( 'edd_metabox_save__edd_bundled_products_conditions', 'edd_sanitize_bundled_products_conditions_save' );
+
+/**
  * Don't save blank rows.
  *
  * When saving, check the price and file table for blank rows.
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -184,22 +184,6 @@ function edd_sanitize_bundled_products_save( $products = array() ) {
 add_filter( 'edd_metabox_save__edd_bundled_products', 'edd_sanitize_bundled_products_save' );
 
 /**
- * Sanitize bundled products conditions on save.
- *
- * @since 3.1
- *
- * @param array $bundled_products_conditions
- * @return array
- */
-function edd_sanitize_bundled_products_conditions_save( $bundled_products_conditions = array() ) {
-	return array_combine(
-		range( 1, count( $bundled_products_conditions ) ),
-		array_values( $bundled_products_conditions )
-	);
-}
-add_filter( 'edd_metabox_save__edd_bundled_products_conditions', 'edd_sanitize_bundled_products_conditions_save' );
-
-/**
  * Don't save blank rows.
  *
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by robincornett — COMMENTED** [review]
I feel like a better change would be to update the `foreach` block for existing products (lines 622-686) to use the `$key` from the saved array.

When I looked into this issue originally, the problem seemed to be that the `name` attribute for the dropdown for existing products doesn't use a defined index key but when there aren't products yet, it does--so existing products end up defaulting to a `0` key/index.

So it seems like we should update the form markup to use the existing array keys instead of an arbitrary index, rather than trying to fix the markup mistakes during the sanitization.

**Comment by cklosowski** [comment]
Take a look at that @robincornett That addresses the issue in the form markup. Actually both initial and existing products were not assigning an index to the name of the dropdown.

**Review by robincornett — CHANGES_REQUESTED** [review]
This is still not working for me, and I think it's because the `edd_sanitize_bundled_products_save` function is removing the array keys on the bundled products, but that's not happening for the bundled product conditions--so the arrays are getting out of sync.

If the keys are irrelevant we need to run the conditions through `array_values` as well, and maybe the `$index` is pointless as it's ignored? If the keys are relevant (which they are for variable pricing but this is different) then we need to remove the `array_values` from that sanitization function, I think.

**Comment by cklosowski** [comment]
@robincornett Okay I added the `array_values` back into the bundled products. Can you give this a go? It all seems to work for me mostly, so not sure where you're seeing the failure.

If this doesn't work can you just adjust as necessary to this branch and push up what is working for you?

**Comment by robincornett** [comment]
This is working better for me, but when it comes to the `get_variable_priced_bundled_downloads` method in the download class, I think things will likely break, because the array keys between the bundled products and the bundled product conditions don't match. The products start counting at `0` because computers and the conditions start at `1` because you've coded them that way.

I think things could work fairly well if the products are also updated to start at `1` like this:
```php
	$products = array_unique( $products );

	return array_combine(
		range( 1, count( $products ) ),
		array_values( $products )
	);
```

It is still possible for the arrays to get out of sync if you accidentally add a duplicate product, because the product sanitization cleans out duplicates and the condition can't do that. It is interesting to me that the duplicate product isn't allowed, as I could see a possibility of wanting to sell three bundle variations and allow a specific product to be delivered for only two of them--but our current code doesn't allow for that.

I'm pushing the change I suggested above.

cklosowski:
I can't review this but @robincornett your changes are working as expected. I'm merging this.

PR #9438, opened 2022-09-16
Trigger edd_load_gateway with timeout

zkawesome:
Event `edd_gateway_loaded` is triggered immediately when the script is loaded, therefore other scripts that rely on this event cannot detect it. 

I have added a 200ms timeout before firing the event.


cklosowski (approved):

robincornett (approved):

cklosowski:
I adjusted the settimeout delay, and once the build passes we can merge and close this out.

PR #9440, opened 2022-09-16
Update download sales/earning calculations

### LINKED ISSUE #9424: Download Report Tiles and Download settings sales and earnings are not accurate
Update get_order_item_count and get_order_item_earnings to be more consistent with dashboard reports.

Both sections should match in this screenshot:
Markup on 2022-09-07 at 16:09:58

https://secure.helpscout.net/conversation/2000520514/2149135?folderId=5747299

robincornett:
Fixes #9424

Proposed Changes:
1. Updates the net sales/earnings calculations method in the download model to query order items that are complete, refunded, and partially refunded, and to use gross order statuses for calculating earnings.
2. Updates the stats class for querying order items to always do an inner join to the orders table, because we need to know what type and status the order has. So the count now gets complete/partially refunded counts for orders of the type sale which are also complete/partially refunded. _Reminder that the stats count is counting rows and the download model is doing quantities._
3. Fixes the order factory for unit tests to set a correct order item status.
4. Updates the stats tests to expect the correct number of order items.

### DIFF
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -930,7 +930,7 @@ class Stats {
 		$this->query_vars['table']             = $this->get_db()->edd_order_items;
 		$this->query_vars['column']            = true === $this->query_vars['exclude_taxes'] ? 'total - tax' : 'total';
 		$this->query_vars['date_query_column'] = 'date_created';
-		$this->query_vars['status']            = array( 'complete' );
+		$this->query_vars['status']            = array( 'complete', 'refunded', 'partially_refunded' );
 
 		// Run pre-query checks and maybe generate SQL.
 		$this->pre_query( $query );
--- a/includes/class-stats.php
+++ b/includes/class-stats.php
@@ -1035,7 +1035,7 @@ class Stats {
 		$this->query_vars['table']             = $this->get_db()->edd_order_items;
 		$this->query_vars['column']            = 'id';
 		$this->query_vars['date_query_column'] = 'date_created';
-		$this->query_vars['status']            = array( 'complete' );
+		$this->query_vars['status']            = array( 'complete', 'refunded', 'partially_refunded' );
 
 		// Run pre-query checks and maybe generate SQL.
 		$this->pre_query( $query );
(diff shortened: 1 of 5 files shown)


includes/class-stats.php
```diff
@@ -930,7 +930,7 @@ public function get_order_item_earnings( $query = array() ) {
 		$this->query_vars['table']             = $this->get_db()->edd_order_items;
 		$this->query_vars['column']            = true === $this->query_vars['exclude_taxes'] ? 'total - tax' : 'total';
 		$this->query_vars['date_query_column'] = 'date_created';
-		$this->query_vars['status']            = array( 'complete' );
+		$this->query_vars['status']            = array( 'complete', 'refunded', 'partially_refunded' );
```
cklosowski, line 933: @robincornett These need to account for renewals somehow as well.

cklosowski (changes requested):

includes/class-stats.php
robincornett: Since this is querying the order items table, renewals are included in this--the order item status for renewal orders is `complete`. However, the inner join for `get_order_item_count` does need to be updated.

cklosowski:
@robincornett something is off here still I think.

I have a product that was purchased 'last month', and refunded 'this month', and my graph says I have 1 sale and negative earnings. The negative earnings is correct, however the 1 sale is not.

It appears the title got the correct state, but the graph did not. I know we originally had this issue to direct at the tile itself, but if we're fixing the tile, we need to fix the graph so they don't show conflicting data.

Graph/Tile:
Screen Shot 2022-09-20 at 6 18 44 AM

Database rows:
```
15500	15435	7832	36906	Simple Shipping Product One	NULL	0	download	complete	-1	5.000000000	-5.000000000	0.000000000	0.000000000	-5.000000000	1.00000	2022-09-17 13:32:35	2022-09-17 13:32:35	urn:uuid:0d130b85-6c1e-4bb0-b8d6-64adf9b24eb9
15435	0	7771	36906	Simple Shipping Product One	NULL	0	download	refunded	1	5.000000000	5.000000000	0.000000000	0.000000000	5.000000000	1.00000	2022-08-26 20:06:50	2022-09-17 13:32:35	urn:uuid:34b95577-8dfe-461c-81b9-8e2ce95ad4e4
```

robincornett (commented):
I can't approve the PR since I authored it originally, but I would if I could. The latest changes you made are working for me, @cklosowski.

cklosowski (approved):

PR #9449, opened 2022-09-21
Fix graph groupings, and the graph should follow the path of the tiles when it comes to calculating earnings #9448

### LINKED ISSUE #9448: Reports Graphs can be inaccurate when grouping by month due to TZ adjustments.
### Expected behavior
When I view a chart by Month (Yearly view) I expect the chart to display correctly.

### Actual behavior
Due to the grouping, we are getting some issues with the query when the first item found in a date grouping is close to the timezone adjustment.

The query groups and then then we build the array for the chart, it groups incorrectly. Here is my output that shows this:
```
string(11) "result data"
Raw result date: string(19) "2022-01-01 10:41:50"
Formatted Date of DB Value: string(7) "2022-01"
Formatted Date of Chart Value: string(7) "2022-01"


string(11) "result data"
Raw result date: string(19) "2022-02-01 05:16:54"
Formatted Date of DB Value: string(7) "2022-01"
Formatted Date of Chart Value: string(7) "2022-01"


string(11) "result data"
Raw result date: string(19) "2022-03-01 05:17:24"
Formatted Date of DB Value: string(7) "2022-02"
Formatted Date of Chart Value: string(7) "2022-01"


string(11) "result data"
Raw result date: string(19) "2022-04-01 09:42:08"
Formatted Date of DB Value: string(7) "2022-04"
Formatted Date of Chart Value: string(7) "2022-01"
```

We can see that when we get to the grouping for results in February of 2022, due to the fact that we've grouped by the date including the hour, it adjusts by the timezone to include it in January. This results in March being missed completely due to this adjustment.

1)
2)
3)
etc, etc

PHP Version:

EDD Version (or branch):

WordPress Version:

cklosowski:
Fixes #9448 

Proposed Changes:
1. On Downloads > Overview, the sales & earnings graph is now following the pattern of the tile when calculating earnings, and including gross statuses in order to properly account for refunds. Since it was not using the `gross` statuses, but was also including `refund` as a type, it was not getting a complete accurate revenue calculation. The sales counts remain the same (just as the tile does).
2. Stops converting to UTC timestamps when grouping data in the array for the graph as it was causing a double adjustment to happen. We've already accounted for the timezone when we call this at the beginning of the chart data gathering `$chart_dates  = Reports\parse_dates_for_range( null, 'now', false );` so there is no need to run it again and 'double adjust'.
3. Applies the logic from 2 above to the refunds graph as well.

I started looking into the other graphs we have in core (download specific graph, gateway graph) and those both seem to use a more 'long winded approach' to these calculations and may need some updates. The downloads graph can be addressed in #9424 which already has a PR and should be fixed up when we include the proper status checks there.


cklosowski:
With these adjustments:
Screen Shot 2022-09-21 at 12 43 31 AM

Without these adjustments:
Screen Shot 2022-09-21 at 12 43 59 AM

cklosowski:
Marking as a Draft, there is some more work here, as I just realized that the counts are slightly different on the yearly graph for each month than if you run the report month by month, and this is due to a DST change between the start of the graph and the end of the graph, which throws off the grouping.

Going to have to make some more adjustments.

cklosowski:
I'm not sure we can get past the oddness of the DST issue. The problem is that we cannot rely on MySQL having TZ info populated, which means we cannot used stringed Timezone Conversion on `CONVERT_TZ`, and we are limited to just using offsets. This, by proxy, means that the only way we can convert timezones is by the current GMT offset, and not the GMT offset for the graph that is being requested.

cklosowski:
Marking for review again, this is as good as we're going to get it for now. It's not 100% accurate when you start spanning DST changes, as it groups things slightly incorrectly but it is only off by the max of the GMT offset.

robincornett (approved):

robincornett (approved):

PR #9457, opened 2022-09-26
Align baseline of multiple Y axes in reports

zkawesome:
Proposed Changes:
- There is now a new option for scales called `hideNegativeTicks` that would hide negative ticks in the Y axis labels. We are using it for the `Sales` Y Axis where ticks cannot be negative.
- We are now adjusting and scaling the Y axis AFTER Chart has been initialized - we want Chart.js to calculate `min` and `max` values for us so that we can calculate the ratio with which we can scale each axis.

Before:

image

After:

image


robincornett:
The graph is looking a lot better on a larger scale...I've done several refunds today to get some negatives happening and when I switch to a "this week" view, the intermediate lines on the sales count graphs are incorrectly cast to 0:
image

PR #9463, opened 2022-10-01
Fix/this month hour by hour

cklosowski:
With the changes in #9338, when you are in the first 2 days of the month, the graph is just outputting two data points.

With 'other' if we have only 2 days time diff, we switch to hour by hour, which we should do here as well.


robincornett (commented):
This seems reasonable but unit tests are failing:
> 1) EDD\Reports\Data\Charts\v2\Manifest_Tests::test_get_chart_options_default_non_pie_options_should_match_and_include_the_endpoint_label

robincornett (approved):

PR #9477, opened 2022-10-05
Fixing chart baseline on high resolutions

### LINKED ISSUE #9355: Reports Dashboard Graph uses a single Y-Axis for both charts

cklosowski:
Additional Fixes #9355


cklosowski:
Before fix:
image

After Fix:
image

PR #9481, opened 2022-10-06
Fix/get activation date

cklosowski:
Due to the load order of `edd_install`, it is possible for the check to get the orders for activation date detection prior to the table existing.

To replicate:
1. Setup a clean WordPress Installation
2. Install EDD from .zip [easy-digital-downloads.zip](https://github.com/awesomemotive/easy-digital-downloads/files/9729258/easy-digital-downloads.zip) but do not activate it.
3. Go to the plugins list and activate EDD.

You will see that it produces a notice about unexpected output.

Adjusts the `edd_get_activation_date` function to first check for the Table existing, and then also updates it to use the `edd_get_orders` function instead of payments.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -1800,29 +1800,36 @@ function edd_get_payment_image( $gateway, $label ) {
  * For existing installs, this option is added whenever the function is first used.
  *
  * @since 2.11.4
+ * @since 3.1 Checks for the table before checking if orders exist.
+ *
  * @return int The timestamp when EDD was marked as activated.
  */
 function edd_get_activation_date() {
 	$activation_date = get_option( 'edd_activation_date', '' );
 	if ( ! $activation_date ) {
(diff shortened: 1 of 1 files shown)


includes/misc-functions.php
```diff
@@ -1800,29 +1800,36 @@ function edd_get_payment_image( $gateway, $label ) {
...
+					'number'  => 1,
+					'orderby' => 'id',
+					'order'   => 'ASC',
+					'fields'  => 'date_created',
+				)
+			);
+			if ( $orders ) {
+				$first_order = reset( $orders );
+				// Use just the date_created, rather than looking for the completed date (first order may not be complete).
+				if ( ! empty( $first_order->date_created ) ) {
```
robincornett, line 1827: This line and the following should be updated to just check `! empty( $first_order )`, right?

robincornett (approved):

PR #9492, opened 2022-10-08
Add current page to edd_global_vars, include in tax recalculation

### LINKED ISSUE #9491: Cart Block - When running any AJAX that returns new cart HTML, the cart is showing the blocks based cart, not the shortcode based cart.
### Expected behavior
When using the existing checkout shortcode and I change taxes or add a dicscount code, I expect them to be applied to the cart and the cart contents updated.

### Actual behavior
When using the existing checkout shortcode, and I adjust my country/region with taxes enabled or try and apply a discount, the cart HTML is replaced with the blocks cart HTML

1) Enable taxes.
2) Add a product to the cart and use the existing shortcode for checkout.
3) Apply a discount or adjust your address so the trigger for recalculating taxes is run.
4) See your cart get replaced.

PHP Version: All

EDD Version (or branch): 3.1

WordPress Version: All

robincornett:
Fixes #9491


spencerfinnell (approved):
Confirmed that I can use the `[download_checkout]` shortcode on the page defined in the EDD settings, as well as the `EDD Checkout` block on an arbitrary page.

PR #9505, opened 2022-10-11
Add a permalink value to the products endpoint #9488

### LINKED ISSUE #9488: Reconsider the usage of guid in the get_product_data API response.
The code we have now and forever https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/api/class-edd-api.php#L1159 uses the guid field in the table.

In some cases that field could be the direct link to a download with no permalinks applied. We should look at using the actual permalink of the product as some users could be using the API to output data in other places.

cklosowski:
Fixes #9488 

Proposed Changes:
1. Adds a new 'permalink' item to the product data for the `products` API Endpoint.
2. Adds unit tests for the `link` and `permalink` data endpoints.

### DIFF
--- a/includes/api/class-edd-api.php
+++ b/includes/api/class-edd-api.php
@@ -1157,7 +1157,7 @@ class EDD_API {
 		$product['info']['modified_date'] = $product_info->post_modified;
 		$product['info']['status']        = $product_info->post_status;
 		$product['info']['link']          = html_entity_decode( $product_info->guid );
-		$product['info']['permalink']     = get_permalink( $product_info->ID );
+		$product['info']['permalink']     = html_entity_decode( get_permalink( $product_info->ID ) );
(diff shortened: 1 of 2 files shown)


includes/api/class-edd-api.php
```diff
@@ -1157,6 +1157,7 @@ public function get_product_data( $product_info ) {
 		$product['info']['modified_date'] = $product_info->post_modified;
 		$product['info']['status']        = $product_info->post_status;
 		$product['info']['link']          = html_entity_decode( $product_info->guid );
+		$product['info']['permalink']     = get_permalink( $product_info->ID );
```
robincornett, line 1160: Should the permalink be run through `html_entity_decode` the same way we are doing for the `guid` link?

robincornett (approved):

PR #9506, opened 2022-10-12
Update user registration messaging

robincornett:
Proposed Changes:
1. This updates the EDD user registration function to match the logic/messaging in the password reset function.
2. It also adds a success message which should show on the registration block after a user is successfully registered (it did not show for me and I'm still thinking through why).


cklosowski:
@robincornett looks like we need to update some unit tests for this:
```


1) Tests_Login_Register::test_process_register_form_empty_fields
--
  | Failed asserting that an array has the key 'empty_password'.
  |  
  | /tmp/Fgl9GFy43XdSauRkVDDgpnpmJWac6mtC/tests/tests-login-register.php:162
  | /tmp/Fgl9GFy43XdSauRkVDDgpnpmJWac6mtC/vendor/phpunit/phpunit/phpunit:51
  |  
  | 2) Tests_Login_Register::test_process_register_form_username_exists
  | Failed asserting that an array has the key 'password_mismatch'.
  |  
  | /tmp/Fgl9GFy43XdSauRkVDDgpnpmJWac6mtC/tests/tests-login-register.php:187
  | /tmp/Fgl9GFy43XdSauRkVDDgpnpmJWac6mtC/vendor/phpunit/phpunit/phpunit:51

```

robincornett:
Digging into why the unit tests were failing, I ended up with commit e09c251, which makes a lot of changes:
1. Updates `edd_process_register_form` to be internally consistent in using the `$data` parameters (currently it uses a mix of `$data` and `$_POST`).
2. Although `$data` and `$_POST` should technically be the same array of data when actually submitting a form, that was not the case for the unit tests. Most of those set the `$_POST` array, but then called the function with a different (sometimes overlapping) array of parameters, which would override the `$_POST` array. This didn't matter when the function was sometimes checking `$_POST` and sometimes checking `$data`, but I think the function was doing things incorrectly/inconsistently, so when I changed the function to always use `$data`, that's when unit tests started breaking.
3. So several unit tests have been updated.

robincornett:
Noting that intermittent tests are still failing but I think that's on Buildkite rather than the code.

cklosowski (dismissed):

includes/users/register.php
```diff
@@ -83,12 +83,18 @@ function edd_process_register_form( $data ) {
...
 	}
 
-	if ( ( ! empty( $_POST['edd_user_pass'] ) && empty( $_POST['edd_user_pass2'] ) ) || ( $_POST['edd_user_pass'] !== $_POST['edd_user_pass2'] ) ) {
-		edd_set_error( 'password_mismatch', __( 'Passwords do not match', 'easy-digital-downloads' ) );
+	if ( empty( $data['edd_user_pass'] ) ) {
+		edd_set_error( 'empty_password', __( 'The password cannot be a space or all spaces.', 'easy-digital-downloads' ) );
+	}
+
+	// Check if password fields do not match.
+	if ( ! empty( $data['edd_user_pass'] ) && trim( $data['edd_user_pass2'] ) !== $data['edd_user_pass'] ) {
```
robincornett, line 96: Reading through this again and this needs to include an `empty( $data['edd_user_pass2'] )` check as well. Updating.

robincornett:
@zkawesome please test with https://github.com/awesomemotive/edd-blocks/pull/117, thanks!

cklosowski (approved):

PR #9509, opened 2022-10-13
Restrict the ajax downloads search to titles only

### LINKED ISSUE #9508: Look into updating edd_ajax_download_search to search titles only
A customer is reporting that the `edd_ajax_download_search` function is returning unexpected, and too many, results. I think the primary reason for this is that it is using the standard search parameter for the core WP function `get_posts` and that kind of query searches not only the title (which is really what would be most useful and appropriate for the product dropdown), but also the excerpt and content. Here's a stripped down example of the query:

```
SELECT wp_posts.ID
FROM wp_posts
WHERE 1=1
AND (((wp_posts.post_title LIKE '%licensed%')
OR (wp_posts.post_excerpt LIKE '%licensed%')
OR (wp_posts.post_content LIKE '%licensed%')))
AND wp_posts.post_type = 'download'
AND ((wp_posts.post_status = 'publish'))
ORDER BY wp_posts.post_date DESC
LIMIT 0, 5
```

### Justification or use case
I think if the product dropdown searched only the titles, it would give more expected results for users, who are probably really just searching by title in this context, and it should be faster.

robincornett:
Fixes #9508

Proposed Changes:
1. Adds a filter on the `WHERE` clause for the ajax downloads search. This restricts the search to run on the download titles only, instead of including the content and excerpt.

### DIFF
--- a/includes/ajax-functions.php
+++ b/includes/ajax-functions.php
@@ -644,13 +644,14 @@ function edd_ajax_download_search() {
 
 	// Default query arguments
 	$args = array(
-		'orderby'        => 'title',
-		'order'          => 'ASC',
-		'post_type'      => 'download',
-		'posts_per_page' => 50,
-		'post_status'    => implode( ',', $status ), // String
-		'post__not_in'   => $excludes,               // Array
-		's'              => $new_search              // String
+		'orderby'          => 'title',
+		'order'            => 'ASC',
+		'post_type'        => 'download',
+		'posts_per_page'   => 50,
+		'post_status'      => implode( ',', $status ), // String
+		'post__not_in'     => $excludes,               // Array
+		'edd_search'       => $new_search,              // String
+		'suppress_filters' => false,
 	);
 
 	// Maybe exclude bundles.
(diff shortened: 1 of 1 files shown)


includes/ajax-functions.php
```diff
@@ -724,9 +727,28 @@ function edd_ajax_download_search() {
...
+ * Filters the WHERE SQL query for the edd_download_search.
+ * This searches the download titles only, not the excerpt/content.
+ *
+ * @since 3.1.0.2
+ * @param string $where
+ * @param WP_Query $wp_query
+ * @return string
+ */
+function edd_ajax_filter_download_where( $where, $wp_query ) {
+	$search = $wp_query->get( 'edd_search' );
```
cklosowski, line 743: We should be use `esc_like` to do proper sanitization of the search query before using it.
https://developer.wordpress.org/reference/classes/wpdb/esc_like/

cklosowski (changes requested):

cklosowski:
This I believe is unrelated as I can produce it on `release/3.1.0.2`, but the `(All Price Options)` is being added to all items after searching, pushed up a change to address that while we're in here making changes.

Initial Page load and opening the chosen box, A Music Album is showing the All option as well as every price option.
image

Once searching I end up with all the options still but they have '(All Options)' added to them:
image

robincornett:
Just noting that I see the issue you've flagged @cklosowski and agree with your fix.

cklosowski (approved):

PR #9512, opened 2022-10-17
Prefix no-js class, add extra edd-has-js rule

robincornett:
Proposed Changes:
1. This updates the `no-js` body class to be EDD specific.
2. This also adds an extra rule for themes which may be inadvertently styling inputs which should be hidden.


cklosowski:
@robincornett can you provide some steps for testing? Ways to replicate what the customers are seeing?

robincornett:
Sure, here are two scenarios I saw yesterday and this PR should fix both:

**Fixing no buttons showing due to a plugin/theme conflict by prefixing the EDD "no-js" body class**
I think multiple themes/plugins which rely on JavaScript detection do something similar to what we are doing, which is to add a `no-js` class to the body by default and then remove/replace it with a `js` class very early. On a site I saw yesterday, another plugin or theme was adding a `no-js` class to the `html` element, but it was not being removed/replaced. Since it's a different element than what our class is on, it ended up being a conflict and problem.

Now, that particular scenario is definitely a bug in something else on that customer's site, but I'm inclined to say that we should prefix our `no-js` class to be defensive about it, a little similar to prefixing functions.

So this one is not as easy to replicate but you can do it manually by adding a `no-js` class to the `html` element in dev tools. On the main branch, the buttons will disappear; on this branch, they will not.

**Fixing double buttons due to a theme with CSS conflicts**
I tested the class switching with multiple themes, including default themes, ours, and some that we've recommended in the past. It worked, but there are some themes out there which have CSS like this in the stylesheet:

```css
.button, .edd-submit, input[type="submit"].edd-submit {
	display: inline-block;
}
```

That's problematic because the inputs which should be hidden if JS is available are displaying, and the CSS is slightly more specific than what is in the main branch (it's both too specific, and not specific enough, as it's targeting JS and non JS inputs). So this branch adds a rule which should be just a little more specific. I considered using `!important` in the EDD CSS but I'm generally opposed to doing that unless I'm trying to override an opinionated plugin.

So to replicate this scenario, I would add the above CSS to the Additional CSS panel for your theme. In the main branch, you'll end up with double buttons; in this branch, you should have just the one correct one.

mihaijoldis:
I am trying to get a copy of a theme from a customer that caused this issue so we can replicate/test. The theme appeared to add the `no-js` to the page and he fixed it by removing that manually in the theme.

cklosowski:
I feel like we're going to need to make an announcement or write up something about this. We changed the `no-js` class name once already, and now we're changing it again, so anyone who adjusted, is going to have to adjust yet again.

What is our mitigation strategy on that user base, and how can we make it as seamless as possible for them?

robincornett:
The `no-js` body class was new in the 3.1 release so I don't think it's being used outside of EDD core yet. I'm proposing we prefix it so that we have control over it, as there are multiple plugins (I myself have at least one using this technique) and probably themes which do similar no-js/js class switching to account for JS specific content. I don't know that it's _necessary_ for us to prefix it as long as other plugins/themes manage how they handle it responsibly, but I did encounter a customer site where another actor was adding the `no-js` class to the `html` element and not removing it, which meant our CSS no longer worked. This would fix an issue like that.

EDD has always had a class `edd-no-js` for element which should show when JS was not enabled, but the classes on the body element are brand new so I feel fairly safe in making this change.

cklosowski:
Thanks @robincornett That's just fine.

cklosowski (approved):

mihaijoldis (approved):

PR #9514, opened 2022-10-17
Update customer address migration

### LINKED ISSUE #9480: Review customer address migration for EDD 3.0
A customer reported [that this step of the EDD 3.0 migration via CLI](https://github.com/awesomemotive/easy-digital-downloads/blob/9eafd82974c6b4eb421f798cfad98da0b893df03/includes/class-edd-cli.php#L1184-L1196) was never completing and locking his site.

He updated the query to this and it was able to complete:
```sql
UPDATE wp_edd_customer_addresses ca
SET ca.is_primary = 1
WHERE ca.id IN (
SELECT MAX(ca2.id)
FROM ( SELECT id, customer_id FROM wp_edd_customer_addresses WHERE type = 'billing' ) ca2
GROUP BY ca2.customer_id
)
```

However, looking at the `customer_addresses` method in the data migrator class, that method already sets the address retrieved from the `_edd_user_address` metadata as the primary, so this query seems potentially redundant.

Additionally, I believe this step is only run as part of a CLI migration, not through the UI migration, so we need to evaluate if it should be added to the UI migration or if it should be removed from the CLI migration.

Additionally, the `customer_addresses` method should more properly be using the `edd_maybe_add_customer_address` method instead of `edd_add_customer_address`, as this migration runs after the orders have migrated and the order addresses have been added to the database. So when the customer address is added at this point, it's actually generally creating a duplicate address, unless the customer updated their address after their most recent order (and therefore the user meta record is not a duplicate of an address stored with an order).

robincornett:
Fixes #9480

Proposed Changes:
This PR does a couple things:
1. Updates the customer address migration to use `edd_maybe_add_customer_address` to prevent duplicate addresses being added. On a test site, this reduced the number of addresses from 22k to 13.7k.
2. Updates the final query to check/update the address type.

### DIFF
--- a/includes/admin/upgrades/v3/class-customer-addresses.php
+++ b/includes/admin/upgrades/v3/class-customer-addresses.php
@@ -92,8 +92,7 @@ class Customer_Addresses extends Base {
 				SET ca.is_primary = 1
 				WHERE ca.id IN (
 					SELECT MAX(ca2.id)
-					FROM ( SELECT * FROM {$this->get_db()->edd_customer_addresses} ) ca2
-					WHERE ca2.type = 'billing'
+					FROM ( SELECT id, customer_id FROM {$this->get_db()->edd_customer_addresses} WHERE type = 'billing' ) ca2
 					GROUP BY ca2.customer_id
 				)
 			";
--- a/includes/admin/upgrades/v3/class-customer-addresses.php
+++ b/includes/admin/upgrades/v3/class-customer-addresses.php
@@ -92,7 +92,7 @@ class Customer_Addresses extends Base {
 				SET ca.is_primary = 1
 				WHERE ca.id IN (
 					SELECT MAX(ca2.id)
-					FROM ( SELECT id, customer_id FROM {$this->get_db()->edd_customer_addresses} WHERE type = 'billing' ) ca2
+					FROM ( SELECT id, customer_id FROM {$this->get_db()->edd_customer_addresses} WHERE ca.type = 'billing' ) ca2
 					GROUP BY ca2.customer_id
 				)
 			";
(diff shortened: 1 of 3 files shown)


mihaijoldis:
Sent this to a customer with the issue and they still reported that it's really really slow and not well optimized. Posting their last message and a possible solution they will use.

Unfortunately, the updated query included still isn't well
optimized in our environment and has to examine an insane number of
rows (as MySQL is running the sub-query for every updated row): 
```
UPDATE
{$wpdb->edd_customer_addresses} ca
SET ca.is_primary = 1
WHERE ca.id IN (
SELECT MAX(ca2.id)
FROM ( SELECT id, customer_id FROM
{$wpdb->edd_customer_addresses} WHERE type = 'billing' ) ca2
GROUP BY ca2.customer_id
)
```
Instead, we may exclude the query when migrating and then run
something like the below (not fully tested yet) to mark the primary
addresses: CREATE TEMPORARY TABLE tmpIds( id INT PRIMARY KEY );

```
INSERT INTO tmpIds
SELECT MAX(ca2.id)
FROM ( SELECT * FROM wp_edd_customer_addresses ) ca2
WHERE ca2.type = 'billing'
GROUP BY ca2.customer_id;

UPDATE wp_edd_customer_addresses ca
SET ca.is_primary = 1
WHERE EXISTS ( SELECT id FROM tmpIds WHERE ca.id = tmpIds.id );

DROP TABLE tmpIds;
```
This approach so far runs in a matter of seconds and
all indications are that it marks the addresses as expected

cklosowski:
It's unfortunate, but we cannot use the create temporary table feature as we cannot rely on every WordPress db user to have this explicit permission, however it looks like in a previous improvement to the migration process we may not need this anymore, anyway.

cklosowski (approved):

PR #9515, opened 2022-10-17
Removed Carbon modify function

### LINKED ISSUE #9502: "Last quarter" showing wrong dates in the reporting
### Expected behavior
If I select the `Last Quarter` period in the reports, I should be getting the correct quarter dates.

### Actual behavior
It should show July 1 to September 30, but instead I am getting:
image

zkawesome:
Fixes #9502 

According to this issue in the Carbon repository: https://github.com/briannesbitt/Carbon/issues/1080#issue-294106621 , the `modify` method inside the Carbon should not be modifying the timezone. They have removed the method here: https://github.com/briannesbitt/Carbon/pull/1070

We have to test this properly. All of our unit tests are passing and I have done multiple tests inside the reports with different timezones selected and have not found any issues so far, but this has to go through multiple rounds of testing.


cklosowski:
@zkawesome @robincornett if we can come up with a few different test cases, including setting specific timezones, etc, and steps to replicate this, let's hand it off to @mihaijoldis @dgoldak to run some tests on it as well.

cklosowski:
Clearing the milestone for this PR, as the issue that is solved with it is already tagged for the milestone.

cklosowski:
Upon initial review, this seems to solve the issue (and likely has a bunch of other small issues it will solve. I'd love to get some more eyes on testing, and once merged, we'll get this on our own data set to see how it's looking, as that is usually a good test to the result in carbon changes.

robincornett:
This seems good to me, but looking at the referenced PR, it looks like this is actually the one which was merged into Carbon: https://github.com/briannesbitt/Carbon/pull/1129 and it includes modifications to the `diffInSeconds` method as well as removing `modify`. Do we need to consider including that change as well?

Testing locally, it seems that the problem demonstrated was for a UTC+ timezone and did not present otherwise. Would it be possible to add a unit test to get the last quarter dates for one of those time zones, which should fail in `main`, and pass in this branch?

cklosowski:
@robincornett I don't think the `diffInSeconds` additions are necessary here as it might cause issues in other places. I think the best solution here is a minimal change to the quarter view and we can adopt more changes when we update carbon in a future release.

@zkawesome it would be great to get some unit tests around this so we don't break it in the future. If you need help writing tests let me or @robincornett know and we can get you started on that path.

cklosowski:
@robincornett @zkawesome I've added two unit tests to this, one for positive GMT offsets and one for negative GMT offsets. I verified that negatives pass on `release/3.1.0.2` and that positive offsets fail on `release/3.1.0.2`

Run on `release/3.1.0.2`
```
1) EDD\Reports\Reports_Functions_Tests::test_parse_dates_for_range_with_last_quarter_range_should_return_those_dates_utc_plus
--- Expected
+++ Actual
@@ @@
 Array (
-    'end' => '2022-09-30 23:59'
+    'end' => '2022-08-31 23:59'
```

cklosowski:
The one filled build was a result of a process failure, not the code.

robincornett:
@cklosowski I'm not seeing your new unit tests please.

robincornett (approved):
Update: I found the new unit tests--didn't realize they were added to the release branch.

Confirmed that the UTC+ test fails on the release branch and passes here. Since that's the failure we can replicate and it's fixed, I'm good with this.

cklosowski (approved):

PR #9523, opened 2022-10-21
Update CLI order migration to allow custom ID parameters

robincornett:
Related to #9358 

Proposed Changes:
This PR adds three new flags to the CLI payments migration:
1. `--id`: run the order migration for just one specific payment ID.
2. `--start`: run the order migration for payments greater than or equal to a given payment ID.
3. `--end`: run the order migration for payments less than or equal to a given payment ID.

`start` and `end` can be used separately or together; setting the `id` will use that ID only. It's assumed that these would be run only with `--force` but there is no guardrail around that.

I've wanted this for a while--there have been times where I've wanted to rerun a migration for just a specific order to troubleshoot something, or a batch of orders for same.


robincornett:
Updated to make some changes:
* Added the `--destroy` option, which can only be used in conjunction with `--force`. Additionally, if both flags are set, a confirmation step is added.
* If the destroy flag is confirmed, existing orders are destroyed before running the migration step.
* If the original flags are set (for the ID, start/end orders), the final steps are updated so that the migration is not marked complete and recalculations are not run, although that's suggested.

My hope with this PR is that users can re-run migrations for specific orders or groups of orders. Hypothetically, it's also possible to migrate a small/test batch of orders before running a full migration.

JJJ:
> If the destroy flag is confirmed, **existing orders** are destroyed **before** running the migration step.

1. Are **all** items destroyed (emptying the table) or only ones matching the `--start` & `--end` (or `--ID`) flags?
2. What happens if start, end, and ID are all passed?
3. What is the default behavior without `--destroy`? Does it check if it was migrated, and skip it if so?
4. Could you query for the existence of existing orders, get a count for possible duplicates, and prompt the CLI for a decision if you know for sure there might be 2 or more and `--force` is not used?
5. I assume any of the `ID` properties require the user of the CLI to already know the IDs in advance. Would it then be necessary to also provide a helper command to ask for any IDs that might not have been migrated (essentially a simple CLI for a complex MySQL query/subquery to check for and report back gaps or problems?)

robincornett:
> Are all items destroyed (emptying the table) or only ones matching the --start & --end (or --ID) flags?
The orders that are destroyed are the ones which are found when querying the database for an existing payment object.

> What happens if start, end, and ID are all passed?
If they are all passed, only `--id` is evaluated (start/end are handled separately and ignored if there is an ID).

> What is the default behavior without --destroy? Does it check if it was migrated, and skip it if so?
Without `--destroy`, the behavior is the current migration behavior: EDD checks to see if the order object exists in the orders table, and if it does, then the migration is not initiated.

> Could you query for the existence of existing orders, get a count for possible duplicates, and prompt the CLI for a decision if you know for sure there might be 2 or more and --force is not used?
> I assume any of the ID properties require the user of the CLI to already know the IDs in advance. Would it then be necessary to also provide a helper command to ask for any IDs that might not have been migrated (essentially a simple CLI for a complex MySQL query/subquery to check for and report back gaps or problems?)
I feel like these are a bit related....my thought on knowing the IDs is that at the point a user is using this tool, they'll have an idea of what payment/order they're wanting to migrate, or what payments might be missing in the case of putting in a range, so even a "ballpark" number would be close enough for the start/end flags (since there will definitely be gaps in the payment IDs due to other post types, logs, etc.).

As far as the `--dry-run` idea goes, I like that, although I'm not sure there is a way to "test" the migration without actually migrating data. The `orders` method doesn't return anything, although I expect we could make that happen easily enough, and then the CLI class could output a line saying "payment XX was not migrated" and give us a starting point for looking at a given payment for troubleshooting.

cklosowski:
@robincornett Can we make a couple adjustments?

1. I'd like to see the variables `$assoc_args['id']`, `$assoc_args['start']`, and `$assoc_args['end']` get assigned to variables so we aren't referencing the `$assoc_args` in multiple places.

2. The values for `id`, `start`, and `end` should be run through some sort of sanitization and validation to ensure they are numeric only. This way someone won't use a SON or ASON prefixed value in here. If we detect a non-numeric value, let's throw an error and tell them to use the IDs from the `post_id` column, not the number they see in the admin area.

3. For the helper information on the id, start, and end, ensure we mention to use the post_id column, not the payment number. This will hopefully help avoid someone trying to use a non-canonical value for these.

4. For the `id` before we start using it in the following section, let's ensure it's actually a payment record, and just exit early so we don't go any further if we don't have a valid payment ID.
```
		if ( ! empty( $assoc_args['id'] ) ) {
			$id             = absint( $assoc_args['id'] );
			$sql_base      .= " AND ID = {$id}";
			$full_migration = false;
```

cklosowski (changes requested):
See requested changes: https://github.com/awesomemotive/easy-digital-downloads/pull/9523#issuecomment-1291094092

robincornett:
Updated to confirm that id, start, end values are numeric before attempting to use, and error messages if not.

Also added the check for the payment existing if the ID is set, although I'm not sure that's necessary as it's technically already handled by the database query in the next steps. It's fine, but redundant, maybe.

cklosowski:
@robincornett while redundant and the query below does get us into the end result, as migrating data is a process that we cannot account for all the aspects and to ensure we don't go any further and rely on 'migration processes', I think it's safer to just ditch if we can't find the ID in the first place.

cklosowski (approved):

PR #9527, opened 2022-10-21
Separate product query for dropdowns

robincornett:
Proposed Changes:
1. Creates a new `get_products` method in the HTML elements class, to allow the products to be retrieved separately from rendering the product dropdown.
2. Implements the new method for the download metabox for bundled products.


mihaijoldis:
Applid this to the user in question. Page load time dropped to a bit over 2 seconds compared to minutes before.
Also noticed queries dropped from 576 to 440. I would say this is a perfect fix!

cklosowski (approved):

PR #9535, opened 2022-10-26
Re: Issue #9443 -- Improved price_id handling

### LINKED ISSUE #9443: edd_get_item_discount_amount assumes discount options have price IDs
### Expected behavior
When I migrate a discount code that has product restrictions, I expect the `edd_get_item_discount_amount` to not throw an error.

### Actual behavior
Because we added `price ID` support to the product restrictions in discounts in 3.0, a code that was migrated for a product that has variable prices doesn't contain the price ID options for the products.

1) Migrate a discount that has a product restriction for a product with variable prices.
2) Attempt to use the discount code ensuring it runs through `edd_get_item_discount_amount`
3) See a notice.

`Undefined index: options in easy-digital-downloads/includes/discount-functions.php on line 1125`

PHP Version: All

EDD Version (or branch): 3.0

WordPress Version: All

StephenGravitt:
- not all cart items are set with an "options" array
- in many cases, cart items instead use a "item_number" array
- safely check if this item has "options" if not, use the latter

Fixes #9443 

Proposed Changes:
1. Check cart item to see if it has 'options' array available
2. If not, grab the 'price_id' by means of the commonly used ['item_number']['options'] array

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1122,7 +1122,11 @@ function edd_get_item_discount_amount( $item, $items, $discounts, $item_unit_pri
 					foreach ( $items as $i ) {
 						if ( ! in_array( $i['id'], $excluded_products ) ) {
 							if ( edd_has_variable_prices( $i['id'] ) ) {
-								$i_amount = edd_get_price_option_amount( $i['id'], $i['options']['price_id'] );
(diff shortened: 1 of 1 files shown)


cklosowski:
Thanks @StephenGravitt. I'm going to user your fix as a basis in another PR.

PR #9538, opened 2022-10-28
Update discount migration to allow for custom metadata

### LINKED ISSUE #9531: Custom discount metadata may not be migrated to new discount object
### Expected behavior
If I register custom metadata on a discount in EDD 2.x, I expect it to be migrated to EDD 3.0.

### Actual behavior
If the custom metadata is prefixed with `_edd_discount`, it is not included in the [array of metadata to migrate](https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/admin/upgrades/v3/class-data-migrator.php#L180-L185). I believe that the migration expects any meta with this prefix to be core metadata which should be mapped to a property.

1. Create a discount in EDD 2.11.
2. Add custom metadata to it and make sure the meta key is prefixed with `_edd_discount`. (See [this gist](https://gist.github.com/polevaultweb/4f595894150d5c06867af032677b2f3a) for an example; the meta key is `_edd_discount_use_renewals`).
3. Migrate the discount.
4. Check the migrated object metadata.

PHP Version: 7.4
EDD Version (or branch): 3.0/main
WordPress Version: 6.0.2

If the `_edd_discount` prefix was intended solely as an internal prefix, it makes sense to not migrate any meta matching it, but I don't think that's been the case, so I think the meta keys should be evaluated instead (or in addition) as to whether they do map to an existing property, and then migrate anything that's left over.

I am opening this as an issue but I think we also need to consider that if someone has custom metadata assigned to something like a discount, they may need to handle the migration themselves, as we do in some of our extensions (Simple Shipping comes to mind). There is a `edd_post_insert_discount` hook which could be used during the migration to convert custom metadata to the new discount object.

robincornett:
Fixes #9531

Proposed Changes:
1. This updates the `discounts` method in the data migrator class to explicitly check any post meta with the `_edd_discount` prefix against the core meta/properties for discounts.
2. If existing meta is not a core property, it should be added to the new discount object's metadata.

To Test:
This could be tested by using the gist linked in the original issue to add some custom meta to a discount for handling discounts on subscriptions.
1. Create the discount in EDD 2.11, making sure that the custom meta is saved to the discount.
2. Migrate the discount to EDD 3.0. In the main branch, that custom meta will be lost; in this branch, it should persist (whether the linked code works in 3.0 is a separate issue as I think it uses some post functions which may not be compatible).
3. This could also be tested with this PR in Recurring: https://github.com/awesomemotive/edd-recurring/pull/1661

### DIFF
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -175,7 +175,7 @@ class Data_Migrator {
 		$args            = array();
 		$meta            = get_post_custom( $data->ID );
 		$meta_to_migrate = array();
-		$meta_to_ignore  = array(
+		$core_meta       = array(
 			'code',
 			'name',
 			'status',
@@ -202,7 +202,7 @@ class Data_Migrator {
 				continue;
 			}
 			$meta_key = str_replace( '_edd_discount_', '', $key );
-			if ( ! in_array( $meta_key, $meta_to_ignore, true ) ) {
+			if ( ! in_array( $meta_key, $core_meta, true ) ) {
 				$meta_to_migrate[ $meta_key ] = $value;
 				continue;
 			}
--- a/includes/admin/upgrades/v3/class-data-migrator.php
+++ b/includes/admin/upgrades/v3/class-data-migrator.php
@@ -175,17 +175,39 @@ class Data_Migrator {
 		$args            = array();
 		$meta            = get_post_custom( $data->ID );
 		$meta_to_migrate = array();
+		$meta_to_ignore  = array(
+			'code',
+			'name',
+			'status',
+			'uses',
+			'max_uses',
+			'amount',
+			'start',
+			'expiration',
+			'type',
+			'min_price',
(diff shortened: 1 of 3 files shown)


robincornett:
Note: This PR can also be tested with All Access as it adds custom metadata to the discounts (`_edd_discount_edd_all_access_discount_restrict`).

cklosowski:
@robincornett this seems to work well. I wonder if there is a way we could 'lazy load' any of the custom meta that didn't get migrated when a discount object is loaded, the first time or maybe update the discount migration to pull this data if we run a `--force` and we don't re-add the discount but do run the custom meta check on it?

cklosowski:
We'd only be able to do that is we haven't removed legacy data already.

robincornett:
We've done something similar in some of our extensions and it was manageable because the extension knew exactly what it was looking for. My concern with trying to do it in core would be that core would have to query all possible legacy data, without knowing exactly what, and that feels expensive.

cklosowski (approved):

PR #9547, opened 2022-11-03
Convert UTC dates to store timezone equivalent

zkawesome:
I believe that we should be converting the UTC timezone to the store timezone via `edd_get_edd_timezone_equivalent_date_from_utc`. I debugged a little bit and found out that UTC+ dates are not converted correctly with current DateTime class.

How to replicate:

- Inside the reports, go into hourly mode (1 or 2 days custom range)
- Your hours of the orders will be in the UTC date instead of store equivalent - test with UTC+ timezones.

If this solution is correct, we have to check other graphs as well.


robincornett (commented):
I think this seems promising. I can set my timezone to one that's affected/wrong in the release branch and the overview graph is correct with this branch (UTC-4).

Also I just merged the release branch into this.

robincornett:
@zkawesome another request to add to this PR: can we add the site timezone setting to the EDD system info file please?

dgoldak:
Update from a customer who applied the fix:
> That does seem to do the trick, although I'm in UTC+1 (The Netherlands)
https://secure.helpscout.net/conversation/2062571257/2281041/

cklosowski:
@zkawesome Let's go ahead and button this up. I like the suggestion from @robincornett to add the timezone to the system info. If we need to change any other graphs let's do that and get this submitted for formal review.

robincornett (approved):
Note: local unit tests against WordPress 5.2 are passing (with the polyfill).

includes/admin/tools.php
```diff
@@ -1350,6 +1350,7 @@ function edd_tools_sysinfo_get() {
 	$return .= 'Language:                 ' . ( ! empty( $locale ) ? $locale : 'en_US' ) . "\n";
 	$return .= 'Permalink Structure:      ' . ( get_option( 'permalink_structure' ) ? get_option( 'permalink_structure' ) : 'Default' ) . "\n";
 	$return .= 'Active Theme:             ' . $theme . "\n";
+	$return .= 'Timezone:                 ' . wp_timezone_string() . "\n";
```
cklosowski, line 1353: Since we're really curious about what timezone EDD is using for charts/graphs, could we include the result from this function as well:
`edd_get_timezone_abbr()`

cklosowski (changes requested):

robincornett (dismissed):

mihaijoldis (approved):

cklosowski:
This is working for using a timezone setting that is UTC based (UTC -6, UTC, UTC +6), but as soon as I switch to a city that is UTC negative, like Chicago (currently UTC -6), I'm getting a date on the 18th of the month, but it's only the 17th in the timezone I've selected:
Graph: `this_month`
Timezone: `Chicago`
Screenshot 2022-11-17 at 1 13 33 PM

Graph: `this_month`
Timezone: `GMT -6`
Screenshot 2022-11-17 at 1 17 27 PM

cklosowski:
@zkawesome that seems to have done it! Nice work :)

cklosowski (approved):

PR #9549, opened 2022-11-05
Improve edd_has_active_discounts for performance and accuracy #9526

### LINKED ISSUE #9526: Limit of 10 discounts when running edd_has_active_discounts can produce inaccurate display of discount field at checkout.
## Bug Report
In [includes/discount-functions.php:395](https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/discount-functions.php#L379) the code loops through the 10 most recent discount codes to decide whether the discount form should be shown on the checkout page. The problem is that if the most recent 10 discount codes have all expired, it won't look for any more and won't show the discount code field.

https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/class-edd-discount.php#L1648

1) Create a working discount code
2) Create 10 more discount codes, with expiry dates in the past
3) Visit the checkout page

### Any other relevant information:

This could be solved in numerous ways. For example, increasing the sample size from 10, or excluding expired codes from the initial query.

cklosowski:
Fixes #9526 

Proposed Changes:
1. Adds 2 parameters to `edd_has_active_discounts`
1a. `$sample_size` (Default: `10`). Serves as a way to increase the number of discounts queried to determine active discounts, depending on the context. 
1b. `$refresh` (Default: `false`). Serves as a way to force updating the new transient (see 2)
2. Adds a new transient `edd_has_active_discounts`. This stores as a `1` or `0` depending on if we find active discounts for the store. The transient has a filterable 5 minute expiration time.
3. Updates All admin based actions for discounts (add, update, activate, deactivate, delete), to force a refresh if the action was successful, with a sample size of `50` to help improve accuracy.
4. Adds a hook into the `edd_after_payment_actions` and if the oder being processed had `discount` adjustments on it, we force the refresh of the transient, with 50 as the sample size.

While the 5 minute transient may seem long, keep in mind a few things.

Any order that is processed will run the after payment actions, and update the transient. Any action by an admin to edit a discount will also refresh the transient.

### DIFF
--- a/includes/admin/discounts/discount-actions.php
+++ b/includes/admin/discounts/discount-actions.php
@@ -55,7 +55,6 @@ function edd_admin_add_discount( $data = array() ) {
 	// Setup default discount values.
 	$to_add            = array();
 	$to_add['status']  = 'active';
-	$current_timestamp = current_time( 'timestamp' );
 
 	$data = array_filter( $data );
 
@@ -142,6 +141,11 @@ function edd_admin_add_discount( $data = array() ) {
 		? 'discount_added'
 		: 'discount_add_failed';
 
+	if ( ! empty( $created ) ) {
+		// Now re-prime the transient for the has_acitve_discounts check
+		edd_has_active_discounts( 50, true );
+	}
+
 	// Redirect.
 	edd_redirect( add_query_arg( 'edd-message', sanitize_key( $arg ), $data['edd-redirect'] ) );
 }
@@ -186,7 +190,6 @@ function edd_admin_edit_discount( $data = array() ) {
 
 	// Prepare update
 	$to_update    = array();
-	$current_time = current_time( 'timestamp' );
 
 	$data = array_filter( $data );
 
@@ -284,6 +287,11 @@ function edd_admin_edit_discount( $data = array() ) {
 		? 'discount_updated'
 		: 'discount_not_changed';
 
+	if ( ! empty( $updated ) ) {
+		// Now re-prime the transient for the has_acitve_discounts check
+		edd_has_active_discounts( 50, true );
+	}
+
 	// Redirect
 	edd_redirect( add_query_arg( 'edd-message', sanitize_key( $arg ), $data['edd-redirect'] ) );
 }
@@ -322,6 +330,11 @@ function edd_admin_delete_discount( $data = array() ) {
 		? 'discount_deleted'
 		: 'discount_deleted_failed';
 
+	if ( ! empty( $deleted ) ) {
+		// Now re-prime the transient for the has_acitve_discounts check
+		edd_has_active_discounts( 50, true );
+	}
+
 	// Redirect
 	edd_redirect( remove_query_arg( 'edd-action', add_query_arg( 'edd-message', sanitize_key( $arg ), $_SERVER['REQUEST_URI'] ) ) );
 }
@@ -355,6 +368,11 @@ function edd_activate_discount( $data = array() ) {
 		? 'discount_activated'
 		: 'discount_activation_failed';
 
+	if ( ! empty( $activated ) ) {
+		// Now re-prime the transient for the has_acitve_discounts check
+		edd_has_active_discounts( 50, true );
+	}
+
 	// Redirect
 	edd_redirect( remove_query_arg( 'edd-action', add_query_arg( 'edd-message', sanitize_key( $arg ), $_SERVER['REQUEST_URI'] ) ) );
 }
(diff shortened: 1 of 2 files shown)


includes/admin/discounts/discount-actions.php
```diff
@@ -142,6 +141,11 @@ function edd_admin_add_discount( $data = array() ) {
 		? 'discount_added'
 		: 'discount_add_failed';
 
+	if ( ! empty( $created ) ) {
+		// Now re-prime the transient for the has_acitve_discounts check
```
robincornett, line 145:
```suggestion
		// Now re-prime the transient for the has_active_discounts check.
```
Typo in each instance of this comment--

robincornett (commented):
I'm confused as to why this is needed in EDD 3.0 actually. Using `edd_get_discounts` with the status set to active should automatically exclude any discounts which are expired or have met their max uses (they will be updated to `expired` and `inactive` accordingly). We should only need to look for any active discount and the number of discounts to query should not matter.

The original issue cites specific lines of code but does not indicate an EDD version. I wonder if this is an issue resolved already in EDD 3.0.

For example, I have 57 discount codes in my database. The most recent 50, I generated via CLI with a past expiration date. Running `edd_has_active_discounts` on the release/3.1.0.3 branch returns true, as four active discounts were found. Every other discount in my database is expired or inactive (hit max uses).

It looks like `edd_get_discounts` loads some compatibility functionality...I wonder if we could accomplish the same functionality and be more efficient, without needing a transient, with just this?

```php
function edd_has_active_discounts() {

	// Query for active discounts.
	$discounts = edd_get_adjustments( array(
		'number' => 1,
		'status' => 'active',
		'type'   => 'discount',
	) );

	return ! empty( $discounts );
}
```

And looping through each found discount object to run the `is_active` helper method should also not be needed.

cklosowski:
@robincornett the main issue is that we don't actually set the discounts to 'inactive' if they are past expiration until they've been loaded.

So if you have 10 discounts created, with a past expiration (not that you just created, but are sitting in the database), until we run `is_active` on them they will return in the query.

Up until that point where we run `$discount->is_active()` the adjustment will be `active` and have an expiration date in the past, but then once `is_active()` is run it will move to `inactive`, but since we only pulled 10 discounts, if all 10 of those are expired (and now inactive that we've run the check of `is_active` it won't actually work until all "expired" discounts are marked as inactive.

includes/admin/discounts/discount-actions.php
cklosowski: Fixed in https://github.com/awesomemotive/easy-digital-downloads/pull/9549/commits/61e661223acf9be5e7f9919354166b21621a5973

includes/discount-functions.php
```diff
@@ -373,31 +373,50 @@ function edd_get_discount_notes( $discount_id = 0 ) {
...
-	foreach ( $discounts as $discount ) {
-		/** @var $discount EDD_Discount */
-		if ( $discount->is_active( false, true ) ) {
-			return true;
+		$has_active_discounts = false;
+	} else {
+		// Check each discount for active status, applying filters, etc...
+		foreach ( $discounts as $discount ) {
+			/** @var $discount EDD_Discount */
+			if ( $discount->is_active( false, true ) ) {
```
robincornett, line 407: I'm wondering if the first parameter should be set to `true` if `$refresh` is true. I've manually set a lot of discounts to active even though their expiration is in the past and just loading the checkout is not getting to an active but unexpired one generally. Deactivating one discount did expire it, and the transient is set to true, but as soon as it expires, it will be false again because the other expired discounts are still active. (Granted this may be an edge case but if the transient is being forcibly refreshed with a larger sample size maybe worth doing?)

robincornett (changes requested):

includes/discount-functions.php
```diff
@@ -373,31 +373,52 @@ function edd_get_discount_notes( $discount_id = 0 ) {
...
 
-	// Check each discount for active status, applying filters, etc...
-	foreach ( $discounts as $discount ) {
-		/** @var $discount EDD_Discount */
-		if ( $discount->is_active( false, true ) ) {
-			return true;
+	$has_active_discounts = false;
+
+	// Query for active discounts.
+	$discounts = edd_get_discounts(
```
robincornett, line 397: This is better, but I'm not sure it's fixing the issue as defined in the replication steps.

This is querying the ten most recent `active` discounts, in this case, and if they're all expired, then the function will return false. I experimented with updating the discounts query to include an end date query, something like this (I know this is hard coded so would obviously have to change):
```
	// Query for active discounts.
	$discounts = edd_get_discounts(
		array(
			'number'         => $sample_size,
			'status'         => 'active',
			'end_date_query' => array(
				'relation' => 'OR',
				array(
					'after'     => array(
						'year'  => 2022,
						'month' => 11,
						'day'   => 16,
					),
					'inclusive' => false,
				),
			),
		)
	);
```
That worked to only get active discounts with future expiration dates, but it fails to get active discounts with no expiration dates, and looking at the `WP_Date_Query` class, I am not sure how to do a combination query for a `null` value in the expiration column. But it seems like if we could do this query, it would solve the reported issue.

PR #9551, opened 2022-11-05
Deprecate old payment actions that are no longer needed #9550

### LINKED ISSUE #9550: Many old legacy hooks running from very old EDD versions.
While researching #9526 I found a large number of old hooks running on very old versions of EDD:
* edd_update_payment_backwards_compat - Added in 2.3
`Listens to the updated_postmeta hook for our backwards compatible payment_meta updates, and runs through them`

* edd_update_old_payments_with_totals - Added in 1.2
`Updates all old payments, prior to 1.2, with new meta for the total purchase amount`

* edd_clear_user_history_cache - Added in 1.2.2
`Flushes the current user's purchase history transient when a payment status is updated`
_I cannot find anywhere that we are actually getting a transient that uses `edd_user_*` in core or extensions._

* edd_cleanup_stats_transients - Added in 2.6.7
`Deletes edd_stats_ transients that have expired to prevent database clogs`
_I'm pretty sure these aren't necessary anymore, with the new query classes and Models for Downloads_

My proposal is to unhook all these, and move them to deprecated functions. Since all of these have hooks, they will impact load time, even if they don't find anything to modify or delete, it's just adding load time and further processing to do nothing.

cklosowski:
Fixes #9550 

Moves the following functions to `deprecated-functions.php` and runs the deprecation function.

* edd_update_payment_backwards_compat - Added in 2.3
`Listens to the updated_postmeta hook for our backwards compatible payment_meta updates, and runs through them`

* edd_update_old_payments_with_totals - Added in 1.2
`Updates all old payments, prior to 1.2, with new meta for the total purchase amount`

* edd_clear_user_history_cache - Added in 1.2.2
`Flushes the current user's purchase history transient when a payment status is updated`
_I cannot find anywhere that we are actually getting a transient that uses `edd_user_*` in core or extensions._

* edd_cleanup_stats_transients - Added in 2.6.7
`Deletes edd_stats_ transients that have expired to prevent database clogs`
_I'm pretty sure these aren't necessary anymore, with the new query classes and Models for Downloads_


robincornett (changes requested):
Unit tests are failing:
- I think the last line needs to be removed from `test_edd_update_payment_status`.
- The `test_edd_upgrade_payments` test may need to be removed entirely.

PR #9554, opened 2022-11-07
Remove currency formatting from File Downloads and New Customers graph

### LINKED ISSUE #9553: File Downloads report graph is formatting numbers as float and currency
### Expected behavior
When I view the file downloads report, I expect the graph to show integers (as you can't process a fractional download) and not be a currency.

### Actual behavior
The values put on the tooltips are formatted as currency and floats:
image

1) Have a site with file downloads
2) View Downloads > Reports > File Downloads

PHP Version: All

EDD Version (or branch): 3.1

WordPress Version: All

zkawesome:
Fixes #9553 

Proposed Changes:
Removed the currency formatting from the File Downloads and New customers graph, so that it shows as an integer.
image

### DIFF
--- a/includes/admin/reporting/reports-callbacks.php
+++ b/includes/admin/reporting/reports-callbacks.php
@@ -118,7 +118,7 @@ function edd_overview_sales_earnings_chart() {
 
 		// Loop through each date there were sales/earnings, which we queried from the database.
 		foreach ( $earnings_results as $earnings_result ) {
-			$date_of_db_value = edd_get_edd_timezone_equivalent_date_from_utc( EDD()->utils->date( $earnings_result->date ) );
+			$date_of_db_value = new DateTime( $earnings_result->date, $timezone );
 
 			// Add any sales/earnings that happened during this hour.
 			if ( $hour_by_hour ) {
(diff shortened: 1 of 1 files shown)


includes/admin/reporting/reports-callbacks.php
```diff
@@ -118,7 +118,7 @@ function edd_overview_sales_earnings_chart() {
 
 		// Loop through each date there were sales/earnings, which we queried from the database.
 		foreach ( $earnings_results as $earnings_result ) {
-			$date_of_db_value = new DateTime( $earnings_result->date, $timezone );
+			$date_of_db_value = edd_get_edd_timezone_equivalent_date_from_utc( EDD()->utils->date( $earnings_result->date ) );
```
robincornett, line 121: I think this change and line 146 are really from a different issue/PR and should be removed from this? Otherwise the PR looks good.

cklosowski: @zkawesome I think robin is right here, looks like this is from the #9278 PR

robincornett (approved):

PR #9577, opened 2022-11-14
Added edd_login_fields_before_submit hook

flowdee:
The hook was inserted identically to the already existing one from the registration form.

This allows, just like in the registration form, to include enhancements such as a captcha.


flowdee:
@cklosowski 🙂🙏

PR #9579, opened 2022-11-15
Add fallback to item_number and set a default i_amount #9443

### LINKED ISSUE #9443: edd_get_item_discount_amount assumes discount options have price IDs

cklosowski:
Fixes #9443 

Proposed Changes:
1. Adds a default `$i_amount` to `0`
2. Adds checks for the `options` array key, else the `item_number - options` array keys.

### DIFF
--- a/includes/discount-functions.php
+++ b/includes/discount-functions.php
@@ -1125,11 +1125,8 @@ function edd_get_item_discount_amount( $item, $items, $discounts, $item_unit_pri
 							$i_amount = 0;
 
 							if ( edd_has_variable_prices( $i['id'] ) ) {
-								if ( isset( $i['options'] ) ) {
-									$i_amount = edd_get_price_option_amount( $i['id'], $i['options']['price_id'] );
-								} elseif ( isset( $i['item_number']['options'] ) ) {
-									$i_amount = edd_get_price_option_amount( $i['id'], $i['item_number']['options']['price_id'] );
-								}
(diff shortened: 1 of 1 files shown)


robincornett (commented):
I think I would feel a little better if the `isset` checks went all the way to the price ID level, so:
```php
isset( $i['options']['price_id'] )
isset( $i['item_number']['options']['price_id'] )
```
But I kind of would even prefer something like this instead:
```php
$price_id = isset( $i['options']['price_id'] ) ? $i['options']['price_id'] : $i['item_number']['options']['price_id'];
$i_amount = edd_get_price_option_amount( $i['id'], $price_id );
```

cklosowski:
@robincornett Fixed up those in https://github.com/awesomemotive/easy-digital-downloads/pull/9579/commits/10d1f2000368a508572235941ceedd11b65e65c6

PR #9580, opened 2022-11-15
Improve local file detection

cklosowski:
Proposed Changes:
1. Updates our `edd_is_local_file` to also check the `home_url` if it is different than the `site_url`
2. Adds a filter allowing custom installations to still report this as local.

### Additional Notes
In 2.11.7 and lower, we used `home_url` as the basis for determining if a file was local. in 3.0+ we switched this to `site_url`. The problem with this is that in some custom setups, these are different and the content might not be in the same path as the WordPress installation.

### DIFF
--- a/includes/process-download.php
+++ b/includes/process-download.php
@@ -402,7 +402,7 @@ function edd_is_local_file( $requested_file ) {
 	 * Allow filtering the edd_is_local_file detection.
 	 *
 	 * EDD tries to identify if a file is hosted locally, so that we can use the proper file delivery method.
-	 * By default we check the site_url, and then the home_url (in the event that those settins are different).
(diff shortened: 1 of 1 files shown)


robincornett (commented):
I think the code is fine, although I'm not completely convinced of the need for the filter at this time (if we know that the `site_url()` generally works, and the `home_url()` will revert to behavior that worked in 2.x). But I know this is also to manage a different kind of scenario so I'll defer judgment on that.

cklosowski:
@robincornett The `home_url()` and `site_url()` should cover MOST cases we run into. The filter allows us to add support for many other types of setups, and allows custom hosting configurations to apply a filter at the host level or managed level if need be.

robincornett (approved):

PR #9583, opened 2022-11-16
Add support for togglable "Advanced Settings" in download file metabox row

### LINKED ISSUE #9261: Add "Show advance settings" support to the file download rows

I need to add an additional data for each downloadable file, so files can be "previewed" before being downloaded (in this case assigning preview mp3s, images, etc). While programatically adding another "row" under the file can be done, it looks ridiculous (and confusing) to the user. So my solution was this:

Screenshot 2022-07-02 at 00 47 10

Ideally, I would like the ability for the file form rows to support "Show advanced options" like the variable price rows. This would allow me to hide these fields on load and make things neater and more accessible (for people who need them). 

### Justification or use case

EDD add-on plugin development.

davidsherlock:
Fixes #9261

Added supported for "advanced settings" in the file metabox row. This was needed as I need to allow the ability to add settings on a per file basis and the row option wasn't a good option from a UX perspective.

See an example of it working here:

Screenshot 2022-11-17 at 01 53 23

### DIFF
--- a/includes/admin/downloads/metabox.php
+++ b/includes/admin/downloads/metabox.php
@@ -790,6 +790,9 @@ add_action( 'edd_meta_box_files_fields', 'edd_render_files_field', 20 );
  * Used to output a table row for each file associated with a download.
  * Can be called directly, or attached to an action.
  *
+ * Added support for advanced settings with the the edd_download_file_option_row
+ * action hook (@since 3.1.0.3).
+ *
  * @since 1.2.2
  * @param string $key Array key
  * @param array $args Array of all the arguments passed to the function
@@ -797,6 +800,7 @@ add_action( 'edd_meta_box_files_fields', 'edd_render_files_field', 20 );
  * @return void
  */
 function edd_render_file_row( $key, $args, $post_id, $index ) {
+	global $wp_filter;
 
 	$args = wp_parse_args( $args, array(
 		'name'           => null,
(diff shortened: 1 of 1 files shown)


includes/admin/downloads/metabox.php
```diff
@@ -806,20 +810,27 @@ function edd_render_file_row( $key, $args, $post_id, $index ) {
...
 
-	$prices           = edd_get_variable_prices( $post_id );
-	$variable_pricing = edd_has_variable_prices( $post_id );
-	$variable_display = $variable_pricing ? '' : ' style="display:none;"';
-	$variable_class   = $variable_pricing ? ' has-variable-pricing' : ''; ?>
+	$prices           		= edd_get_variable_prices( $post_id );
+	$variable_pricing 		= edd_has_variable_prices( $post_id );
+	$variable_display 		= $variable_pricing ? '' : ' style="display:none;"';
+	$variable_class   		= $variable_pricing ? ' has-variable-pricing' : '';
+	$custom_file_options 	= isset( $wp_filter['edd_download_file_option_row'] ) ? true : false; ?>
```
robincornett, line 817:
```suggestion
	$custom_file_options = has_action( 'edd_download_file_option_row' );
	?>
```
And the global can be removed from line 803. Also please fix the alignment/spaces in the previous variables (some tabs have gotten mixed in).

includes/admin/downloads/metabox.php
```diff
@@ -806,20 +810,27 @@ function edd_render_file_row( $key, $args, $post_id, $index ) {
...
 
 	<div class="edd-repeatable-row-header edd-draghandle-anchor">
 		<span class="edd-repeatable-row-title" title="<?php _e( 'Click and drag to re-order files', 'easy-digital-downloads' ); ?>">
 			<?php printf( esc_html__( '%1$s file ID: %2$s', 'easy-digital-downloads' ), esc_html( edd_get_label_singular() ), '<span class="edd_file_id">' . esc_html( $key ) . '</span>' ); ?>
 			<input type="hidden" name="edd_download_files[<?php echo esc_attr( $key ); ?>][index]" class="edd_repeatable_index" value="<?php echo esc_attr( $index ); ?>"/>
 		</span>
+		<?php
+		$actions = array();
+		if ( $custom_file_options ) {
+			$actions['show_advanced'] = '<a href="#" class="toggle-custom-price-option-section">' . __( 'Show advanced settings', 'easy-digital-downloads' ) . '</a>';
```
robincornett, line 827:
```suggestion
			$actions['show_advanced'] = '<a class="toggle-custom-price-option-section">' . __( 'Show advanced settings', 'easy-digital-downloads' ) . '</a>';
```
This really should be a button instead of a link but I know this is following the pattern set by the existing "button"...but the `href` should be removed, pending fixing all of the buttons in the metaboxes.

includes/admin/downloads/metabox.php
```diff
@@ -887,7 +898,21 @@ function edd_render_file_row( $key, $args, $post_id, $index ) {
 		<?php do_action( 'edd_download_file_table_row', $post_id, $key, $args ); ?>
 
 	</div>
-<?php
+
+	<?php
+	if ( $custom_file_options ) {
+		?>
+
+		<div class="edd-custom-price-option-sections-wrap">
```
robincornett, line 906: I know this is using the class that it is to inherit the JS behavior without needing to modify the JS, but I would suggest adding an additional file specific class here so that it's a little less confusing when reading the code/markup.

robincornett (changes requested):

includes/admin/downloads/metabox.php
davidsherlock: @robincornett I actually badly hurt my hand at the weekend, so typing (and coding) is a pain. Would you be able to make the required changes here please? Would help me a lot, as I am taking a break from coding until my hand recovers (also upgrading my mouse and keyboard to help).

PR #9586, opened 2022-11-17
Update receipt link email tag rendering

### LINKED ISSUE #9414: Logging in from the {receipt_link} page simply refreshes the login form
### Expected behavior
When I view the {receipt_link} page as not logged in I am presented with a login form. When entering my credentials the page should log me in and show the receipt page.

### Actual behavior
The login form simply refreshes and I am not logged in.

1) Add {receipt_link} to your purchase receipt and make a purchase.
2) From the email, make sure you are not logged in and access that link
3) enter your login details on the form

PHP Version: 7+

EDD Version (or branch): main

WordPress Version: latest

Reported on .org here https://wordpress.org/support/topic/login-for-view-receipt-not-working/
I replicated it locally and the page simply refreshes. The functionality is in the 'edd_receipt_shortcode' from line 702. I also think the `$edd_login_redirect = edd_get_current_page_url();` code might not be right as that for me returns `https://www.localhost.com/edd/` instead of the full receipt view URL

robincornett:
Fixes #9414

Proposed Changes:
1. Updates the `edd_receipt_shortcode` parameters to include a redirect attribute, set to the `edd_get_current_page_url()` by default.
2. Updates `edd_render_receipt_in_browser()`:
    * To use the `$data` parameter provided by the EDD hook.
    * To return early if the login submission is detected so that the `edd_post_actions` will take precedence instead.
    * To add the payment ID/EDD action to a URL to add to the shortcode as a custom redirect since this is not part of the `$wp->request`.

### DIFF
--- a/includes/emails/template.php
+++ b/includes/emails/template.php
@@ -259,12 +259,38 @@ function edd_render_receipt_in_browser( $data ) {
 <!DOCTYPE html>
 <html lang="en">
 	<head>
-		<title><?php _e( 'Receipt', 'easy-digital-downloads' ); ?></title>
+		<title><?php esc_html_e( 'Receipt', 'easy-digital-downloads' ); ?></title>
 		<meta charset="utf-8" />
 		<meta name="robots" content="noindex, nofollow" />
 		<?php wp_head(); ?>
+		<style>
+			body.edd_receipt_page {
+				margin: 12px auto;
+				align-items: center;
+				border: 1px solid #cfcfcf;
+				max-width: fit-content;
+				padding: 12px 24px;
+				border-radius: 8px;
+			}
+
+			.edd_receipt_page #edd_login_form fieldset {
+				border: none;
+				display: grid;
+			}
+
+			.edd_receipt_page #edd_login_form label,
+			.edd_receipt_page #edd_login_form input[type=text],
+			.edd_receipt_page #edd_login_form input[type=password]{
+				display: block;
+				width: 100%;
+			}
+
+			.edd_receipt_page th {
+				text-align: left;
+			}
+		</style>
 	</head>
-<body class="<?php echo apply_filters('edd_receipt_page_body_class', 'edd_receipt_page' ); ?>">
+<body class="<?php echo esc_attr( apply_filters( 'edd_receipt_page_body_class', 'edd_receipt_page' ) ); ?>">
 	<div id="edd_receipt_wrapper">
 		<?php do_action( 'edd_render_receipt_in_browser_before' ); ?>
 		<?php echo do_shortcode( '[edd_receipt payment_key=' . $key . ']' ); ?>
--- a/includes/emails/template.php
+++ b/includes/emails/template.php
@@ -234,13 +234,25 @@ function edd_get_sale_notification_body_content( $payment_id = 0, $payment_data 
  *
  * @since 1.5
  * @author Sunny Ratilal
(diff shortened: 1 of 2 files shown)


mihaijoldis:
So I opened the page and put in random user/password. The URL of the site is the same but I was somehow redirected to my main homepage and that's where the error shows https://a.supportally.com/i/NB5JPy

Logging in with a different user/pasword than the purchase shows https://a.supportally.com/i/pELdPR which I assume is correct but It does look weird to not have some sort of header/footer at all in this view

Logging in with the right account shows the receipt https://a.supportally.com/i/pq6xzi but again it looks really weird.

robincornett:
Yes, I agree it looks weird but that appears to have been the intent for this view--it is not loading WP, just rendering the receipt. I didn't change how it looks/loads.

Personally I would advocate for using the new `{receipt}` email tag which will use the receipt page if it's set (or confirmation page if not), and that should render as you expect. With the receipt block instead of the shortcode, it will support viewing the receipt as a guest as well.

robincornett:
@zkawesome is that with the shortcode or block please? If the block is also affected by this issue we should try to fix that in the 2.0.2 release for blocks to have it in this version.

robincornett:
@zkawesome I've pushed a couple more commits which should fix this. Instead of adding a new parameter to the shortcode I'm using `edd_get_receipt_page_uri( $order->id );` in the shortcode form. (Right now my latest is not showing here for me for some reason but I pushed it.)

I will make a new issue/PR in blocks for the same treatment.

cklosowski:
While this seems to work I think we can do a little better with the styling of the login form. Since the receipt page loads with no template, we're left with an awkward login view:
image


I applied some CSS to the `edd_receipt_page` body class and it seems to make it at least look a little less broken:
```
    margin: 12px auto !important;
    align-items: center;
    border: 1px solid #cfcfcf !important;
    max-width: fit-content;
    padding: 12px 24px;
    border-radius: 8px;
```

Screenshot 2022-11-21 at 1 11 43 AM

cklosowski:
Here are the Twenty Twenty Two views without and with these styles:
Screenshot 2022-11-21 at 1 13 20 AM

Screenshot 2022-11-21 at 1 13 31 AM

robincornett:
So I've pushed another commit to add styling to the page that EDD makes for this. However, I'm realizing that the changes I made on Friday are changing the behavior of this receipt/login process.

What this email tag has _historically_ done is to direct the user to a stripped down page on the website showing just the receipt. There are a couple of hooks here that aren't anywhere else, and none of the site's navigation, etc. will show on this page.

The changes I made on Friday, for the shortcode redirect to always go to the receipt page, end up rendering the standard receipt page if someone does go through the login process, so it's actually a pretty significant behavior change.

To keep the original behavior we probably need to go back to something that honors the receipt link URL, without necessarily adding a new parameter to the shortcode. My thought would be to define the global `$edd_login_redirect` in the `edd_render_receipt_in_browser` function and then only override it in the shortcode if it's not already defined.

If we do that, Chris' suggested styling makes the rendered receipt a little more palatable, although it's still very stripped down.

cklosowski:
For some reason the PR isn't showing all the changes, you can view them here: https://github.com/awesomemotive/easy-digital-downloads/compare/release/3.1.0.3...issue/9414

cklosowski (approved):

cklosowski (approved):

PR #9589, opened 2022-11-21
Update early return in edd_build_order

### LINKED ISSUE #9570: PayPal Buy now functionality no longer works for guests
### Expected behavior
Be able to use buy now as a guest
### Actual behavior
Modal window closes with an error `An unexpected error occurred. Please try again.`
1) Use buy now buttons or the purchase_link shortcode with the `direct=true` parameter
2) Try to purchase as a guest

PHP Version:7+

WordPress Version: latest

part of it is coming from a change we made here: https://github.com/awesomemotive/easy-digital-downloads/issues/9157 where we return false and do not create an order at all if there is no email address

robincornett:
Fixes #9570

Proposed Changes:
1. Removes the user info/email check from `edd_build_order`.
2. Adds an early return if no order ID was found or the order was not successfully added.

### DIFF
--- a/includes/orders/functions/orders.php
+++ b/includes/orders/functions/orders.php
@@ -649,13 +649,6 @@ function edd_build_order( $order_data = array() ) {
 		return false;
 	}
 
-	$order_data = wp_parse_args(
-		$order_data,
-		array(
-			'user_info' => array(),
-		)
-	);
-
 	/* Order recovery ********************************************************/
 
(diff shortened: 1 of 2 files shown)


mihaijoldis:
So this seems to be working fine but noticed something but then maybe it always worked like that.
When Auto Register is active and you make a guest purchase shouldn't be a user account created?

cklosowski (approved):

PR #9619, opened 2023-02-08
Better structured data filters

leewillis77:
This PR modifies the class responsible for producing LD+JSON structured data as follows
1. Corrects the schema type of the brand attribute from `Thing` to `Brand` as `Thing` is invalid 
2. Introduces two new filters (`edd_generate_download_structured_data_variable_price_offer` / `edd_generate_download_structured_data_offer`) for the Offer elements of the structured data that pass the `EDD_Download` being processed as well as the specific `price` being processed for variable priced items
3. Passes the `EDD_Download` being processed to the `edd_generate_download_structured_data` filter so that filters know which download is being processed without having to guess from the calculated SKU and potentially re-load the `EDD_Download` instance

### DIFF
--- a/includes/class-structured-data.php
+++ b/includes/class-structured-data.php
@@ -165,34 +165,43 @@ class Structured_Data {
 			$offers = array();
 
 			foreach ( $variable_prices as $price ) {
-				$offers[] = array(
-					'@type'           => 'Offer',
-					'price'           => $price['amount'],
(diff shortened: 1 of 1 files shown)


robincornett:
@leewillis77 thank you for submitting the PR--we've incorporated your filters into the 3.1.4 release.

PR #9683, opened 2023-10-23
Don't assign default type in Downloads' "Files" metabox

goncalovf:
Fixes https://github.com/awesomemotive/easy-digital-downloads/issues/9682


robincornett:
Hi @goncalovf! Thank you for the report and the PR. This is the change we are intending to implement in EDD 3.2.4.

goncalovf:
This change was in fact implemented in EDD 3.2.4, so I'm going to go ahead and close this PR.

Issue #9699, opened 2023-11-28
Database error when performing an Order query containing a date_query and a product_id


jkudish:
### Expected behavior

Performing a Order query containing a date_query and a product_id should return results

### Actual behavior

Performing a Order query containing a date_query and a product_id produces the following database error: `WordPress database error Column 'date_created' in where clause is ambiguous for query SELECT...`

### Steps to reproduce the behavior

Perform a query similar to this:

```php
$query = array(
    'product_id' => 1
    'date_query' => array(
        array(
            'after'     => '2023-11-01 00:00:00',
            'before'    => '',
            'inclusive' => true,
        ),
    ),
);

// this will produce a database error:
// WordPress database error Column 'date_created' in where clause is ambiguous for query SELECT ... 
// this is due to the JOIN on the order items table
$orders = edd_get_orders( $query );
```

For ease of reproduction I've made a minimum viable reproduction plugin in the form of a WP_CLI command availalble here: https://gist.github.com/jkudish/e5cd9254abfa4e4903dd837c72513c8f

### Information (if a specific version is affected):

PHP Version: any
EDD Version (or branch): 3.0+
WordPress Version: any

I will submit a PR that fixes this shortly

PR #9700, opened 2023-11-28
Use table alias/name when building date query

jkudish:
This PR attempts to fix the issue described in #9699 

Use the received table name & table alias arguments in the `get_sql` function to set those as properties on the Date class. Note that `get_sql` already gets called with those params as per the line below. However the `get_sql` function oddly wasn't defining those arguments in its definition.

 https://github.com/awesomemotive/easy-digital-downloads/blob/715373d18971c8b12e949bb3fd5d6c40aac6c234/includes/database/engine/class-query.php#L1363 

Then, when building the SQL query, use the alias or name (only if available) to prepend to the column name in the query. This prevents ambiguity issues when performing the SQL query

This prevents and fixes the issue described in https://github.com/awesomemotive/easy-digital-downloads/issues/9699

I am open to receiving feedback on the solution proposed and working on an updated solution.

Thanks!

### DIFF
--- a/includes/database/engine/class-date.php
+++ b/includes/database/engine/class-date.php
@@ -137,6 +137,19 @@ class Date extends Base {
 		'AND'
 	);
 
+
+	/**
+	 * @since 3.2.6
+	 * @var string|null Table name
+	 */
+	public $table_name = null;
+
+	/**
+	 * @since 3.2.6
+	 * @var string|null Table alias
+	 */
+	public $table_alias = null;
+
 	/**
 	 * Constructor.
 	 *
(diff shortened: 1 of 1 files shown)


robincornett:
Hi @jkudish, and thank you so much for looking into this and submitting a PR! I'll check it over locally, but I wanted to ask if you would be willing to also submit the issue and PR over in the [BerlinDB repository](https://github.com/berlindb/core)? Berlin is the package/library we are using to manage our custom database tables, and based on what you're saying, it looks like this is a fix that should be applied in there.

jkudish:
Apologies for the delay. I've moved this over to https://github.com/berlindb/core/pull/164 as suggested

PR #9706, opened 2024-01-03
Fix casing in the change name form

chriscct7:
Fixes #

Proposed Changes:
1.
2.
3.


chriscct7:
will merge

PR #9734, opened 2024-05-27
Fix a typo class-edd-notices.php

slaFFik:
"permenently" => "permanently"

I also improved the translators experience when working with the nginx-related notice.

I merged 2 translatable strings into one.


cklosowski:
Thanks @slaFFik We'll get this fixed up in the next major release when we do a hand-full of translation improvements. I'm going to close this PR as we'll update it in the private repo.

PR #9735, opened 2024-05-31
Trigger JS event on checkout success.

zutigrm:
Fixes #

Proposed Changes:
1. There is a JS event triggered in case of checkout failure `edd_checkout_error`, but not the one for successful purchase. It will be very handy to have one, to align with other plugins that need to listen to this action as well, like in our case for custom tracking.


cklosowski:
@zutigrm I think this has some potential but we need to do some testing and compatibility research for all different gateway situations (off-site, on-site, etc) as well as ensuring that having this trigger doesn't open up the potential for breaking the checkout processing if the Javascript that hooks into the `trigger` fails.

We'll look at this and do some testing to see what we feel comfortable with.

PR #9747, opened 2024-07-09
Updated: hook

### LINKED ISSUE #9746: Support post type admin column for `hierarchical` post type download
## Support Post Type Admin Column for hierarchical Post Type

Currently, the plugin uses the action hook[ `manage_posts_custom_column`](https://github.com/awesomemotive/easy-digital-downloads/blob/main/includes/admin/downloads/dashboard-columns.php#L109). This hook only supports the default case. No data will appear in the post type admin column if a developer changes the post type attribute to `hierarchical`. It can be solved either by changing the hook to `manage_download_posts_custom_column` or adding the hook `add_action( 'manage_pages_custom_column', 'edd_render_download_columns', 10, 2 );`. The first one is recommended.

codersantosh:
Updated: Changed the hook `manage_posts_custom_column` to `manage_download_posts_custom_column`

Fixes #9746

This update will support both the default case and hierarchical post types for the `download`.

### DIFF
--- a/includes/admin/downloads/dashboard-columns.php
+++ b/includes/admin/downloads/dashboard-columns.php
@@ -106,7 +106,7 @@ function edd_render_download_columns( $column_name, $post_id ) {
 			break;
 	}
 }
-add_action( 'manage_posts_custom_column', 'edd_render_download_columns', 10, 2 );
+add_action( 'manage_download_posts_custom_column', 'edd_render_download_columns', 10, 2 );
 
 /**
  * Registers the sortable columns in the list table


JJJ (changes requested):
This seems like a good idea.

Recommend to remove this check at the same time:

https://github.com/awesomemotive/easy-digital-downloads/blob/cbe963bd61d6de5ead0eacb8b7cf1ae276c68284/includes/admin/downloads/dashboard-columns.php#L52-L56

PR #9752, opened 2024-08-05
Emails: Fix wrong filter name in new_user email

Giordanidis:


cklosowski:
Thanks @Giordanidis We've resolved this in version 3.3.4.

Issue #9757, opened 2024-08-28
miss $user_login and $user_data for retrieve_password_title filter


AkramiPro:
## Bug Report

**it get error when set ```$accepted_args``` as 3 in ```add_filter```**

```EDD\Emails\Templates\PasswordReset::get_default_subject()```

![image](https://github.com/user-attachments/assets/cb1b4c00-2032-4536-bc3a-1bf70a78f844)

![image](https://github.com/user-attachments/assets/91717fc0-87e1-431a-be5d-1a593781260f)

cklosowski:
This has been resolved with version 3.3.4.

PR #9760, opened 2024-08-31
Fix `retrieve_password_title` filter args

AkramiPro:
Add `$user_login` and `$user_data` to `retrieve_password_title` filter in password reset emails. #9757


cklosowski:
Thanks @AkramiPro We've resolved this with version 3.3.4.

